PR #1404, opened 2014-12-08
Updates to /status API

beav:
This patch returns additional data from the /status API call. Specifically, it
returns a worker list, the message broker connection status, the database
connection status, and the Pulp version (RHBZ #704184).

There is a small behavior change to include the resource manager and scheduler
as part of the worker heartbeat list. Previously, these were not included in
the list.

### DIFF
--- a/bindings/pulp/bindings/server_info.py
+++ b/bindings/pulp/bindings/server_info.py
@@ -17,6 +17,7 @@ Handles calls to the server that query the plugin and type capabilities.
 
 from pulp.bindings.base import PulpAPI
 
+
 class ServerInfoAPI(PulpAPI):
     def __init__(self, pulp_connection):
         super(ServerInfoAPI, self).__init__(pulp_connection)
@@ -50,13 +51,4 @@ class ServerInfoAPI(PulpAPI):
         @return: Response
         """
         path = self.base_path + 'distributors/'
-        return self.server.GET(path)
-
-    def ping(self):
-        """
-        Retrieves basic status information from the server.
-
-        @return: Response
-        """
-        path = '/v2/services/status/'
         return self.server.GET(path)
--- a/server/pulp/server/async/worker_watcher.py
+++ b/server/pulp/server/async/worker_watcher.py
@@ -28,21 +28,6 @@ from pulp.server.managers import resources
 _logger = logging.getLogger(__name__)
 
 
-def _is_resource_manager(event):
-    """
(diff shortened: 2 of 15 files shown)


server/pulp/server/async/worker_watcher.py
```diff
@@ -132,10 +113,6 @@ def handle_worker_offline(event):
     """
     event_info = _parse_and_log_event(event)
 
-    # if this is the resource_manager do nothing
-    if _is_resource_manager(event):
-        return
-
     msg = _("Worker '%(worker_name)s' shutdown") % event_info
     _logger.info(msg)
     _delete_worker(event_info['worker_name'], normal_shutdown=True)
```
beav: This was not being done correctly in the original PR. I've since updated.

beav:
@rbarlow I think I got everything. I saw that you LGTM'd but if you don't mind, can you re-review?

I will merge this to 2.6-testing as well when I merge.

bindings/pulp/bindings/server_info.py
```diff
@@ -51,12 +52,3 @@ def get_distributors(self):
...
         return self.server.GET(path)
-
-    def ping(self):
-        """
-        Retrieves basic status information from the server.
-
-        @return: Response
-        """
-        path = '/v2/services/status/'
-        return self.server.GET(path)
```
rbarlow: Since the bindings could be considered part of our API, we shouldn't remove this until 3.0.0. However, we could insert a DeprecationWarning.

rbarlow:
Re-LGTM, except that I think we should deprecate that binding rather than remove it.

bindings/pulp/bindings/server_info.py
beav: `DeprecationWarning` and `PendingDeprecationWarning` are ignored by default in Python 2.7+ unless the user adds `-Wd` to the command line.

I will add a bug to remove this in 3.0 and mention the bug and deprecation in the method's docblock.

bmbouter: @beav you are correct this should just be removed. The webcode handler is there, but it hasn't been wired up in application.py in a long time. There is no way for a web request to reach this code.

beav:
Closing PR, will re-open a new one against `2.6-testing` and merge since this is already LGTM'd.

beav:
merged in #1478

PR #1445, opened 2014-12-18
1171278 - additional helper method for finding units

beav:
The fix for RHBZ 1171278 in pulp_rpm needs to be able to find a unit based on the
unit key, regardless of which repo its in. This patch adds
`find_unit_by_unit_key` to the SearchUnitsMixin to facilitiate this.

**NOTE**: this should not be merged until after 2.4.4 is released!


rbarlow:
Can you resubmit this PR against 2.4-dev? Then it'll be safe to merge now since 2.4.4 isn't out the door yet.

beav:
closing to re-open against 2.4-dev

PR #1447, opened 2014-12-18
port to gofer 2.1.

jortel:
Updated to work with gofer 2.3.  Latest upstream and in Fedora 21.
- improved support for multiple messaging adapters (formerly transports).
- standardized url which includes messaging adapter.  format: *adapter*+*scheme*://*user*:*password*@*host*:*port*/*virtual-host*  instead of the *transport* option.
- hundreds of new unit tests.


jortel:
Should be merged to 2.6-testing.  Replaced by #1509.

PR #1463, opened 2014-12-24
PEP-8 pulp.server.db.model.dispatch and its tests.

rbarlow:
It turned out that the tests for pulp.server.db.model.dispatch were split between two files. This PR also merges these tests together into one file.


beav:
LGTM

PR #1470, opened 2014-12-24
Run flake8 in run-tests.py.

rbarlow:
This commit adds support for flake8 to run-tests.py. It currently only
checks the files that I've already made sure were PEP-8 to catch PEP-8
regressions.

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -11,6 +11,16 @@ from pulp.devel.test_runner import run_tests
 PROJECT_DIR = os.path.dirname(__file__)
 subprocess.call(['find', PROJECT_DIR, '-name', '*.pyc', '-delete'])
 
+# Check for style
+config_file = os.path.join(PROJECT_DIR, 'flake8.cfg')
+# These paths should all pass PEP-8 checks
+paths_to_check = [
+    'server/pulp/server/tasks/',
+    'server/test/unit/server/']
+paths_to_check = [os.path.join(PROJECT_DIR, p) for p in paths_to_check]
+command = ['flake8', '--config', config_file]
+command.extend(paths_to_check)
+flake8_exit_code = subprocess.call(command)
 
 PACKAGES = [
     os.path.dirname(__file__),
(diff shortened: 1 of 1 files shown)


run-tests.py
```diff
@@ -11,6 +11,15 @@
 PROJECT_DIR = os.path.dirname(__file__)
 subprocess.call(['find', PROJECT_DIR, '-name', '*.pyc', '-delete'])
 
+# Check for style
+config_file = os.path.join(PROJECT_DIR, 'flake8.cfg')
+# These paths should all pass PEP-8 checks
+paths_to_check = [
+    os.path.join(PROJECT_DIR, 'server', 'pulp', 'server', 'tasks'),
+    os.path.join(PROJECT_DIR, 'server', 'test', 'unit', 'server')]
```
beav: Can this be a list of strings instead? That would be easier to edit as we fix stuff, even if it's just like

```python
paths_to_check [os.path.join(PROJECT_DIR, partial_path) for partial_path in list_of_pep8_dirs)]
```

run-tests.py
```diff
@@ -37,4 +46,6 @@
 dir_safe_all_platforms = [os.path.join(os.path.dirname(__file__), x) for x in TESTS_ALL_PLATFORMS]
 dir_safe_non_rhel5 = [os.path.join(os.path.dirname(__file__), x) for x in TESTS_NON_RHEL5]
 
-sys.exit(run_tests(PACKAGES, dir_safe_all_platforms, dir_safe_non_rhel5))
+tests_exit_code = run_tests(PACKAGES, dir_safe_all_platforms, dir_safe_non_rhel5)
+
+sys.exit(flake8_exit_code or tests_exit_code)
```
beav: cool!

run-tests.py
rbarlow: On 01/02/2015 04:18 PM, Chris Duryee wrote:
> Can this be a list of strings instead? That would be easier to edit as
> we fix stuff, even if it's just like
> 
> paths_to_check [os.path.join(PROJECT_DIR, partial_path) for partial_path in list_of_pep8_dirs)]

Sure!

run-tests.py
rbarlow: On 01/02/2015 04:19 PM, Chris Duryee wrote:
> cool!

Thanks! Could you LGTM this for me?

beav:
LGTM.  `paths_to_check` change is only suggested but not required

rbarlow:
I've updated this script with the change you suggested. I cannot merge this PR until these two get reviewed:

https://github.com/pulp/pulp/pull/1463

and

https://github.com/pulp/pulp/pull/1459

PR #1475, opened 2015-01-03
Updated epel.rpm link, please note this will keep getting outdated as ep...

Jmainguy:
...el link will keep changing

### DIFF
--- a/docs/sphinx/user-guide/installation.rst
+++ b/docs/sphinx/user-guide/installation.rst
@@ -85,7 +85,7 @@ Repositories
 
    RHEL7::
 
-    $ sudo rpm -Uvh https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-2.noarch.rpm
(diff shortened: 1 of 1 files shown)


barnabycourt:
Is there a clean way to document this that will not keep getting outdated?

Jmainguy:
We could provide a static epel rpm (Version does not really matter, older versions will just update when they yum update anyways)

We could try and get fedora to start using a static link which always post to their latest rpm. (Would be really helpful if they did this)

We could point users to go to this page 
http://download-i2.fedoraproject.org/pub/epel/7/x86_64/repoview/epel-release.html
And then tell them to rpm -Uvh the latest rpm on that page 

We could write a long string which curls the page, greps for rpm, returns the rpm, and have rpm -Uvh run against it, probably prone to error and not the best idea.

Of those options to any seem attractive to y'all, or y'all have any better ideas?

beav:
I'd be :+1: for "provide static epel version and let it get updated via yum update"

dkliban:
I didn't realize that this pull request existed.  I submitted my own yesterday with a note on how to find out the proper version.  It got merged yesterday.  https://github.com/pulp/pulp/pull/1621

PR #1476, opened 2015-01-05
1111591 - require passwords to not be empty strings

asmacdo:
[BZ-1111591](https://bugzilla.redhat.com/show_bug.cgi?id=1111591)

Okaara cannot recognize an empty string as a password, so we should not allow the user to get into a broken state.

Reopened from https://github.com/pulp/pulp/pull/1367

### DIFF
--- a/client_admin/test/unit/test_auth.py
+++ b/client_admin/test/unit/test_auth.py
@@ -11,8 +11,7 @@
 # have received a copy of GPLv2 along with this software; if not, see
 # http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
 
-import os
-import sys
+import unittest
 
 import mock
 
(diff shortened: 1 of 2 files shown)


asmacdo:
@rbarlow, since you reviewed the WIP, could you take a look at this?

rbarlow:
This PR needs some unit tests to go with the new change. Let me know when you've added a test to make sure the empty password is handled correctly, and I'll check it out again.

asmacdo:
@rbarlow, there were very minimal tests of the affected functions. I have expanded these tests to cover all affected code.

client_admin/test/unit/test_auth.py
```diff
@@ -55,29 +54,92 @@ def test_invalid_input(self):
...
-class TestAuthUser(base_builtins.PulpClientTests):
+class TestAuthUser(unittest.TestCase):
 
-    USER_LOGIN = 'test-login'
-    PASSWORD = 'test-password'
-    NAME = 'test-name'
+    def test__prompt_password_given(self):
+        """
+        Test that prompt password will ask for a password until the password
+        is not blank
```
rbarlow: This docstring doesn't seem accurate for what this test is doing. It looks like the tests just makes sure the password works in the normal case.

rbarlow:
LGTM.

PR #1477, opened 2015-01-05
1066022 - remove incorrect role update doc

beav:
Roles are updated via permissions now, but the doc still referenced an old way
of doing it. This patch removes the role update API document and refers people
to the permissions API.

### DIFF
--- a/docs/sphinx/dev-guide/integration/rest-api/role/cud.rst
+++ b/docs/sphinx/dev-guide/integration/rest-api/role/cud.rst
@@ -1,5 +1,8 @@
-Create, Update, and Delete
-==========================
+Create and Delete
+=================
+
+This page details role creation and deletion. Roles are not updated per-se;
+please see the :doc:`Permissions API <../permission/actions>` for more detail.
 
 Create a Role
 -------------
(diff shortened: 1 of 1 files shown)


mhrivnak:
LGTM

ipanova:
I think this bug created a lot of confusion, so I wanted to clarify that it is still *possible* to update role.
You can update with PUT method the `description` and `display name` of the role but you cannot add `permissions` in this call. After Saily's fix it explcitly tells that:
```
 "error": {
        "code": "PLP0000",
        "data": {},
        "description": "Update Keyword [permissions] is not supported",
        "sub_errors": []
    },
```
So there is need only to remove this line https://github.com/pulp/pulp/pull/1477/files#diff-756b4ce388bdb7c489851208b9489e46L65 and update the sample request that will exclude the permissions.

skarmark:
Ya, like what Ina said, we can still update display name and description of a role through this API. I am not sure whether it is that useful though, so we can potentially get rid of the API. Either we should do that or update the docs to just remove permissions from valid params and add a link to permissions doc to update permissions on roles (Chris has already taken care of the link part which looks good)

beav:
Closing this PR and opening a new one so I can point it to `2.6-testing`.

PR #1478, opened 2015-01-05
Updates to /status API

beav:
This patch returns additional data from the /status API call. Specifically, it
returns a worker list, the message broker connection status, the database
connection status, and the Pulp version (RHBZ #704184).

There is a small behavior change to include the resource manager and scheduler
as part of the worker heartbeat list. Previously, these were not included in
the list. These entries are filtered out when Pulp requests a free worker.

A few PEP8 fixes are included as well for files that were touched in this patch.

Additionally, the `ping()` method in ServerInfoAPI has been removed; this
called a URL that no longer exists.

### DIFF
--- a/bindings/pulp/bindings/server_info.py
+++ b/bindings/pulp/bindings/server_info.py
@@ -17,6 +17,7 @@ Handles calls to the server that query the plugin and type capabilities.
 
 from pulp.bindings.base import PulpAPI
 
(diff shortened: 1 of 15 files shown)


beav:
This was LGTM'd as part of #1404 . I put in a new PR just to switch the branch.

PR #1482, opened 2015-01-06
1174283 - bump python-requests to 2.4.3

asmacdo:
[BZ-1174283](https://bugzilla.redhat.com/show_bug.cgi?id=1174283)

This change already [exists in nectar](https://github.com/pulp/nectar/blob/master/python-nectar.spec#L19), but it needs to be copied over to the pulp deps for building.

### DIFF
--- a/deps/python-nectar/python-nectar.spec
+++ b/deps/python-nectar/python-nectar.spec
@@ -16,7 +16,7 @@ BuildArch:      noarch
 BuildRequires:  python-setuptools
 
 Requires:       python-isodate >= 0.4.9
-Requires:       python-requests >= 2.2.1
+Requires:       python-requests >= 2.4.3
 
 %description
 Nectar is a download library that abstracts the workflow of making and tracking


asmacdo:
@beav, should this go to 2.6-testing instead?

rbarlow:
See my comment on the BZ. I'm not convinced that we need Pulp's require statement to be updated. As far as I am aware, Pulp is not using any features that require 2.4 of requests, and so its require statement for 2.2 is fine and actually correct.

asmacdo:
@rbarlow, the change is in the python-nectar spec.

rbarlow:
@asmacdo I'm not sure I'm following. Are you saying that you agree that we don't need this same change here?

asmacdo:
@rbarlow, sorry for not being clear. The original change was to nectar, which does use 2.4.3 functionality. I changed the python-nectar.spec file in the nectar repo.

The change here is in not in the pulp require statements, it is in the nectar require statements in the pulp deps. The python-nectar.spec should be identical (and will be after this change) to the python-nectar spec file in the nectar repo.

rbarlow:
Man, I feel dumb now. I thought this change was in Pulp's spec file, but I see now that it's the nectar spec in deps. Sorry for the confusion, and LGTM!

PR #1484, opened 2015-01-06
Support flake8 as part of the devel test runner

barnabycourt:


devel/pulp/devel/test_runner.py
```diff
@@ -60,9 +69,28 @@ def run_tests(packages, tests_all_platforms, tests_non_rhel5):
...
     # don't run the server or plugins tests in RHEL5.
+    flake8_exit_code = 0
     if sys.version_info >= (2, 6):
+        # make sure we test everything
         args.extend(tests_non_rhel5)
+
+        # Check the files for coding conventions
+        if run_flake8:
+            if not flake8_paths:
+                flake8_paths = ['.']
```
rbarlow: We should probably find a better way of choosing the path. This assumes that your CWD is a particular value, but the run-tests.py script does not otherwise assume that. You may need to have this method require these paths to be set to avoid this problem since it will be difficult to determine otherwise.

devel/pulp/devel/test_runner.py
```diff
@@ -60,9 +69,28 @@ def run_tests(packages, tests_all_platforms, tests_non_rhel5):
...
+
+            flake8_args = ['flake8', '--config', 'flake8.cfg', ]
+            if flake8_excludes:
+                joined_flake8_excludes = ','.join(flake8_excludes)
+                flake8_args.extend(['--excludes', joined_flake8_excludes])
+
+            for flake_path in flake8_paths:
+                flake8_command = copy.copy(flake8_args)
+                flake8_command.append(flake_path)
+                flake8_exit_code = (flake8_exit_code or subprocess.call(flake8_command))
```
rbarlow: flake8 accepts multiple paths as arguments, so rather than calling it several times like this you could just append all the args to flake_args and run it once. This is how run-tests.py currently works.

rbarlow: Also, running it this way will not show all the errors at once (it's failfast).

devel/pulp/devel/test_runner.py
```diff
@@ -60,9 +69,28 @@ def run_tests(packages, tests_all_platforms, tests_non_rhel5):
...
+
     else:
+        flake8_exit_code = 0
```
rbarlow: This was already set to 0 at the beginning of this method.

devel/pulp/devel/test_runner.py
rbarlow: The other option is to require these paths to be absolute paths, and to require them to be present. Thus, it is the caller's responsibility to make sure the paths do not depend on CWD.

run-tests.py
```diff
@@ -11,16 +11,10 @@
 PROJECT_DIR = os.path.dirname(__file__)
 subprocess.call(['find', PROJECT_DIR, '-name', '*.pyc', '-delete'])
 
-# Check for style
-config_file = os.path.join(PROJECT_DIR, 'flake8.cfg')
 # These paths should all pass PEP-8 checks
 paths_to_check = [
     'server/pulp/server/tasks/',
     'server/test/unit/server/']
-paths_to_check = [os.path.join(PROJECT_DIR, p) for p in paths_to_check]
```
rbarlow: By dropping this, you cause this script to depend on CWD when it was CWD agnostic before. I think it's important to be able to call the tests with any path from any working directory. If we simply require this param, expect the caller to pass it as absolute paths, and keep this line, everything will continue to work.

run-tests.py
```diff
@@ -11,16 +11,10 @@
 PROJECT_DIR = os.path.dirname(__file__)
 subprocess.call(['find', PROJECT_DIR, '-name', '*.pyc', '-delete'])
 
-# Check for style
-config_file = os.path.join(PROJECT_DIR, 'flake8.cfg')
```
rbarlow: You've also assumed CWD by dropping this.

devel/pulp/devel/test_runner.py
```diff
@@ -60,9 +69,28 @@ def run_tests(packages, tests_all_platforms, tests_non_rhel5):
...
     if sys.version_info >= (2, 6):
+        # make sure we test everything
         args.extend(tests_non_rhel5)
+
+        # Check the files for coding conventions
+        if run_flake8:
+            if not flake8_paths:
+                flake8_paths = ['.']
+
+            flake8_args = ['flake8', '--config', 'flake8.cfg', ]
```
rbarlow: This is assuming that CWD has flake8.cfg.

rbarlow:
I don't think we should merge this until it works no matter which directory the user runs it from.

barnabycourt:
Agree'd, I'll think through the CWD issue

rbarlow:
Another thing that we could do as part of this PR, or later if you prefer, is to support the -x flag on flake8. I added -x a while back - it's basically a failfast on any tests that fail. For flake8, we could exit immediately if the -x flag was set and there were any flake8 problems, rather than running the tests after flake8. If you'd prefer to do that as a separate effort, let me know and I'll file a task about it.

PR #1487, opened 2015-01-06
do not throw exception information away

asmacdo:
Pulp caught exceptions (Exception) and then reraised as a PulpDataException. Additionally, errors are sometimes thown that contain other errors inside e.extra_data. This information was thrown away and could not be used.

Solution renders all error data recursively and allows the exception to bubble up rather than reraising as a different exception.

Reopened from https://github.com/pulp/pulp/pull/1392

This is necessary for the right information to get back to the user after we merge https://github.com/pulp/pulp_docker/pull/52

### DIFF
--- a/server/test/unit/test_repo_distributor_manager.py
+++ b/server/test/unit/test_repo_distributor_manager.py
@@ -237,7 +237,7 @@ class RepoDistributorManagerTests(base.PulpServerTests):
         try:
             self.distributor_manager.add_distributor('rohan', 'mock-distributor', {}, True)
             self.fail('Exception expected')
-        except exceptions.PulpDataException:
+        except Exception, e:
             pass
 
         # Cleanup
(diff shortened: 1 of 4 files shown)


server/test/unit/test_repo_distributor_manager.py
```diff
@@ -237,7 +237,7 @@ def test_add_distributor_validate_raises_error(self):
         try:
             self.distributor_manager.add_distributor('rohan', 'mock-distributor', {}, True)
             self.fail('Exception expected')
-        except exceptions.PulpDataException:
+        except Exception, e:
```
asmacdo: The reason for this is that the exception that bubbles up is a generic Exception. Previously we caught this and reraised more specific exception.

asmacdo:
@rbarlow since you looked at this before, would you mind taking another?

asmacdo:
@rbarlow, there we go. Ready for review.

rbarlow:
LGTM.

PR #1488, opened 2015-01-06
Fixes the mutation of queue name to get it right on all hostnames

bmbouter:
https://bugzilla.redhat.com/show_bug.cgi?id=1179463

The issue was that strip() treats the string passed into it as an iterator, which was stripping extra characters on hostnames that end in 'd', 'q', or '.'. The incorrect resultant queue name caused the task dispatched to not go to the correct worker.

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -271,7 +271,8 @@ class Task(CeleryTask, ReservedTaskMixin):
         ### This is a temporary workaround for https://github.com/celery/celery/issues/2216 ###
         queue_name_to_save = queue
         if kwargs.get('exchange', None) == DEDICATED_QUEUE_EXCHANGE:
-            kwargs['queue'] = queue.rstrip('.dq')
+            # Strip off '.dq' from the end of the queue name
+            kwargs['queue'] = queue[:-3]
         #######################################################################################
 
         async_result = super(Task, self).apply_async(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -271,7 +271,8 @@ def apply_async(self, *args, **kwargs):
         ### This is a temporary workaround for https://github.com/celery/celery/issues/2216 ###
         queue_name_to_save = queue
         if kwargs.get('exchange', None) == DEDICATED_QUEUE_EXCHANGE:
-            kwargs['queue'] = queue.rstrip('.dq')
+            # Strip off '.dq' from the end of the queue name
+            kwargs['queue'] = queue[:-3]
```
mhrivnak: I assume this means you're 100% confident that the queue value will always end in ``.dq``. If not, you could use something like ``if foo.endswith('.dq'):`` or a regex-based substitution.

mhrivnak:
LGTM

server/pulp/server/async/tasks.py
bmbouter: I am very, very confident that when the exchange specifies the dedicated queue exchange, the queue name contains a .dq at the end of it introduced in the assigned_queue name on [L#126](https://github.com/bmbouter/pulp/blob/1179463/server/pulp/server/async/tasks.py#L126). Pulp only uses the dedicated queue exchange for tasks dispatched using appy_async_with_reservation.

PR #1489, opened 2015-01-06
PEP-8 pulp.server.content.sources.

rbarlow:
I ended up also PEP-8ing some Nodes modules that I had to modify due to
changes in this package as well.


barnabycourt:
LGTM

PR #1491, opened 2015-01-07
1123509 - deprecated ssl_ca_certificate setting in server.conf

skarmark:
https://bugzilla.redhat.com/show_bug.cgi?id=1123509

### DIFF
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -101,6 +101,9 @@
 # ssl_ca_certificate: full path to the CA certificate used to sign the Pulp
 #     server's SSL certificate; consumers will use this to verify the
 #     Pulp server's SSL certificate during the SSL handshake
+#     Deprecated! - Please note that this setting will be removed in the next
+#     release and will be replaced by a setting in the consumer.conf file
+#     that will allow the user to specify a directory path containing certificates.
 #
 # user_cert_expiration: number of days a user certificate is valid
 #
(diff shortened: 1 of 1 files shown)


server/etc/pulp/server.conf
```diff
@@ -110,7 +113,7 @@
 [security]
 # cacert: /etc/pki/pulp/ca.crt
 # cakey: /etc/pki/pulp/ca.key
-# ssl_ca_certificate: /etc/pki/pulp/ssl_ca.crt
+# ssl_ca_certificate: /etc/pki/pulp/ssl_ca.crt #Deprecated! See above description for details.
```
asmacdo: Obviously very minor, but I think that since the majority of our code is python, we should use PEP-8 compliant comments where it makes sense. So it would be my preference, though not required, that we use 2 spaces before the second `#` and one following before the comment.

server/etc/pulp/server.conf
```diff
@@ -101,6 +101,9 @@
 # ssl_ca_certificate: full path to the CA certificate used to sign the Pulp
 #     server's SSL certificate; consumers will use this to verify the
 #     Pulp server's SSL certificate during the SSL handshake
+#     Deprecated! - Please note that this setting will be removed in the next
+#     release and will be replaced by a setting in the consumer.conf file
```
asmacdo: Might be nice to specify the next X release since this would not be removed in a hypothetical 2.7.

asmacdo:
LGTM

PR #1493, opened 2015-01-07
1171278 - additional helper method for finding units

beav:
The fix for 1171278 in pulp_rpm needs to be able to find a unit based on the
unit key, regardless of which repo its in. This patch adds
`find_unit_by_unit_key` to the SearchUnitsMixin to facilitiate this.

### DIFF
--- a/server/pulp/plugins/conduits/mixins.py
+++ b/server/pulp/plugins/conduits/mixins.py
@@ -221,6 +221,29 @@ class SearchUnitsMixin(object):
             logger.exception('Exception from server requesting all units of type [%s]' % type_id)
             raise self.exception_class(e), None, sys.exc_info()[2]
 
+    def find_unit_by_unit_key(self, type_id, unit_key):
+        """
+        Finds a unit based on its unit key. If more than one unit comes back,
+        an exception will be raised.
+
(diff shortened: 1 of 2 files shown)


server/pulp/plugins/conduits/mixins.py
```diff
@@ -221,6 +221,33 @@ def search_all_units(self, type_id, criteria):
...
+            existing_unit = content_query_manager.get_content_unit_by_keys_dict(type_id, unit_key)
+            type_def = types_db.type_definition(type_id)
+            plugin_unit = common_utils.to_plugin_unit(existing_unit, type_def)
+            return plugin_unit
+        except MissingResource:
+            return None
+        except Exception, e:
+            logger.exception('Exception from server requesting unit of type [%s] with key [%s]'
+                             % (type_id, unit_key))
+            raise self.exception_class(e), None, sys.exc_info()[2]
```
barnabycourt: Why mask the original exception?  I could perhaps see wrapping it inside a coded exception so you had both available but masking seems unwise

beav: I don't remember why I did this originally to be honest, it was before break :smile:

I will update.

barnabycourt:
LGTM

PR #1494, opened 2015-01-08
PEP-8 pulp.server.agent.

rbarlow:


server/pulp/server/managers/consumer/agent.py
```diff
@@ -10,7 +10,8 @@
 from pulp.plugins.loader import api as plugin_api, exceptions as plugin_exceptions
 from pulp.plugins.model import Consumer as ProfiledConsumer
 from pulp.plugins.profiler import Profiler, InvalidUnitsRequested
-from pulp.server.agent import Context, PulpAgent
```
asmacdo: I agree that this kind of change is ok, but I don't think it should be in a PEP-8 pull request. I think we want to minimize the possibility that these changes have unwanted side effects.

server/pulp/server/managers/consumer/content.py
```diff
@@ -41,7 +24,7 @@ def install(self, id, units, options={}):
         @type options: dict
```
asmacdo: Do you want to switch over to `:param:` during style changes?

asmacdo:
LGTM

server/pulp/server/managers/consumer/agent.py
rbarlow: On 01/08/2015 04:55 PM, Austin wrote:
>> -from pulp.server.agent import Context, PulpAgent
> 
> I agree that this kind of change is ok, but I don't think it should be
> in a PEP-8 pull request. I think we want to minimize the possibility
> that these changes have unwanted side effects.

In this case, I grepped to find out who was importing this stuff, and it
was only used inside the platform.

Also, flake8 will get mad about those imports in __init__.py, so we
either have to remove them or add #noqa to them. Since they really don't
belong there anyway, I went for the former option.

server/pulp/server/managers/consumer/content.py
rbarlow: On 01/08/2015 04:55 PM, Austin wrote:
> Do you want to switch over to |:param:| during style changes?

No. Once I finish PEP-8'ing our code, PEP-257 is next!

PR #1495, opened 2015-01-08
1168331 - Give index pages useful sidebars.

rbarlow:
https://bugzilla.redhat.com/show_bug.cgi?id=1168331

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -132,7 +132,10 @@ html_static_path = ['_static']
 #html_use_smartypants = True
 
 # Custom sidebar templates, maps document names to template names.
-#html_sidebars = {}
+html_sidebars = {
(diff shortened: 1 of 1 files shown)


asmacdo:
@rbarlow, unfortunately, I think that this change only affects the default sphinx theme, not the sphinx_rtd_theme that read the docs uses.

rbarlow:
On 01/08/2015 04:52 PM, Austin wrote:
> @rbarlow <https://github.com/rbarlow>, unfortunately, I think that this
> change only affects the default sphinx theme, not the sphinx_rtd_theme
> that read the docs uses.

Perhaps it is possible for us to get that RTD theme in our conf.py so
that WYSIWYG.

asmacdo:
@rbarlow, I actually made that change, but it is only in master ATM.

rbarlow:
@asmacdo I installed the theme (via yum in F20!) and it looks like you are correct ☹.

PR #1497, opened 2015-01-09
1172745 - celerybeat service incorrectly reports broker on startup.

ipanova:
https://bugzilla.redhat.com/show_bug.cgi?id=1172745
Issue consisted in the fact that celerybeat was using the unconfigured
celery app. That's why the broker string was incorrect. By adding --app
option it will use the configured app and the startup info will be correct.


bmbouter:
LGTM

PR #1500, opened 2015-01-09
add read the docs theme to sphinx

asmacdo:
Backported from https://github.com/pulp/pulp/pull/1362.

This change uses the sphinx_rtd_theme if it is installed, otherwise it uses the default. This makes it simpler to make docs changes in 2.6.

The theme is available with pip:

    pip install sphinx_rtd_theme

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,7 +11,13 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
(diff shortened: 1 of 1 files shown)


rbarlow:
FYI, you can also yum install python-sphinx_rtd_theme on F2{0,1}.

rbarlow:
LGTM.

PR #1501, opened 2015-01-09

dkliban:


beav:
Is 2.4-dev the correct branch for this? I'm guessing it should be either 2.6-testing, 2.6-dev or master.

PR #1502, opened 2015-01-09

dkliban:


server/test/unit/server/webservices/test_application.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pulp.server.webservices.application import wsgi_application
+from pulp.server.initialization import InitializationException
+
+class TestApplication(unittest.TestCase):
+
+    @patch('pulp.server.webservices.application._initialize_pulp')
+    def test_wsgi_application_exception(self, mock_initialize_pulp):
+        mock_initialize_pulp.side_effect = Exception
+        with self.assertRaises(Exception):
```
bmbouter: I suggest using `self.assertRaises(Exception, wsgi_application)` instead of the 'with' operator.

server/test/unit/server/webservices/test_application.py
```diff
@@ -0,0 +1,20 @@
...
+            wsgi_application()
+
+    @patch('pulp.server.webservices.application._initialize_pulp')
+    def test_wsgi_application_initilization_exception(self, mock_initialize_pulp):
+        mock_initialize_pulp.side_effect = InitializationException('blah')
+        with self.assertRaises(InitializationException):
```
bmbouter: Same as L#13

bmbouter:
Not all the tests pass in jenkins and also it needs to be squashed

dkliban:
Closing this one and opening another one.

PR #1511, opened 2015-01-12
add ropeproject to gitignore

asmacdo:
Rope is a tool used by vim/emacs for refactoring and ensuring imports. I use this, and would like it in the gitignore.


barnabycourt:
LGTM

PR #1517, opened 2015-01-13
Fix agent plugin packaging.

jortel:
Fixes a variety of pulp 2.6 / gofer 2.4 integration issus:

- agent plugin not being installed in site-packages.
- pulp needs to explicitly declare the pulp.task queue (not longer implicitly created by gofer consumer)
- need to use the get_url() *workaround* when starting the reply consumer.  get_url() goes away in 3.0.

### DIFF
--- a/server/pulp/server/agent/direct/services.py
+++ b/server/pulp/server/agent/direct/services.py
@@ -34,7 +34,7 @@ class Services(object):
 
     @staticmethod
     def start():
-        url = config.get('messaging', 'url')
+        url = Services.get_url()
         Services.reply_handler = ReplyHandler(url)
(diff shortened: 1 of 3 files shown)


server/pulp/server/agent/direct/services.py
```diff
@@ -34,7 +34,7 @@ def init():
 
     @staticmethod
     def start():
-        url = config.get('messaging', 'url')
+        url = Services.get_url()
```
jortel: Using get_url() is a temporary work around until pulp 3.0 when we can change the pulp configuration files.

barnabycourt:
LGTM

PR #1518, opened 2015-01-13
1078833 - Updating and deleting importers now returns 404 when appropriate.

jeremycline:
https://bugzilla.redhat.com/show_bug.cgi?id=1078833

Furthermore, deleting an importer now requires the
DELETE permission instead of the UPDATE permission.

### DIFF
--- a/server/pulp/server/webservices/controllers/repositories.py
+++ b/server/pulp/server/webservices/controllers/repositories.py
@@ -6,7 +6,8 @@ import sys
 
 import web
 
-from pulp.common import constants, dateutils, tags
+from pulp.common import constants, dateutils, error_codes, tags
+from pulp.plugins.loader import api as plugin_api
 from pulp.server.auth.authorization import CREATE, DELETE, EXECUTE, READ, UPDATE
 from pulp.server.db.model.criteria import Criteria, UnitAssociationCriteria
 from pulp.server.db.model.repository import Repo, RepoContentUnit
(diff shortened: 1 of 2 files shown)


server/pulp/server/webservices/controllers/repositories.py
```diff
@@ -362,8 +363,21 @@ def GET(self, repo_id, importer_id):
...
+        :param repo_id:     The id of the repository to remove the importer from
+        :type  repo_id:     str
+        :param importer_id: The id of the importer to remove from the given repository
+        :type  importer_id: str
+        """
+        # Raise a MissingResource exception if the repo or the importer doesn't exist
+        importer_manager = manager_factory.repo_importer_manager()
+        importer = importer_manager.get_importer(repo_id)
+        if importer['id'] != importer_id:
+            raise exceptions.MissingResource(importer_id=importer_id)
```
beav: This may need to be a coded exception but I'm not 100% sure

server/pulp/server/webservices/controllers/repositories.py
```diff
@@ -374,8 +388,15 @@ def DELETE(self, repo_id, importer_id):
 
     @auth_required(UPDATE)
     def PUT(self, repo_id, importer_id):
+        # Raise a MissingResource exception if the repo or the importer doesn't exist
+        importer_manager = manager_factory.repo_importer_manager()
+        importer = importer_manager.get_importer(repo_id)
+        if importer['id'] != importer_id:
+            raise exceptions.MissingResource(importer_id=importer_id)
```
beav: same here re: coded exceptions

beav:
LGTM aside from coded exception question.

Please be sure to set the BZ to the proper target release as well :koala:

server/pulp/server/webservices/controllers/repositories.py
jeremycline: @beav Although this isn't a subclass of PulpCodedException, @barnabycourt did add the 'error' field with an error code, PLP0009, so as a user will see something like 
`{"http_request_method": "GET", ...., "error": {"code": "PLP0009", "data": {"resources": {"resource_id": "notarepo"}}}`

beav: ah, ok cool

PR #1522, opened 2015-01-13
Remove missing dirs from PEP-8 checks.

rbarlow:


skarmark:
Looks good. Merge away.

PR #1524, opened 2015-01-14
create base for plugins view module

asmacdo:
Creates a skeleton for all of the plugin urls to be converted to django. Each url branch will branch from this to prevent merge conflicts.


rbarlow:
LGTM.

PR #1525, opened 2015-01-14
1182322 - handle case where PLAIN is used with saslwrapper

beav:
There is an error string we did not handle correctly when PLAIN sasl mech was
used, python-saslwrapper was installed, and cyrus-sasl-plain was *not*
installed.

This patch contains a patch to Pulp's python-kombu to handle this scenario.


beav:
I will tito tag after this is merged.

Also, I will put in an upstream PR for kombu after this bz is set to verified

rbarlow:
LGTM.

PR #1527, opened 2015-01-15

dkliban:
https://bugzilla.redhat.com/show_bug.cgi?id=1177018

### DIFF
--- a/server/test/unit/server/webservices/test_application.py
+++ b/server/test/unit/server/webservices/test_application.py
@@ -1,0 +1,19 @@
+import unittest
+
+from mock import patch
+
+from pulp.server.webservices.application import wsgi_application
+from pulp.server.initialization import InitializationException
+
+
+class TestApplication(unittest.TestCase):
+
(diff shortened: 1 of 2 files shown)


server/test/unit/server/webservices/test_application.py
```diff
@@ -0,0 +1,19 @@
...
+
+from pulp.server.webservices.application import wsgi_application
+from pulp.server.initialization import InitializationException
+
+
+class TestApplication(unittest.TestCase):
+
+    @patch('pulp.server.webservices.application._initialize_pulp')
+    def test_wsgi_application_exception(self, mock_initialize_pulp):
+        mock_initialize_pulp.side_effect = Exception
```
bmbouter: L#13 is a class type and L#18 is a specific instantiated exception. Newer versions of mock use one or the other, but I think older ones expected a instance not a class. It's probably good to have side_effect be set to an instance. Not a required change, but more of a style thing.

bmbouter:
Please run this on Jenkins before it is merged. If Jenkins passes on all platforms then LGTM.

PR #1528, opened 2015-01-15
1182335 - Fixes username and password auth for mongoDB connection

bmbouter:
https://bugzilla.redhat.com/show_bug.cgi?id=1182335

The way we did username and password auth would have worked if all of our code uses mongoengine, but because some of the code uses PyMongo directly the auth wasn't working. It really should have worked with how it was written IMHO, but it doesn't because of a strange mongoengine behavior. I've filed an [upstream bug with mongoengine](https://github.com/MongoEngine/mongoengine/issues/851) on that.

I've hand tested this also to verify that Pulp works correctly with username and password auth for MongoDB.

I also ensured that all tests pass on all platforms using [Jenkins job 89](https://pulp-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/unittest-pulp/89/). There was one test failure on EL5, but that is unrelated to this PR.

### DIFF
--- a/server/pulp/server/db/connection.py
+++ b/server/pulp/server/db/connection.py
@@ -8,7 +8,7 @@ from gettext import gettext as _
 
 import mongoengine
 from pymongo.collection import Collection
-from pymongo.errors import AutoReconnect
+from pymongo.errors import AutoReconnect, OperationFailure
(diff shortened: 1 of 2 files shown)


dkliban:
LGTM

PR #1529, opened 2015-01-15
1145723 - touch and chown log file before writing to it

beav:
The previous fix for RHBZ #1145723 did not work properly for freshly installed
systems with more than one Pulp worker. In these cases, only the last log file
would get its permissions set properly because the ``chown`` command was done
outside of the loop over each worker.

This patch moves the ``chown`` inside the loop. I also moved it above the log
file writing operation in a few other places to make the ``chown`` command more
obvious.


rbarlow:
LGTM.

PR #1530, opened 2015-01-15
Make auth required easy to mock

bmbouter:
This PR will allow auth_required statement able to easily have assertions made about them.

Docs on how to use this can be found here:  https://fedorahosted.org/pulp/wiki/ReplacewebpywithDjango#UnitTesting

### DIFF
--- a/server/pulp/server/auth/authorization.py
+++ b/server/pulp/server/auth/authorization.py
@@ -1,19 +1,9 @@
 # -*- coding: utf-8 -*-
 
-# Copyright © 2010-2011 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
-# License as published by the Free Software Foundation; either version
-# 2 of the License (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
-
 """
 Utility functions to manage permissions and roles in pulp.
 """
+
 import logging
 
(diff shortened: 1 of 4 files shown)


rbarlow:
Hey that's nice!

server/pulp/server/auth/authorization.py
```diff
@@ -1,19 +1,5 @@
...
-# 2 of the License (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
-
-"""
-Utility functions to manage permissions and roles in pulp.
-"""
```
rbarlow: PEP-257 requires modules to have docblocks too, and once I finish PEP-8, PEP-257 is next. Plz to add back ☺

bmbouter:
I think I need to better handle the case when @auth_required(). Fixing it now.

server/pulp/server/auth/authorization.py
bmbouter: I did not know that. I will bring it back with the next commit.

bmbouter:
I documented how to use this in the porting to Django guide.

https://fedorahosted.org/pulp/wiki/ReplacewebpywithDjango#UnitTesting

bmbouter:
I will squash this before I merge it.

asmacdo:
Very clever @bmbouter! Also simple to use and will require minimal changes. 

I have only done minimal hand testing, but we will be putting it through the ringer as we merge the PR backlog, so I think we should go ahead an merge it.  LGTM. 

I think we should also add unit tests for `_verify_auth`, but I am ok with adding this later so we can begin the merges.

bmbouter:
Squashed and Merging.

PR #1531, opened 2015-01-15
Add an API for storing custom Unit metadata.

rbarlow:
https://pulp.plan.io/issues/86

closes #86

### DIFF
--- a/docs/dev-guide/integration/rest-api/content/index.rst
+++ b/docs/dev-guide/integration/rest-api/content/index.rst
@@ -7,6 +7,6 @@ Content Manipulation APIs
    upload
    associate
    orphan
-   retrieval
+   units
    source
    catalog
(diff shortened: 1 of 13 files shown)


rbarlow:
This very accurately describes what it was like to work on this story:

Working on Pulp

mhrivnak:
I assume that when units are retrieved from requests such to URIs such as ``/pulp/api/v2/repositories/<repo_id>/search/units/``, the response will include this new field. It would be good to update those examples in the REST API docs to at least include the default value.

mhrivnak:
LGTM

rbarlow:
I've fixed the other documentation examples that involve units, thanks for noting that! I also filed a story to add some automation to test our documentation examples out so that it will be easier to find problems like this:

https://pulp.plan.io/issues/113

PR #1535, opened 2015-01-16
PEP-8 pulp.plugins.util.

rbarlow:


barnabycourt:
ok test

asmacdo:
Assigning @rbarlow until conficts are resolved.  (So we don't keep checking this one for review.)

rbarlow:
@asmacdo I don't think merge conflicts should be a reason to avoid review. The PEP-8 PRs will all conflict immediately when others are merged since they all edit the list of flake8 paths in run-tests.py. This means that every time I merge one, all the others will become conflicted. Thus, it's not scalable to require PRs to have the "OK to merge" button. I will resolve conflicts when I merge it by hand.

asmacdo:
@rbarlow, fair point. LGTM

PR #1536, opened 2015-01-16
update external deps file for docker and crane

beav:
In the past we had to tag pulp-docker and python-crane in manually to the
correct koji tag. This is a step that was frequently forgotten.

Instead, use external-deps.json which automatically pulls the correct versions
into the specificed tag.


rbarlow:
LGTM

PR #1537, opened 2015-01-16
django conversion /plugins/types/

asmacdo:
I will rebase when ready to merge.

### DIFF
--- a/server/pulp/server/webservices/views/plugins.py
+++ b/server/pulp/server/webservices/views/plugins.py
@@ -1,11 +1,10 @@
-import json
-
-from django.http import HttpResponse
 from django.views.generic import View
 
 from pulp.server.auth import authorization
 from pulp.server.managers import factory
 from pulp.server.webservices.controllers.decorators import auth_required
+from pulp.server.webservices.views.util import (generate_json_response,
+                                                generate_json_response_with_pulp_encoder)
 
 
 class DistributorResourceView(View):
@@ -29,4 +28,26 @@ class TypeResourceView(View):
 
 
 class TypesView(View):
-    pass
+    """
+    View for dealing with all plugin types.
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request):
+        """
+        Get all type definitions
+
+        :param request: WSGI Request obect
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :return       : Serialized list of objects representing all available content types
+        :rtype        : HttpResponse
+        """
+        manager = factory.plugin_manager()
+        type_defs = manager.types()
+
+        for type_definition in type_defs:
+            href = {'_href': '/'.join([request.get_full_path().rstrip('/'),
+                                       type_definition['id'], ''])}
+            type_definition.update(href)
+
+        return generate_json_response_with_pulp_encoder(type_defs)
--- a/server/test/unit/server/webservices/test_urls.py
+++ b/server/test/unit/server/webservices/test_urls.py
@@ -46,6 +46,9 @@ def assert_url_match(expected_url, url_name, *args, **kwargs):
 
 
 class TestDjangoPluginsUrls(unittest.TestCase):
+    """
+    Test url matching for plugins urls.
+    """
 
     def test_match_distributor_resource_view(self):
         pass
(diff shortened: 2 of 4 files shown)


server/pulp/server/webservices/views/plugins.py
```diff
@@ -29,4 +30,27 @@ class TypeResourceView(View):
 
 
 class TypesView(View):
-    pass
+    """
+    View for dealing with all plugin types.
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request, *args, **kwargs):
```
bmbouter: I don't think this should have *args and **kwargs. If this view does receive a positional or keyword argument in the future this signature can be adjusted.

server/pulp/server/webservices/views/plugins.py
```diff
@@ -29,4 +30,27 @@ class TypeResourceView(View):
...
+        """
+        Get all type definitions
+
+        :param request: WSGI Request obect
+        :type  request: WSGIRequest
+        :return       : Serialized list of obects representing all available content types
```
bmbouter: Style thing, but I generally don't align the colons vertically because then if one variable name changes more lines than necessary need to be modified. This will be less independent from a maintenance perspective. Not a required change.

server/test/unit/server/webservices/views/test_plugins.py
```diff
@@ -30,4 +31,33 @@ class TestTypeResourceView(unittest.TestCase):
...
+        mock_plugin_manager = mock.MagicMock()
+        mock_plugin_manager.types.return_value = [{'id': 'mock_id'}]
+        mock_factory.plugin_manager.return_value = mock_plugin_manager
+        request = mock.MagicMock()
+        request.get_full_path.return_value = '/mock/path'
+
+        types_view = TypesView()
+        response = types_view.get(request)
+
+        self.assertTrue(isinstance(response, HttpResponse))
```
bmbouter: If this gets refactored to use the helper method then these sorts of tests can be put on the helper function instead of this specific view. At that point this specific view could mock the helper function and assert that this view uses the helper function.

server/test/unit/server/webservices/test_urls.py
```diff
@@ -4,6 +4,9 @@
 
 
 class TestDjangoPluginsUrls(unittest.TestCase):
+    """
+    Test url matching for /plugins/* urls.
```
bmbouter: I probably wouldn't write the URL here specifically in case it changes in the future.

server/test/unit/server/webservices/test_urls.py
```diff
@@ -4,6 +4,9 @@
 
 
 class TestDjangoPluginsUrls(unittest.TestCase):
```
bmbouter: Consider using a single big class for all URLs tests. 1 test per url would be 102 methods on the class. I think having all of those tests in one place is better than spreading them out everywhere. @asmacdo @mhrivnak What do you think?

server/test/unit/server/webservices/test_urls.py
```diff
@@ -21,4 +24,26 @@ def test_match_type_resource_view(self):
...
     def test_match_types_view(self):
-        pass
+        """
+        Test url matching for /plugins/types/.
+        """
+        match = resolve('/v2/plugins/types/')
+
+        self.assertEqual(match.url_name, 'plugin_types')
+
+        should_not_match = [
```
bmbouter: Based on yesterday's convo I don't think we should do negative matching but we should ensure:

expected view name + specific args --> produces correct URL
AND
that correct url when resolved will produce the correct view name.

server/test/unit/server/webservices/test_urls.py
asmacdo: @bmbouter, the advantage of breaking it up is that the tests are a little more organized and the individual tests do not need to be namespaced as carefully. I did them this way since so many PRs will go in side by side to prevent accidental clash between them. Also, since these are the only tests in this file, they will be contained together.

That being said, those reasons are minor and it does make sense to test them together since the urls are all contained in a single class.

asmacdo:
@bmbouter, ready for review

bmbouter:
Hand tested using `./web_compare.sh GET /v2/plugins/types/ True`. Both headers and output were identical.

bmbouter:
The coverage report looks good for the code introduced.

bmbouter:
@asmacdo LGTM

PR #1540, opened 2015-01-16
django conversion /plugins/importers/

asmacdo:


asmacdo:
Hand testing:

`./web_compare.sh GET /v2/plugins/importers/ True` http://fpaste.org/173661/22032329/

dkliban:
LGTM

PR #1541, opened 2015-01-16
django conversion /plugins/distributors/

asmacdo:


asmacdo:
Hand testing
`/web_compare.sh GET /v2/plugins/distributors/ True` http://fpaste.org/173679/34590142/

dkliban:
LGTM

PR #1543, opened 2015-01-16
django conversion /content/uploads/<upload_id>/<offset>/

asmacdo:


server/pulp/server/webservices/urls.py
```diff
@@ -4,16 +4,20 @@
...
                                                    TypeResourceView, TypesView)
 from pulp.server.webservices.views.tasks import TasksView
+from pulp.server.webservices.views.content import UploadSegmentResourceView
 from pulp.server.webservices.views.repositories import RepoSync
 
 
 urlpatterns = patterns('',
+    url(r'^v2/content/uploads/(?P<upload_id>[^/]+)/(?P<offset>[^/]+)/$',
+        UploadSegmentResourceView.as_view(),
+        name='upload_segment_resource'),
```
asmacdo: change to content_upload_segment_resource

server/test/unit/server/webservices/test_urls.py
```diff
@@ -3,6 +3,45 @@
 from django.core.urlresolvers import resolve, Resolver404
 
 
+class TestDjangoContentUrls(unittest.TestCase):
+
+    def test_match_upload_segment_resource_view(self):
```
asmacdo: add doc block

asmacdo:
PUT with valid id and valid offset
`/v2/content/uploads/e5b2973f-b5d3-4c36-9da4-8461c1405d5a/0/` http://fpaste.org/175046/22281786/
PUT with valid id and invalid offset
`/v2/content/uploads/e5b2973f-b5d3-4c36-9da4-8461c1405d5a/invalid/` http://fpaste.org/175050/14222820/
PUT with invalid id and invalid offset
`v2/content/uploads/invalid/invalid/` http://fpaste.org/175052/22822211/
PUT with invalid id and valid offset
`/v2/content/uploads/invalid/0/` http://fpaste.org/175053/22823301/

server/test/unit/server/webservices/views/test_content.py
```diff
@@ -0,0 +1,54 @@
...
+        mock_upload_manager = mock.MagicMock()
+        mock_factory.content_upload_manager.return_value = mock_upload_manager
+        request = mock.MagicMock()
+
+        upload_segment_resource = UploadSegmentResourceView()
+
+        try:
+            upload_segment_resource.put(request, 'mock_id', 'invalid_offset')
+        except InvalidValue:
+            pass
```
dkliban: I would use:

self.assertRaises(InvalidValue(['offset']), upload_segment_resource.put, request, 'mock_id', 'invalid_offset')

Also, should probably add a test to check that you get a MissingResource exception if the resource id is bad.

asmacdo: I agree about assertRaises.

I don't think I can really test that a MissingResource exception is raised since that is actually a side effect of the manager; this code doesn't interact with it at all. I think I should just remove the `raises MissingResource` statement from the comments. What do you think @dkliban?

dkliban:
LGTM

PR #1549, opened 2015-01-19
Introducing Consumer group view

ipanova:
~~1 tests should be added( test_urls.py + test_consumer_group.py)~~
~~2 docs for every class and method should be added~~
~~3 add manual testing fpaste~~

### DIFF
--- a/server/pulp/server/webservices/urls.py
+++ b/server/pulp/server/webservices/urls.py
@@ -1,5 +1,12 @@
 from django.conf.urls import patterns, url
 
+from pulp.server.webservices.views.consumer_groups import (ConsumerGroupView,
+                                                           ConsumerGroupAssociateActionView,
+                                                           ConsumerGroupBindingView,
+                                                           ConsumerGroupBindingsView,
+                                                           ConsumerGroupContentActionView,
+                                                           ConsumerGroupResourceView,
+                                                           ConsumerGroupUnassociateActionView)
 from pulp.server.webservices.views.content import UploadSegmentResourceView
 from pulp.server.webservices.views.plugins import (DistributorResourceView, DistributorsView,
                                                    ImporterResourceView, ImportersView,
@@ -15,6 +22,20 @@ from pulp.server.webservices.views.tasks import TasksView
 
 urlpatterns = patterns('',
     url(r'^v2/actions/login/$', LoginView.as_view(), name='login'),
+    url(r'^v2/consumer_groups/$', ConsumerGroupView.as_view(), name='consumer_group'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/$',
+        ConsumerGroupResourceView.as_view(), name='consumer_group_resource'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/associate/$',
+        ConsumerGroupAssociateActionView.as_view(), name='consumer_group_associate'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/unassociate/$',
+        ConsumerGroupUnassociateActionView.as_view(), name='consumer_group_unassociate'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/content/(?P<action>[^/]+)/$',
+        ConsumerGroupContentActionView.as_view(), name='consumer_group_content'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/bindings/$',
+        ConsumerGroupBindingsView.as_view(), name='consumer_group_bind'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)' +
+        r'/bindings/(?P<repo_id>[^/]+)/(?P<distributor_id>[^/]+)/$',
+        ConsumerGroupBindingView.as_view(), name='consumer_group_unbind'),
     url(r'^v2/content/uploads/(?P<upload_id>[^/]+)/(?P<offset>[^/]+)/$',
         UploadSegmentResourceView.as_view(), name='content_upload_segment_resource'),
     url(r'^v2/plugins/distributors/$', DistributorsView.as_view(), name='plugin_distributors'),
--- a/server/pulp/server/webservices/views/consumer_groups.py
+++ b/server/pulp/server/webservices/views/consumer_groups.py
@@ -1,0 +1,381 @@
+import json
+
+from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect, HttpResponseBadRequest
+from django.views.generic import View
+
+from pulp.server import exceptions as pulp_exceptions
+from pulp.server.auth import authorization
+from pulp.server.db.model.consumer import ConsumerGroup
+from pulp.server.db.model.criteria import Criteria
+from pulp.server.managers import factory
+from pulp.server.managers.consumer.group.cud import bind, unbind
+from pulp.server.webservices.controllers.base import json_encoder as pulp_json_encoder
+from pulp.server.webservices.controllers.decorators import auth_required
+from pulp.server.webservices.views.util import (generate_json_response,
+                                                generate_json_response_with_pulp_encoder)
+
+
+class ConsumerGroupView(View):
+    """
+    Views for consumer groups.
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request):
+        """
+        List the available consumer groups.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :return: Response containing a list of consumer groups
+        :rtype: django.http.HttpResponse
+        """
+        collection = ConsumerGroup.get_collection()
+        cursor = collection.find({})
+        groups = []
+        for group in cursor:
+            link = {"_href": reverse('consumer_group_resource',
+                    kwargs={'consumer_group_id': group['id']})}
+            group.update(link)
+            groups.append(group)
+        return generate_json_response_with_pulp_encoder(groups)
+
+    @auth_required(authorization.CREATE)
+    def post(self, request):
+        """
+        Create a consumer group and return a serialized object containing just created group.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :return: Response containing the consumer group
+        :rtype: HttpResponseRedirect
+        :raises: MissingValue if group ID is not provided
+        :raises: InvalidValue if some parameters are invalid
+        """
+        params = request.body_as_json
+        group_id = params.pop('id', None)
+        if group_id is None:
+            raise pulp_exceptions.MissingValue(['id'])
+        display_name = params.pop('display_name', None)
+        description = params.pop('description', None)
+        consumer_ids = params.pop('consumer_ids', None)
+        notes = params.pop('notes', None)
+        if params:
+            raise pulp_exceptions.InvalidValue(params.keys())
+        manager = factory.consumer_group_manager()
+        group = manager.create_consumer_group(group_id, display_name, description, consumer_ids,
+                                              notes)
+        link = {"_href": reverse('consumer_group_resource',
+                kwargs={'consumer_group_id': group['id']})}
+        group.update(link)
+        return HttpResponseRedirect(
+            link["_href"],
+            json.dumps(group, default=pulp_json_encoder),
+            content_type="application/json",
+            status=201,
+            )
+
+
+class ConsumerGroupResourceView(View):
+    """
+    Views for a specific consumer group.
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request, consumer_group_id):
+        """
+        Return a serialized object representing the requested group.
+
+        :param request: WSGI request object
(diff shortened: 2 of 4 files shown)


ipanova:
`GET /v2/consumer_groups/` http://fpaste.org/174901/22272356/

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,372 @@
...
+    @auth_required(authorization.DELETE)
+    def delete(self, request, consumer_group_id):
+        """
+        Delete a specified consumer group.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :param consumer_group_id: id for the requested group
+        :type consumer_group_id: string
+        :return: Response containing null
```
bmbouter: s/Response containing null/An empty response.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,372 @@
...
+    @auth_required(authorization.CREATE)
+    def post(self, request, consumer_group_id, action):
+        """
+        Content actions.
+        """
+        method = getattr(self, action, None)
+        if method:
+            return method(request, consumer_group_id)
+        else:
+            raise BadRequest()
```
bmbouter: What is BadRequest? It is not defined in this module.

server/test/unit/server/webservices/test_urls.py
```diff
@@ -107,3 +107,80 @@ def test_match_tasks_url(self):
...
+        url_name = 'consumer_group_bind'
+        assert_url_match(url, url_name, consumer_group_id='test-group')
+
+    def test_match_consumer_group_binding_view(self):
+        """
+        Test url matching for consumer_groups binding removal
+        """
+        url = '/v2/consumer_groups/test-group/bindings/repo1/dist1/'
+        url_name = 'consumer_group_unbind'
+        assert_url_match(url, url_name, consumer_group_id='test-group', repo_id='repo1', distributor_id='dist1')
```
bmbouter: flake8 says this line is too long

bmbouter:
There are a lot of PEP-8 errors in this file `server/test/unit/server/webservices/views/test_consumer_groups.py`

bmbouter:
There are also some in URLs.py

bmbouter:
Before merging it would be good to rebase this against the latest feature branch and then squash these commits

server/test/unit/server/webservices/views/test_consumer_groups.py
```diff
@@ -0,0 +1,302 @@
+import os
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pulp.server.webservices.settings'
```
bmbouter: These statements should not be here. If you rebase against the latest-feature branch then your ./run-tests.py will set this for you

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,375 @@
+import json
+
+from web.webapi import BadRequest
```
bmbouter: We are eventually going to delete web.py so we can't use BadRequest as a type. Anything that imports starting with web.* is contained within web.py

server/pulp/server/webservices/urls.py
```diff
@@ -1,6 +1,13 @@
 from django.conf.urls import patterns, url
 
 from pulp.server.webservices.views.content import UploadSegmentResourceView
+from pulp.server.webservices.views.consumer_groups import (ConsumerGroupView,
```
bmbouter: This should come before pulp.server.webservices.views.content for alphabetical imports

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+from django.views.generic import View
+
+from pulp.server import exceptions as pulp_exceptions
+from pulp.server.auth import authorization
+from pulp.server.db.model.consumer import ConsumerGroup
+from pulp.server.db.model.criteria import Criteria
+from pulp.server.managers import factory
+from pulp.server.managers.consumer.group.cud import bind, unbind
+from pulp.server.webservices.controllers.decorators import auth_required
+from pulp.server.webservices.controllers.base import json_encoder as pulp_json_encoder
```
bmbouter: This and line 13 should be swapped for alphabetical imports since base comes before decorators.

server/pulp/server/webservices/urls.py
```diff
@@ -15,6 +22,20 @@
...
+        ConsumerGroupResourceView.as_view(), name='consumer_group_resource'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/associate/$',
+        ConsumerGroupAssociateActionView.as_view(), name='consumer_group_associate'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/unassociate/$',
+        ConsumerGroupUnassociateActionView.as_view(), name='consumer_group_unassociate'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/actions/content/(?P<action>[^/]+)/$',
+        ConsumerGroupContentActionView.as_view(), name='consumer_group_content'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)/bindings/$',
+        ConsumerGroupBindingsView.as_view(), name='consumer_group_bind'),
+    url(r'^v2/consumer_groups/(?P<consumer_group_id>[^/]+)' + \
```
bmbouter: flake8 says: "E502 the backslash is redundant between brackets" Since its within parenthes it should not require the backslash.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+        notes = params.pop('notes', None)
+        if params:
+            raise pulp_exceptions.InvalidValue(params.keys())
+        manager = factory.consumer_group_manager()
+        group = manager.create_consumer_group(group_id, display_name, description, consumer_ids,
+                                              notes)
+        link = {"_href": reverse('consumer_group_resource',
+                kwargs={'consumer_group_id': group['id']})}
+        group.update(link)
+        # TODO location is shown with domain, should be without
```
bmbouter: This todo should be removed. I think we resolved that it was ok that Django builds the domain. I documented it [here some](https://fedorahosted.org/pulp/wiki/ReplacewebpywithDjango#Behavioralchangesthatarebeingleftdifferent).

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+
+    @auth_required(authorization.READ)
+    def get(self, request, consumer_group_id):
+        """
+        Return a serialized object representing the requested group.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :param consumer_group_id: id for the requested group
+        :type consumer_group_id: string
```
bmbouter: Do you mean string or basestring? If its string then unicode characters will not be allowed.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+
+class ConsumerGroupContentActionView(View):
+    """
+    Views for content manipulation on consumer group.
+    """
+
+    @auth_required(authorization.CREATE)
+    def post(self, request, consumer_group_id, action):
+        """
+        Content actions.
```
bmbouter: Parameter documentation and an improved docstring are needed for this function.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+        task = factory.consumer_group_manager().uninstall_content(consumer_group_id,
+                                                                  units, options)
+        raise pulp_exceptions.OperationPostponed(task)
+
+
+class ConsumerGroupBindingsView(View):
+    """
+    Views for repository binding to the group.
+    """
+    @auth_required(authorization.CREATE)
```
bmbouter: Maybe add a space above this line?

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+        Designed to be idempotent so only MissingResource is expected to
+        be raised by manager.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :param consumer_group_id: The consumer group ID to bind.
+        :type consumer_group_id: str
+        :raises: MissingResource if group id does not exist
+        :raises: InvalidValue some parameters are invalid
+        :raises: OperationPostponed
```
bmbouter: Can you say a little more about OperationPostponed? For intsance it is what is raised under normal operation. Making any similar fixes throughout the PR would be good too.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+                           binding_config, options)
+        async_task = bind.apply_async(bind_args_tuple)
+        raise pulp_exceptions.OperationPostponed(async_task)
+
+
+class ConsumerGroupBindingView(View):
+    """
+    Represents a specific consumer group binding.
+    """
+    @auth_required(authorization.DELETE)
```
bmbouter: Maybe add a space above here?

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+    def delete(self, request, consumer_group_id, repo_id, distributor_id):
+        """
+        Delete a bind association between the consumers belonging to the specified
+        consumer group and repo-distributor. Designed to be idempotent.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :param consumer_group_id: A consumer group ID.
+        :type consumer_group_id: str
```
bmbouter: I don't think str is an actual python type. It probably should be basestring. Similar fixed throughout the PR.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -0,0 +1,373 @@
...
+        :param repo_id: A repo ID.
+        :type repo_id: str
+        :param distributor_id: A distributor ID.
+        :type distributor_id: str
+        :raises: MissingResource if repo, distributor, group id is missing
+        :raises: OperationPostponed
```
bmbouter: More detail on the OperationPostponed similar to other areas.

server/test/unit/server/webservices/views/test_consumer_groups.py
```diff
@@ -0,0 +1,296 @@
...
+        request = mock.MagicMock()
+        consumer_group = ConsumerGroupResourceView()
+        response = consumer_group.get(request, 'foo')
+
+        self.assertTrue(isinstance(response, HttpResponse))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response._headers.get('content-type'),
+                         ('Content-Type', 'application/json'))
+        content = json.loads(response.content)
+        expected_cont = {'id': 'foo', '_href': '/v2/consumer_groups/foo/'}
```
bmbouter: I think this PR should introduce and use one or two helper methods similar to the helper methods @asmacdo  introduced: `generate_json_response_with_pulp_encoder()` and `generate_json_response()`

bmbouter:
@ipanova: I'm done reviewing this PR. If you have any questions about the comments let me know. LGTM

bmbouter:
The code in consumer_groups.py is missing tests. Currently it is at 83%, but it needs to be at 100%. You can generate the coverage report using `./run-tests.py --enable-coverage` and then read the report in the pulp checkout `cover/index.html`. In the report you'll see which lines in pulp.server.webservices.views.consumer_group are not covered. You should write new tests that use mock to cause those codepaths to be taken but you only need to make assertions on the returned output.

bmbouter:
I'm done reviewing this. It needs more tests, and there are some other suggestions and comments.

server/pulp/server/webservices/views/consumer_groups.py
ipanova: I tried to create the group via cli and api( webpy) with unicode characters and did not succeed. So i guess the right type would be string.

server/pulp/server/webservices/views/consumer_groups.py
rbarlow: On 01/27/2015 03:15 PM, bmbouter wrote:
> I don't think str is an actual python type.

str is a built-in Python type:

https://docs.python.org/2/library/functions.html#str

However, I usually list basestring as my accepted input types, if only
because most of the code I use would work the same on unicode and str types.

bmbouter:
Nice, you are at 100% coverage!

bmbouter:
I think since string isn't a formal type the docstrings should refer to 'str' instead. After that go ahead and merge!

rbarlow:
On 01/29/2015 11:21 AM, bmbouter wrote:
> I think since string isn't a formal type the docstrings should refer to
> 'str' instead.

basestring is a good type name to use. Both str and unicode are
instances of basestring, and for most of our functions/methods we are
really depending on the basestring API and not specifically str or
unicode. Of course, if you find yourself doing things where the
difference between unicode and str matter, you should be explicit.

PR #1551, opened 2015-01-19
create base for repo_groups django conversion

asmacdo:
Creates a skeleton for the django conversion of all repo_groups urls.


ipanova:
LGTM

PR #1552, opened 2015-01-20
PEP-8 pulp.plugins.loader.

rbarlow:


rbarlow:
@pulpbot is lying - it seems that he just doesn't have the right version of gofer. The tests passed on my checkout.

server/pulp/plugins/loader/manager.py
```diff
@@ -1,27 +1,13 @@
...
-# implied warranties of MERCHANTABILITY, NON-INFRINGEMENT, or FITNESS FOR A
-# PARTICULAR PURPOSE.
-# You should have received a copy of GPLv2 along with this software; if not,
-# see http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
-
-import copy
-import logging
 from gettext import gettext as _
 from pprint import pformat
+import copy
```
asmacdo: I would prefer this to be alphabetical.

https://github.com/pulp/pulp_docker/pull/52#discussion_r22557739

asmacdo:
LGTM

server/pulp/plugins/loader/manager.py
rbarlow: I also prefer them to be alphabetical, but in addition I prefer to group by the two styles of import (all from's together, all import's together, and from before import).

PR #1553, opened 2015-01-20
create utility for the generation of a django response

asmacdo:
@mhrivnak, assuming that this is what you had in mind, I would like to get this merged so we can refactor the existing django work to use this.


server/pulp/server/webservices/views/utils.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+
+def generate_django_response(response, content=None, default=None, content_type='application/json'):
```
mhrivnak: I would put "json" in the name, since serializing to json is the main point of having this. Maybe ``generate_json_response``.

mhrivnak: I'd call ``response`` instead ``response_class``, and give it a default value. Also I'd make ``content`` the first argument, to accomodate the most common use case of:

```
return generate_json_response(my_data)
```

server/pulp/server/webservices/views/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    :param content_type: type of returned content
+    :type  content_type: str
+
+    :raises            : TypeError if response is not a valid Django response object
+    :return            : response containing the serialized content
+    :rype              : HttpResponse
+    """
+    json_obj = json.dumps(content, default=default)
+    return response(json_obj, content_type=content_type)
+
```
mhrivnak: Given the suggestions above, you could make a "shortcut" that uses the optional serializer function.

```
generate_json_response_with_magic = functools.partial(generate_json_response, default=someimportedfunc)
```

You could also just make it a separate function if you prefer that calls the above function.

server/test/unit/server/webservices/views/test_utils.py
```diff
@@ -0,0 +1,46 @@
...
+class TestWebservicesUtils(unittest.TestCase):
+    """
+    Test webservices utilities.
+    """
+
+    def test_generate_django_response_default_params(self):
+        """
+        Make sure that the response is correct under normal conditions.
+        """
+        test_content = """{'foo': 'bar'}"""
```
mhrivnak: I think this might not be doing quite what you intended. Your sample content looks suspiciously like serialized json. That will work, but it's really just the same as being any old string. I think what you might actually want is a regular python object here, such as

``test_content = {'foo': 'bar'}``

which will get serialized into a json string during your test.

asmacdo:
@mhrivnak ready for rereview

server/pulp/server/webservices/views/util.py
```diff
@@ -0,0 +1,32 @@
...
+
+    :raises               : TypeError if response does not implement Django response object API
+    :return               : response containing the serialized content
+    :rtype                : HttpResponse or subclass
+    """
+    json_obj = json.dumps(content, default=default)
+    return response_class(json_obj, content_type=content_type)
+
+# Shortcut function to generate a json response using the in house json_encoder
+generate_json_response_with_pulp_encoder = functools.partial(
```
bmbouter: This could use a more pronounced comment. I really like how this is done!

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        response = util.generate_json_response(HttpResponseNotFound)
+        self.assertTrue(isinstance(response, HttpResponse))
+        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response._headers.get('content-type'),
+                         ('Content-Type', 'application/json'))
+
+    def test_generate_json_response_invalid_response_class(self):
+        """
+        generate_django_response should raise a TypeError for any responses that are not a
```
bmbouter: Only the first line is displayed when tests are run with -v. If a test has a comment then the first line only is shown. I usually have a short single line sentence, and then use detail like this as another paragraph below in the docblock.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -0,0 +1,57 @@
...
+                         ('Content-Type', 'application/json'))
+        response_content = json.loads(response.content)
+        self.assertEqual(response_content, test_content)
+
+    def test_generate_json_response_not_found(self):
+        """
+        Test that response is correct for non-base HttpResponses
+        """
+        response = util.generate_json_response(HttpResponseNotFound)
+        self.assertTrue(isinstance(response, HttpResponse))
```
bmbouter: I expected this would assert HttpResponseNotFound as the type since that is what we asked generate_json_response() to make in L#33.

bmbouter:
I think the assertion to HttpResponseNotFound should be fixed and the more pronounced comment on generate_json_response_with_pulp_encoder. Other than those LGTM.

server/pulp/server/webservices/views/util.py
```diff
@@ -0,0 +1,39 @@
...
+    :param content_type   : type of returned content
+    :type  content_type   : str
+
+    :raises               : TypeError if response does not implement Django response object API
+    :return               : response containing the serialized content
+    :rtype                : HttpResponse or subclass
+    """
+    json_obj = json.dumps(content, default=default)
+    return response_class(json_obj, content_type=content_type)
+
```
bmbouter: 2 spaces for PEP8

server/pulp/server/webservices/views/util.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+Shortcut function to generate a json response using the in house json_encoder.
+
+This function is equivalent to:
+`generate_json_response(content, default=pulp_json_encoder)`
```
bmbouter: backticks?

asmacdo: markdown habbit :) I will remove them.

PR #1558, opened 2015-01-21
PEP-8 pulp.plugins.types.

rbarlow:


asmacdo:
https://github.com/rbarlow/pulp/blob/pep8-plugins_types/server/pulp/plugins/types/parser.py#L188

Should be:

    error_descriptors = [d for d in descriptors if 'types' not in d.parsed]

run-tests.py
```diff
@@ -13,6 +13,8 @@
 
 # These paths should all pass PEP-8 checks
 paths_to_check = [
+    'server/pulp/plugins/file/',
```
asmacdo: This addition introduces two issues:

    server/pulp/plugins/file/distributor.py:104:17: E265 block comment should start with '# '
    server/pulp/plugins/file/distributor.py:226:13: E265 block comment should start with '# '

asmacdo:
After addressing those 2 things, LGTM.

run-tests.py
rbarlow: I'm surprised at this. On my dev box, I do not see these particular errors, but I did find that file to have these problems. I wonder why my own machine isn't finding these problems?

Anyways, I fixed it.

PR #1560, opened 2015-01-22
1183700 - Update pulp.spec to overwrite Celery related init scripts

bmbouter:
https://bugzilla.redhat.com/show_bug.cgi?id=1183700

This will cause pulp-server to overwrite the pulp_celerybeat, pulp_workers, and pulp_resource_manager Upstart scripts. I tested rebuild Pulp with this new spec file and the build worked.


beav:
LGTM :koala:

PR #1561, opened 2015-01-22
agent validate registration using rest api.

jortel:
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1159303

In the past, the agent determined consumer registration based on the existence of a valid consumer certificated.  When unregistered from the consumer, the certificate is deleted which is detected by the agent.  When a consumer is unregistered from the server and the agent is running, the server sends the agent notification of the un-registration and the agent deletes the certificate.  Because were never deleting the agent queue, the agent would always get this notification.  However, in the sprint, we plan to have the server delete agent queues after un-registration.  This opens the door for the case that the agent is not running on un-registration so it never gets the notification.  When it starts, the certificate will still exist so the agent will think it's still registered.  But it's not.

The fix here, is to determine registration using a REST call to the server.

Normally I'm very opposed to module global variables but decided to make an exception in this case.

### DIFF
--- a/agent/pulp/agent/gofer/pulpplugin.py
+++ b/agent/pulp/agent/gofer/pulpplugin.py
@@ -5,12 +5,15 @@ Contains recurring actions and remote classes.
 
 import os
 
+from time import sleep
+from threading import Thread
+from gettext import gettext as _
 from logging import getLogger
 
 from M2Crypto import RSA, BIO
 from M2Crypto.X509 import X509Error
 
-from gofer.decorators import *
+from gofer.decorators import initializer, remote, action
 from gofer.agent.plugin import Plugin
 from gofer.pmon import PathMonitor
 from gofer.agent.rmi import Context
@@ -22,6 +25,7 @@ from pulp.agent.lib.dispatcher import Dispatcher
 from pulp.agent.lib.conduit import Conduit as HandlerConduit
 from pulp.bindings.server import PulpConnection
 from pulp.bindings.bindings import Bindings
+from pulp.bindings.exceptions import NotFoundException
 from pulp.client.consumer.config import read_config
 
 
@@ -39,6 +43,9 @@ path_monitor = PathMonitor()
 # this plugin object
 plugin = Plugin.find(__name__)
 
+# registration status
+registered = False
+
 
 @initializer
 def init_plugin():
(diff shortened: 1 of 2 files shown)


beav:
Is this related to https://bugzilla.redhat.com/show_bug.cgi?id=1159303?

jortel:
Yes.  Related.

jortel:
Thinking about this again today and realized that I should add retry logic to handle cases where pulp/satellite rest api is temporarily unavailable.  Stay tuned.

jortel:
Done.  Review away ..

agent/pulp/agent/gofer/pulpplugin.py
```diff
@@ -118,6 +151,35 @@ def get_secret():
...
+    def run(self):
+        while True:
+            try:
+                validate_registration()
+                if registered:
+                    update_settings()
+                    plugin.attach()
+                else:
+                    plugin.detach()
+                break
```
beav: maybe add a comment here to make it more obvious that this breaks out of the "while True" loop. It took me some time to figure this out when reviewing :cat2:

agent/test/unit/test_agent.py
```diff
@@ -99,12 +105,111 @@ def load_plugin(mock_read):
...
     def setUp(self):
         self.plugin = PluginTest.load_plugin()
 
 
+class TestValidateRegistration(PluginTest):
+
+    @patch('pulp.agent.gofer.pulpplugin.PulpBindings')
+    @patch('pulp.agent.gofer.pulpplugin.ConsumerX509Bundle')
+    def test_registered(self, bundle, bindings):
+        uid = 'Z28'
```
beav: :car:

agent/pulp/agent/gofer/pulpplugin.py
```diff
@@ -118,6 +151,35 @@ def get_secret():
...
+class Attach(Thread):
+    """
+    This thread (task) persistently:
+      - validates the registration status
+      - if registered, updates the plugin settings and attach.
+      - if not registered, detach the plugin.
+    The reason for doing this in a thread is that we don't
+    want to block in the initializer.
+    """
+
```
beav: There is some extra whitespace here that will make flake8 complain, and shows up as red in `git diff HEAD^`. The extra whitespace is in a few other places as well.

beav:
LGTM after adding BZ number to commit message and removing extra whitespace before newlines in a few areas.

jortel:
flake8 happy.

PR #1562, opened 2015-01-22
Move repo auth functionality to platform

beav:
This code previously existed only in the RPM plugin but other plugins need to
use it as well.

This patch does the following:

 * moves code and tests from pulp_rpm.repo_auth to pulp.repoauth
 * moves the repo_auth wsgi script and conf file to platform
 * updates the pulp spec file
 * adds a new package, python-pulp-repoauth
 * updates pulp-dev.py
 * updates to the playpen scripts to create test certs
 * adds unit tests to improve coverage of repoauth module
 * adds an optional entry to repo_auth.conf to handle multiple URL prefixes
 * release notes for 2.7.x (assuming this goes into a Y release)


barnabycourt:
ok test

beav:
Is a dep not available on the PR builder?

from the build log:
```
ImportError: No module named rhsm
```

barnabycourt:
If this module depends on python-rhsm then that will have to be added as a requires line for the spec file for pulp-server

beav:
@barnabycourt  Ah, that would do it. This was added to the spec in this patch:

```
Requires: python-rhsm >= 1.8.0
```
This PR is against master instead of 2.6-dev though, so I don't know if it makes sense to add the Requires statement to 2.6-dev. Could the PR builder manually `yum install install python-rhsm` for now during setup? It is not an ideal solution but I'm not sure if a better fix is available.

beav:
There is a test failure on this after rebasing that isn't related to `python-rhsm`. I am looking into this now.

barnabycourt:
ok test

repoauth/pulp/repoauth/auth_handler_framework.py
```diff
@@ -0,0 +1,93 @@
+from pulp.repoauth import oid_validation, auth_enabled_validation
+
+try:
+    from mod_python import apache
```
mhrivnak: oh wow, we really need to clean out this whole repoauth module some time.

repoauth/pulp/repoauth/protected_repo_utils.py
```diff
@@ -0,0 +1,175 @@
...
+This module contains utilities for manipulating a local store of protected
+repo information.
+'''
+
+import os
+from threading import RLock
+
+# -- constants ----------------------------------------------------------------------
+
+WRITE_LOCK = RLock()
```
mhrivnak: :scream_cat:

repoauth/setup.py
```diff
@@ -0,0 +1,17 @@
...
+
+setup(
+    name='pulp-repoauth',
+    version='2.6.0',
+    license='GPLv2+',
+    packages=find_packages(exclude=['test']),
+    author='Pulp Team',
+    author_email='pulp-list@redhat.com',
+    entry_points={
+        'console_scripts': [
```
mhrivnak: does this really belong here? I would expect it to only come with the server package.

mhrivnak:
Merge when ready.

beav:
ok test

repoauth/pulp/repoauth/protected_repo_utils.py
rbarlow: On 02/04/2015 10:06 AM, Michael Hrivnak wrote:
>> +WRITE_LOCK = RLock()
> 
> :scream_cat:

Hahahaha, what the heck is going on in here‽

PR #1566, opened 2015-01-23
BZ1120671 - scheduled sync now have repo id and action associated with them in the task list

dkliban:
This PR introduces a dispatch_sync_with_auto_publish task. This task dispatched the actual sync_with_auto_publish task using apply_async_with_reservation. This mimics the behavior in the webservices controller responsible for repo syncs.

### DIFF
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -6,6 +6,7 @@ import logging
 import platform
 import threading
 import time
+import uuid
 
 from celery import beat
 from celery.result import AsyncResult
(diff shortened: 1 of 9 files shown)


mhrivnak:
Can you describe why the original plan did not work? I'm surprised this required changes to the scheduler itself.

To quickly recap, the plan as I understood it was to make ``pulp.server.yasks.sync_with_auto_publish`` look like ``pulp.server.tasks.publish``, remove the ``base=Task`` argument from both of their decorators, and move most of the pre-existing logic of ``pulp.server.yasks.sync_with_auto_publish`` into its manager.

dkliban:
Removing `base=Task` from the `sync_with_auto_publish` would break the `pulp-admin rpm repo sync run --repo-id zoo`.  pulp-admin expects to receive the ID of a task it can track.  However, if the API returns an ID of a task that kicked off the real `sync`, the real `sync` tasks can not be monitored by pulp-admin.   This is the behavior of `pulp.server.tasks.repository.publish`.  

The scheduler uses send_task method to dispatch our tasks from the `scheduled_calls` collection.   This meant that in order to kick off the sync task using `apply_async_with_reservation` we needed a dispatch task.  In order to hand the failure handler the right task ID to monitor, we added the ability to specify a task id for the inner (sync) task as an argument for the new `dispatch_sync_with_auto_publish` task.

barnabycourt:
ok test

PR #1572, opened 2015-01-26
Downgrade python-qpid requirement to match EL5.

jortel:


deps/gofer/gofer.spec
```diff
@@ -7,7 +7,7 @@
 
 Name: gofer
 Version: 2.4.0
-Release: 1%{?dist}
+Release: 2%{?dist}
```
barnabycourt: Should this be release 2.pulp?

barnabycourt:
LGTM

PR #1578, opened 2015-01-28
Disable StrictMode for ssh

AndreaGiardini:
Since the dev-script.sh sets apache rights for the home directory the publickey authentication for ssh doesn't work anymore. I also enable the services at boot.

Example: after running dev-script on a vagrant machine i'm not able to log in the machine anymore.

Tell me if i need to prompt for authorization to the user and i will edit the PR

### DIFF
--- a/playpen/dev-setup.sh
+++ b/playpen/dev-setup.sh
@@ -35,6 +35,23 @@ else
     esac
 fi
 
+if ! sudo grep -q "StrictModes no" /etc/ssh/sshd_config; then
+    echo -e "\nIn some setups (es. Vagrant machine) is it necessary to disable"
+    echo "Ssh strict key check to access the machine."
+    echo "Strict Key check will be disabled now and for future restart"
+    while true; do
+        read -p "Do you want to proceed?(y/n)" yn
+        case $yn in
+            [Yy])
+                echo "Disabling StrictHostCheck"
+                sudo sed -i 's/#StrictModes yes/StrictModes no/' /etc/ssh/sshd_config
+                sudo systemctl restart sshd
+                break;;
+            [Nn]) break;;
+        esac
+    done
+fi 
+
 echo "Disabling selinux for dev install"
 # dev setup does not work with selinux at this time
 sudo setenforce 0
@@ -80,9 +97,9 @@ pushd pulp
 setfacl -m user:apache:rwx .
 popd; popd
 
-
 echo "Starting services, this may take a few minutes due to mongodb journal allocation"
 for s in qpidd goferd mongod; do
+  sudo systemctl enable $s
   sudo systemctl start $s
 done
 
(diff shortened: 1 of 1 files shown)


rbarlow:
On 01/28/2015 06:51 AM, Andrea Giardini wrote:
> Since the dev-script.sh sets apache rights for the home directory the
> publickey authentication for ssh doesn't work anymore.

I believe I ran into this issue with my PR that adds RHEL 7 support, and
I think I may have gotten it working there. Though, I am not sure. My
memory of last week is already hazy! I know that I did change this area
of code just a bit, at least.

I'm busy on something else ATM, but I'll take a look at this when I get
a chance to respond to the comments on mine. It'll probably be next week
though.

playpen/dev-setup.sh
```diff
@@ -80,10 +80,14 @@ pushd pulp
 setfacl -m user:apache:rwx .
 popd; popd
 
+echo "Disabling strict mode for ssh"
+echo 'StrictModes no' | sudo tee -a /etc/ssh/sshd_config
+sudo systemctl restart sshd
```
rbarlow: The rest of this file has been adjusted to be idempotent, as of a PR I just merged yesterday. Would you mind adding a check that this setting is not already configured before appending it to the file and restarting ssh?

playpen/dev-setup.sh
```diff
@@ -80,10 +80,14 @@ pushd pulp
...
 
 echo "Starting services, this may take a few minutes due to mongodb journal allocation"
 for s in qpidd goferd mongod; do
   sudo systemctl start $s
+  sudo systemctl enable $s
```
rbarlow: In master, I've added this line just before the start. If you wouldn't mind, move it to the same place so we won't double enable it when this PR gets merged forward to master.

playpen/dev-setup.sh
```diff
@@ -92,6 +96,7 @@ sudo -u apache pulp-manage-db
 echo "Starting more services"
 for s in httpd pulp_workers pulp_celerybeat pulp_resource_manager; do
   sudo systemctl start $s
+  sudo systemctl enable $s
```
rbarlow: It looks like I forgot to do this on my change to master, nice!

playpen/dev-setup.sh
rbarlow: I do think it would be good if the script prompted the user with a y/n before doing this, since it is a security change. Something similar to the selinux prompt. Also, the prompt should not happen if StrictModes is already set to no.

rbarlow:
Thanks for the patch! Please do these things:

1) Make the sshd_config change idempotent, so that re-running this script doesn't add that setting repeatedly. One way this might be easy to accomplish is to use sed to s/#StrictModes yes/StrictModes no/ on the file. Or you could do a different test and use tee like you did if the test determines that setting not to be set.
2) Prompt the user before making the ssh change, but only if the change needs to be made. If the user declines, it's ok to skip and continue.
3) Move that one enable to be above the start line so that it doesn't get double enabled in master.

Let me know when it's ready for another look, and thanks!

rbarlow:
ok test

AndreaGiardini:
Should be ok now :)

rbarlow:
Thanks @AndreaGiardini!

PR #1580, opened 2015-01-28
1186420 - reading cert_t for custom SSL config

lzap:
https://bugzilla.redhat.com/show_bug.cgi?id=1186420

Gives read permissions to all cert_t certs. Usually certs are kept with cert_t anyway so Celery should be allowed to read them.


bmbouter:
LGTM

PR #1583, opened 2015-01-28
1159303 - delete agent queues on consumer unregistration.

jortel:
https://bugzilla.redhat.com/show_bug.cgi?id=1159303

The NotFound exception is new in gofer 2.5 which contains support for amqp 1.0 (proton).  We can delay merging until that work is complete and gofer 2.5 is released.

### DIFF
--- a/server/pulp/server/agent/direct/pulpagent.py
+++ b/server/pulp/server/agent/direct/pulpagent.py
@@ -19,6 +19,7 @@ Agent request flow:
 from logging import getLogger
 
 from gofer.proxy import Agent
+from gofer.messaging import Queue, NotFound
 
 
 log = getLogger(__name__)
@@ -75,6 +76,23 @@ class PulpAgent(object):
 
         admin = agent.Admin()
         admin.cancel(criteria=criteria)
+
+    @staticmethod
+    def delete_queue(url, name):
+        """
+        Purge and delete the agent queue.
+        :param url: The broker URL.
+        :type url: str
+        :param name: The queue name.
+        :type name: str
+        """
+        try:
+            queue = Queue(name)
+            queue.purge(url)
+            queue.delete(url)
+        except NotFound:
+            # queue may not exist
+            pass
 
 
 # --- Agent Capabilities -----------------------------------------------------
--- a/server/pulp/server/managers/consumer/agent.py
+++ b/server/pulp/server/managers/consumer/agent.py
@@ -14,10 +14,13 @@ Contains agent management classes
 """
 
 import sys
+
 from logging import getLogger
 from uuid import uuid4
+from gettext import gettext as _
 
-from pulp.common import constants
+from celery import task
+
 from pulp.common import tags
 from pulp.plugins.conduits.profiler import ProfilerConduit
 from pulp.plugins.loader import api as plugin_api
@@ -32,6 +35,11 @@ from pulp.server.async.task_status_manager import TaskStatusManager
 from pulp.server.agent import Context
 
 
+QUEUE_DELETE_DELAY = 600  # 10 min.
+QUEUE_DELETE_FAILED = _('queue %(name)s cannot be deleted: %(reason)s')
+QUEUE_DELETED = _('queue %(name)s deleted')
+
+
 logger = getLogger(__name__)
 
 
@@ -40,11 +48,13 @@ class AgentManager(object):
     The agent manager.
     """
 
-    def unregistered(self, consumer_id):
+    @staticmethod
+    def unregistered(consumer_id):
         """
         Notification that a consumer (agent) has
         been unregistered.  This ensure that all registration
-        artifacts have been cleaned up.
+        artifacts have been cleaned up.  Then, we fire off a task to lazily
+        delete the agent queue.
         :param consumer_id: The consumer ID.
         :type consumer_id: str
         """
@@ -53,6 +63,9 @@ class AgentManager(object):
         context = Context(consumer)
         agent = PulpAgent()
         agent.consumer.unregistered(context)
+        url = context.url
+        name = context.route.split('/')[-1]
+        delete_queue.apply_async(args=[url, name, consumer_id], countdown=QUEUE_DELETE_DELAY)
 
     @staticmethod
     def bind(consumer_id, repo_id, distributor_id, options):
@@ -404,6 +417,47 @@ class AgentManager(object):
             agent_bindings.append(agent_binding)
         return agent_bindings
 
+    @staticmethod
+    def delete_queue(url, name, consumer_id):
+        """
+        Delete the agent queue.
+        :param url: The broker URL.
+        :type url: str
+        :param name: The queue name.
+        :type name: str
+        :param consumer_id: The consumer ID.
+        :type consumer_id: str
+        """
+        try:
+            manager = managers.consumer_manager()
+            manager.get_consumer(consumer_id)
+            return  # still registered (abort)
+        except MissingResource:
+            # expected
+            pass
+        agent = PulpAgent()
+        agent.delete_queue(url, name)
+
+
+@task
+def delete_queue(url, name, consumer_id):
+    """
+    Task to delete the agent queue.
+    :param url: The broker URL.
+    :type url: str
+    :param name: The queue name.
+    :type name: str
+    :param consumer_id: The consumer ID.
+    :type consumer_id: str
+    """
+    try:
+        AgentManager.delete_queue(url, name, consumer_id)
+    except Exception, e:
+        delete_queue.retry(countdown=QUEUE_DELETE_DELAY)
+        logger.error(QUEUE_DELETE_FAILED, {'name': name, 'reason': e})
+    else:
+        logger.info(QUEUE_DELETED, {'name': name})
+
 
 class Units(dict):
     """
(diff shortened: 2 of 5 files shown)


server/pulp/server/managers/consumer/agent.py
```diff
@@ -14,10 +14,13 @@
 """
 
 import sys
+
 from logging import getLogger
 from uuid import uuid4
+from gettext import gettext as _
```
bmbouter: should come before logging import to be alphabetical

server/pulp/server/managers/consumer/agent.py
```diff
@@ -30,6 +33,12 @@
 from pulp.server.managers import factory as managers
 from pulp.server.async.task_status_manager import TaskStatusManager
 from pulp.server.agent import Context
+from pulp.server.async.tasks import Task
+
+
+QUEUE_DELETE_DELAY = 600  # 10 min.
+QUEUE_DELETE_FAILED = _('queue %(name)s cannot be deleted: %(reason)s')
```
bmbouter: nice internationalization

server/pulp/server/managers/consumer/agent.py
```diff
@@ -30,6 +33,12 @@
 from pulp.server.managers import factory as managers
 from pulp.server.async.task_status_manager import TaskStatusManager
 from pulp.server.agent import Context
+from pulp.server.async.tasks import Task
+
+
+QUEUE_DELETE_DELAY = 600  # 10 min.
```
bmbouter: Do we want this configurable? I'm not sure myself but I want to as the question.

server/pulp/server/managers/consumer/agent.py
```diff
@@ -404,6 +418,46 @@ def _unbindings(bindings):
...
+            manager.get_consumer(consumer_id)
+            return  # still registered (abort)
+        except MissingResource:
+            # expected
+            pass
+        agent = PulpAgent()
+        agent.delete_queue(url, name)
+
+
+@task(base=Task)
```
bmbouter: This will cause the task to be recorded with a task status. That is what you want right? Can you verify that the task_status records that get created look right? If you want it to occur but not be user facing then you can not use `base=Task`

server/pulp/server/managers/consumer/agent.py
```diff
@@ -404,6 +418,46 @@ def _unbindings(bindings):
...
+        Delete the agent queue.
+        :param url: The broker URL.
+        :type url: str
+        :param name: The queue name.
+        :type name: str
+        :param consumer_id: The consumer ID.
+        :type consumer_id: str
+        """
+        try:
+            manager = managers.consumer_manager()
```
bmbouter: Could this statement raise the MissingResource exception? If not it should be moved outside the try/except block

server/pulp/server/agent/direct/pulpagent.py
```diff
@@ -76,6 +77,23 @@ def cancel(context, task_id):
...
+    def delete_queue(url, name):
+        """
+        Purge and delete the agent queue.
+        :param url: The broker URL.
+        :type url: str
+        :param name: The queue name.
+        :type name: str
+        """
+        try:
+            queue = Queue(name)
```
bmbouter: Which of these statements can raise NotFound? The others should be moved outside of the try/except block.

server/pulp/server/managers/consumer/agent.py
```diff
@@ -404,6 +418,46 @@ def _unbindings(bindings):
             agent_bindings.append(agent_binding)
         return agent_bindings
 
+    @staticmethod
+    def delete_queue(url, name, consumer_id):
+        """
+        Delete the agent queue.
+        :param url: The broker URL.
```
bmbouter: I usually put a whitespace line between the description and the parameters for readability.

server/pulp/server/managers/consumer/agent.py
```diff
@@ -404,6 +418,46 @@ def _unbindings(bindings):
...
+    :param url: The broker URL.
+    :type url: str
+    :param name: The queue name.
+    :type name: str
+    :param consumer_id: The consumer ID.
+    :type consumer_id: str
+    """
+    try:
+        AgentManager.delete_queue(url, name, consumer_id)
+        logger.info(QUEUE_DELETED, {'name': name})
```
bmbouter: This probably isn't a line you are guarding for exceptions. This should be moved to an 'else' clause so this would have try/except/else as codepath branches.

server/pulp/server/managers/consumer/agent.py
```diff
@@ -416,6 +470,7 @@ def __init__(self, units):
         :param units: A list of content units.
         :type units: list
         """
+        super(Units, self).__init__()
```
bmbouter: Why is this needed? How does this impact the code behaviors?

server/test/unit/server/agent/direct/test_agent.py
```diff
@@ -28,8 +28,44 @@ def test_capabilities(self):
...
+
+    @patch('pulp.server.agent.direct.pulpagent.Queue')
+    @patch('pulp.server.agent.direct.pulpagent.NotFound', NotFound)
+    def test_delete_queue_not_found(self, queue):
+        url = 'test-url'
+        name = 'test-queue'
+        queue.return_value.purge.side_effect = NotFound
+
+        # test
+        agent = PulpAgent()
```
bmbouter: If the code is broken then an exception will be raised here. consider putting it in a try/except block and in the except block putting a self.fail(). That will cause it to be marked as a failure which is more accurate than an Error.

server/test/unit/server/managers/consumer/test_agent_manager.py
```diff
@@ -536,6 +543,75 @@ def test_get_distributor():
...
+    @patch('pulp.server.managers.consumer.agent.managers.consumer_manager')
+    @patch('pulp.server.managers.consumer.agent.PulpAgent')
+    def test_delete_queue(self, agent, consumer_manager):
+        url = 'test-url'
+        queue = 'test-queue'
+        consumer_id = 'test-consumer'
+        consumer_manager.return_value.get_consumer.side_effect = MissingResource
+
+        # test
+        AgentManager.delete_queue(url, queue, consumer_id)
```
bmbouter: Similar try/except suggested here as [this comment](https://github.com/pulp/pulp/pull/1583/files#r23771860).

bmbouter:
flake8 all pass, tests all pass for me, and all new modules have 100% code coverage! Good job. Implementing the suggestions before merging would be good. Besides code cleanup there are questions regarding the task_status data created by the delete_queue task and configurability of the the queue delete delay.

bmbouter:
Also I see that 2.6-testing was merged in after your commit. For working branches the best is to rebase your commit onto 2.6-testing again so that your commit will come after HEAD on 2.6-testing. We had a lot of challenges in the Django branching stuff because we were mixing merge and rebase. We should have used rebase, and so I'm passing along this info. Ask me if its not clear why rebase is better than merge.

bmbouter:
The 'jortel-' in the branch name is redundant since it shows the PR from your fork already. Then it reads as 'jortel:jortel-1159303'

server/pulp/server/managers/consumer/agent.py
jortel: alphabetical?  I'm not aware of this requirement in pep8.  can you point me to the text?

server/pulp/server/managers/consumer/agent.py
jortel: I'm not sure either but leaning against.  Figured that when in doubt, error on the side of less complexity.

server/pulp/server/managers/consumer/agent.py
jortel: Good catch.  Nope, I do not want this recorded in our task status collection.

server/pulp/server/managers/consumer/agent.py
jortel: when overriding constructors, the superclass constructor should always be called first unless the explicit intent is to completely replace its behavior (and done with extreme caution).  Not calling the superclass constructor can result in partially initialized objects.

server/pulp/server/managers/consumer/agent.py
bmbouter: It's not a PEP8 requirement, just a healthy convention. Notice that almost all imports are alphabetized in large portions of the codebase.

server/pulp/server/managers/consumer/agent.py
bmbouter: I think this is fine for now. The issue was first noticed by QE so in terms of tests it may be too long. This is already a huge improvement so lets leave it alone for now.

server/pulp/server/managers/consumer/agent.py
bmbouter: Agreed. This change is more correct so leave it. I just hope it doesn't cause an unexpected change.

server/pulp/server/managers/consumer/agent.py
rbarlow: On 01/29/2015 10:29 AM, Jeff Ortel wrote:
> alphabetical? I'm not aware of this requirement in pep8. can you point
> me to the text?

Agreed with what @bmbouter said, and I'll add that it's specifically
helpful because it groups imports from the same module and nearby
modules together.

jortel:
ok test

jortel:
ok test

beav:
ok test

PR #1584, opened 2015-01-29
1186420 - fixed reading cert_t for custom SSL config

lzap:
This fixes my previous commit. I've tested this on RHEL6 and RHEL7. Modern
Fedoras should go well as well.

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -52,7 +52,11 @@ allow celery_t pulp_cert_t:dir { getattr search };
 allow celery_t pulp_cert_t:file { read write getattr open };
 
 # custom SSL certificates reading
-miscfiles_read_generic_certs(celery_t)
+ifdef(`distro_rhel6', `
+    miscfiles_read_certs(celery_t)
+',`
+    miscfiles_read_generic_certs(celery_t)
+')
 
 ######################################
 


bmbouter:
LGTM

lzap:
We have other similar issue, now it requests writing: http://pastebin.test.redhat.com/261336

`miscfiles_manage_cert_files(celery_t)`

Is it possible celery does *write* to the katello-default-ca.crt file?

bmbouter:
@lzap When I search that pastebin for 'write' I get no results. I do see a lot of read denials though:  avc:  denied  { read } for  pid=8285 comm="celery" name="katello-default-ca.crt"

I don't see that celery is being denied write access, do you?

lzap:
If you use audit2allow on this, you will get the manage_cert_files above. I bet Celery is opening in the "rw" mode instead of "r" only. Can you check the codebase please? This should be fixed in Celery rather than in SELinux. Maybe we found a bug :-)

lzap:
Wait, there had to be write I missed to pastebin. Let me doublecheck.

lzap:
Disregard it is really `allow celery_t cert_t:file { read getattr open };`

bmbouter:
@lzap the `{ read getattr and open}` should be allowed on celery_t with `miscfiles_read_generic_certs(celery_t)` statement. If it's going to read it needs `open` and `read` by definition. `getattr` is also likely included. The paste you sent has a lot of 'read' denials in it, but I don't see any for `getattr` so I'm not sure where that is coming from. Perhaps the system this was tested on did not have a recompiled version with the `miscfiles_read_generic_certs` statement in it? 

It would have had to have been compiled by either downstream or manually recompiled. Did one of those things happen? It hasn't been included in a Pulp upstream beta so the version being tested couldn't have come from upstream Pulp.

Send more thoughts and we can discuss. We can also discuss on IRC in #pulp.

PR #1587, opened 2015-01-29
Django conversion cleanup

asmacdo:


bmbouter:
flake8 of all remaining files looks good, and coverage of webservices.views is at 100%! Merge it!

PR #1588, opened 2015-01-29
1120671 - scheduled tasks show correct resources in task list

mhrivnak:
I also rearranged some code to make the sync workflow consistent with the way
other "itineraries" work.

https://bugzilla.redhat.com/show_bug.cgi?id=1120671

### DIFF
--- a/server/pulp/server/tasks/repository.py
+++ b/server/pulp/server/tasks/repository.py
@@ -16,11 +16,9 @@ import logging
 import celery
 
 from pulp.common.error_codes import PLP0002, PLP0003, PLP0007
-from pulp.common.tags import action_tag, resource_tag, RESOURCE_REPOSITORY_TYPE
 from pulp.server.async.tasks import Task, TaskResult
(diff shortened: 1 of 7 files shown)


server/pulp/server/tasks/repository.py
```diff
@@ -168,7 +166,7 @@ def distributor_update(repo_id, distributor_id, config, delta):
     return TaskResult(distributor, bind_error, additional_tasks)
 
 
-@celery.task(base=Task)
+@celery.task()
 def publish(repo_id, distributor_id, overrides=None):
```
dkliban: Is this publish task still needed?  Where is it used?

dkliban:
Everything else looks good to merge.

server/pulp/server/tasks/repository.py
mhrivnak: Yep, this is what the scheduler uses if you have a scheduled publish.

PR #1590, opened 2015-01-30
refactor consumer_groups tests to use generate_json_response

asmacdo:


bmbouter:
flake8 is good, all tests pass, and 100% coverage still. LGTM

PR #1591, opened 2015-01-30
django conversion content/units/<type_id>/<unit_id>/pulp_user_metadata

asmacdo:
Hand testing:
http://fpaste.org/178382/26274381/


rbarlow:
LGTM.

PR #1593, opened 2015-01-30
Add default database definition so EL6 tests are happy

bmbouter:
Fixes build failures on EL6.

### DIFF
--- a/server/pulp/server/webservices/views/consumer_groups.py
+++ b/server/pulp/server/webservices/views/consumer_groups.py
@@ -1,8 +1,5 @@
-import json
-
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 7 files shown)


server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -69,12 +70,10 @@ def post(self, request):
         link = {"_href": reverse('consumer_group_resource',
                 kwargs={'consumer_group_id': group['id']})}
         group.update(link)
-        return HttpResponseRedirect(
```
asmacdo: Hand verified:

`POST /v2/consumer_groups/`
http://fpaste.org/178741/22653021/

PR #1576, opened 2015-01-27
WIP Introducing roles view

ipanova:
~~1  Two Unit tests should be added.~~
~~2 Manual testing example should be added.~~

PR #1597, opened 2015-02-02
Introducing roles view

ipanova:
Manual Testing:
GET /pulp/api/v2/roles/ http://fpaste.org/178273/14226139/
POST /pulp/api/v2/roles/ http://fpaste.org/178275/42261425/
GET /pulp/api/v2/roles/<role_id>/ http://fpaste.org/178276/14386142/
GET nonexistent role_id http://fpaste.org/178278/14484142/
DELETE /pulp/api/v2/roles/<role_id>/ http://fpaste.org/178279/14226145/
PUT /pulp/api/v2/roles/<role_id>/ http://fpaste.org/178283/14740142/
GET /pulp/api/v2/roles/<role_id>/users/ http://fpaste.org/178284/61484314/
POST /pulp/api/v2/roles/<role_id>/users/ http://fpaste.org/178286/61502814/
POST nonexistent login http://fpaste.org/178287/14226151/
DELETE /pulp/api/v2/roles/<role_id>/users/<login>/ http://fpaste.org/178288/26151951/

reference :https://github.com/pulp/pulp/pull/1576


server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
+import json
+
+from django.core.urlresolvers import reverse
+from django.utils.encoding import iri_to_uri
+from django.views.generic import View
+from django.http import HttpResponseRedirect
+from pulp.server import exceptions as pulp_exceptions
```
mhrivnak: blank line above for pep8

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+        user_query_manager = factory.user_query_manager()
+        permissions_manager = factory.permission_manager()
+        roles = role_query_manager.find_all()
+        for role in roles:
+            role['users'] = [u['login'] for u in
+                             user_query_manager.find_users_belonging_to_role(role['id'])]
+
+            resource_permission = {}
+            # isolate schema change
+            if role['permissions']:
```
mhrivnak: This is completely correct and fine, but just so you're aware of another stylistic option, some people would write these two lines as just:

``for item in role['permissions'] or []:``

which saves a level of indentation and only accesses the value in ``role`` once. But it's really just a stylistic preference, and you're welcome to keep what you have here if you like. I personally don't feel strongly either way.

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+                for item in role['permissions']:
+                    resource = item['resource']
+                    operations = item.get('permission', [])
+                    resource_permission[resource] = [permissions_manager.operation_value_to_name(o)
+                                                     for o in operations]
+
+            role['permissions'] = resource_permission
+        for role in roles:
+            link = {"_href": reverse('role_resource',
```
mhrivnak: better to stick with just single quotes for consistency.

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+            # isolate schema change
+            if role['permissions']:
+                for item in role['permissions']:
+                    resource = item['resource']
+                    operations = item.get('permission', [])
+                    resource_permission[resource] = [permissions_manager.operation_value_to_name(o)
+                                                     for o in operations]
+
+            role['permissions'] = resource_permission
+        for role in roles:
```
mhrivnak: You already iterate through the ``roles`` on line 35, so do you actually need a second pass of iteration here? Or can you put the steps below in the loop above?

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+        role = manager.create_role(role_id, display_name, description)
+        link = {"_href": reverse('role_resource',
+                kwargs={'role_id': role['id']})}
+        role.update(link)
+        response = generate_json_response_with_pulp_encoder(role)
+        response['Location'] = iri_to_uri(link['_href'])
+        response.status_code= 201
+        response.reason_phrase= 'CREATED'
+        return response
+        #return HttpResponseRedirect(
```
mhrivnak: Can we just remove this? It's better to not pollute the code base with commented-out or otherwise disabled code.

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+        display_name = role_data.get('display_name', None)
+        description = role_data.get('description', None)
+        manager = factory.role_manager()
+        role = manager.create_role(role_id, display_name, description)
+        link = {"_href": reverse('role_resource',
+                kwargs={'role_id': role['id']})}
+        role.update(link)
+        response = generate_json_response_with_pulp_encoder(role)
+        response['Location'] = iri_to_uri(link['_href'])
+        response.status_code= 201
```
mhrivnak: space before the ``=``

mhrivnak: I like using the values from httplib. I this case, ``httplib.CREATED``. It's a little more descriptive, and thus helps developers avoid choosing the wrong code based on a misunderstanding. It also makes it slightly easier to search for uses of that actual HTTP response code.

Although admittedly, neither of those are especially strong arguments. I can also imagine some people like the simplicity of just the int. I can go either way on it.

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,229 @@
...
+        response.reason_phrase= 'CREATED'
```
mhrivnak: Ok, I should have kept reading before commenting on the line above. I'm surprised django doesn't default this to the value in ``httplib.responses``. It's worth making sure; you might find that django does use that as a default. If it doesn't, it would be a great RFE and contribution to django, but you should in the mean time do this:

```
response.status_code = httplib.CREATED
response.reason_phrase = httplib.responses[response.status_code]
```

mhrivnak:
Looks good! Let me know if you have any questions (perhaps ping me just to make sure I see comments here), or otherwise merge when ready.

server/pulp/server/webservices/views/roles.py
ipanova: If I will do it in this way then in the header I will have
`HTTP/1.1 201 Created` instead of `HTTP/1.1 201 CREATED`, in django seems like that the reason is written in uppercase.

bmbouter:
@ipanova looking at http://fpaste.org/178275/42261425/

It looks like Location is pretty different. The web.py one doesn't really make sense, but the Django one looks right. Is there a bug with the web.py version? If so it probably should be filed and this change can be marked as its fix to send it to QE.

What do you think?

bmbouter:
@ipanova I looked at all the URLs and [this is the one that has problems](https://github.com/pulp/pulp/pull/1597#issuecomment-73956071). flake8 looks good. Test coverage is 100% and tests pass on all platforms. I'm assigning it back to you.

ipanova:
@bmbouter the location in web.py was from the begging wrong, so the django now handles it in the right way. i will report the bug and will refer this pr as the fix.

server/pulp/server/webservices/views/roles.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @auth_required(authorization.CREATE)
+    def post(self, request):
+        """
+        Create a new role.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :return: Response containing the role
+        :rtype: HttpResponseRedirect
```
asmacdo: using the new util, isn't this is an HttpResponse now?

bmbouter:
@ipanova regarding [BZ 1191972](https://bugzilla.redhat.com/show_bug.cgi?id=1191972) that is the right thing to do. Thanks!

server/pulp/server/webservices/views/roles.py
bmbouter: @ipanova asmacdo is right, these docblocks need to be updated to specify the return type as `HttpResponse`. please updat.e

bmbouter:
ok test

ipanova:
ok test

PR #1599, opened 2015-02-02
Documenting a quirk of this function that could lead to clobbering

mhrivnak:
pre-existing unit metadata, and how to avoid that.

re #101 

https://pulp.plan.io/issues/101


beav:
LGTM

PR #1601, opened 2015-02-02
1174361 - Revert patch introduced with b0f2319. It is not needed.

bmbouter:
https://bugzilla.redhat.com/show_bug.cgi?id=1174361

This patch is no longer needed at all. Once this is merged forward and the dep is rebuilt without it I am going to CLOSE the BZ.


beav:
LGTM

PR #1602, opened 2015-02-03
remove controllers that have been replaced by django views

asmacdo:
Removes webpy controllers that have been replaced by django views.


server/test/unit/test_plugin_manager.py
```diff
@@ -73,6 +73,7 @@ def test_types(self):
         """
 
         # Setup
+        types_db.clean()
```
beav: Is this change intentional?

asmacdo: Yes, I should have mentioned that. One of the tests that I deleted cleans up after one of the tests that remains. When I removed the webpy tests, it caused this test to fail because it was operating on an unclean database.

beav:
This looks ok to me but someone with more familiarity with the Django work should probably review as well.

bmbouter:
@asmacdo there are some flake8 issues with these. would you run flake8 on all the files and push a fix for those?

bmbouter:
Besides the flake8 LGTM

bmbouter:
@asmacdo also can you track this issue here. https://pulp.plan.io/issues/148

PR #1603, opened 2015-02-03
Bug 1182279 - DEBUG level logs database password

ipanova:
https://bugzilla.redhat.com/show_bug.cgi?id=1182279


bmbouter:
two small things:
1) commit labels should be in the form 'XXXXXXX - <message>'. No need to write Bug at the front.
2) Also the branch name does not need to include bug so 'bug-1182279' can just be '1182279'

Fix issue (1) on this PR and do future PRs according to (2).

bmbouter:
@ipanova I don't mean the title of the PR, but the commit message itself. That commit message gets included in the changelog of the RPM in the final build. The title of the PR is a convention, but it never is used anywhere unlike the commit messages.

bmbouter:
LGTM after fixing the commit message. Good job!

beav:
This was merged to 2.6-dev but the BZ is for 2.6.0 (2.6-testing). I am not sure which release it is meant for, but there is a mismatch somewhere :koala:

bmbouter:
This was later merged to 2.6-testing with PR #1605 so the BZ target release should be right.

PR #1608, opened 2015-02-03
1184262: do not download signed RHEL5 rpms when downloading the mash

beav:


docs/user-guide/installation.rst
```diff
@@ -84,6 +84,10 @@ Repositories
  * Fedora: http://repos.fedorapeople.org/repos/pulp/pulp/fedora-pulp.repo
  * RHEL: http://repos.fedorapeople.org/repos/pulp/pulp/rhel-pulp.repo
 
+.. note::
+   If you would like to install the RHEL 5 Client for Pulp, please use the
+   `RHEL 5 repository <http://repos.fedorapeople.org/repos/pulp/pulp/rhel5-pulp.repo>`_.
```
rbarlow: We keep the other two repo files in the rel-eng directory in this repo. Can you add this rhel5-pulp.repo file there too? Otherwise, fedorapeople is the only official copy of the file.

Also, I noticed that only the testing and beta repos were defined in the RHEL 5 file. It would be good to go ahead and add the pulp-2-stable and pulp-stable yum repos there so that when we release it's already there and ready.

rbarlow:
LGTM.

docs/user-guide/installation.rst
beav: added

PR #1612, opened 2015-02-04
Add support for using mock objects in the SideEffect

barnabycourt:


devel/pulp/devel/unit/util.py
```diff
@@ -1,5 +1,5 @@
 import os
-
```
jortel: pep8

jortel:
Good addition.  Thanks! @barnabycourt.
LGTM

barnabycourt:
ok test

PR #1615, opened 2015-02-06
PEP-8 tests for pulp.server.event.amqp.

rbarlow:


server/test/unit/server/event/test_amqp.py
```diff
@@ -1,19 +1,6 @@
...
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
-
 import mock
 
-import base
+from ... import base
```
beav: Can this be changed to an absolute import?

beav:
ok test

rbarlow:
ok test

server/test/unit/server/event/test_amqp.py
rbarlow: I don't believe so, because base does not live in an installed package. I don't necessarily think there's much of an advantage to making base an installed package either.

beav:
LGTM

PR #1616, opened 2015-02-07
Do not log the length of the database password.

rbarlow:
This commit also fixes several unit tests that were assuming that the database was
local and unauthenticated.

### DIFF
--- a/server/test/unit/server/db/test_connection.py
+++ b/server/test/unit/server/db/test_connection.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python
-
 from ConfigParser import NoOptionError
 import unittest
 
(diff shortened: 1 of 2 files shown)


server/test/unit/server/db/test_connection.py
```diff
@@ -340,12 +372,16 @@ def test_initialize_username_and_password(self, mock_mongoengine):
         mock_mongoengine_instance = mock_mongoengine.connect.return_value
         mock_mongoengine_instance.server_info.return_value = {"version":
                                                               connection.MONGO_MINIMUM_VERSION}
-        config.config.set('database', 'username', 'admin')
-        config.config.set('database', 'password', 'admin')
+        config.config.set('database', 'name', 'nbachamps')
+        config.config.set('database', 'username', 'larrybird')
+        config.config.set('database', 'password', 'celtics1981')
```
beav: :four_leaf_clover:

beav:
Can you make a BZ for this issue?

beav:
after the bz is created, the change itself is LGTM

rbarlow:
On 02/07/2015 03:46 PM, Chris Duryee wrote:
> Can you make a BZ for this issue?

I can, but I don't believe this code was ever in a Pulp release. Do you
think it's necessary given that?

beav:
I think yes in this particular case, just because it was introduced by a change that had a bz, rhbz 1182279. You could just re-use that bz # I suppose.

rbarlow:
I created this bug for it:

https://bugzilla.redhat.com/show_bug.cgi?id=1190824

beav:
LGTM after build completes

PR #1617, opened 2015-02-07
PEP-8 tests for pulp.server.managers.consumer.bind.

rbarlow:


barnabycourt:
Looking at the console log it appears to be a pep8 item
11:52:30 server/pulp/server/managers/auth/cert/certificate.py:939:1: E402 module level import not at top of file

barnabycourt:
I've got a patch that should take care of the the one pep8 issue in https://github.com/pulp/pulp/pull/1619  

You did not introduce this error.  LGTM

PR #1618, opened 2015-02-09
1185011 - pulp-admin says sync schedule deleted succesfully for non-existing schedule id

ipanova:
https://bugzilla.redhat.com/show_bug.cgi?id=1185011

### DIFF
--- a/server/pulp/server/managers/schedule/utils.py
+++ b/server/pulp/server/managers/schedule/utils.py
@@ -1,15 +1,4 @@
 # -*- coding: utf-8 -*-
-#
-# Copyright © 2012 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public License as
-# published by the Free Software Foundation; either version 2 of the License
-# (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied, including the
-# implied warranties of MERCHANTABILITY, NON-INFRINGEMENT, or FITNESS FOR A
-# PARTICULAR PURPOSE.
-# You should have received a copy of GPLv2 along with this software; if not,
-# see http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
 
 from gettext import gettext as _
 import itertools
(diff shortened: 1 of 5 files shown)


bmbouter:
There are a lot of flake8 errors in these files. I didn't look if you introduced them, but would you want to fix them?

server/pulp/server/managers/schedule/utils.py
```diff
@@ -104,11 +94,17 @@ def delete(schedule_id):
...
     :type  schedule_id: basestring
     """
+
     try:
-        ScheduledCall.get_collection().remove({'_id': ObjectId(schedule_id)}, safe=True)
+        spec = {'_id': ObjectId(schedule_id)}
     except InvalidId:
         raise exceptions.InvalidValue(['schedule_id'])
 
+    schedule = ScheduledCall.get_collection().find_and_modify(
```
bmbouter: atomic, nice!

server/pulp/server/managers/schedule/utils.py
```diff
@@ -104,11 +94,17 @@ def delete(schedule_id):
     :param schedule_id: a unique ID for a schedule
     :type  schedule_id: basestring
     """
+
     try:
-        ScheduledCall.get_collection().remove({'_id': ObjectId(schedule_id)}, safe=True)
```
bmbouter: @ipanova this should have worked. Do you know why it didn't work?

bmbouter: A similar line is present in line 116.

bmbouter:
I hand tested this and it fixes the issue. I was also able to reproduce the issue on the parent commit.

bmbouter:
@ipanova  LGTM, but I'm concerned that because the removed line 108 didn't work we may have problems elsewhere get_collection().remove() is called. What do you think?

bmbouter:
@ipanova after some IRC convo I see this works because the remove() occured but didn't validate. Thanks for the explanation. LGTM

PR #1619, opened 2015-02-09
Print helpful messages before starting the flake8

barnabycourt:
Print messages before starting flake8 and the unit tests to make it easier to follow the jenkins log files


beav:
LGTM

It may make sense to squash the two commits into one but the change itself looks good.

PR #1622, opened 2015-02-10
1190756 - Send SIGQUIT to the workers instead of the proxy.

rbarlow:
https://bugzilla.redhat.com/show_bug.cgi?id=1190756


asmacdo:
LGTM

PR #1623, opened 2015-02-10
Create 1.4 compatible HttpResponseRedirect object.

ipanova:
https://pulp.plan.io/issues/149

### DIFF
--- a/server/test/unit/server/webservices/views/test_util.py
+++ b/server/test/unit/server/webservices/views/test_util.py
@@ -54,3 +54,18 @@ class TestWebservicesUtils(unittest.TestCase):
         test_content = {'foo': 'bar'}
         util.generate_json_response_with_pulp_encoder(test_content)
         mock_json.dumps.assert_called_once_with(test_content, default=pulp_json_encoder)
+
+    @mock.patch('pulp.server.webservices.views.util.iri_to_uri')
+    def test_generate_redirect_response(self, mock_iri_to_uri):
+        """
+        Test HttpResponseRedirect.
+        """
+        test_content = {'foo': 'bar'}
+        href = '/some/url/'
+        response = HttpResponse(content=test_content)
+        redirect_response = util.generate_redirect_response(response, href)
+        self.assertEqual(redirect_response.status_code, 201)
+        self.assertEqual(redirect_response.reason_phrase, 'CREATED')
+        self.assertEqual(redirect_response._headers['location'][1],
+                         str(mock_iri_to_uri.return_value))
+        mock_iri_to_uri.assert_called_once_with(href)
--- a/server/pulp/server/webservices/views/util.py
+++ b/server/pulp/server/webservices/views/util.py
@@ -2,6 +2,7 @@ import functools
 import json
 
 from django.http import HttpResponse
+from django.utils.encoding import iri_to_uri
 
 from pulp.server.webservices.controllers.base import json_encoder as pulp_json_encoder
 
@@ -38,3 +39,10 @@ generate_json_response_with_pulp_encoder = functools.partial(
     generate_json_response,
     default=pulp_json_encoder,
 )
+
+
+def generate_redirect_response(response, href):
+    response['Location'] = iri_to_uri(href)
+    response.status_code = 201
+    response.reason_phrase = 'CREATED'
+    return response


bmbouter:
I expected the branch name to be '149' but instead it is 'task149'. All tasks, issues, stories, etc in Redmine share the same numbering sequence so '149' as the branch name specifies some clearly. This does not need to be adjusted for this PR, just something I observed.

bmbouter:
Typically the redmine issue is linked to in the PR description.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -54,3 +54,14 @@ def test_generate_json_response_with_pulp_encoder(self, mock_json):
         test_content = {'foo': 'bar'}
         util.generate_json_response_with_pulp_encoder(test_content)
         mock_json.dumps.assert_called_once_with(test_content, default=pulp_json_encoder)
+
+    def test_generate_response_redirect(self):
+        """
+        Test HttpResponseRedirect.
+        """
+        test_content = {'foo': 'bar'}
+        response = util.generate_json_response_with_pulp_encoder(test_content)
```
bmbouter: Can this call to `utils.generate_json_response_with_pulp_encoder` build the object manually? A simple HttpResponse(content=test_content) should suffice. This way the unit test won't be affected by breakages in `utils.generate_json_response_with_pulp_encoder`.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -54,3 +54,14 @@ def test_generate_json_response_with_pulp_encoder(self, mock_json):
...
+        response = util.generate_redirect_response(response, '/some/url/')
```
bmbouter: Generally I try to avoid the reassignment of a variable. At different times in the code `response` will take on different states which make it slightly more difficult to read.

bmbouter: The string `'/some/url/'` could be de-duplicated as an variable in this function and then used again on L#67.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -54,3 +54,14 @@ def test_generate_json_response_with_pulp_encoder(self, mock_json):
         test_content = {'foo': 'bar'}
         util.generate_json_response_with_pulp_encoder(test_content)
         mock_json.dumps.assert_called_once_with(test_content, default=pulp_json_encoder)
+
+    def test_generate_response_redirect(self):
```
bmbouter: Consider mocking iri_to_uri() so that the unit test does not actually call into Django.

bmbouter:
flake8 and tests look good. I left some comments. Please make some edits and ping me for re-review.

server/test/unit/server/webservices/views/test_util.py
ipanova: nice catch, thank you :)

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -54,3 +54,16 @@ def test_generate_json_response_with_pulp_encoder(self, mock_json):
...
+        """
+        Test HttpResponseRedirect.
+        """
+        test_content = {'foo': 'bar'}
+        href = '/some/url/'
+        response = HttpResponse(content=test_content)
+        redirect_response = util.generate_redirect_response(response, href)
+        self.assertEqual(redirect_response.status_code, 201)
+        self.assertEqual(redirect_response.reason_phrase, 'CREATED')
+        mock_iri_to_uri.assert_called_once_with(href)
```
bmbouter: Try adding an assert statement to verify that the Location header is set to the return value of mock_iri_to_uri. From my own debugging I think the header is stored at redirect_response._headers['location']. I recommend letting mock auto-create the return_value which you can access with mock_iri_to_uri.return_value.

bmbouter:
@ipanova if you can add that assert statement and the tests pass on all platforms then LGTM.

rbarlow:
On 02/10/2015 08:04 AM, bmbouter wrote:
> I expected the branch name to be '149' but instead it is 'task149'. All
> tasks, issues, stories, etc in Redmine share the same numbering sequence
> so '149' as the branch name specifies some clearly. This does not need
> to be adjusted for this PR, just something I observed.

I don't think we need to have a policy about what branch names are used.
For one, the branch name resides only in the developer's fork of Pulp.
Secondly, onethe commit is merged the branch name is not part of the
repository history.

As long as we format our commit messages well, I think the developer can
use whatever branch names they like.

bmbouter:
@rbarlow I agree with you, but [the docs document](http://pulp.readthedocs.org/en/latest/dev-guide/contributing/branching.html#bug-fix-branches) that we do assert a branch name convention. Should we adjust the docs?

bmbouter:
@ipanova LGTM, great job in resolving this.

rbarlow:
On 02/10/2015 10:13 AM, bmbouter wrote:
> @rbarlow <https://github.com/rbarlow> I agree with you, but the docs
> document
> <http://pulp.readthedocs.org/en/latest/dev-guide/contributing/branching.html#bug-fix-branches>
> that we do assert a branch name convention. Should we adjust the docs?

Yes, I also commented on the pull request that created that doc saying I
disagreed with it then. There really is no reason to have such a policy.
It's needlessly strict.

PR #1624, opened 2015-02-10
Gofer 2.5

jortel:


beav:
LGTM once jenkins build passes.

Do we need to go through the karma process for gofer 2.5 for fedora 21?

PR #1632, opened 2015-02-11
PEP-8 distributor mixin tests.

rbarlow:


server/test/unit/plugins/conduits/test_mixins.py
```diff
@@ -2,14 +2,73 @@
...
+                                          ImporterScratchPadMixin, ImporterConduitException)
 from pulp.server.db.model.repository import Repo
 from pulp.server.managers.repo.cud import RepoManager
+from pulp.server.managers.repo.distributor import RepoDistributorManager
 from pulp.server.managers.repo.importer import RepoImporterManager
 import pulp.plugins.types.database as types_database
 import pulp.server.managers.factory as manager_factory
 
 
+class DistributorScratchpadMixinTests(base.PulpServerTests):
```
bmbouter: @rbarlow Do you think future versions of Pulp will want class docblocks and function docblocks for test code too? Not necessary for this commit, but I am asking in general.

bmbouter:
flake8 looks good on all changed files. LGTM

server/test/unit/plugins/conduits/test_mixins.py
rbarlow: On 02/11/2015 08:34 AM, bmbouter wrote:
> @rbarlow <https://github.com/rbarlow> Do you think future versions of
> Pulp will want class docblocks and function docblocks for test code too?
> Not necessary for this commit, but I am asking in general.

I think it's important. We have too many tests that are impossible to
tell what they are actually testing. Docblocks are the natural place to
describe what the test is doing in a human readable way.

Once I finish PEP-8ing the code, I plan to apply PEP-257 next which is
going to be strict about this.

bmbouter: sounds good. :palm_tree: :cactus: :camel: :camera:

PR #1627, opened 2015-02-10
Adding back urls deleted because of compatibility issues.

ipanova:
GET /pulp/api/v2/consumer_groups/ http://fpaste.org/184316/67088614/
POST /pulp/api/v2/consumer_groups/ http://fpaste.org/184326/67157314/
POST missing required param /pulp/api/v2/consumer_groups/  http://fpaste.org/184318/36710061/
POST invalid param /pulp/api/v2/consumer_groups/  http://fpaste.org/184320/36711661/
GET /pulp/api/v2/content/uploads/ http://fpaste.org/184323/71301142/
POST /pulp/api/v2/content/uploads/ http://fpaste.org/184324/67144414/

PR #1635, opened 2015-02-11
Introducing permissions view.

ipanova:
GET /pulp/api/v2/permissions/ http://fpaste.org/184277/42366756/
GET /pulp/api/v2/permissions for particular resource http://fpaste.org/184276/42366749/
POST /pulp/api/v2/permissions/actions/grant_to_user/ http://fpaste.org/184289/14236692/
POST /pulp/api/v2/permissions/actions/grant_to_role/ http://fpaste.org/184292/36693731/
POST /pulp/api/v2/permissions/actions/revoke_from_role/  http://fpaste.org/184297/69689142/
POST /pulp/api/v2/permissions/actions/revoke_from_user/ http://fpaste.org/184298/23669796/
POST invalid params /pulp/api/v2/permissions/actions/revoke_from_user/  http://fpaste.org/184301/23669940/  (rest of the cases are the same)


bmbouter:
@ipanova I looked at all the fpastes and they all look good except for this one: http://fpaste.org/184301/23669940/

This one seems to have the same extra slash problem with _href similar to [the other code](https://github.com/pulp/pulp/pull/1627#issuecomment-73947500) that uses that redirect helper function. Fixing it in the helper function would probably be good so that one fix fixes all of these cases.

bmbouter:
flake8 looks good, test coverage is 100% and they pass on all platforms.

server/pulp/server/webservices/views/permissions.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request):
+        """
+        Retrieve permissions for all resources or for a particular resource.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :return: Response containing a list of permissions for resource/s
```
bmbouter: I'll suggest an extra line of whitespace between the return params (:return, :rtype) and the argument params for readability. Not required, but a stylistic suggestion.

server/pulp/server/webservices/views/permissions.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        query_params = request.GET
+        resource = query_params.get('resource', None)
+
+        permissions = []
+        if resource is None:
+            permissions = factory.permission_query_manager().find_all()
+        else:
+            permission = factory.permission_query_manager().find_by_resource(resource)
+            if permission is not None:
```
bmbouter: FYI I think this can also be rewritten as `if not permission:`

server/pulp/server/webservices/views/permissions.py
```diff
@@ -0,0 +1,189 @@
...
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :return: Response containing a list of permissions for resource/s
+        :rtype: django.http.HttpResponse
+        """
+        query_params = request.GET
+        resource = query_params.get('resource', None)
+
+        permissions = []
+        if resource is None:
```
bmbouter: FYI this could be written as `if not resource:`, but I like your way better.

server/test/unit/server/webservices/views/test_permissions.py
```diff
@@ -0,0 +1,186 @@
...
+    def test_grant_to_user_invalid_param(self):
+        """
+        Test grant permissions to user with invalid params.
+        """
+        request = mock.MagicMock()
+        request.body_as_json = {'operations': ['READ'], 'resource': '/v2/some/'}
+        grant = GrantToUserView()
+        try:
+            grant.post(request)
+        except InvalidValue, response:
```
bmbouter: This would be better written as assertRaises(InvalidValue, grant.post, request). It would collapse 6 lines down to 1.

bmbouter: This may be problematic because you won't have a reference to the 'response' object anymore. Perhaps that is why it is written this way. It's up to you how best to fix it (if any).

bmbouter:
The tests look like the mock things well, good job.

bmbouter:
@ipanova I'm done with this so I am assigning it back to you. The only issue that really needs fixing is the [extra-slash regression issue](https://github.com/pulp/pulp/pull/1635#issuecomment-73958684).

server/pulp/server/webservices/views/permissions.py
ipanova: done

server/test/unit/server/webservices/views/test_permissions.py
ipanova: i will leave it as it is, since i am using the response object to check different things.

ipanova:
@bmbouter I looked closer to the extra slash issue - seems like that in the curl I provided url with extra slash. So the problem is not in the redirect helper function, because here i do not even use it, but in the typo during the input. Here the right example http://fpaste.org/184587/14237336/

ipanova:
ok test

ipanova:
ok test

PR #1636, opened 2015-02-11
1185868 - allow as_generator to be passed in to get_source_units 

beav:
do_get_repo_units() supports `as_generator` but previously we were unable to
pass this in via get_source_units.


mhrivnak:
LGTM

PR #1638, opened 2015-02-12
Remove dist_list that is no longer used

barnabycourt:


beav:
@barnabycourt this is for 2.6-dev and not testing?

barnabycourt:
@beav The dist_list was created for nodes in the 2.6-testing branch so there is no need to remove it there.

beav:
oh i see what you mean

LGTM

PR #1641, opened 2015-02-12
Removes working directory cleanup from repo group delete

dkliban:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1192177

### DIFF
--- a/server/pulp/server/managers/repo/group/cud.py
+++ b/server/pulp/server/managers/repo/group/cud.py
@@ -1,7 +1,4 @@
-from gettext import gettext as _
 import logging
-import os
-import shutil
 import sys
 
 from celery import task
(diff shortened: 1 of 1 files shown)


bmbouter:
@dkliban Where is the code that does the cleanup now?

dkliban:
https://github.com/pulp/pulp/blob/master/server/pulp/server/async/tasks.py#L374
https://github.com/pulp/pulp/blob/master/server/pulp/server/async/tasks.py#L401

bmbouter:
@dkliban cool. Is there another place where cleanup happens if neither of these handlers run?

dkliban:
@bmbouter  https://github.com/pulp/pulp/blob/master/server/pulp/server/async/tasks.py#L126

bmbouter:
Cool, the removing of the code is all we needed here. I don't think any new tests are needed since there this code is going to be rewritten with mongoengine in the not-too-distant future. LGTM.

PR #1643, opened 2015-02-13
PEP-8 test_manager_factory.py.

rbarlow:


server/test/unit/server/managers/test_factory.py
```diff
@@ -101,7 +84,7 @@ def test_get_manager_invalid_type(self):
             self.fail('Invalid manager type did not raise an exception')
         except factory.InvalidType, e:
             self.assertEqual(e.type_key, 'foo')
-            print(e)
+            self.assertEqual(str(e), 'Invalid manager type requested [foo]')
```
beav: good catch :panda_face:

beav:
LGTM

PR #1645, opened 2015-02-16
WIP Introducing users view.

ipanova:
`POST /pulp/api/v2/users/` http://fpaste.org/186531/41838331/
`POST /pulp/api/v2/users/ with missing required key 'login'`  http://fpaste.org/186541/18470114/ ( i changed the error from InvalidValue-->MissingValue since i think it is more proper in this case)
`POST /pulp/api/v2/users/ with invalid key ` http://fpaste.org/186545/18495914/ ( bug reported against webpy https://bugzilla.redhat.com/show_bug.cgi?id=1193059)
`PUT /pulp/api/v2/users/user_id/` http://fpaste.org/186550/18548314/
`DELETE /pulp/api/v2/users/user_id/` http://fpaste.org/186553/18590514/
`GET /pulp/api/v2/users/` http://fpaste.org/186554/14241859/
`GET /pulp/api/v2/users/user_id/` http://fpaste.org/186556/41860861/
`GET /pulp/api/v2/users/user_id/ non-existent user` http://fpaste.org/186560/14241862/

This PR also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1193059
and solves issue of https://github.com/pulp/pulp/pull/1627#issuecomment-73948238


ipanova:
ok test

ipanova:
ok test

ipanova:
ok test

ipanova:
ok test

ipanova:
@bmbouter @asmacdo $ curl -i -H "Accept: application/json" -H "WebFrameworkSwitch: webpy" -X POST -k -u admin:admin -d '{}' "https://localhost/pulp/api/v2/users/"
HTTP/1.1 400 Bad Request
Date: Wed, 18 Feb 2015 11:35:19 GMT
Server: Apache/2.4.10 (Fedora) OpenSSL/1.0.1e-fips mod_wsgi/3.5 Python/2.7.5
Content-Encoding: utf-8
Content-Length: 362
Connection: close

{"http_request_method": "POST", "exception": null, "error_message": "Invalid properties: ['login']", "property_names": ["login"], "_href": "/pulp/api/v2/users/", "http_status": 400, "error": {"code": "PLP0015", "data": {"property_names": ["login"], "properties": "['login']"}, "description": "Invalid properties: ['login']", "sub_errors": []}, "traceback": null}[ipanova@ina pulp]$ 
[ipanova@ina pulp]$ curl -i -H "Accept: application/json" -H "WebFrameworkSwitch: django" -X POST -k -u admin:admin -d '{}' "https://localhost/pulp/api/v2/users/"
HTTP/1.1 400 BAD REQUEST
Date: Wed, 18 Feb 2015 11:35:23 GMT
Server: Apache/2.4.10 (Fedora) OpenSSL/1.0.1e-fips mod_wsgi/3.5 Python/2.7.5
Content-Encoding: utf-8
Content-Length: 1608
Connection: close

{"http_request_method": "POST", "exception": ["MissingValue: Missing values for: ['login']\n"], "error_message": "Missing values for: ['login']", "_href": "/pulp/api/v2/users/", "http_status": 400, "error": {"code": "PLP0016", "data": {"property_names": ["login"], "properties": "['login']"}, "description": "Missing values for: ['login']", "sub_errors": []}, "missing_property_names": ["login"], "traceback": ["  File \"/usr/lib/python2.7/site-packages/django/core/handlers/base.py\", line 112, in get_response\n    response = wrapped_callback(request, *callback_args, **callback_kwargs)\n", "  File \"/usr/lib/python2.7/site-packages/django/views/generic/base.py\", line 69, in view\n    return self.dispatch(request, *args, **kwargs)\n", "  File \"/usr/lib/python2.7/site-packages/django/views/generic/base.py\", line 87, in dispatch\n    return handler(request, *args, **kwargs)\n", "  File \"/home/ipanova/pulp_development/pulp/server/pulp/server/webservices/controllers/decorators.py\", line 241, in _auth_decorator\n    return _verify_auth(self, operation, super_user_only, method, *args, **kwargs)\n", "  File \"/home/ipanova/pulp_development/pulp/server/pulp/server/webservices/controllers/decorators.py\", line 195, in _verify_auth\n    value = method(self, *args, **kwargs)\n", "  File \"/home/ipanova/pulp_development/pulp/server/pulp/server/webservices/views/util.py\", line 89, in wrapper\n    return func(*args, **kwargs)\n", "  File \"/home/ipanova/pulp_development/pulp/server/pulp/server/webservices/views/users.py\", line 61, in post\n    raise pulp_exceptions.MissingValue(['login'])\n"]}[ipanova@ina pulp]$

ipanova:
@asmacdo @bmbouter  after this commit the behavior in django changed and it is different from webpy:
POST /pulp/api/v2/users/ with missing required key 'login' http://fpaste.org/186541/18470114/ ( i changed the error from InvalidValue-->MissingValue since i think it is more proper in this case)

Now it looks like this : http://fpaste.org/186957/14242595/ basically it contains the traceback. is this the desired behavior?
Now every exception contains the traceback http://fpaste.org/186959/42425966/

Do we need that much information?

ipanova:
ok test

server/pulp/server/webservices/views/users.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Grant permissions
+        permission_manager = factory.permission_manager()
+        permission_manager.grant_automatic_permissions_for_resource(link['_href'])
+
+        response = generate_json_response_with_pulp_encoder(user)
+        return generate_redirect_response(response, link['_href'])
+
+
+class UserView(View):
```
asmacdo: Maybe change to UserResource? That has been the convention, and User vs Users is not visually obvious.

ipanova:
ok test

ipanova:
ok test

server/pulp/server/webservices/urls.py
```diff
@@ -85,5 +86,7 @@
     url(r'^v2/roles/(?P<role_id>[^/]+)/users/$', RoleUsersView.as_view(), name='role_users'),
     url(r'^v2/roles/(?P<role_id>[^/]+)/users/(?P<login>[^/]+)/$', RoleUserView.as_view(), name='role_user'),
     url(r'^v2/tasks/$', TaskCollectionView.as_view(), name='task_collection'),
-    url(r'^v2/tasks/(?P<task_id>[^/]+)/$', TaskResourceView.as_view(), name='task_resource')
+    url(r'^v2/tasks/(?P<task_id>[^/]+)/$', TaskResourceView.as_view(), name='task_resource'),
+    url(r'^v2/users/$', UsersView.as_view(), name='users'),
+    url(r'^v2/users/(?P<login>[^/]+)/$', UserResourceView.as_view(), name='user_resource')
```
asmacdo: no change necessary, but i think we should put commas even on the last line. This way will have smaller diffs when moving things around.

ipanova:
ok test

server/pulp/server/webservices/views/util.py
```diff
@@ -61,7 +82,8 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
 
         try:
-            request.body_as_json = json.loads(request.body)
+            request_json = json.loads(request.body)
+            request.body_as_json = _ensure_input_encoding(request_json)
```
asmacdo: Can you move this line to an `else`? We are trying to have only one line in a `try` so it is clear what caused the exception.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -5,7 +5,7 @@
 from django.http import HttpResponse, HttpResponseNotFound
 
 from pulp.server.webservices.controllers.base import json_encoder as pulp_json_encoder
-from pulp.server.exceptions import PulpCodedValidationException
+from pulp.server.exceptions import PulpCodedValidationException, InputEncodingError
```
asmacdo: Please put InputEncodingErrror first so it is alphabetical

asmacdo: Please put InputEncodingErrror first so it is alphabetical

ipanova:
ok test

server/pulp/server/webservices/views/util.py
rbarlow: -----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

On 02/18/2015 11:33 AM, Austin wrote:
> Can you move this line to an |else|? We are trying to have only one
> line in a |try| so it is clear what caused the exception.

I'm not sure I agree with being so strict about this. The point of the
try/Except block is to handle exceptional situations. If the handling
needs to be done the same for a group of lines that could all raise
the same Exception, the DRY principle says you should group them. For
example, suppose one(), two(), and three() might all raise a
ValueError, and suppose that if any one of them do this, I need to
handle it the same way. This proposed policy would mean I'd have to
write this:

```
try:
    one()
Except ValueError:
    remote_connection.close()
    _logger.info("Closed remote connection")
    local_file.close()
try:
    two()
Except ValueError:
    remote_connection.close()
    _logger.info("Closed remote connection")
    local_file.close()
try:
    three()
Except ValueError:
    remote_connection.close()
    _logger.info("Closed remote connection")
    local_file.close()
```

I think this is much cleaner, and you really don't lose anything that
is important:

```
try:
    one()
    two()
    three()
Except ValueError:
    remote_connection.close()
    _logger.info("Closed remote connection")
    local_file.close()
```

My point is not that we should always do this latter pattern, it's
that we shouldn't be so quick to make strict rules about it that don't
have any clear benefit. I think we should do what makes the most sense
in each individual situation.
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v1

iQIcBAEBAgAGBQJU5M2aAAoJEIyFaKUJtmpiarYP/RBMsfNTrnaje4v/5J/B7oeT
wl9H8tDSZ0ELGaNx3dZjiJhkUzPfCsX22/6AFNBk1n+U0WPebPKJ1/8+YfXehTrM
LuC8Vxn1Zm8rhT0puVmHh6UzvWP7l10ucca8F4g+wbm1yhQth0dIQkRue7ByOvE9
EvUyU3k8pmfaLGfESUSSIxnD8AjyR3mosev2DfKrJXcrW3LiS9AJD7+nO8GitRbS
hpDTfM++1hcODriHskUplkZ4GHwETha37EShCVv+oMvfeVW3LX1XaJaPa3HdewXB
FmQTYuhBhI+nSzek1vCvhOzhNSiNidYcpnlLmr1VQink3wHk9YWyAhVEwRqjYqas
juBGITZtzFxGp30cKVR3lOehrz1KEbClVisiAPCBM67ZmVqwOmfm2/FMPm4aLala
xLIS+Oq4vTSokaNSyV5YVvApn8A+mpO8/Q2l+PEZuaNzTtCS9tx2+2kC3S9QIOwm
Gz6iXIlyc07lYvFTA23Rj4jsQ+Kj+NGXkWEQ2T5nxjpGBjhz7xV4YWpUmV3hvm/d
ialFHYTpoYJ9Q1vOBXaU/EOHnUJ7C2FVBWQRh62BX0PVvXgHpOPZvCdg+imNe3ig
AIg2ZqmZPp2npgU02mwBUQOQ0/6AjFXigyx9RlNVZkJwI8x7lnkWASdkczsKvAth
iSZgWWZfDGwTZ1BdaUKM
=iKTz
-----END PGP SIGNATURE-----

bmbouter: I think the important thing is to not include statements in try blocks accidentally. If you want multiple statements to each be guarded for the same exception type being raised then I think that makes sense.

A bit unrelated, but the thing we should never write is

```
try:
    <thing>
except Exception:
    pass
```

Catching all Exception types and silencing them is a great way to cause really hard to find bugs. That is one rule that I think we should be strict on.

rbarlow: -----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

Even worse is catching an Exception, and then transforming it into a
PulpSomethingSomethingException and raising that, but trying to keep
the same traceback and other frame information. This happened to me
while trying to hunt down a very mysterious test failure last night,
and it was the worst. The Exeption that actually made it to the top
had no information to help me debug, because it was all hidden by that
bit that transformed it. Unfortunately, this pattern happens all over
our code from the old days ☹
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v1

iQIcBAEBAgAGBQJU5NpfAAoJEIyFaKUJtmpi8t4P/0ZKfemOuIkxSmqMz2hnvJrs
km0LOzMKassFu6NIosc161/DbaIU2tuRYEhuzR5wJh7H0N0/Qijnc80TjvLFpMZQ
pzg4+eneUsL3E5ahTPOP/SdlIOPxCU76IKtXLRPWynQ4oG3MRufuPBkvFYHUHoVY
CHQuTN07bwSWj/SBlaxhFbrhZOe8QKDf8XtuPVMP3JnHedC5d4M1HNyivMthMGNr
bCxfM7QnShXgbI/xat96GqOKcDj950hae3qRQnFhoo1kBzLtNCvF7QYHgFJhVPo7
6jVv1jzeXZsgzq2/C+pM3rhZrXeHlNs2GBIQfeLe0mIHCP+xq1kDAJbboxdLtrWD
cmlgplpPL+18DNmPQ4MCtCJSOHHOeRppmXz5QdZO2sltHJeHVhyd3rFnCb3UjDPw
wLNvmSUEziGGrzCVI40Xt5jb2nCExRU2IAYiOVJy63BPtghSPAJ9BxCy+EEy/+NI
YY1G5dRXqT5LgApwaxjcBFrHKS1UlZRDtLNeHdKEaKMuO71rMffkA3fgDyruno6K
rKU2iF8b5LbSsw2Ra5Gj+PcuDY3kYV+ghacolNdGKhzsasGF0r9OpdPOwi2/TB8c
DEZKzFTb2iaE1YqxYA9rDBJB7g6vmno6vEkH6lWAAeJiEIo0m+IjsPpA2hz+WYeX
q3gvq+f0pVe5VsUA5WTI
=91IQ
-----END PGP SIGNATURE-----

asmacdo: All the points made here are valid, but are not relevant to this line. `_ensure_input_encoding` can raise, but not a ValueError, so that line should be moved.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -165,3 +165,27 @@ def mock_function(self, request):
...
+class TestEnsureInputEncoding(unittest.TestCase):
+
+    def test_ensure_invalid_input_encoding(self):
+        """
+        Test invalid input encoding
+        """
+        input_data = {"invalid": "json\x81"}
+        try:
+            util._ensure_input_encoding(input_data)
+        except InputEncodingError:
```
asmacdo: Unless you want to save the exception in this line, you can use a `self.assertRaises` instead. The pattern that does not work in older versions of python is:

    with self.asserRaises(blah):
        do_stuff()

server/pulp/server/exceptions.py
```diff
@@ -422,7 +422,7 @@ class InputEncodingError(PulpDataException):
     """
 
     def __init__(self, value):
-        super(DuplicateResource, self).__init__(value)
+        super(InputEncodingError, self).__init__(value)
```
asmacdo: Whoa.

server/pulp/server/managers/auth/user/cud.py
```diff
@@ -111,35 +111,40 @@ def update_user(login, delta):
...
-                    if old_role not in delta['roles']:
+                    if old_role not in roles:
                         role_manager.remove_user_from_role(old_role, login)
-                user['roles'] = delta['roles']
+                user['roles'] = roles
 
         if invalid_values:
             raise InvalidValue(invalid_values)
+        if delta:
+            raise InvalidValue(delta.keys())
```
asmacdo: It looks like this introduces new functionality. Many of the endpoints did not enforce that extra values cannot be passed with webpy, why did you want to enforce that for this one?

server/pulp/server/webservices/views/users.py
```diff
@@ -0,0 +1,182 @@
...
+    @json_body_required
+    def post(self, request):
+        """
+        Create a new user.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+
+        :return: Response containing the user
+        :rtype: django.http.HttpResponse
```
asmacdo: Add the `raises`

server/pulp/server/webservices/views/users.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Retrieve a specific user.
+
+        :param request: WSGI request object
+        :type request: django.core.handlers.wsgi.WSGIRequest
+        :param login: login for the requested user
+        :type login: str
+
+        :return: Response containing the user
+        :rtype: django.http.HttpResponse
```
asmacdo: Add the `raises` here also. I might be mistaken here, but I think the idea is that we include it if we raise in in the view, and not if raised in a manager.

server/test/unit/server/webservices/views/test_users.py
```diff
@@ -0,0 +1,183 @@
...
+    @mock.patch('pulp.server.webservices.views.users.Permission.get_collection')
+    @mock.patch('pulp.server.webservices.views.users.generate_json_response')
+    @mock.patch('pulp.server.webservices.views.users.factory')
+    def test_delete_single_user(self, mock_factory, mock_resp, mock_perm):
+        """
+        Test user deletion.
+        """
+        mock_factory.user_manager.return_value.delete_user.return_value = None
+        mock_perm.return_value.find_one.return_value = 'some'
+        mock_perm.return_value.remove.return_value = None
```
asmacdo: I don't think this line is necessary.

server/pulp/server/webservices/views/users.py
```diff
@@ -0,0 +1,182 @@
...
+        delta = user_data.get('delta', None)
+
+        # Perform update
+        manager = factory.user_manager()
+        result = manager.update_user(login, delta)
+        process_dictionary_against_whitelist(result, USER_WHITELIST)
+        return generate_json_response_with_pulp_encoder(result)
+
+
+def add_link(user):
```
asmacdo: This block is getting coverage from the fact that it is being used black box style in the other tests, but I think this should also get its own unit test.

server/test/unit/server/webservices/views/test_util.py
ipanova: nice catch, thanks

server/pulp/server/webservices/views/users.py
ipanova: yes, you are completely right

server/pulp/server/managers/auth/user/cud.py
ipanova: we have in ours docs:
400 - if one or more of the parameters is invalid
in my understanding when i mistype a value(or write in the data something that is not specified in the docs, like some random field) i would be prompted about that.
example: "error_message": "Invalid properties: ['pasword']"

Webpy is very inconsistent in this functionality, but most of the code is checking the params.
For example here:  https://github.com/pulp/pulp/blob/master/server/pulp/server/webservices/controllers/consumer_groups.py#L51
or here:
https://github.com/pulp/pulp/blob/master/server/pulp/server/managers/consumer/group/cud.py#L99
If I will remove that line and accidentally have some typo in the data, I  will have something like:

`$ curl -i -H "Accept: application/json" -H "WebFrameworkSwitch: django" -X PUT -k -u admin:admin -d '{"delta":{"pasword":"xxx"}}' "https://localhost/pulp/api/v2/users/user3/"
HTTP/1.1 200 OK
Date: Thu, 19 Feb 2015 10:13:21 GMT
Server: Apache/2.4.10 (Fedora) OpenSSL/1.0.1e-fips mod_wsgi/3.5 Python/2.7.5
Content-Length: 109

{"_id": {"$oid": "54e4686245ef48558f24fd70"}, "name": "user3", "roles": [], "_ns": "users", "login": "user3"}[ipanova@ina pulp]$`

The user will be prompted OK and he will even not notice that he has a typo in the 'password' and actually it was not updated.
In addition to this, that line fixes the bug  https://bugzilla.redhat.com/show_bug.cgi?id=1193059

ipanova:
ok test

server/pulp/server/managers/auth/user/cud.py
asmacdo: My concern is that you could pass `{"user": "asmacdo", "password": "verysecure", "extra": "whoops"}` and that would now fail. I think that the change that you propose is reasonable, but I am not sure if it would be a bugfix or a new feature. @bmbouter?

ipanova: it is a bugfix, because previously you would not be able to make such situation were you would have 400 in the response( in users api)

asmacdo: Ok, makes sense. Since the docs say "400 - if one or more of the parameters is invalid", I suppose an extra field would count.

asmacdo:
LGTM

server/pulp/server/managers/auth/user/cud.py
ipanova: I think when you mistype some field it is counted as extra/unknown field,
that's why when you have *'pasword'* it will tell you Invalid properties: ['paswod']"

asmacdo: :+1:

ipanova:
ok test

ipanova:
ok test

asmacdo:
ok test

ipanova:
ok test

ipanova:
ok test

ipanova:
ok test

ipanova:
ok test

ipanova:
ok test

bmbouter:
ok test

ipanova:
ok test

PR #1648, opened 2015-02-16
BZ1182760 - Add proper SELinux rules for PuppetInstallDistributor

dkliban:
Some puppet modules have symlinks.  celery_t context needs to have
permissions to manage creation and removal of symlinks.


bmbouter:
@dkliban Are there supposed to be docs fixes with this?

bmbouter:
The BZ needs a target release set.

bmbouter:
Once the docs get updated it probably should be reviewed. I don't think this should be merged without docs changes since we'll forget what to write once we go onto other work. As far as the bugfix that looks right. I'm assigning back to you.

bmbouter:
LGTM

PR #1650, opened 2015-02-17
Move request.body_as_json from middleware to decorator

asmacdo:
Since the middleware runs on every endpoint, an empty or invalid body was defaulting to `{}`. When a body is required, an invalid or absent body should be raising a 400 and giving a helpful error. Instead it was running with a default value. This change moves this logic from the middleware to 2 decorators. `json_body_required` raises a 400 if the body does not contain valid JSON, `json_body_allow_empty` allows only an empty body or valid JSON. This does technically change the API because webpy was raising a 500 rather than a 400 in at least some of these cases, which I think represents a bug fix rather than a backwards incompatible change, particularly because it is listed this way in our docs.

Additionally, this creates a problem in the tests because the decorator cannot be mocked. The solution is to pass valid JSON through the decorator rather than setting the request.body_as_json directly. 

We will need to do some hand testing before merge to ensure that the JSON body is actually required for each endpoint. If not, the decorator can be switched to `json_body_allow_empty`. See the `DeleteOrphansActionView` to see how this should work.

If a request body is not passed, the value of request.body_as_json will be `None`. If this should be an empty list or dict, this assignment should take place in the View.

Additionally this will solve https://bugzilla.redhat.com/show_bug.cgi?id=1191417


asmacdo:
Altered endpoints:

`PUT /v2/consumer_groups/<consumer_group_id>/`
`POST /v2/consumer_groups/<consumer_group_id>/actions/associate/`
`POST '^v2/consumer_groups/<consumer_group_id>/actions/unassociate/`
`POST ^v2/consumer_groups/<consumer_group_id>/actions/content/<action>/`
`POST ^v2/consumer_groups/<consumer_group_id>/bindings/`
`POST v2/content/actions/delete_orphans/`
`PUT ^v2/content/units/<type_id>/<unit_id>/pulp_user_metadata/`
`POST v2/events/`
`PUT '^v2/events/<event_listener_id>/`
`POST '^v2/permissions/`
`POST v2/permissions/actions/revoke_from_user/`
`POST r'^v2/permissions/actions/grant_to_role/`
`POST '^v2/permissions/actions/revoke_from_role/`
`POST r'^v2/roles/`
`PUT '^v2/roles/<role_id>/`
`POST r'^v2/roles/<role_id>/users/'`

common/pulp/common/error_codes.py
```diff
@@ -96,3 +96,4 @@
                              " directory:  %(path)s"), ['path'])
 PLP1008 = Error("PLP1008", _("The importer type %(importer_type_id)s does not exist"),
                 ['importer_type_id'])
+PLP1009 = Error("PLP1009", _("Request data is not valid JSON"), [])
```
bmbouter: data is kind of an ambiguous term in HTTP land.

s/Request data is not valid JSON/The request body does not contain valid JSON/

server/pulp/server/webservices/middleware/parse_body.py
```diff
@@ -1,16 +0,0 @@
-import json
-
-
-class ParseBodyMiddleware(object):
```
bmbouter: yes!

server/pulp/server/webservices/settings.py
```diff
@@ -22,7 +22,6 @@
 MIDDLEWARE_CLASSES = (
     'django.middleware.http.ConditionalGetMiddleware',
     'pulp.server.webservices.middleware.exception.DjangoExceptionHandlerMiddleware',
-    'pulp.server.webservices.middleware.parse_body.ParseBodyMiddleware',
```
bmbouter: yes!

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -10,7 +10,8 @@
 from pulp.server.managers.consumer.group.cud import bind, unbind
 from pulp.server.webservices.controllers.decorators import auth_required
 from pulp.server.webservices.views.util import (generate_json_response,
-                                                generate_json_response_with_pulp_encoder)
+                                                generate_json_response_with_pulp_encoder,
+                                                must_have_json_body)
```
bmbouter: consider renaming this to `json_body_required`

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -56,6 +57,7 @@ def delete(self, request, consumer_group_id):
         result = manager.delete_consumer_group(consumer_group_id)
         return generate_json_response(result)
 
+    @must_have_json_body
```
bmbouter: Please reorder to make the `auth_required` statements always the top-most decorator. If a request isn't authorized I don't want to tell the user anything until they come back with proper authorization. The way this is written will have auth_required will have validation occur before authentication+authorization.

server/pulp/server/webservices/views/content.py
```diff
@@ -167,7 +170,7 @@ def post(self, request):
 
         :raises: OperationPostponed when an async operation is performed
         """
-        all_orphans = request.body_as_json
+        all_orphans = request.body_as_json or {}
```
bmbouter: Do you want this to be treated as an expression? Maybe you want this to act like a ternary operator, but this is going to return True/False based on the or statement. Is that what you want?

server/pulp/server/webservices/views/events.py
```diff
@@ -102,6 +104,7 @@ def delete(self, request, event_listener_id):
         return generate_json_response(None)
 
     @auth_required(authorization.UPDATE)
+    @must_have_json_body
```
bmbouter: This is the correct decorator ordering.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -69,3 +72,96 @@ def test_generate_redirect_response(self, mock_iri_to_uri):
...
         mock_iri_to_uri.assert_called_once_with(href)
+
+
+class TestMustHaveJSONBody(unittest.TestCase):
+
+    def test_must_have_json_body_valid(self):
+        mock_request = mock.MagicMock()
+        mock_request.body = '{"Valid": "JSON"}'
+
+        @must_have_json_body
```
bmbouter: I never considered using the decorator to decorate a function and then test it all within a test. neat.

server/pulp/server/webservices/views/consumer_groups.py
asmacdo: good catch!

server/pulp/server/webservices/views/content.py
asmacdo: This actually does work as a ternary assignment, and is considered idiomatic python I believe.

    In [1]: request = None

    In [2]: x = request or {}

    In [3]: x
    Out[3]: {}

A chain of `or`s will return the first truthy value, if none are truthy, it will return the last.

Likewise a chain of `and`s will return the last truthy value, if none are truthy it returns the first.

This works because the truthyness of the value stored in the variable will always match the boolean value of the conditional.

bmbouter: This makes sense, but it still looks like an expression. It's fine to leave as-is; you've shown it produces the correct result. I learned something. I'm more used to this style:  https://docs.python.org/2.6/reference/expressions.html#conditional-expressions

asmacdo: I have actually just realized that this conditional will have to be used for every instance where we allow an empty body, so I am going to change this do be the default in the util rather than in each use.

`s/request.body_as_json = None/request.body_as_json = {}/`

asmacdo:
@bmbouter, ready for rereview.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -10,7 +10,9 @@
 from pulp.server.managers.consumer.group.cud import bind, unbind
 from pulp.server.webservices.controllers.decorators import auth_required
 from pulp.server.webservices.views.util import (generate_json_response,
-                                                generate_json_response_with_pulp_encoder)
+                                                generate_json_response_with_pulp_encoder,
+                                                json_body_allow_empty,
```
bmbouter: Maybe s/json_body_allow_empty/json_body_required_allow_empty/   At face value it seems that it just allows empty with no requirement enforced.

server/pulp/server/webservices/views/util.py
```diff
@@ -46,3 +49,22 @@ def generate_redirect_response(response, href):
...
+            return func(*args, **kwargs)
+
+        try:
+            request.body_as_json = json.loads(request.body)
+        except ValueError:
+            raise PulpCodedValidationException(error_code=error_codes.PLP1009)
+        return func(*args, **kwargs)
+    return wrapper
+
+json_body_allow_empty = functools.partial(json_body_required, allow_empty=True)
```
bmbouter: great use of partial

server/pulp/server/webservices/views/util.py
```diff
@@ -46,3 +49,22 @@ def generate_redirect_response(response, href):
...
+    def wrapper(*args, **kwargs):
+        request = args[1]
+        if allow_empty and not request.body:
+            request.body_as_json = {}
+            return func(*args, **kwargs)
+
+        try:
+            request.body_as_json = json.loads(request.body)
+        except ValueError:
+            raise PulpCodedValidationException(error_code=error_codes.PLP1009)
```
bmbouter: It's strange to me that the coded validation exceptions aren't raised as: `raise PulpException1009`

bmbouter:
@asmacdo Do you think tests should be written that pass in invalid json and empty json and assert that the API does the right thing? I see the decorator code itself is tested which is good, but we don't assert that the right decorators are set on the right methods. Some invalid and empty tests could do that since they flow through the decorator. What do you think?

bmbouter:
After some IRC conversation, we can introduce those other tests some time in the future. We have never done those before so it would be a new thing versus the porting to Django that we are focused on. LGTM

PR #1651, opened 2015-02-17
Bump patch version in dev branch

barnabycourt:


beav:
is the `1a1` needed? [PEP-396](https://www.python.org/dev/peps/pep-0396/) just mentions x.y.z versions, maybe we can just do that to save some updates.

barnabycourt:
ok test

barnabycourt:
@beav This follows the StrictVersion from [PEP-0386](https://www.python.org/dev/peps/pep-0386/).  If we only specify x.y.z how do we know that something is from an alpha or beta build?

beav:
Ah, I was not able to find that PEP in my orignial search :koala:

This change LGTM personally but @rbarlow had comments in other PRs that were related to this one. Maybe we should discuss this after scrum?

PR #1652, opened 2015-02-17
[WIP] Moved content type loading to a migration

dkliban:
The developer docs still need to be updated, but I wanted to get some input on the implementation.


dkliban:
ok test

PR #1654, opened 2015-02-17
add traceback and exception to django exception handler

asmacdo:
These fields were included in webpy so we should include them in django as well.


bmbouter:
LGTM

PR #1655, opened 2015-02-18
Pep8 r tests

rbarlow:
The two commits in this PR are intentional, and I think it makes sense to keep them both despite our usual policy, as they are both atomic and they make separate, unrelated changes. The first moves the tests into the logical locations they should be in, and the second applies the PEP-8 style to them.

This PR took an unusual amount of time, as it caused a test failure due to changing the run time order of tests. Tracking that down was not an enjoyable exercise, but it turned out to be due to the use of a non-standard mocking style that caused other tests not to be using the objects they thought they were using. I've fixed one place that causes this, but there are others.


rbarlow:
@pulpbot, why you gotta be a weirdo? You said it passed and then it failed, and the details link is a 404? Whaaaaaaa? ☺

rbarlow:
ok test

asmacdo:
LGTM

asmacdo:
@rbarlow, btw, thanks for the commit structure, made review very simple.

PR #1656, opened 2015-02-18
BZ1193689 - Removed getting of working directories where they are not used

dkliban:
https://bugzilla.redhat.com/show_bug.cgi?id=1193689


asmacdo:
LGTM

dkliban:
ok test

asmacdo:
@dkliban, the name of this PR should be changed.

Do you have a way to confirm that the working directory is not being used in each of these cases?

asmacdo:
Since these are not allowed anyway, I think we are good to merge this. LGTM

dkliban:
ok test

PR #1659, opened 2015-02-18
Pep8 s tests

rbarlow:
This PR has two commits on purpose. They are related, but are also separate and make sense to be two commits in master. The first commit moves some test modules to the locations they ought to be, and the second makes sure they all pass flake8.


server/test/unit/server/managers/consumer/test_profile.py
```diff
@@ -284,7 +266,7 @@ def test_consumer_deleted(self):
         profiles = list(cursor)
         self.assertEquals(len(profiles), 0)
 
-    @patch('pulp.server.managers.factory.consumer_agent_manager')
+    @mock.patch('pulp.server.managers.factory.consumer_agent_manager')
```
asmacdo: Thanks!

PR #1660, opened 2015-02-18
Set the PR remote to "upstream" instead of "origin".

rbarlow:
The pull request remote was accidentally set to origin instead of upstream in the dev-setup.sh
script. This PR fixes the PRs to use the same remote as the other upstream branches.


asmacdo:
I noticed this earlier. Thanks for fixing!

asmacdo:
LGTM

PR #1661, opened 2015-02-18
Removed dropping of indeces in pulp-manage-db

dkliban:


dkliban:
ok test

docs/user-guide/tuning.rst
```diff
@@ -176,6 +176,12 @@ dedicated hardware for larger production deployments. At this time, Pulp can be
 used with `replication <http://docs.mongodb.org/manual/replication/>`_ but does
 not support sharding.
 
+If searches for content are performing poorly, performance could be improved by adding an index
```
rbarlow: s/could/may/

rbarlow:
I believe it was agreed that we did need to drop the indices and recreate them one last time in a migration. There has been some work that drops indices in the past, and users who upgrade from an old Pulp to 2.7 will miss these index removals (thus the need for one last removal). So I think we need a migration that drops and adds one last time.

Let me know when you've done that and I can re-review.

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

beav:
This appears to address @rbarlow 's comment and looks good overall.

LGTM

PR #1662, opened 2015-02-19
Add install_requires arguments to our setup.py files.

rbarlow:
This commit also removes some incorrect Requires lines in our spec file, and adds
python-gofer-qpid to our server qpid group.

https://pulp.plan.io/issues/278

re #278

This is necessary to satisfy Travis in https://github.com/pulp/pulp_python/pull/20

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -305,11 +305,9 @@ Requires: python-pymongo >= 2.5.2
 Requires: python-mongoengine >= 0.7.10
 Requires: python-setuptools
 Requires: python-webpy
-Requires: python-okaara >= 1.0.32
 Requires: python-oauth2 >= 1.5.211
 Requires: python-httplib2
 Requires: python-isodate >= 0.5.0-1.pulp
-Requires: python-BeautifulSoup
 Requires: python-qpid
 Requires: python-nectar >= 1.1.6
 Requires: httpd
@@ -318,7 +316,6 @@ Requires: openssl
 Requires: nss-tools
 Requires: python-ldap
 Requires: python-gofer >= %{gofer_version}
-Requires: python-gofer-qpid >= %{gofer_version}
 Requires: crontabs
 Requires: acl
 Requires: mod_wsgi >= 3.4-1.pulp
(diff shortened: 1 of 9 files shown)


rbarlow:
I can confirm that the current state of this PR did pass a test Travis build on the Python plugins.

pulp.spec
```diff
@@ -305,11 +305,9 @@ Requires: python-pymongo >= 2.5.2
 Requires: python-mongoengine >= 0.7.10
 Requires: python-setuptools
 Requires: python-webpy
-Requires: python-okaara >= 1.0.32
 Requires: python-oauth2 >= 1.5.211
 Requires: python-httplib2
 Requires: python-isodate >= 0.5.0-1.pulp
-Requires: python-BeautifulSoup
```
asmacdo: I wondered why that was in there.

server/setup.py
```diff
@@ -13,5 +13,8 @@
             '../libexec/pulp-manage-workers = pulp.server.async.manage_workers:main',
             'pulp-manage-db = pulp.server.db.manage:main',
         ]
-    }
+    },
+    install_requires=[
+        'celery >=3.1.0, <3.2.0', 'httplib2', 'iniparse', 'isodate>=0.5.0', 'm2crypto',
+        'mongoengine>=0.7.10', 'oauth2>=1.5.211', 'pymongo>=2.5.2', 'setuptools', 'web.py']
```
asmacdo: When you merge this forward, should you add Django here?

asmacdo:
LGTM, but I think someone with more experience with builds should also take a look.

server/setup.py
rbarlow: On 02/20/2015 05:24 PM, Austin wrote:
> When you merge this forward, should you add Django here?

Yep!

bmbouter: @rbarlow  How is the inclusion of Django here related to this bug?

https://bugzilla.redhat.com/show_bug.cgi?id=1195254

rbarlow: On 02/23/2015 10:39 AM, bmbouter wrote:
> @rbarlow <https://github.com/rbarlow> How is the inclusion of Django
> here related to this bug?
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=1195254

Adding Django here will cause Django to get installed via pip by the
dev-setup.sh script, but once we fix that bug it will get installed via
yum instead.

We should aim for nothing but the repos themselves getting installed by pip.

beav:
this LGTM as well

rbarlow:
On 02/24/2015 11:43 AM, Chris Duryee wrote:
> this LGTM as well

Thanks for the review! I updated the Python pull request to use only
Pulp master now. Hopefully Travis will think everything is cool now.

PR #1663, opened 2015-02-19
Pep8 t tests

rbarlow:


server/test/unit/plugins/types/test_parser.py
```diff
@@ -1,25 +1,11 @@
...
-
 import unittest
 
 from pulp.plugins.types import parser, model
 
-# -- test data ---------------------------------------------------------------
 
-VALID_DESCRIPTOR_1 = model.TypeDescriptor('valid_descriptor_1',
-"""{"types": [
+VALID_DESCRIPTOR_1 = model.TypeDescriptor(
```
asmacdo: Doesnt fail flake8, but the indention of this block doesn't seem right.

PR #1665, opened 2015-02-19
Fix a failing PEP-8 test.

rbarlow:
I noticed after merging one of my PEP-8 PRs that due to other changes in master, my merge introduced a failure. I usually run the tests before pushing merges, but this time I forgot, and it bit me!


barnabycourt:
LGTM

PR #1669, opened 2015-02-20
Gofer 2.5.3

jortel:
Fixes an issue that is exacerbated by SSL reconfiguration.  Basically, goferd gets tied up with reconnect issues when responding to requests that it cannot route to a plugin.


barnabycourt:
LGTM as long as the PR builder passes

PR #1670, opened 2015-02-20
django conversion - repo_groups urls

asmacdo:
Convert repo groups urls.

Hand testing, all identical where expected:
Includes a fix for [BZ-1194421](https://bugzilla.redhat.com/show_bug.cgi?id=1194421)

`GET /v2/repo_groups/`
http://fpaste.org/185364/38587891/
`POST /v2/repo_groups/`
valid: http://fpaste.org/185372/59647142/
dupe: http://fpaste.org/185376/59826142/
invalid: http://fpaste.org/185377/85992414/

`GET /v2/repo_groups/<repo_group_id>/`
valid: http://fpaste.org/185379/86021314/
invalid: http://fpaste.org/185381/38602891/

`DELETE /v2/repo_groups/<repo_group_id>/`
valid: http://fpaste.org/185383/60646142/
invalid: http://fpaste.org/185386/23860786/

`PUT /v2/repo_groups/<repo_group_id>/`
valid id: http://fpaste.org/185390/38613831/
invalid id: http://fpaste.org/185392/23861529/
invalid params: http://fpaste.org/185393/61678142/

`POST /v2/repo_groups/<repo_group_id>/actions/associate/`
http://fpaste.org/185407/38634491/
`POST /v2/repo_groups/<repo_group_id>/actions/unassociate/`
http://fpaste.org/185409/42386356/
Test function
http://fpaste.org/185406/23863285/

`GET /v2/repo_grous/<repo_group_id>/distributors/`
This one includes a bugfix for [BZ-1194421](https://bugzilla.redhat.com/show_bug.cgi?id=1194421):
http://fpaste.org/188365/63810142/
`POST /v2/repo_grous/<repo_group_id>/distributors/`
http://fpaste.org/187792/43762811/

`GET /v2/repo_grous/<repo_group_id>/distributors/<distributor_id>/`
http://fpaste.org/187798/76869142/
`GET /v2/repo_grous/<repo_group_id>/distributors/<distributor_id>/` with fake id
http://fpaste.org/187802/14243770/

`PUT /v2/repo_grous/<repo_group_id>/distributors/<distributor_id>/`
Valid: http://fpaste.org/187820/14243794/
Invalid body: http://fpaste.org/187848/38209114/
Invalid repo group id or dist id: http://fpaste.org/188373/64561142/

`DELETE /v2/repo_grous/<repo_group_id>/distributors/<distributor_id>/`
http://fpaste.org/188224/44922014/

`POST /v2/repo_groups/<repo_group_id>/actions/publish/`
http://fpaste.org/188339/44601801/


server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        collection = RepoGroupModel.get_collection()
+        cursor = collection.find({})
+        groups = []
+        for group in cursor:
+            group['_href'] = reverse('repo_group_resource', kwargs={'repo_group_id': group['id']})
+            groups.append(group)
+        return generate_json_response_with_pulp_encoder(groups)
+
+    @auth_required(authorization.CREATE)
+    @json_body_allow_empty
```
ipanova: I think in this case it should require the body, since for repo group creation you need at least the 'id' which is the required key

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+
+        :return: Response containing serialized dict of the specified group
+        :rtype: django.http.HttpResponse
+        :raises pulp_exceptions.MissingResource: if repo_group_id is not found
+        """
+        collection = RepoGroupModel.get_collection()
+        group = collection.find_one({'id': repo_group_id})
+        if group is None:
+            raise pulp_exceptions.MissingResource(repo_group=repo_group_id)
+        group['_href'] = request.get_full_path()
```
ipanova: Why do you use here the get_full_path() and not the reverse?

ipanova: Why do you use her the get_full_path() and not the reverse?

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+    def delete(self, request, repo_group_id):
+        """
+        Delete the specified repo group.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_group_id: id of repo group to delete
+        :type  repo_group_id: str
+
+        :return: Response containing None
```
ipanova: Maybe 'An empty response' will be better?

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        Get information about a single repo group distributor association.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_group_id: repo group the distributor is associated with
+        :type  repo_group_id: str
+        :param distributor_id: distributor to get
+        :type  distributor_id: str
+
+        :return: response containing a serialized dict of the added distributor
```
ipanova: response containing a serialized dict of requested distributor

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        :rtype: django.http.HttpResponse
+        """
+        distributor_manager = managers_factory.repo_group_distributor_manager()
+        dist = distributor_manager.get_distributor(repo_group_id, distributor_id)
+        dist['_href'] = request.get_full_path()
+        return generate_json_response_with_pulp_encoder(dist)
+
+    @auth_required(authorization.DELETE)
+    @json_body_allow_empty
```
ipanova: not needed here

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        Disassociate a distributor from a repository group.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_group_id: repo group the distributor is associated with
+        :type  repo_group_id: str
+        :param distributor_id: distributor to disassociate
+        :type  distributor_id: str
+
+        :return: response containing None
```
ipanova: 'An empty response'

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        :rtype: django.http.HttpResponse
+        """
+        params = request.body_as_json
+        force = params.get('force', False)
```
ipanova: do we support the 'force' option when deleting the distributor from repo_group?

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        return generate_json_response_with_pulp_encoder(result)
+
+
+class RepoGroupPublishView(View):
+    """
+    View to trigger the publish of a repo group.
+    """
+
+    @auth_required(authorization.EXECUTE)
+    @json_body_allow_empty
```
ipanova: I think here it should require the body since in the data we should specify the required key 'id' of distributor

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        Dispatch a task to publish content from the repo group using the distributor specified by
+        the params.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_group_id: repo group to publish
+        :type  repo_group_id: str
+
+        :raises OperationPosponed: dispatch a task
```
ipanova: add raises MissingValue

server/test/unit/server/webservices/views/test_repo_groups.py
```diff
@@ -0,0 +1,506 @@
...
+
+class TestRepoGroupsView(unittest.TestCase):
+    """
+    Tests for RepoGroupsView.
+    """
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repo_groups.reverse')
+    @mock.patch('pulp.server.webservices.views.repo_groups.RepoGroupModel')
```
ipanova: I think you can mock directly here 'pulp.server.webservices.views.repo_groups.RepoGroupModel.get_collection'

server/pulp/server/webservices/views/repo_groups.py
asmacdo: You are right, it isn't mentioned in the docs, but the manager does take a force parameter and has some logic based it.

server/pulp/server/webservices/views/repo_groups.py
asmacdo: Even if we don't need to enforce anything about the json, this decorator is required to make the `request.body_as_json` field available.

server/pulp/server/webservices/views/repo_groups.py
asmacdo: In order to maintain the current API, I let an empty body pass through the decorator so it can be handled by the view. In this case, it will be caught by the logic of the view and will raise a `MissingValue` with a 400. Using `json_body_required` would also return a 400, but with a different error.

server/pulp/server/webservices/views/repo_groups.py
asmacdo: Here, I want to return the same path that was just called. This can be rebuilt using `reverse()`, but that introduces possibility for error, particularly if the code changes at some time in the future. I figure that this is just simpler.

server/pulp/server/webservices/views/repo_groups.py
ipanova: i think that this is just deprecated https://github.com/pulp/pulp/blob/master/server/pulp/server/webservices/controllers/repo_groups.py#L195
as `force = params.get('force', False)` implies the fact that some data was transferred in the request 

$ curl -i -H "Accept: application/json" -H "WebFrameworkSwitch: webpy" -X DELETE -k -u admin:admin -d '{"force":"not-sure-what-it-is"}' "https://localhost/pulp/api/v2/repo_groups/group1/distributors/85bdfe00-df97-41ca-a1c2-f053e71090d8/"
HTTP/1.1 200 OK
Date: Mon, 23 Feb 2015 15:00:59 GMT
Server: Apache/2.4.10 (Fedora) OpenSSL/1.0.1e-fips mod_wsgi/3.5 Python/2.7.5
Content-Length: 4

null[ipanova@ina pulp]$ 

I think we just have to hardcode  the force parameter in https://github.com/pulp/pulp/blob/master/server/pulp/server/webservices/controllers/repo_groups.py#L198

ipanova: or actually just skip it ,as by default it has False value https://github.com/pulp/pulp/blob/master/server/pulp/server/managers/repo/group/distributor.py#L188

asmacdo: By allowing empty, a body is not necessary and will follow the default path of `force=false`. You could also send a DELETE request with the body `{"force": true}`.

server/pulp/server/webservices/views/repo_groups.py
asmacdo: Same as https://github.com/pulp/pulp/pull/1670#discussion_r25169256.

server/pulp/server/webservices/views/repo_groups.py
ipanova: I do not have a strong opinion here but 1) if we started to use reverse() let's just use it through all the code. 2) if we will introduce in the future query parameters then we would have to change get_full_path back to reverse https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpRequest.get_full_path

it is up to you to decide what you fancy more

server/pulp/server/webservices/views/repo_groups.py
```diff
@@ -0,0 +1,378 @@
...
+
+        :return: Response containing a serialized list of dicts, one for each repo group
+        :rtype: django.http.HttpResponse
+        """
+
+        collection = RepoGroupModel.get_collection()
+        cursor = collection.find({})
+        groups = []
+        for group in cursor:
+            group['_href'] = reverse('repo_group_resource', kwargs={'repo_group_id': group['id']})
```
ipanova: by the way, since reverse of 'repo_group_resource' is repeated 4 times you can do something like here https://github.com/pulp/pulp/commit/c30f0962563dbed33eabf3523f0ba2930658f341#diff-113ebb31dc6faa5d855700dc67532dddR162

same with 'repo_group_distributor_resource'

server/test/unit/server/webservices/views/test_repo_groups.py
```diff
@@ -0,0 +1,505 @@
...
+    """
+    Tests for RepoGroupsView.
+    """
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repo_groups.reverse')
+    @mock.patch('pulp.server.webservices.views.repo_groups.RepoGroupModel.get_collection')
+    @mock.patch('pulp.server.webservices.views.repo_groups' +
+                '.generate_json_response_with_pulp_encoder')
```
ipanova: From my point of view this way more readable:
@mock.patch(
        'pulp.server.webservices.views.repo_groups.generate_json_response_with_pulp_encoder')
 def test_get_repo_groups(self, mock_resp, mock_collection, mock_reverse):

server/test/unit/server/webservices/views/test_repo_groups.py
```diff
@@ -0,0 +1,505 @@
...
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repo_groups' +
+                '.generate_json_response_with_pulp_encoder')
+    @mock.patch('pulp.server.webservices.views.repo_groups.managers_factory')
+    def test_get_repo_group_distributor(self, mock_factory, mock_resp):
+        """
+        Get a single repo_groups distributor.
+        """
+        mock_request = mock.MagicMock()
+        mock_request.get_full_path.return_value = '/mock/path/'
```
ipanova: update tests with reverse, since there is no get_full_path anymore

server/pulp/server/webservices/views/repo_groups.py
asmacdo: Makes sense but it is replacing a one line call that is unlikely to change. I would prefer to leave as since it will be more transparent.

ipanova:
LGTM.

PR #1677, opened 2015-02-25
Prevent nodes files from showing up in the pulp-server rpm

barnabycourt:


beav:
Is there a redmine issue for this? I am not familiar with the problem it is fixing.

barnabycourt:
@beav There is no open issue for it that I know of.  It was brought up in conversation.

beav:
I see, it removes these files:

```bash
$ rpm -ql pulp-server | grep nodes
/usr/lib/pulp/plugins/types/nodes.json
/var/lib/pulp/nodes
/var/lib/pulp/nodes/published
/var/lib/pulp/nodes/published/http
/var/lib/pulp/nodes/published/https
```

LGTM

PR #1678, opened 2015-02-26
PEP-8 pulp.agent.

rbarlow:


devel/pulp/devel/mock_handlers.py
```diff
@@ -241,7 +241,6 @@ def deploy(self):
         sys.path.insert(0, self.SITE_PACKAGES)
         for handler in (RPM, SRPM, SECTION_MISSING, CLASS_NDEF, NO_MODULE):
             self.__deploy(handler)
-        print 'deployed'
```
asmacdo: Good change, but the rest of this file has PEP-8 issues. Why this line?

asmacdo:
LGTM

devel/pulp/devel/mock_handlers.py
rbarlow: On 03/02/2015 05:46 PM, Austin wrote:
>> -        print 'deployed'
> 
> Good change, but the rest of this file has PEP-8 issues. Why this line?

This line prints obnoxious messages during test runs, which made
debugging some tests difficult.

PR #1681, opened 2015-02-27
Add blinker dependency to setup.py

beav:
Pulp_python unit tests were emitting error messages due to this dep being
missing. Blinker is needed for mongoengine signals to work.


bmbouter:
LGTM

PR #1683, opened 2015-03-02
668 - 500 error is raised when attempting to create consumer group with already existing id

ipanova:
closes [668] (https://pulp.plan.io/issues/668)

### DIFF
--- a/server/pulp/server/managers/consumer/group/cud.py
+++ b/server/pulp/server/managers/consumer/group/cud.py
@@ -1,18 +1,6 @@
-# -*- coding: utf-8 -*-
-#
-# Copyright © 2012 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public License as
-# published by the Free Software Foundation; either version 2 of the License
-# (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied, including the
-# implied warranties of MERCHANTABILITY, NON-INFRINGEMENT, or FITNESS FOR A
-# PARTICULAR PURPOSE.
-# You should have received a copy of GPLv2 along with this software; if not,
-# see http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
-
 import logging
 import re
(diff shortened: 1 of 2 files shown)


server/pulp/server/managers/consumer/group/cud.py
```diff
@@ -78,9 +66,7 @@ def create_consumer_group(group_id, display_name=None, description=None, consume
         try:
             collection.insert(consumer_group, safe=True)
         except DuplicateKeyError:
-            raise pulp_exceptions.PulpCodedValidationException(
-                [PulpCodedException(error_codes.PLP1004, type=ConsumerGroup.collection_name,
-                                    object_id=group_id)])
+            raise pulp_exceptions.DuplicateResource(group_id), None, sys.exc_info()[2]
```
asmacdo: @barnabycourt, it looks like this was the only place in the code base that PLP1004 was used. Is it ok to remove that code?

asmacdo:
@ipanova, with the redmine switch, I think the commit message should be a `include closes 885`. http://pulp.readthedocs.org/en/latest/dev-guide/contributing/merging.html#pull-requests

asmacdo:
LGTM

server/pulp/server/managers/consumer/group/cud.py
ipanova: the problem itself was in the `except` line, it would look like this http://fpaste.org/192664/42538499/ but i decided to go with the 2nd option since it is a pattern in pulp

asmacdo: @ipanova, sorry I was unclear. I think this is a good change, I meant that if you remove this, maybe we should remove this as well:
https://github.com/pulp/pulp/blob/25ae34a7a5db783eb3cac6d10e9080ba813fab6c/common/pulp/common/error_codes.py#L92

ipanova: ah yeah, good catch

PR #1685, opened 2015-03-02
Fix error during pulp-selinux uninstall

beav:
The `uninstall.sh` script was removing two `.if` files which were also
controlled by RPM. These should only be cleaned up by RPM itself.

re #554


ipanova:
LGTM

PR #1687, opened 2015-03-03
gracefully handle no dev log for pulp-manage-db, closes 639

asmacdo:
If `/dev/log` doesn't exist, don't try to use it. Instead, print a helpful message and exit.

Closes: https://pulp.plan.io/issues/639


ipanova:
LGTM

PR #1690, opened 2015-03-03
Added SELinux rules to restrict pid and tmp file management

dkliban:
This patch also adds back some lines of pulp.spec file that were
accidently removed in
https://github.com/pulp/pulp/commit/f4d2de05d58f71e9baccd18e0ce8020c3e80d6fb

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -174,12 +174,12 @@ mkdir -p %{buildroot}/%{_sysconfdir}/default/
 mkdir -p %{buildroot}/%{_sysconfdir}/httpd/conf.d/
 mkdir -p %{buildroot}/%{_usr}/lib/%{name}/plugins
 mkdir -p %{buildroot}/%{_usr}/lib/%{name}/plugins/types
-mkdir -p %{buildroot}/%{_var}/lib/%{name}/celery
 mkdir -p %{buildroot}/%{_var}/lib/%{name}/uploads
 mkdir -p %{buildroot}/%{_var}/lib/%{name}/published
 mkdir -p %{buildroot}/%{_var}/lib/%{name}/static
 mkdir -p %{buildroot}/%{_var}/www
 mkdir -p %{buildroot}/%{_var}/cache/%{name}
+mkdir -p %{buildroot}/%{_var}/run/%{name}
 # These directories are used for Nodes
 mkdir -p %{buildroot}/%{_var}/lib/%{name}/nodes/published/http
 mkdir -p %{buildroot}/%{_var}/lib/%{name}/nodes/published/https
@@ -384,13 +384,13 @@ Pulp provides replication, access, and accounting for software repositories.
 # - apache:apache
 %defattr(-,apache,apache,-)
 %dir %{_var}/lib/%{name}
-%{_var}/lib/%{name}/celery
 %{_var}/lib/%{name}/published
 %{_var}/lib/%{name}/static
 %{_var}/lib/%{name}/uploads
 %dir %{_var}/log/%{name}
 %{_var}/www/pub
 %{_var}/cache/%{name}/
+%{_var}/run/%{name}/
 # Install the docs
 %defattr(-,root,root,-)
 %doc README LICENSE COPYRIGHT
--- a/server/etc/default/upstart_pulp_celerybeat
+++ b/server/etc/default/upstart_pulp_celerybeat
@@ -9,7 +9,7 @@ CELERYBEAT_LOG_LEVEL="INFO"
 # CELERYBEAT_PID_FILE="/var/run/pulp/celerybeat.pid"
 
 # The user that Celerybeat runs as must be able to write to this folder
-CELERYBEAT_OPTS="--workdir=/var/lib/pulp/celery/"
+CELERYBEAT_OPTS="--workdir=/var/run/pulp/"
 
 # Configure Python's encoding for writing all logs, stdout and stderr
 PYTHONIOENCODING="UTF-8"
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -8,12 +8,19 @@ init_daemon_domain(celery_t, celery_exec_t)
 type pulp_var_cache_t;
 files_type(pulp_var_cache_t)
 
+type pulp_var_run_t;
+files_pid_file(pulp_var_run_t)
+
+type pulp_tmp_t;
+files_tmp_file(pulp_tmp_t)
+
 require {
         type celery_t;
+        type pulp_var_run_t;
+        type pulp_tmp_t;
         type pulp_var_cache_t;
         type pulp_cert_t;
         type puppet_etc_t;
-        type var_run_t;
         class dir { getattr search write remove_name create add_name rmdir };
         class file { lock rename write setattr getattr read create unlink open };
         class process { setsched signal signull };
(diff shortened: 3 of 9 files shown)


lzap:
For the record, I think you compile the policies twice in the build section. I noticed this the other day. Minor issue.

lzap:
There is a loop or something, `make compile` compiles all policies.

pulp.spec
```diff
@@ -390,6 +390,8 @@ Pulp provides replication, access, and accounting for software repositories.
 %{_var}/lib/%{name}/uploads
 %dir %{_var}/log/%{name}
 %{_var}/www/pub
+%{_var}/cache/%{name}/
```
bmbouter: Do we want to chown this to apache? It seems right, but what does Pulp write into this directory?

server/etc/default/upstart_pulp_celerybeat
```diff
@@ -9,7 +9,7 @@ CELERYBEAT_LOG_LEVEL="INFO"
 # CELERYBEAT_PID_FILE="/var/run/pulp/celerybeat.pid"
 
 # The user that Celerybeat runs as must be able to write to this folder
-CELERYBEAT_OPTS="--workdir=/var/lib/pulp/celery/"
+CELERYBEAT_OPTS="--workdir=/var/run/pulp/"
```
bmbouter: yes!

server/selinux/server/pulp-celery.te
```diff
@@ -74,7 +90,7 @@ manage_files_pattern(celery_t, pulp_var_cache_t, pulp_var_cache_t)
 
 ######################################
 
-allow celery_t var_run_t:file { write getattr read create unlink open };
+allow celery_t celery_var_run_t:file { write getattr read create unlink open };
```
bmbouter: Isn't this statement redundant with L#57? I think the only one not picked up by L#57 is `unlink` but pid files don't use symlinks so it shouldn't need unlink. I think this line is safe to remove, but it would need to be retested.

server/selinux/server/pulp-celery.te
```diff
@@ -91,7 +106,6 @@ logging_send_syslog_msg(celery_t)
 miscfiles_read_localization(celery_t)
 miscfiles_manage_cert_dirs(celery_t)
 sysnet_read_config(celery_t)
-
```
bmbouter: I think this whitespace was useful. It also is consistent with the rest of the policy to have a line of whitespace after a policy block.

server/selinux/server/pulp-celery.te
```diff
@@ -40,8 +47,17 @@ allow celery_t self:unix_dgram_socket { create connect };
 #
 
 allow celery_t self:process { setsched };
-allow celery_t tmp_t:dir { write remove_name create add_name rmdir setattr };
-allow celery_t tmp_t:file { rename create unlink setattr write read getattr open lock };
+
+allow celery_t celery_tmp_t:dir { write remove_name create add_name rmdir setattr };
+files_tmp_filetrans(celery_t, celery_tmp_t, file)
+
+allow celery_t celery_tmp_t:file { rename create unlink setattr write read getattr open lock };
```
bmbouter: I think this statement should be:

`allow celery_t celery_tmp_t:file manage_file_perms;`

I looked up what that macro expands into in `/usr/share/selinux/devel/include/support/obj_perm_sets.spt` and it says: { create open getattr setattr read write append rename link unlink ioctl lock }

This expands the permissions some (append, ioctl, link), but for tmp files (celery_tmp_t) context I this is the right thing to do. This needs to be checked that it compiles on FC20, EL6, and EL7.

server/selinux/server/pulp-celery.te
```diff
@@ -40,8 +47,17 @@ allow celery_t self:unix_dgram_socket { create connect };
 #
 
 allow celery_t self:process { setsched };
-allow celery_t tmp_t:dir { write remove_name create add_name rmdir setattr };
-allow celery_t tmp_t:file { rename create unlink setattr write read getattr open lock };
+
+allow celery_t celery_tmp_t:dir { write remove_name create add_name rmdir setattr };
```
bmbouter: I think this statement should be:

`allow celery_t celery_tmp_t:dir manage_dir_perms;`

I looked up what that macro expands into in `/usr/share/selinux/devel/include/support/obj_perm_sets.spt` and it says you'll get:

`{ create open getattr setattr read write link unlink rename search add_name remove_name reparent rmdir lock ioctl }`

You'll get the obj_perm_sets.spt file if you install `selinux-policy-devel`

This expands the permissions by adding `{ open getattr read link unlink rename search reparent lock ioctl }`, but for the tmp dirs context (`celery_tmp_t`) I think this is the right thing to do. This needs to be checked that it compiles on FC20, EL6, and EL7 to esnure this macro is defined in those versions of `selinux-policy-devel`.

server/selinux/server/pulp-celery.te
```diff
@@ -8,12 +8,19 @@ init_daemon_domain(celery_t, celery_exec_t)
 type pulp_var_cache_t;
 files_type(pulp_var_cache_t)
 
+type celery_var_run_t;
```
bmbouter: I think `celery_var_run_t` should be called `pulp_var_run_t`. After looking at a system with SELinux enabled the convention seems to be to have the software itself (not its dependency) carry the context name. Even the `celery_t` context should probably be named `pulp_t` or `pulp_celery_t` but that is a [different issue](https://pulp.plan.io/issues/563).

This should be renamed to `pulp_var_run_t`.

server/selinux/server/pulp-celery.te
```diff
@@ -8,12 +8,19 @@ init_daemon_domain(celery_t, celery_exec_t)
...
+files_pid_file(celery_var_run_t)
+
+type celery_tmp_t;
```
bmbouter: Similar here as with L#11. I think this should be `pulp_tmp_t`.

pulp.spec
bmbouter: This looks right, but what are the contents of this directory? I looked on an installed Pulp system, and this folder didn't even exist.

bmbouter:
Nice job [writing some tasks](https://pulp.plan.io/issues/594#note-3) that QE can use to help verify. I think they are missing a step that says that a worker needs to be running after the change to `tasks/__init__.py`

bmbouter:
@dkliban I'm assigning back to you. I think this needs some adjustments. They aren't too big, but I do think they are important.

dkliban:
I also updated the gist.  It now includes instructions to start or restart workers after editing the file.

dkliban:
https://gist.github.com/dkliban/18989c57bae02a3d466d

server/selinux/server/pulp-celery.te
```diff
@@ -40,8 +47,17 @@ allow celery_t self:unix_dgram_socket { create connect };
 #
 
 allow celery_t self:process { setsched };
-allow celery_t tmp_t:dir { write remove_name create add_name rmdir setattr };
-allow celery_t tmp_t:file { rename create unlink setattr write read getattr open lock };
+
+allow celery_t pulp_tmp_t:dir manage_dir_perms;
+files_tmp_filetrans(celery_t, pulp_tmp_t, file)
```
bmbouter: This and the whitespace line below can be removed. It is redundant with L#55

bmbouter: This and L#55 should be swapped so the file and dirs are grouped together

bmbouter:
One small line swap fix, and then go ahead and merge. Assigning back to you.

PR #1691, opened 2015-03-03
Upstart use stop/start instead of restart_workers.

rbarlow:
This commit updates our Upstart init scripts to call stop_workers and
then start_workers, instead of restart_workers(). There seems to be an
issue with Celery multi, restarting, and Pulp. This commit works around
that issue by fully stopping the workers instead of issuing a restart to
them.

https://pulp.plan.io/issues/686

closes #686


beav:
LGTM

PR #1692, opened 2015-03-04
Node distributor removal now deletes published data.

mhrivnak:
re #681

https://pulp.plan.io/issues/681

### DIFF
--- a/nodes/parent/pulp_node/distributors/http/distributor.py
+++ b/nodes/parent/pulp_node/distributors/http/distributor.py
@@ -13,6 +13,7 @@
 
 from gettext import gettext as _
 from logging import getLogger
+import os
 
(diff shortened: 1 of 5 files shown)


mhrivnak:
As a quick explanation, I did a small bit of re-arranging so that the publish dir for a given repo is only calculated in one place, which is now on the distributor's ``_get_publish_dir`` method.

jortel:
LGTM

jortel:
Since 2.6.0 has not been released.  shouldn't this go against 2.6-testing?

mhrivnak:
I don't think it's worth delaying 2.6.0 for this. It can go in 2.6.1, and anyone who wants it sooner can cherry-pick it. I branched from 2.6-testing so it would cherry-pick cleanly.

PR #1694, opened 2015-03-04
fallback to config wrap width when unable to access tty size, closes 335

asmacdo:
Using an ssh client to execute pulp-admin commands caused a traceback if `wrap_to_terminal` in  `/etc/pulp/admin/admin.conf` was set to `true`. There is no way to get the terminal size in this case because the command is being run without a tty device, so we should fall back to the `wrap_width` specified in `admin.conf`


beav:
LGTM once jenkins passes

beav:
ok test

PR #1701, opened 2015-03-05
Synchronous agent attach to prevent race conditions.

jortel:
re:#734

The asynchronous attach (using a thread) was designed to prevent the pulp plugin *initializer* from blocking the loading of other plugins if the registration validation was blocked.  The problem with doing this is that - when there are pending requests, the gofer scheduler will attempt to dispatch to a plugin that has not completely loaded.  This fails.  So, this PR goes back the synchronous way of doing things (except the addition of making the API call.

This matches changes made to the Katello agent.

Gofer 2.6+ provides for more alternatives here but for now, we need to do this.

If gofer survives the chopping block, I will suggest a few things to clean up the pulp agent:

- Refactor the pulpplugin.py into several modules (something like):
  - plugin.py (remote API)
  - auth.py  (message authentication)
  - profile.py (profile reporting action)
  - registration (validate registration)

- Get rid of having the agent pragmatically update the settings.  The pulp-consumer can simply update
write the plugin configuration at registration.

### DIFF
--- a/agent/pulp/agent/gofer/pulpplugin.py
+++ b/agent/pulp/agent/gofer/pulpplugin.py
@@ -6,7 +6,6 @@ Contains recurring actions and remote classes.
 import os
 
 from time import sleep
-from threading import Thread
 from gettext import gettext as _
 from logging import getLogger
 
@@ -47,6 +46,13 @@ plugin = Plugin.find(__name__)
 registered = False
 
 
+class ValidateRegistrationFailed(Exception):
+    """
+    The REST call to the server to validate registration failed.
+    """
+    pass
+
+
 @initializer
 def init_plugin():
     """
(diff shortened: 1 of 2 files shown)


agent/pulp/agent/gofer/pulpplugin.py
```diff
@@ -55,12 +61,21 @@ def init_plugin():
...
-    attach = Attach()
-    attach.start()
+    while True:
+        try:
+            validate_registration()
+            if registered:
+                update_settings()
+            # DONE
+            break
+        except ValidateRegistrationFailed:
```
mhrivnak: This seems like a good opportunity to log something informative.

agent/pulp/agent/gofer/pulpplugin.py
```diff
@@ -115,13 +130,25 @@ def certificate_changed(path):
...
+        try:
+            validate_registration()
+            if registered:
+                update_settings()
+                plugin.attach()
+            else:
+                plugin.detach()
+            # DONE
+            break
+        except ValidateRegistrationFailed:
```
mhrivnak: same here.

agent/pulp/agent/gofer/pulpplugin.py
jortel: The validation failure is logged where it is raised.  Logging something here would be redundant.

PR #1704, opened 2015-03-09
158 - Add python-qpid 0.18-1 to EL5 external_deps.json

barnabycourt:
Also, remove the builder.py since that code now lives in pulp_packaging

closes #158


beav:
Can we schedule a deep dive on the new build style? :swimmer:

The change itself is LGTM.

PR #1705, opened 2015-03-09
Agent related 2.6.0 release notes.

jortel:
Add agent related release notes.

### DIFF
--- a/docs/user-guide/release-notes/2.6.x.rst
+++ b/docs/user-guide/release-notes/2.6.x.rst
@@ -83,6 +83,17 @@ You can see the complete list of bugs that were
 Known Issues
 ------------
 
+* An issue in the pulp (gofer) agent plugin *can* cause in-progress RMI requests to be
+  discarded when `goferd` is restarted. Should this occur, an entry is written to the system log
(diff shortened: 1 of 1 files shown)


docs/user-guide/release-notes/2.6.x.rst
```diff
@@ -83,6 +83,16 @@ You can see the complete list of bugs that were
...
 ------------
 
+* An issue in the pulp (gofer) agent plugin *can* cause in-progress RMI requests to be
+  discarded when `goferd` is restarted. Should this occur, an entry is written to the system log
+  on the consumer. On the Pulp server, the associated task will appear to never complete.
+  This has been fixed in Pulp 2.6.1.
+
+* Version 2.5 of the python-gofer-amqp messaging adapter, which is used to support RabbitMQ,
+  contains a regression. It pertains to the reconnect logic. Depending on how a connection
+  error manifests itself, it *can* result in infinite recursion during reconnect. This issue has
```
mhrivnak: It may be helpful to be more specific about the behavior that will ensue. Since python doesn't allow infinite recursion, I'm guessing an exception gets raised pretty quickly. Does the process barf and exit? Something else?

PR #1707, opened 2015-03-09
split database query in publish_step into multiple smaller ones

beav:
For large numbers of available packages this step breaks with an assertion in
mongodb (assertion 13385 combinatorial limit of $in partitioning of result set
exceeded).  I fixed it with running the database query multiple times for
smaller subsets of our available units.

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -1055,12 +1055,14 @@ class GetLocalUnitsStep(PluginStep):
         # any units that are already in pulp
         units_we_already_had = set()
 
-        # for any unit that is already in pulp, save it into the repo
-        for unit_dict in self.content_query_manager.get_multiple_units_by_keys_dicts(
-                self.unit_type, self.parent.available_units, self.unit_key_fields):
-            unit = self._dict_to_unit(unit_dict)
-            self.get_conduit().save_unit(unit)
-            units_we_already_had.add(unit)
+        # mongodb throws exceptions for too big queries, so we spilt it into multiple smaller ones
+        for page in misc.paginate(self.parent.available_units, 50):
+            # for any unit that is already in pulp, save it into the repo
+            for unit_dict in self.content_query_manager.get_multiple_units_by_keys_dicts(
+                    self.unit_type, page, self.unit_key_fields):
+                unit = self._dict_to_unit(unit_dict)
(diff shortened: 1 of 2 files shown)


beav:
ok test

server/pulp/plugins/util/publish_step.py
```diff
@@ -1055,12 +1055,14 @@ def process_main(self):
...
         units_we_already_had = set()
 
-        # for any unit that is already in pulp, save it into the repo
-        for unit_dict in self.content_query_manager.get_multiple_units_by_keys_dicts(
-                self.unit_type, self.parent.available_units, self.unit_key_fields):
-            unit = self._dict_to_unit(unit_dict)
-            self.get_conduit().save_unit(unit)
-            units_we_already_had.add(unit)
+        # mongodb throws exceptions for too big queries, so we spilt it into multiple smaller ones
+        for page in misc.paginate(self.parent.available_units, 50):
```
rbarlow: Do you think we should go with something larger than 50 units per page? This might cause a lot of queries, which might also not be optimal.

I'm not sure what value I would suggest, maybe 500?

If you prefer to keep 50, that's fine with me.

server/test/unit/plugins/util/test_publish_step.py
```diff
@@ -1180,12 +1180,16 @@ def test_no_available_units(self, mock_get_multiple):
...
+    def test_calls_get_multiple(self, mock_paginate, mock_get_multiple):
+        """
+        ensure that paginate is used
+        """
+        mock_paginate.return_value = []
 
         self.step.process_main()
 
-        mock_get_multiple.assert_called_once_with('fake_unit_type', [], ['foo'])
+        mock_paginate.assert_called_once_with([], 50)
```
rbarlow: Should we assert that self.parent.available_units was specifically used as the first arg to paginate?

rbarlow:
LGTM!

server/pulp/plugins/util/publish_step.py
beav: I will test out 50 vs 500 and see what the performance difference is.

beav:
ok test

beav:
ok to test

beav:
ok test

beav:
Merging patch now.

Big thanks to @aschieb for this patch!

PR #1719, opened 2015-03-12
Create /var/run/pulp in pulp-dev.py.

rbarlow:


pulp-dev.py
```diff
@@ -38,13 +37,13 @@
...
     '/usr/lib/pulp/consumer/extensions',
     '/usr/lib/yum-plugins/',
+    '/var/cache/pulp',
     '/var/lib/pulp',
     '/var/lib/pulp_client',
     '/var/lib/pulp_client/admin',
     '/var/lib/pulp_client/admin/extensions',
     '/var/lib/pulp_client/consumer',
     '/var/lib/pulp_client/consumer/extensions',
-    '/var/cache/pulp',
```
dkliban: This is where workers create working directories. please don't remove that.

rbarlow: On 03/12/2015 03:51 PM, Dennis Kliban wrote:
>> -    '/var/cache/pulp',
> 
> This is where workers create working directories. please don't remove that.

It wasn't removed, it was moved to be in alphabetical order (see above).
This list was alphabetized otherwise, and it makes it easier to group
nearby/related folders together.

-- 
Randy Barlow

dkliban:
LGTM

PR #1730, opened 2015-03-18
make uri_path exclude get parameters

asmacdo:
This corrects a subtle exception handling difference between Webpy and Django .

For example, if you cause an errorr with a request url of `/v2/repositories/zoo/history/sync/?limit=notanint` you will get json objects that contain the the href of the call.

Django:

    "_href": "/pulp/api/v2/repositories/zoo/history/sync/?limit=notanint"
Webpy:

    "_href": "/pulp/api/v2/repositories/zoo/history/sync/"

### DIFF
--- a/server/pulp/server/webservices/http.py
+++ b/server/pulp/server/webservices/http.py
@@ -219,7 +219,7 @@ def uri_path():
     @return: full current URI path
     """
     if _is_django():
-        return unicode(_get_wsgi_environ()['REQUEST_URI'])
+        return unicode(_get_wsgi_environ()['REQUEST_URI']).split("?")[0]
     else:
         return web.http.url(web.ctx.path)
 


rbarlow:
LGTM.

asmacdo:
@ipanova, before I merge this I would like to see if/how this affects your stuff.

ipanova:
@asmacdo  Only Consumer history has same behaviour. It does not affect me, maybe some conflicts but that's resolvable. The question is - is this really an issue? I think it is more correct to have in the _href path + query params.  The uri_path() returns current URI path which can contain the query params. http://en.wikipedia.org/wiki/URI_scheme#Examples
Anyway, I am ok with both options

asmacdo:
@ipanova, thanks for the feedback. I discussed with @bmbouter, and we agree with you. I have updated the [wiki](https://fedorahosted.org/pulp/wiki/ReplacewebpywithDjango#Behavioralchangesthatarebeingleftdifferent) to include this change.

PR #1731, opened 2015-03-19
Introducing Status view.

ipanova:
`GET /pulp/api/v2/status/ ` http://fpaste.org/200093/26782045/


server/pulp/server/webservices/views/status.py
```diff
@@ -0,0 +1,43 @@
+"""
+Unauthenticated status API so that other can make sure we're up (to no good).
```
rbarlow: Hahahahahahaha, this is great. :+1:

server/pulp/server/webservices/views/status.py
```diff
@@ -0,0 +1,43 @@
...
+import pulp.server.managers.status as status_manager
+from pulp.server.webservices.views.util import generate_json_response_with_pulp_encoder
+
+
+class StatusView(View):
+    """
+    View for server status
+    """
+
+    def get(self, request):
```
rbarlow: Should we require authentication on this view? I don't know for sure one way or another, but if this view causes a lot of queries, it might be a DOS vector.

I assume the old view did not require this, so perhaps we could consider this in a separate effort. What do you think?

rbarlow:
Looks great to me!

rbarlow:
Will you remove the old handlers in a separate PR? That makes sense to me, just wanted to make sure we don't forget. Though, I suppose we will also just do a big sweep in the end anyway.

ipanova:
@rbarlow I wanted to remove the webpy code in the second commit in this PR, but i can also do it in a separate PR. not sure what choice is better

server/pulp/server/webservices/views/status.py
ipanova: I think that's a valid problem, I will fill in a separate issue.

rbarlow:
On 03/23/2015 05:19 AM, Ina Panova wrote:
> @rbarlow <https://github.com/rbarlow> I wanted to remove the webpy code
> in the second commit in this PR, but i can also do it in a separate PR.
> not sure what choice is better

I think either way is fine, so whichever you prefer.

-- 
Randy Barlow

PR #1732, opened 2015-03-19
refs #99 - add ostree web prefix.

jortel:
Putting plugin specific values here is a total hack but reworking how repo-auth works is outside the scope of this pr.


beav:
LGTM for the code itself once travis passes. If you don't mind, can you create an issue in redmine to fix this in a better way?

Originally I had this being set via `repo_auth.conf` but I don't think that is a good way to do it either since it requires an extra step after installing a plugin RPM.

PR #1735, opened 2015-03-20
Use epel-release in dev-setup.sh if the OS is CentOS.

rbarlow:
This also installs bash completion.

### DIFF
--- a/playpen/dev-setup.sh
+++ b/playpen/dev-setup.sh
@@ -96,7 +96,11 @@ else
         sudo wget -q https://repos.fedorapeople.org/repos/pulp/pulp/rhel-pulp.repo
     fi
     if ! rpm -q epel-release; then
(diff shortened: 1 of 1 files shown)


playpen/dev-setup.sh
```diff
@@ -104,7 +108,7 @@ sudo yum-config-manager --enable pulp-2.6-testing > /dev/null
 popd
 
 echo "installing some dev tools"
-sudo yum install -y vim-enhanced python-virtualenvwrapper
+sudo yum install -y vim-enhanced python-virtualenvwrapper bash-completion python-django-bash-completion
```
beav: For readability, can this be on two lines, using `\`? :eyeglasses:

beav:
LGTM, newline request is optional

playpen/dev-setup.sh
rbarlow: Fixed.

PR #1725, opened 2015-03-17
Add a SearchView and convert TaskSearch and UserSearch to use it.

rbarlow:
https://pulp.plan.io/issues/94

closes #94

PR #1736, opened 2015-03-20
Remove the webpy controllers for users and tasks.

rbarlow:
There is also a small fix in here to appease flake8. It's unrelated, but the tests were failing on my dev box without it.

This cannot be merged until https://github.com/pulp/pulp/pull/1725 is merged, or we won't have an API for searching users and tasks ☺

https://pulp.plan.io/issues/94

re #94


ipanova:
LGTM after https://github.com/pulp/pulp/pull/1725 is merged.

asmacdo:
Hijacked by #1776 and merged there.

PR #1739, opened 2015-03-23
540 - Check previous version of pulp-server SELinux policy at ugrade

dkliban:
This patch enhances the pulp-selinux RPM install process. The version of the
pulp-server SELinux policy is recorded in /var/lib/rpm-state/pulp before the new
policy is installed. Then that version is passed as an argument to relabel.sh.
The relabel script then only relabels the files that have not been properly labeled
in the previously installed version. If no version is passed in, all files
are relabeled.

Fixes #540

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -815,6 +815,13 @@ Requires(postun): /usr/sbin/semodule
 %description    selinux
 SELinux policy for Pulp's components
 
+%pre selinux
+# Record old version so we can limit which restorecon statement are executed later
+test -e %{_localstatedir}/lib/rpm-state/%{name} || mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
+oldversion=$(semodule -l | grep pulp-server)
+echo ${oldversion:12} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
+
+exit 0
 %post selinux
 # Enable SELinux policy modules
 if /usr/sbin/selinuxenabled ; then
@@ -825,7 +832,8 @@ fi
 # Spacewalk saw same issue and filed BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=505066
 %posttrans selinux
 if /usr/sbin/selinuxenabled ; then
- %{_datadir}/pulp/selinux/server/relabel.sh %{_datadir}
+ cat %{_localstatedir}/lib/rpm-state/%{name}/old-version | xargs %{_datadir}/pulp/selinux/server/relabel.sh
+ rm %{_localstatedir}/lib/rpm-state/%{name}/old-version
 fi
 
 %preun selinux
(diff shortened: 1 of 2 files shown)


pulp.spec
```diff
@@ -815,6 +815,13 @@ Requires(postun): /usr/sbin/semodule
 %description    selinux
 SELinux policy for Pulp's components
 
+%pre selinux
+# Record old version,if any, so we can limit which restorecon statement are executed later
```
bmbouter: s/,if any,//

pulp.spec
```diff
@@ -815,6 +815,13 @@ Requires(postun): /usr/sbin/semodule
...
+test -e %{_localstatedir}/lib/rpm-state/%{name} || mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
+oldversion=$(semodule -l | grep pulp-server)
+echo ${oldversion:12} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
+
+exit 0
```
bmbouter: is this supposed to be here?

server/selinux/server/relabel.sh
```diff
@@ -1,13 +1,20 @@
...
-/sbin/restorecon -i -R /etc/pki/pulp
-/sbin/restorecon -i -R /srv/pulp
-/sbin/restorecon -i -R /usr/bin/pulp-admin
-/sbin/restorecon -i -R /usr/bin/pulp-consumer
-/sbin/restorecon -i -R /var/lib/pulp
-/sbin/restorecon -i -R /var/log/pulp
-
-# Relabel the celery binary
-/sbin/restorecon -i -R /usr/bin/celery
+# If upgrading from before 2.4.0
```
bmbouter: This needs to be reworked so that if no argument is passed that all restorecon statements in history are run.

pulp.spec
```diff
@@ -815,6 +815,13 @@ Requires(postun): /usr/sbin/semodule
 %description    selinux
 SELinux policy for Pulp's components
 
+%pre selinux
+# Record old version,if any, so we can limit which restorecon statement are executed later
+test -e %{_localstatedir}/lib/rpm-state/%{name} || mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
```
bmbouter: When the mkdir runs will it give it the right permissions and ownership? Maybe use `-m` on mkdir

bmbouter: I guess it doesn't matter since its deleted below. If that is correct then don't worry about it.

bmbouter:
Changing the behavior when no argument is passed is important. Also removing `exit 0` if it is no longer needed. Other than those two things, LGTM.

server/selinux/server/relabel.sh
dkliban: When no argument is returned, $1 resolves to an empty string which makes the statement [$1 < 'any_string'] return true. As a result all the statements will get run.

pulp.spec
dkliban: According to https://fedoraproject.org/wiki/Packaging:ScriptletSnippets#Syntax it does need to be there.

server/selinux/server/relabel.sh
dkliban: if you save this little bit as an executable you can try this for yourself. http://fpaste.org/201685/14271384/

bmbouter: Oh yeah this is good.

PR #1740, opened 2015-03-23
Added message about new documentation

dkliban:
Closes #693


docs/sphinx/dev-guide/index.rst
```diff
@@ -1,14 +1 @@
...
-
-   contributing/index
-   architecture
-   conventions/index
-   policies/index
-   newtypesupport/index
-   integration/index
-   glossary
-   troubleshooting
+The latest documentation has been moved to `here <http://pulp.readthedocs.org/en/latest/>`_.
```
bmbouter: I think this should have the full URL written out to the user. Same for below.

bmbouter: Also https should be used instead

bmbouter: Also it can just end in / instead of /en/latest/. RTD automaticaly redirects to /latest/ and selects the language based on the client so we should specify english.

bmbouter: It would be good to make this a lot bigger also

bmbouter: and centered

bmbouter:
I built this with the RTD theme and it looks good.

PR #1742, opened 2015-03-24
django conversion - repositories

asmacdo:
closes [#707](https://pulp.plan.io/issues/707)
closes [#749](https://pulp.plan.io/issues/749)
closes [#756](https://pulp.plan.io/issues/756)
closes [#804](https://pulp.plan.io/issues/804)
closes [#805](https://pulp.plan.io/issues/805)
closes [#812](https://pulp.plan.io/issues/812)


asmacdo:
`GET /v2/repositories/`
http://fpaste.org/195909/92468314/
`GET /v2/repositories/?details=true`
http://fpaste.org/195911/14259248/
`GET /v2/repositories/?details=false` 
*includes bugfix - https://pulp.plan.io/issues/707
http://fpaste.org/195922/25227142/
`GET /v2/repositories/?importers=true` 
http://fpaste.org/195925/25925310/
`GET /v2/repositories/?distributors=true`
http://fpaste.org/195927/59253681/

asmacdo:
`POST /v2/repositories/` 
normal conditions
http://fpaste.org/195947/42592707/
duplicate resource
http://fpaste.org/195946/25927019/
missing required field "id"
http://fpaste.org/195950/14259271/

asmacdo:
`GET /v2/repositories/<repo_id>/`
existing repo
http://fpaste.org/195957/92821314/
nonexisting repo
http://fpaste.org/195960/92827114/

asmacdo:
`GET /v2/repositories/<repo_id>/?details=true`
http://fpaste.org/195963/28546142/
`GET /v2/repositories/<repo_id>/?details=false`
includes [bugfix #707](https://pulp.plan.io/issues/707)
http://fpaste.org/195966/28627142/
`GET /v2/repositories/<repo_id>/?importers=true`
http://fpaste.org/195968/25928726/
`GET /v2/repositories/<repo_id>/?distributors=true`
http://fpaste.org/195970/42592880/

asmacdo:
`DELETE /v2/repositories/<repo_d>/`
existing repo
http://fpaste.org/196031/14259368/
nonexisting repo
http://fpaste.org/196029/36801142/

asmacdo:
`PUT /v2/repositoreis/<repo_id>/`
no data
http://fpaste.org/196258/42599951/
delta only
http://fpaste.org/196344/42601119/
update importer_config
http://fpaste.org/196448/26024185/
update distributor_configs (spawn a task)
http://fpaste.org/196445/26023855/

asmacdo:
`GET /v2/repositories/<repo_id>/importers/`
existing repo
http://fpaste.org/196664/84728142/
nonexisting repo
http://fpaste.org/196665/84788142/

asmacdo:
`POST /v2/repositories/<repo_id>/importers/`
existing repo/importer_type_id
http://fpaste.org/196687/26086696/
existing repo, nonexisting importer
http://fpaste.org/196689/60868011/
nonexisting repo
http://fpaste.org/196691/60869121/

asmacdo:
`GET /v2/repositories/<repo_id>/importers/<importer_id>/`
existing repo and importer
http://fpaste.org/196779/96127142/
existing repo nonexisting importer 
*[includes bugfix #749](https://pulp.plan.io/issues/749)
http://fpaste.org/196793/60970031/
nonexisting repo
http://fpaste.org/196795/14260970/

asmacdo:
`DELETE /v2/repositories/<repo_id>/importers/<importer_id>/`
existing and nonexisting
http://fpaste.org/196807/09847414/

asmacdo:
`PUT /v2/repositories/<repo_id>/importers/<importer_id>/`
existing repo and importer, correct params
http://fpaste.org/197094/26168218/
existing repo and importer, missing `importer_config` param
http://fpaste.org/197096/26168353/
existing repo, nonexisting importer
http://fpaste.org/197097/26168458/
nonexisting repo
http://fpaste.org/197098/16854514/

asmacdo:
`GET /v2/repositories/<repo_id>/importers/<importer_id>/schedules/sync/`
as expected
http://fpaste.org/197142/42617288/
invalid importer_id
http://fpaste.org/197144/17302114/
invalid repo_id
http://fpaste.org/197146/61730641/

asmacdo:
`POST /v2/repositories/<repo_id>/importers/<importer_id>/schedules/sync/`
minimal config
http://fpaste.org/197170/75405142/
extra params
http://fpaste.org/197172/75495142/
missing schedule
http://fpaste.org/197179/14261760/
bad importer
http://fpaste.org/197180/14261763/
bad repo
http://fpaste.org/197195/26177487/

asmacdo:
`GET /v2/repositories/<repo_id>/importers/<importer_id>/schedules/sync/<schedule_id>/`
as expected
http://fpaste.org/197592/26257239/
invalid schedule_id
*includes bugfix [redmine-756](https://pulp.plan.io/issues/756)
http://fpaste.org/197614/14262591/
nonexistent but possibly valid schedule_id
http://fpaste.org/197615/14262592/
nonexistent importer_id
http://fpaste.org/197617/59287142/
nonexistent repo_id
http://fpaste.org/197619/14262593/

asmacdo:
`DELETE /v2/repositories/<repo_id>/importers/<importer_id>/schedules/sync/<schedule_id>/`
existing and no-longer-existing schedule_id
http://fpaste.org/197621/62598911/
invalid schedule_id
*includes bugfix [redmine-756](https://pulp.plan.io/issues/756)
http://fpaste.org/197622/14262601/
invalid importer_id
http://fpaste.org/197626/42626021/
invalid repo_id
http://fpaste.org/197630/62602601/

asmacdo:
`PUT /v2/repositories/<repo_id>/importers/<importer_id>/schedules/sync/<schedule_id>/`
minimal body
http://fpaste.org/197676/14262667/
make a change
http://fpaste.org/197679/67071142/

asmacdo:
`GET /v2/repositories/<repo_id>/distributors/`
as expected
http://fpaste.org/197741/27496914/
nonexisting repo-id
http://fpaste.org/197743/42627503/

asmacdo:
`POST /v2/repositories/<repo_id>/distributors/`
as expected
http://fpaste.org/197740/42627490/
missing data
http://fpaste.org/197748/75346142/
missing repo_id
http://fpaste.org/197753/27571014/

asmacdo:
`GET /v2/repositories/<repo_id>/distributors/<distributor_id>/`
existing distributor
http://fpaste.org/197789/80100142/
nonexisting distributor
http://fpaste.org/197791/26280180/
nonexisting repo
http://fpaste.org/197793/62802281/

asmacdo:
`GET /v2/repositories/<repo_id>/distributors/<distributor_id>/`
existing distributor
http://fpaste.org/197789/80100142/
nonexisting distributor
http://fpaste.org/197791/26280180/
nonexisting repo
http://fpaste.org/197793/62802281/

asmacdo:
`DELETE /v2/repositories/<repo_id>/distributors/<distributor_id>/`
as expected
http://fpaste.org/198621/14265303/
invalid distributor
http://fpaste.org/198622/30519142/
missing repo
http://fpaste.org/198623/65305921/

asmacdo:
`PUT /v2/repositories/<repo_id>/distributors/<distributor_id>/`
as expected
http://fpaste.org/197798/81178142/
without "distributor_config" in request body
http://fpaste.org/197799/42628126/

asmacdo:
`GET /v2/repositories/<repo_id>/distributors/<distributor_id/schedules/publish/`
as expected
http://fpaste.org/198659/14265351/
invalid distributor
http://fpaste.org/198660/26535270/
invalid repo
http://fpaste.org/198663/65353881/

asmacdo:
`POST /v2/repositories/<repo_id>/distributors/<distributor_id/schedules/publish/`
as expected
http://fpaste.org/198627/65317601/

asmacdo:
`GET /v2/repositories/<repo_id>/distributors/<distributor_id/schedules/publish/<schedule_id/`
as expected
http://fpaste.org/198679/53725714/
missing schedule_id
http://fpaste.org/198680/42653736/
invalid schedule_id
[includes bugfix #756](https://pulp.plan.io/issues/756)
http://fpaste.org/198681/26537399/
missing distributor
http://fpaste.org/198684/26537572/

asmacdo:
`PUT /v2/repositories/<repo_id>/distributors/<distributor_id/schedules/publish/<schedule_id/`
as expected
http://fpaste.org/198694/38397142/
invalid schedule id
http://fpaste.org/198947/60033214/

asmacdo:
`DELETE /v2/repositories/<repo_id>/distributors/<distributor_id/schedules/publish/<schedule_id/`
http://fpaste.org/198696/14265386/

asmacdo:
`POST /v2/repositories/actions/content/regenerate_applicability/`
missing body param repo_criteria
http://fpaste.org/199143/14266199/
repo_criteria does not contain a dict
http://fpaste.org/199154/20594142/
as expected
http://fpaste.org/199158/26620885/

asmacdo:
`GET /v2/repositories/<repo_id>/history/sync/`
as expected
http://fpaste.org/199715/04873142/
missing repo_id
http://fpaste.org/199718/70495114/
with get params:
limit is valid
http://fpaste.org/199721/70505114/
limit is invalid
there is a subtle difference - see https://github.com/pulp/pulp/pull/1730 and [wiki](https://fedorahosted.org/pulp/wiki/ReplacewebpywithDjango#Behavioralchangesthatarebeingleftdifferent)
http://fpaste.org/199722/42670522/

asmacdo:
`GET /v2/repositories/<repo_id>/history/publish/<distributor_id>/`
as expected
http://fpaste.org/200203/42679279/
missing distributor_id
http://fpaste.org/200204/14267929/
missing repo_id
http://fpaste.org/200205/42679297/
with get param
http://fpaste.org/200206/42679306/
with invalid limit
there is a subtle difference - see https://github.com/pulp/pulp/pull/1730 and [wiki]
http://fpaste.org/200208/79311614/

asmacdo:
`POST /v2/repositories/<repo_id>/actions/sync/`
as expected
http://fpaste.org/200229/67947651/
with override config
http://fpaste.org/200232/26794994/
missing repo
http://fpaste.org/200233/79525814/

asmacdo:
`POST /v2/repositories/<repo_id>/actions/publish/`
as expected
http://fpaste.org/200234/26795373/
missing distributor
http://fpaste.org/200237/14267955/
missing repo
http://fpaste.org/200236/42679551/

asmacdo:
`POST /v2/repositories/<repo_id>/actions/associate/`
as expected
http://fpaste.org/201482/12300414/
missing destination repo
http://fpaste.org/201485/71230971/
missing source repo
http://fpaste.org/201486/14271231/
bad criteria
Includes bugfix for [#804](https://pulp.plan.io/issues/804)
http://fpaste.org/201632/42713469/
missing source_repo_id
http://fpaste.org/201521/27125004/

asmacdo:
`POST /v2/repositories/<repo_id>/actions/unassociate/`
bad criteria
Includes bugfix for [#804](https://pulp.plan.io/issues/804)
http://fpaste.org/201744/14271426/
as expected
http://fpaste.org/201748/14271432/

asmacdo:
`POST /v2/repositories/<repo_id>/actions/import_upload/`
as expected
http://fpaste.org/201786/14701414/
missing required params
contains [bugfix #805](https://pulp.plan.io/issues/805)
http://fpaste.org/201780/46551142/

asmacdo:
~~`POST /v2/repositories/<repo_id>/actions/resolve_dependencies/`
missing required params
includes [bugfix #806](https://pulp.plan.io/issues/806)
http://fpaste.org/201791/71482251/~~

resolve_depenencies has been removed, fixes [#812](https://pulp.plan.io/issues/812)

server/pulp/server/webservices/urls.py
```diff
@@ -26,6 +26,13 @@
     RepoGroupAssociateView, RepoGroupDistributorResourceView, RepoGroupDistributorsView,
     RepoGroupPublishView, RepoGroupResourceView, RepoGroupsView, RepoGroupUnassociateView
 )
+from pulp.server.webservices.views.repositories import(
+    ReposView, ContentApplicabilityRegenerationView, RepoDistributorsView,
+    RepoDistributorResourceView, RepoImporterResourceView, RepoImportersView,
+    RepoResourceView, RepoPublishScheduleResourceView, RepoPublishScheduleCollectionView,
+    RepoSyncScheduleCollectionView, RepoSyncScheduleResourceView, RepoSyncHistory,
+    RepoPublishHistory, RepoSync, RepoPublish, RepoAssociate, RepoUnassociate, RepoImportUpload,
+)
```
ipanova: please make it in alphabetical order

server/pulp/server/webservices/urls.py
```diff
@@ -101,6 +108,40 @@
...
+    url(r'^v2/repositories/$', ReposView.as_view(), name='repos'),
+    url(r'^v2/repositories/actions/content/regenerate_applicability/$',
+        ContentApplicabilityRegenerationView.as_view(), name='repo_content_app_regen'),
+    url(r'^v2/repositories/(?P<repo_id>[^/]+)/$', RepoResourceView.as_view(), name='repo_resource'),
+    url(r'^v2/repositories/(?P<repo_id>[^/]+)/importers/$', RepoImportersView.as_view(),
+        name='repo_importers'),
+    url(r'^v2/repositories/(?P<repo_id>[^/]+)/importers/(?P<importer_id>[^/]+)/$',
+        RepoImporterResourceView.as_view(), name='repo_importer_resource'),
+    url(r'^v2/repositories/(?P<repo_id>[^/]+)/importers/(?P<importer_id>[^/]+)/schedules/sync/$',
+        RepoSyncScheduleCollectionView.as_view(), name='repo_sync_schedule_collection'),
```
ipanova: maybe better 'repo_sync_schedules'?

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+
+    @staticmethod
+    def _process_repos(repos, importers=False, distributors=False):
+        """
+        Apply standard processing to a collection of repositories being returned to a client. Adds
+        the object link and optionally adds related importers and distributors.
+
+        :param repos: collection of repositories
+        :type  repos: list, tuple
+        :param importers: iff True, adds related importers under the attribute "importers".
```
ipanova: s/iff/if

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :type  importers: bool
+        :param distributors: iff True, adds related distributors under the attribute "distributors"
```
ipanova: same here

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        Looks for query parameters 'importers' and 'distributors', and will add
+        the corresponding fields to the repository returned. Query parameter
+        'details' is equivalent to passing both 'importers' and 'distributors'.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: id of requested repository
+        :type  repo_id: str
+
+        :return: Response containing a serialized dict for the created repo.
```
ipanova: s/for the created repo/for the requested repo

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :rtype : django.http.HttpResponse
+        :raises pulp_exceptions.MissingResource if repo cannot be found
```
ipanova: i guess the second colon should be after the exception

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+    def delete(self, request, repo_id):
+        """
+        Dispatch a task to delete a repository.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: id of repository to be removed
+        :type  repo_id: str
+
+        :return: Response containing a serialized dict for the created repo.
```
ipanova: remove this line

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        Remove an importer from a repository.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: The id of the repository to remove the importer from
+        :type  repo_id: str
+        :param importer_id: The id of the importer to remove from the given repository
+        :type  importer_id: str
+
+        :raises pulp_exceptions.MissingResource if importer cannot be found for this repo
```
ipanova: colon after the exception

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :raises pulp_exceptions.OperationPostponed to dispatch a task to delete the importer
```
ipanova: same here

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+
+    @auth_required(authorization.UPDATE)
+    @json_body_required
+    def put(self, request, repo_id, importer_id):
+        """
+        Associate an importer to a repository.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: The id of the repository to associate with
```
ipanova: s/The id of the repository to associate with/The id of the repository

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :type  repo_id: str
+        :param importer_id: The id of the importer to associate
```
ipanova: s/The id of the importer to associate/The id of the importer to update

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: id of the repository
+        :type  repo_id: str
+        :param importer_id: create a new scheduled sync for this importer
+        :type  importer_id: str
+
+        :return: Response containing a serialized dict of the new  scheduled sync
+        :rtype : django.http.HttpResponse
+        :raises pulp_exceptions.UnsupportedValue if there are unsupported request body params
```
ipanova: colon after exception

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: id of repository to be updated
+        :type  repo_id: str
+
+        :return: Response containing a serialized dict for the updated repo.
+        :rtype : django.http.HttpResponse
+        :raises pulp_exceptions.OperationPostponed: if a distributor is updated, dispatch a task
+        """
+
+        delta = request.body_as_json.get('delta', None)
```
ipanova: would not it be better to have params = request.body_as_json and then just params.get() instead of calling 'request.body_as_json.get' everytime?

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :param repo_id: id of the repository
+        :type  repo_id: str
+        :param importer_id: remove the scheduled sync from this importer
+        :type  importer_id: str
+        :param schedule_id: id of the scheduled repository sync to delete
+        :type  schedule)id: str
+
+        :return: An empty response
+        :rtype: django.http.HttpResponse
+        :raises pulp_exceptions.MissingResource: if schedule_id does not exist for the importer
```
ipanova: if schedule_id/importer_id/repo_id do not exist

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        """
+
+        try:
+            self.manager.delete(repo_id, importer_id, schedule_id)
+        except pulp_exceptions.InvalidValue:
```
ipanova: except pulp_exceptions.MissingResource

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+            raise pulp_exceptions.MissingResource(schedule_id=schedule_id)
+        return generate_json_response(None)
+
+    @auth_required(authorization.UPDATE)
+    @json_body_required
```
ipanova: from the docs i can see that all params are optional, therefore we should use here @json_body_allow_empty
http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/sync.html#updating-a-scheduled-sync
check this please

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        should have a boolean value.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: id of the repository the distributor is associated with
+        :type  repo_id: str
+        :param distributor_id: id of the distributor instance to update.
+        :type  distributor_id: str
+
+        :raises pulp_exceptions.MissingValue: dispatch a task
```
ipanova: s/:raises pulp_exceptions.MissingValue: dispatch a task/raises pulp_exceptions.MissingValue: if distributor_config is none

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        """
+
+        self.manager.validate_distributor(repo_id, distributor_id)
+        resource_href = reverse('repo_publish_schedule_resource', kwargs={
+            'repo_id': repo_id, 'distributor_id': distributor_id, 'schedule_id': schedule_id
+        })
+        return self._get(schedule_id, resource_href)
+
+    @auth_required(authorization.UPDATE)
+    @json_body_required
```
ipanova: from the docs i can see that all params are optional, therefore we should use here @json_body_allow_empty
Same like for scheduled sync update.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+            updates['iso_schedule'] = updates.pop('schedule')
+        schedule = self.manager.update(repo_id, distributor_id, schedule_id, updates)
+        ret = schedule.for_display()
+        ret['_href'] = reverse('repo_publish_schedule_resource', kwargs={
+            'repo_id': repo_id, 'distributor_id': distributor_id, 'schedule_id': schedule_id
+        })
+        return generate_json_response(ret)
+
+    @auth_required(authorization.DELETE)
+    def delete(self, request, repo_id, distributor_id, schedule_id):
```
ipanova: would you implement same try/except like in scheduled sync removal? https://github.com/pulp/pulp/pull/1742/files#diff-69de4359dc90d2ab726dbca11e4f32b4R565

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        return generate_json_response_with_pulp_encoder(entries)
+
+
+class RepoSync(View):
+    """
+    View for syncing a repository.
+    """
+
+    @auth_required(authorization.EXECUTE)
+    @json_body_allow_empty
```
ipanova: would you change the docs and specify that the 'override_config' is optional param?
http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/sync.html#sync-a-repository

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        raise pulp_exceptions.OperationPostponed(async_result)
+
+
+class RepoPublish(View):
+    """
+    View for publishing a repository.
+    """
+
+    @auth_required(authorization.EXECUTE)
+    @json_body_allow_empty
```
ipanova: Houston we have a problem :)
seems like we have to put here the @json_body_required as if we don't specify any params in the task result we have an error http://fpaste.org/202731/29680914/
So the solution would be to make the 'id' as required' param i.e. make a check and  raise a MissingValue.

ipanova: or, dig into the code and see if there is an option make possible publish for all distributors if you do not specify anything the body.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        raise pulp_exceptions.OperationPostponed(async_result)
+
+
+class RepoUnassociate(View):
+    """
+    View to unassociate a unit from a repository.
+    """
+
+    @auth_required(authorization.UPDATE)
+    @json_body_allow_empty
```
ipanova: i think we have to make criteria as required key, as if you do not specify any body, *all* units after this call will be removed. i do not think that would make happy our users. http://fpaste.org/202761/99384142/

ipanova: actually even if specifying the criteria but leaving it empty would delete all the units http://fpaste.org/202762/42729945/
seems like we would need to check if the value of criteria is not empty dict.
@asmacdo what do you think?

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param dest_repo_id: id of the repository content will be copied to
+        :type  dest_repo_id: str
+
+        :raises pulp_exceptions.MissingValue: if required param source_repo_id is not passed
+        :raises pulp_exceptions.InvalidValue: if source_repo_id is not found or if criteria
+                                              params cannot be parsed
+        :raises pulp_exceptions.OperationPostponed: dispatch a publish repo task
+        :raises pulp_exceptions.PulpDataException: if there is a problem parsing the criteria object
```
ipanova: this line is not needed

server/pulp/server/webservices/views/repositories.py
asmacdo: since body_as_json is an attribute, there wouldn't be a performance difference between the 2, so I think its just a style thing. I find this way slightly more readable, but I don't feel strongly about it.

server/pulp/server/webservices/views/repositories.py
asmacdo: @ipanova, yeah, I wasn't sure exactly how to do this one. All params are optional, but you should have at least one of them, otherwise you are doing an update request without updating.

server/pulp/server/webservices/views/repositories.py
asmacdo: same response as above.

server/pulp/server/webservices/views/repositories.py
asmacdo: The sync/importer stuff is actually a little strange. The exceptions here should bubble up properly. See the comment on sync/importer for the weirdness that forced the try/except.

server/pulp/server/webservices/views/repositories.py
asmacdo: InvalidValue can be raised if the schedule_id is not a valid bson ObjectId, but in that case we should still return an InvalidValue.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1214 @@
...
+        # Catch MissingResource because this is body data, raise 400 rather than 404
+        try:
+            repo_query_manager.get_repository(source_repo_id)
+        except pulp_exceptions.MissingResource:
+            raise pulp_exceptions.InvalidValue(['source_repo_id'])
+
+        if criteria_body:
+            try:
+                criteria = UnitAssociationCriteria.from_client_input(criteria_body)
+            except pulp_exceptions.InvalidValue, e:
```
asmacdo: That code was eating all exceptions and returning an unhelpful one. 
https://pulp.plan.io/issues/804

server/pulp/server/webservices/views/repositories.py
asmacdo: @ipanova, looks reasonable. In fact, the docs have id as required. Good catch.
http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/publish.html#publish-a-repository

server/pulp/server/webservices/views/repositories.py
asmacdo: @ipanova, I also dislike this, but I think this is a general problem. I think we should not make this change as part of the conversion, because this could be done generally in the criteria models I think. I think there may be a bug about this?

server/pulp/server/webservices/views/repositories.py
asmacdo: yup, good catch

server/pulp/server/webservices/views/repositories.py
ipanova: yes you are right

server/pulp/server/webservices/views/repositories.py
ipanova: it is exactly the same case like with repo update - all params are optional, so if do not specify any of them you will do an update with actually updating anything 
https://github.com/pulp/pulp/pull/1742/files#diff-69de4359dc90d2ab726dbca11e4f32b4R254
i think the @json_body_allow_empty is right in this case, because if you put the @json_body_required it will contradict with the docs, which say that all params are optional and none of them are required.

server/pulp/server/webservices/views/repositories.py
ipanova: sweet

server/pulp/server/webservices/views/repositories.py
ipanova: if it is a criteria issue then i think we should fill a bug for it separetely

ipanova: yes, it is a general criteria issue, i just checked consumer group unassociate members, which has same logic as repo unassociate units, and same issue http://fpaste.org/203243/73815491/
I think this is strange, because seems to me that previously it was working normally( i am not sure), maybe a regression bug?

server/pulp/server/webservices/views/repositories.py
asmacdo: :+1:

server/test/unit/server/webservices/views/test_schedule.py
```diff
@@ -0,0 +1,60 @@
+import unittest
```
asmacdo: @ipanova, I missed these tests on the first go around. Please be sure to have a look at this file too!

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+    def test__merge_related_objects_no_objects(self):
+        """
+        Test that merge happens correctly when there are no objects to merge.
+        """
+
+        mock_repos = [{'id': 'mock1'}, {'id': 'mock2'}]
+        mock_importers = []
+
+        mock_importer_manager = mock.MagicMock()
+        mock_importer_manager.find_by_repo_list.return_value = mock_importers
```
ipanova: i think you directly assign the return_value to an empty list

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        ret = repositories._merge_related_objects('importers', mock_importer_manager, mock_repos)
+
+        self.assertTrue(len(ret) == 2)
+        self.assertEqual(map(itemgetter('id'), ret), ['mock1', 'mock2'])
+        mock1_expected_importers = []
+        mock2_expected_importers = []
+        mock1_importers = ret[map(itemgetter('id'), ret).index('mock1')]['importers']
+        mock2_importers = ret[map(itemgetter('id'), ret).index('mock2')]['importers']
+        self.assertEqual(mock1_importers, mock1_expected_importers)
+        self.assertEqual(mock2_importers, mock2_expected_importers)
```
ipanova: same here, you can assert directly to [] instead of consuming a var

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        """
+        Test _process_repos without optional args, assert that processing was called for each repo.
+        """
+
+        mock_reverse.return_value = '/mock/path/'
+        mock_repos = [{'id': 'mock1'}, {'id': 'mock2'}]
+        repos_view = ReposView()
+        repos_view._process_repos(mock_repos, False, False)
+        mock_convert_date.assert_has_calls(
+            [mock.call({'id': 'mock1', '_href': '/mock/path/'}),
```
ipanova: i do not really like the fact that the path is mocked, and it is the same for different repos. how can you know that reverse works properly and you did not mess with the kwargs. i think leaving reverse not mocked is not that bad idea, so then you can assert and check the expected _href and you will be sure that the _href itself is correct. what is your opinion on this?

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+             mock.call({'id': 'mock2', '_href': '/mock/path/'})]
+        )
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repositories.ReposView._process_repos')
+    @mock.patch('pulp.server.webservices.views.repositories.RepoModel.get_collection')
```
ipanova: i think it would be better if you would also mock the generate_json_response_with_pulp_encoder and then assert that the response was called with proper expected content

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        """
+
+        mock_repos = [{'mock_repo_1': 'somedata'}, {'mock_repo_2': 'moredata'}]
+        mock_collection.return_value.find.return_value = mock_repos
+        mock_request = mock.MagicMock()
+        mock_request.GET = {'details': 'true'}
+        repos_view = ReposView()
+
+        repos_view.get(mock_request)
+        mock_process.assert_called_once_with(mock_repos, True, True)
```
ipanova: same here, i think it would be useful to test the structure and the content of the response body

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repositories.ReposView._process_repos')
+    @mock.patch('pulp.server.webservices.views.repositories.RepoModel.get_collection')
+    def test_get_repos_with_false(self, mock_collection, mock_process):
```
ipanova: nice, also would need to add such a testcase to the consmers

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        mock_request.body = json.dumps(mock_repo)
+        mock_reverse.return_value = '/mock/path/'
+        mock_repo_manager = mock_factory.repo_manager.return_value
+        mock_repo_manager.create_and_configure_repo.return_value = mock_repo
+
+        repos_view = ReposView()
+        response = repos_view.post(mock_request)
+
+        mock_repo['_href'] = '/mock/path/'
+        mock_resp.assert_called_once_with(mock_repo)
```
ipanova: in my opinion it is better to use three vars in this case mock_repo for the request body, then mock_repo1 for the return_value of mock_repo_manager and mock_repo2 when you do the assertion here - mock_resp.assert_called_once_with.
reason why it is better - because usually the response body is different from request body,plus you will be sure that mock_repo content will not be changed somewhere in the middle if you forgot something to mock

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+class TestRepoResourceView(unittest.TestCase):
+    """
+    Tests for RepoResoureceView.
+    """
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
+    @mock.patch('pulp.server.webservices.views.repositories.manager_factory')
```
ipanova: just to simplify you can do @mock.patch('pulp.server.webservices.views.repositories.manager_factory.repo_query_manager')

and then just :
mock_factory.return_value.find_by_id.return_value = mock_repo

From my point of view it is more readable, but is might be just a style thing

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        mock_repo = {'mock_repo': 'somedata'}
+        mock_q_manager = mock_factory.repo_query_manager.return_value
+        mock_q_manager.find_by_id.return_value = mock_repo
+        mock_request = mock.MagicMock()
+        mock_request.GET = {}
+
+        repos_resource = RepoResourceView()
+        response = repos_resource.get(mock_request, 'mock_repo')
+
+        mock_resp.assert_called_once_with(mock_repo)
```
ipanova: same as previously, it is better to have mock_repo and expected_content:

expected_content = {'mock_repo': 'somedata', '_href' : '/some/href/'}
mock_resp.assert_called_once_with(expected_content)

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: thick line is not needed

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        repos_resource = RepoResourceView()
+        try:
+            repos_resource.get(mock_request, 'mock_repo')
+        except pulp_exceptions.MissingResource, response:
+            pass
+        else:
+            raise AssertionError("MissingResource should be raised for a nonexisting repository")
+
+        self.assertEqual(response.http_status_code, 404)
+        self.assertTrue(response.error_code is error_codes.PLP0009)
```
ipanova: you could also add `self.assertEqual(response.error_data['resources'], {'repo': 'mock_repo'})`

server/test/unit/server/webservices/views/test_repositories.py
ipanova: were not you supposed also to mock _convert_repo_dates_to_strings?

ipanova: ah i see, anyway no harm would be done here as last_unit_added is None same as last_unit_removed

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        """
+        Retrieve an existing repository with details.
+        """
+
+        mock_repo = {'mock_repo': 'somedata'}
+        mock_q_manager = mock_factory.repo_query_manager.return_value
+        mock_q_manager.find_by_id.return_value = mock_repo
+        mock_request = mock.MagicMock()
+        mock_request.GET = {'details': 'true'}
+        mock_merge.side_effect = lambda x, y, z: z
```
ipanova: why side_effect is used here instead of return_value? i thought that side_effect is used when some exception is raised

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        mock_manager = mock_factory.repo_manager.return_value
+        mock_task_result = mock_manager.update_repo_and_plugins.return_value
+        mock_task_result.spawned_tasks = 'distributor'
+
+        mock_request = mock.MagicMock()
+        mock_data = {'distributor_configs': 'distributor_data'}
+        mock_request.body = json.dumps(mock_data)
+
+        repos_resource = RepoResourceView()
+        try:
```
ipanova: i think you can avoid try/except block just by doing 
self.assertRaises(OperationPostponed, repos_resource.put, request, 'mock_repo')

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        mock_factory.repo_importer_manager.return_value.get_importers.return_value = mock_importer
+        repo_importers = RepoImportersView()
+        response = repo_importers.get(mock_request, 'mock_repo')
+        mock_resp.assert_called_once_with(mock_importer)
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_CREATE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: this mock is not needed

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        repos_resource = RepoResourceView()
+        response = repos_resource.put(mock_request, 'mock_repo')
+
+        mock_resp.assert_called_once_with({'mock': 'repo'})
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_UPDATE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: not needed

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        repo_importer = RepoImporterResourceView()
+        response = repo_importer.get(mock_request, 'mock_repo', 'mock_importer')
+
+        mock_resp.assert_called_once_with(mock_importer)
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: not needed

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        else:
+            raise AssertionError("MissingResource should be raised if importer does not exist")
+
+        self.assertEqual(response.http_status_code, 404)
+        self.assertTrue(response.error_code is error_codes.PLP0009)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_DELETE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: not need it, check this throughout the pr

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        self.assertEqual(response.http_status_code, 202)
+        mock_update_importer.apply_async_with_reservation.assert_called_once_with(
+            mock_tags.RESOURCE_REPOSITORY_TYPE, 'mock_repo', ['mock_repo'],
+            {'importer_config': 'test'}, tags=mock_task
+        )
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_UPDATE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: remove unnecessary mock like generate_json_response_with_pulp_encoder, tags and update_importer_config

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        self.assertEqual(response.http_status_code, 404)
+        self.assertTrue(response.error_code is error_codes.PLP0009)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_UPDATE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.repo_importer_manager.update_importer_config')
+    @mock.patch('pulp.server.webservices.views.repositories.tags')
```
ipanova: same here

server/test/unit/server/webservices/views/test_repositories.py
ipanova: this is what i was talking about https://github.com/pulp/pulp/pull/1742/files#diff-6e1391b92ad7c95e1b6f7b958e10bdddR972

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+            'some iso8601', 2, False
+        )
+        mock_resp.assert_called_once_with({'_id': 'mock_schedule', '_href': '/mock/path/'})
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_CREATE())
+    @mock.patch('pulp.server.webservices.views.repositories.generate_json_response')
+    @mock.patch('pulp.server.webservices.views.repositories.reverse')
+    @mock.patch('pulp.server.webservices.views.repositories.manager_factory')
```
ipanova: remove all not needed mock here

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+            'mock_repo', 'mock_importer', 'mock_schedule')
+        mock_resp.assert_called_once_with(None)
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_DELETE())
+    @mock.patch('pulp.server.webservices.views.repositories.generate_json_response')
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.manager_factory.repo_sync_schedule_manager')
+    def test_delete_sync_schedule_invalid_schedule(self, mock_manager, mock_resp):
```
ipanova: would you add a testcase where schedule_id is invalid ObjectId so InvalidValue is raised?

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+    def test_update_sync_schedule_no_schedule_param(self, mock_manager, mock_reverse, mock_resp):
+        """
+        Attempt to update a dschedueld sync while missing the required schedule parameter.
+        """
+
+        mock_request = mock.MagicMock()
+        mock_request.body = json.dumps({'some': 'data'})
+        sync_resource = RepoSyncScheduleResourceView()
+        response = sync_resource.put(mock_request, 'mock_repo', 'mock_importer', 'mock_schedule')
+        sync_resource.manager.update.assert_called_once_with(
```
ipanova: is not is supposed to be mock_manager.update?

server/pulp/server/webservices/views/repositories.py
```diff
@@ -0,0 +1,1219 @@
...
+    def post(self, request, repo_id):
+        """
+        Associate a distributor with a repository.
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param repo_id: The id of the repository to associate with
+        :type  repo_id: str
+
+        :raises pulp_exceptions.MissingResource: if repo_id does not exist
```
ipanova: remove this line

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_CREATE())
+    @mock.patch('pulp.server.webservices.views.repositories.generate_redirect_response')
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
+    @mock.patch('pulp.server.webservices.views.repositories.reverse')
+    @mock.patch('pulp.server.webservices.views.repositories.manager_factory')
+    def test_post_distributors(self, mock_factory, mock_rev, mock_resp, mock_redir):
+        """
+        Associate a distributor to a repository with minimal options.
```
ipanova: empty body cannot be minimal options here, because in this call there are some required params that cannot be omitted. http://fpaste.org/203748/27459187/

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+        repo_dist = RepoDistributorResourceView()
+        response = repo_dist.get(mock_request, 'mock_repo', 'mock_distributor')
+
+        mock_resp.assert_called_once_with(mock_dist)
+        self.assertTrue(response is mock_resp.return_value)
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_DELETE())
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.generate_json_response_with_pulp_encoder')
```
ipanova: remove this line

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+    @mock.patch('pulp.server.webservices.views.repositories.reverse')
+    @mock.patch('pulp.server.webservices.views.repositories.manager_factory')
+    def test_post_new_scheduled_publish(self, mock_factory, mock_reverse, mock_resp, mock_redir):
+        """
+        Create a new scheduled publish.
+        """
+
+        mock_request = mock.MagicMock()
+        mock_request.body = json.dumps({})
```
ipanova: body cannot be empty in this call

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -0,0 +1,1977 @@
...
+
+    @mock.patch('pulp.server.webservices.controllers.decorators._verify_auth',
+                new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.repositories.RepoPublishScheduleResourceView._get')
+    @mock.patch(
+        'pulp.server.webservices.views.repositories.manager_factory.repo_publish_schedule_manager')
+    @mock.patch('pulp.server.webservices.views.repositories.reverse')
+    def test_get_publish_schedule(self, mock_rev, mock_manager, mock_get):
+        """
+        Test retrieval of a single scheduled publis
```
ipanova: is not it supposed to be mock_manager.validate_distributor ?

ipanova: please update/add necessary testcase/s if you are going to implement in the code the try/except block

ipanova: Test regenerate content applicability with empty repo_criteria

ipanova: this is not needed

ipanova: raise AssertionError('MissingValue should be raised if repo_criteria is None')

ipanova: i think you can get rid of mock_get_repo() just by doing:
mock_factory.repo_query_manager.return_value.get_repository.side_effect = MissingResource()

ipanova: you can directly write here json.dumps({'source_repo_id': 'mock_source_repo'}) since you are not using source_repo var anywhere else

ipanova: update the message

ipanova: not needed

ipanova: update the message

ipanova:
@asmacdo  Well done! Seems like after this PR you will be master level 80 of such huge PRs. ^_^
i think i am done with your pr, please take a look at all my comments, update the code + tests then for the next round i will re-review again code + tests + manual testing.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: I can see what you mean, this testing does not guarantee against the wrong arguments are being passed to reverse. I am leaning towards agreeing with you, but since it will take some time to make the change I want to think about it a little more. Also, I think that we may have talked about this early on. @bmbouter, do you remember?

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: Yeah, I will do that here. The others however would be redundant since they would test exactly the same thing.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: since the only difference is the value of `include_importers` and `include_distributors`, I think the response body is not necessary.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: I am not sure what you mean, I think that this is that test?

It is also covered by the hand testing:
invalid schedule_id
*includes bugfix [redmine-756](https://pulp.plan.io/issues/756)
http://fpaste.org/197622/14262601/

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: `mock_manager.update` is not actually called. The manager is set in the parent class when `RepoSyncScheduleResourceView` is initialized, making this a different object. 

I could use `mock_manager().update` but I wanted to be clear that it was the mock that belongs to the `sync_resource` instance.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: I meant minimal to pass through the view. The missing params would be caught in the manager, which is mocked.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: same as above -- I wanted to be clear that this is a method that belongs to the instance.

asmacdo: good catch.

asmacdo: I actually need source_id for `mock_get_repo`.

asmacdo: The problem is that this `repo_query_manager.get_repository` is run on dest_repo_id and source_repo_id. I want to make sure that dest_repo_id does not raise, but source_id does. Its not pretty, but it works :)

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: Wow. These are everywhere! Thanks for noticing :+1:

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: mock_resp is used here. the others were not necessary.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: side_effect can be used to execute a function rather than returning the function, and it overrides return_value. In this case, I wanted `mock_merge(name, manager, repos)` to return repos.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: I think it is needed, can you elaborate?

server/pulp/server/webservices/views/repositories.py
asmacdo: Done.

server/test/unit/server/webservices/views/test_repositories.py
asmacdo: This change has been made, but a GitHub bug has moved the comment.

server/test/unit/server/webservices/views/test_repositories.py
bmbouter: This is related to how pure of a unit testing methodology we want to have. Pure unit tests don't rely on third party units for their correctness to happen. As such, a unit under test should be fully mocked so that it doesn't make calls outside of the unit under test. Note that unit tests cannot guarantee overall correctness so that is OK. It only asserts correctness on the component under test.

Not mocking `reverse` would be more of an integration test not a unit test because it would test the integration between two components (the unit under test and the reverse component). I advise a pure unit test methodology and also having completely separate set of integration tests. We have those integration tests separate so I think these should be pure unit tests.

I think `reverse` should be mocked because then we can...
* assert that it is called
* assert that we are passing it the right parameters
* assert that the result is returned as _href. Currently `reverse.return_value` returns a string but it could be better to let it return a mock() object as mocks normally do and then assert that the _href response `is reverse.return_value`.

asmacdo: @ipanova, I will redo these tests with the strategy above. I think that the addition of `assert_called_once_with()`  should cover the case that you were concerned about.

ipanova:
@asmacdo  all tests are passing but coverage is at 99%. https://github.com/pulp/pulp/pull/1742/files#diff-69de4359dc90d2ab726dbca11e4f32b4R1081 seems like you forgot to add the tescase after some changes in the code, please add this test and feel free to remove the webpy code and then merge!

server/pulp/server/webservices/views/repositories.py
asmacdo: Redmine 826: [:beetle:](https://pulp.plan.io/issues/826)

asmacdo:
ok test

asmacdo:
@ipanova, bug filed and additional test case added. The second commit is for your convenience, I will squash it after you (and pulp-bot) approve.

PR #1743, opened 2015-03-24
convert "workers" collection to mongoengine

beav:
This patch converts the `workers` collection to mongoengine. It also moves some
functionality into tasks.py that previously lived in the worker manager.

One thing to note, it appears that the existing workers collection did not have
a `_ns` field. Thus, I did not add this into the new collection.

This commit includes a fix to CriteriaQuerySet's `update()` method which was
incorrectly firing the `post_save()` hook on TaskStatus objects, even when the
Document was not a task status.

A spec file requirement for `python-blinker` was also added to support
mongoengine signals.


beav:
I get two unexpected results in the unit tests when run on mongoengine 0.7. I am working on this now.

beav:
big thaks to @rbarlow for helping me sort out a tricky issue that was caught by @pulpbot

server/pulp/server/async/scheduler.py
```diff
@@ -229,9 +228,7 @@ def check_workers(self):
             'Looking for workers missing for more than %s seconds') % self.WORKER_TIMEOUT_SECONDS
         _logger.debug(msg)
         oldest_heartbeat_time = datetime.utcnow() - timedelta(seconds=self.WORKER_TIMEOUT_SECONDS)
-        worker_criteria = Criteria(filters={'last_heartbeat': {'$lt': oldest_heartbeat_time}},
-                                   fields=('_id', 'last_heartbeat'))
-        worker_list = list(resources.filter_workers(worker_criteria))
+        worker_list = Worker.objects(last_heartbeat__lt=oldest_heartbeat_time)
```
mhrivnak: So much nicer!!!

server/pulp/server/async/tasks.py
```diff
@@ -84,6 +84,70 @@ def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
...
+    :raises NoWorkers: If all workers have reserved_resource entries associated with them.
+
+    :returns:          The Worker instance that has no reserved_resource
+                       entries associated with it.
+    :rtype:            pulp.server.db.model.resources.Worker
+    """
+
+    # Build a mapping of queue names to Worker objects
+    workers_dict = dict((worker['name'], worker) for worker in Worker.objects())
+    worker_names = [name for name in workers_dict.keys()]
```
mhrivnak: You can skip the list comprehension and just use ``workers_dict.keys()``

server/pulp/server/async/tasks.py
```diff
@@ -84,6 +84,70 @@ def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
...
+    reserved_names = [r['worker_name'] for r in ReservedResource.get_collection().find()]
+
+    # Find an unreserved worker using set differences of the names, and filter
+    # out workers that should not be assigned work.
+    # NB: this is a little messy but set comprehensions are in python 2.7+
```
mhrivnak: I actually think this is very clear and readable.

server/pulp/server/async/tasks.py
```diff
@@ -108,17 +171,16 @@ def _delete_worker(name, normal_shutdown=False):
         _logger.error(msg)
 
     # Delete the worker document
-    worker_list = list(resources.filter_workers(Criteria(filters={'_id': name})))
-    if len(worker_list) > 0:
-        worker_document = worker_list[0]
-        worker_document.delete()
+    worker = Worker.objects(name=name).first()
```
mhrivnak: I think you can just call ``.delete()`` instead of ``.first()``, and ditch the lines below.

server/pulp/server/db/model/workers.py
```diff
@@ -0,0 +1,40 @@
...
+class Worker(Document):
+    """
+    Represents a worker.
+
+    This inherits from mongoengine.Document and defines the schema for the documents
+    in the worker collection.
+
+    :ivar worker_name:    worker name, in the form of "worker_type@hostname"
+    :type worker_name:    basestring
+    :ivar last_heartbeat:  ISO8601 representation of the time the task started executing
```
mhrivnak: This looks like the wrong description. Copy-and-paste error perhaps?

server/pulp/server/db/model/workers.py
```diff
@@ -0,0 +1,40 @@
...
+    :type last_heartbeat:  basestring
+    """
+    name = StringField(primary_key=True)
+    last_heartbeat = DateTimeField()
```
mhrivnak: Just checking- this is going to be a datetime object on an instance, right?

server/pulp/server/async/scheduler.py
beav: :panda_face:

(options for emoji related to mongoengine are lacking)

server/pulp/server/async/tasks.py
beav: updated :key: :musical_keyboard:

server/pulp/server/async/tasks.py
beav: updated

server/pulp/server/db/model/workers.py
beav: good catch, `worker_name` should be just `name`. Both got written in the db as `_id`.

server/pulp/server/db/model/workers.py
beav: correct, this is not an iso8601 string but a "regular" datetime.datetime. I made a copy/paste error in the docstring above.

beav:
@mhrivnak pr is updated for comments, reassigning back for review

PR #1745, opened 2015-03-25
Fixes epel url changing

Jmainguy:
RHEL/Centos do not need to grab the rpm from the website, they can just run yum install epel-release

### DIFF
--- a/docs/user-guide/installation.rst
+++ b/docs/user-guide/installation.rst
@@ -88,27 +88,18 @@ Repositories
    If you would like to install the RHEL 5 Client for Pulp, please use the
    `RHEL 5 repository <http://repos.fedorapeople.org/repos/pulp/pulp/rhel5-pulp.repo>`_.
 
-2. For RHEL and CentOS systems, the EPEL repositories are required. Following commands will add the
+.. note::
+   The epel-release package is provided in the ``extras`` repositry for RHEL.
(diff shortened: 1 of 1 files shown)


beav:
This didn't work for me on RHEL. Is it a CentOS-only thing?

Jmainguy:
@beav updated PR to reflect that the Extras repo needs to be enabled for yum install epel-release to work.

beav:
Can you confirm the "extras" repo is available in el6? We may need a different step for 6 vs 7.

Jmainguy:
I can confirm it worked on centos6 and centos7. Would need access to a normal rhel subscribed host to check rhel6 and rhel7, although I imagine they should be identical to centos in this regard.

beav:
I enabled `rhel-6-server-extras-rpms` but the `epel-release` rpm doesn't seem to be in there.

rbarlow:
On 03/26/2015 11:00 AM, Chris Duryee wrote:
> I enabled |rhel-6-server-extras-rpms| but the |epel-release| rpm doesn't
> seem to be in there.

I was under the impression that CentOS has that package, but RHEL does
not. I'm not sure if that is true or not.

-- 
Randy Barlow

Jmainguy:
Confirmed, this does not work on RHEL6, only Centos6 and Centos7. Closing this out.

PR #1748, opened 2015-03-27
pulp-qpid-ssl-cfg - More robust hostname default, prompt for input, support relative paths for INST_DIR

vitaminmoo:
Remote hosts are more likely to be connecting to qpidd using the fully
qualified hostname, especially in complex environments where not all
machines share the same domain/search suffix.

Sometimes the fqdn is not properly set up, and the hostname contains the
entire fqdn. This change addresses this, while also allowing the user to
specify in case a DNS alias or IP address is preferred.

It also fills the example configs out with the hostname used as the CN
in the cert, as that is the only string that will actually work due to
server certificate verification.

Support relative paths for INST_DIR

Previously relative path usage was not supported for INST_DIR due to the
script cding to the temp directory. Use readlink to make it an absolute
path.

### DIFF
--- a/server/bin/pulp-qpid-ssl-cfg
+++ b/server/bin/pulp-qpid-ssl-cfg
@@ -82,7 +82,7 @@ echo ""
 echo "Please enter the hostname clients will use to connect to the qpid server."
 echo "If not specified, '$HOST' will be used."
 echo ""
-read -sp "Enter a hostname:" ans
+read -p "Enter a hostname:" ans
 if [ "${#ans}" -gt 0 ]
 then
   HOST=$ans
--- a/server/bin/pulp-qpid-ssl-cfg
+++ b/server/bin/pulp-qpid-ssl-cfg
@@ -14,7 +14,11 @@
 #
 
 DIR="/tmp/tmp$RANDOM"
-HOST=`hostname`
+HOST=`hostname -f`
+if [ -z "$HOST" ]
+then
+  HOST=`hostname`
+fi
 PWDFILE="password"
 SEEDFILE="seed"
 INST_DIR='/etc/pki/pulp/qpid'
@@ -38,7 +42,7 @@ echo ""
 read -p "Enter a directory [$INST_DIR]:" ans
 if [ "${#ans}" -gt 0 ]
 then
-  INST_DIR=$ans
+  INST_DIR=`readlink -f "$ans"`
 fi
 echo $INST_DIR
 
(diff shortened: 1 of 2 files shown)


dkliban:
ok test

dkliban:
ok to test

barnabycourt:
ok to test

barnabycourt:
ok to test

server/bin/pulp-qpid-ssl-cfg
```diff
@@ -14,7 +14,11 @@
 #
 
 DIR="/tmp/tmp$RANDOM"
```
jortel: Recommend a space here and
```
# prompt user for hostname
```
Here for consistency.

server/bin/pulp-qpid-ssl-cfg
```diff
@@ -14,7 +14,11 @@
...
-HOST=`hostname`
+HOST=`hostname -f`
+if [ -z "$HOST" ]
+then
+  HOST=`hostname`
+fi
```
jortel: Recommend a blank link here.

server/bin/pulp-qpid-ssl-cfg
```diff
@@ -14,7 +14,11 @@
 #
 
 DIR="/tmp/tmp$RANDOM"
-HOST=`hostname`
```
jortel: Since you have introduced logic here, I'd like to see this moved under the ``setup`` section with the others.  This part of the script is for static variables.

server/bin/pulp-qpid-ssl-cfg
```diff
@@ -14,7 +14,11 @@
...
+HOST=`hostname -f`
+if [ -z "$HOST" ]
```
jortel: Why the logic here?

server/bin/pulp-qpid-ssl-cfg
```diff
@@ -73,6 +77,19 @@ then
   echo "Using CA key: $CA_KEY_PATH"
 fi
 
+# prompt user for this server's hostname, defaulting to the local fqdn
+echo ""
+echo "Please enter the hostname clients will use to connect to the qpid server."
+echo "If not specified, '$HOST' will be used."
+echo ""
+read -sp "Enter a hostname:" ans
```
jortel: Suggest ``read -p`` without the ``-s`` so the user can see what they're typing.

jortel:
Nice patch.  Thanks!

server/bin/pulp-qpid-ssl-cfg
vitaminmoo: It's pretty common in my experience for the fqdn to not resolve properly locally, resulting in `hostname -f` outputting an error to STDERR and nothing to STDOUT. In this case, there are two situations I can think of:

- Non-fqdn hostname contains the fqdn
- Non-fqdn hostname contains just a shortname
- Hostname is completely misconfigured.

This attempts to handle the possibilities as gracefully as possible by preferring the DNS-resolved FQDN, and falling back to what the system thinks its hostname is, which may actually be resolvable within the same domain/search suffix.

jortel:
LGTM

jortel:
Tested on F21.

jortel:
Just remembered to check the documentation for this script.  Please update the documentation to reflect changes to the script.

https://github.com/pulp/pulp/blob/master/docs/user-guide/qpid.rst

vitaminmoo:
Docs are updated.

jortel:
Thank You!
LGTM

PR #1750, opened 2015-03-30
Allow passwordless mongodb connections to be configured

beav:
MongoDB allows passwordless logins. However, Pulp was checking that both the
username *and* password were set before attempting a login.

This patch alters the user/pass check to fail only if a password is set sans
username. If a username is set sans password, a MongoDB connection will be
attempted.

Note that all of the auth still happens in Mongo. This was just a "pre-check"
that Pulp was doing to give a clear error message.

fixes #708

### DIFF
--- a/server/pulp/server/db/connection.py
+++ b/server/pulp/server/db/connection.py
@@ -84,13 +84,13 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
         # attempt to authenticate to the database
         username = config.config.get('database', 'username')
         password = config.config.get('database', 'password')
-        if username and password:
+        if username:
             _logger.debug(_('Attempting username and password authentication.'))
             connection_kwargs['username'] = username
             connection_kwargs['password'] = password
-        elif (username and not password) or (password and not username):
-            raise Exception(_("The server config specified username/password authentication but "
-                            "is missing either the username or the password"))
+        elif password and not username:
+            raise Exception(_("The server config specified a database password, but is "
+                              "missing a database username."))
 
         shadow_connection_kwargs = copy.deepcopy(connection_kwargs)
         if connection_kwargs.get('password'):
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/connection.py
```diff
@@ -84,13 +84,13 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
         username = config.config.get('database', 'username')
         password = config.config.get('database', 'password')
-        if username and password:
+        if username:
             _logger.debug(_('Attempting username and password authentication.'))
             connection_kwargs['username'] = username
             connection_kwargs['password'] = password
-        elif (username and not password) or (password and not username):
+        elif (password and not username):
             raise Exception(_("The server config specified username/password authentication but "
```
bmbouter: I'll suggest the error should read:  "The server config specified a database password, but is missing a database username."

server/pulp/server/db/connection.py
```diff
@@ -84,13 +84,13 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
         # attempt to authenticate to the database
         username = config.config.get('database', 'username')
         password = config.config.get('database', 'password')
-        if username and password:
+        if username:
             _logger.debug(_('Attempting username and password authentication.'))
             connection_kwargs['username'] = username
             connection_kwargs['password'] = password
-        elif (username and not password) or (password and not username):
+        elif (password and not username):
```
bmbouter: I don't think this needs the () around it anymore. Previously that enforced an order of evaluation, but we can use default left to right evaluation here.

bmbouter:
Can a test be added that asserts that setting a password without specifying a username will raise an exception. Otherwise that branch lack test coverage.

bmbouter:
I think the three comments here should be implemented, but then LGTM.

server/pulp/server/db/connection.py
beav: updated :tennis:

server/pulp/server/db/connection.py
beav: updated, good catch :fish:

beav:
A test for password but no username already exists in `TestDatabaseAuthentication.test_initialize_password_no_username()`, it is not on the diff.

beav:
Marking as LGTM per @bmbouter 's comment.

PR #1759, opened 2015-04-02
make critera required on repo associate and unassociate

asmacdo:
If criteria is allowed to default to `None`, it is initialized as an empty criteria query:

`{'criteria': {}}` which returns all possible results. This is unexpected, so we should simply enforce that criteria must exist.

closes #826

### DIFF
--- a/server/pulp/server/webservices/views/repositories.py
+++ b/server/pulp/server/webservices/views/repositories.py
@@ -1130,7 +1130,7 @@ class RepoAssociate(View):
                 invalid_criteria.add_child_exception(e)
                 raise invalid_criteria
         else:
-            criteria = None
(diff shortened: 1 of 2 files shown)


mhrivnak:
Since this hasn't been triaged yet, I think we should hold this PR until the bug finishes the triage process. I'd be less concerned if it was a backward-compatible change.

asmacdo:
@mhrivnak, works for me. Assigning to myself to prevent anyone from trying to review.

mhrivnak:
I suggest closing this for now, and make sure there's a link to it in the redmine issue. It can then be referenced whenever we're ready to make the backward-incompatible change.

PR #1761, opened 2015-04-02
Added upgrade instruction from 2.6.x

dkliban:


bmbouter:
Why do all tasks need to be cancelled for a 2.6 upgrade to the next version?

dkliban:
@bmbouter I copied that from our instructions for upgrading from 2.5 to 2.6. It was necessary there because of changes to the queues.  So perhaps we don't need to do that here.

bmbouter:
There was a specific change from 2.5 -> 2.6 that required tasks to be cancelled due to changes in the tasking system. Generally from 2.6 forward, tasks should not need to be cancelled.

dkliban:
ok test

docs/user-guide/release-notes/master.rst
```diff
@@ -55,6 +55,27 @@ Known Issues
 Upgrade Instructions for 2.6.x --> master
 -----------------------------------------
 
+Upgrade the packages using:
```
bmbouter: you can make these `:::` and it will allow you to remove all this whitespace.

docs/user-guide/release-notes/master.rst
```diff
@@ -55,6 +55,27 @@ Known Issues
...
+
+::
+
+    sudo -u apache pulp-manage-db
+
+After migrating the database reload the systemd process using:
+
+::
+
+    sudo systemctl daemon-reload
```
bmbouter: I think this will fail on EL6 so it needs to say something like "if you are running systemd ....". Even if it doesn't fail it's not necessary on EL6. Maybe add it as a ..note block or something since it's not a required step for all platforms.

bmbouter: Also this should not generally be needed, is there something about a change that is only on master currently that requires it? If so can you link to it?

dkliban: https://github.com/pulp/pulp/commit/e6ea3131796df60de3f7e3b98acc2ef108f517f7

The changes to the manage workers scripts require the reloading of systemd.

docs/user-guide/release-notes/master.rst
bmbouter: I wasn't very clear here. I mean replace `Upgrade the packages using:` with `Upgrade the packages using:::`

The `:::` additionally below will show as a single colon on a line when rendered.

docs/user-guide/release-notes/master.rst
```diff
@@ -55,6 +55,22 @@ Known Issues
 Upgrade Instructions for 2.6.x --> master
 -----------------------------------------
 
+Upgrade the packages using:
+:::
```
bmbouter: I recommended the wrong fix in my earlier review. This doesn't compile correctly either. Try using `::` throughout you changes like this:

```
Upgrade the packages using::

    sudo yum update
```

docs/user-guide/release-notes/master.rst
```diff
@@ -55,6 +55,22 @@ Known Issues
...
+    sudo yum update
+
+After yum completes you should migrate the database using:
+:::
+    sudo -u apache pulp-manage-db
+
+.. note::
```
bmbouter: Nice note

docs/user-guide/release-notes/master.rst
```diff
@@ -55,6 +55,22 @@ Known Issues
...
+    If using systemd, you need to reload the systemd process before restarting services. This can
+    be done using::
+        sudo systemctl daemon-reload
```
bmbouter: This looks great compiled

bmbouter:
After the colon fixes, LGTM

PR #1763, opened 2015-04-02
Add warning in nodes setup to not copy/paste values

beav:
If `oauth_key` is used instead of `key` under `[parent_oauth]`, users will get
a confusing error during node sync.

fixes #528


ipanova:
@beav  do you think it will be also valuable to add this note directly into the nodes.conf?

beav:
@ipanova I thought this as well, but I did not want to make a conf file change in 2.6.1 since it would result in a `.rpmnew` file being created.

PR #1764, opened 2015-04-03
Some consumer history event types are not implemented

ipanova:
Closes # 829
https://pulp.plan.io/issues/829

### DIFF
--- a/server/test/unit/server/managers/consumer/test_agent.py
+++ b/server/test/unit/server/managers/consumer/test_agent.py
@@ -231,6 +231,8 @@ class TestAgentManager(TestCase):
         mock_context.assert_called_with(consumer, task_id=task_id, consumer_id=consumer['id'])
         mock_profiler.install_units.assert_called_with(consumer, [unit], options, {}, ANY)
         mock_agent.install.assert_called_with(mock_context.return_value, [unit], options)
+        mock_factory.consumer_history_manager().record_event.assert_called_with(
+            consumer['id'], 'content_unit_installed', {'units': [unit]})
 
     @patch('pulp.server.managers.consumer.agent.uuid4')
     @patch('pulp.server.managers.consumer.agent.TaskStatus')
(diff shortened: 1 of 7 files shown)


bmbouter:
I don't know if the 'Closes # 829' will be recognized by Redmine with the space and the capital. I would use 'closes #829' in the commit message. Github puts the commit message as the PR comment, but Redmine only looks at the commit message.

server/test/unit/server/managers/consumer/test_agent.py
```diff
@@ -231,6 +231,8 @@ def test_install_content(self, *mocks):
         mock_context.assert_called_with(consumer, task_id=task_id, consumer_id=consumer['id'])
         mock_profiler.install_units.assert_called_with(consumer, [unit], options, {}, ANY)
         mock_agent.install.assert_called_with(mock_context.return_value, [unit], options)
+        mock_factory.consumer_history_manager().record_event.assert_called_with(
```
bmbouter: I think you need a similar mock assertion for the `history_manager.record_event(consumer_id, 'content_unit_uninstalled', {'units': units})` in agent.py

bmbouter:
I see that you added test coverage for the rest of the consumer history events, great job.

bmbouter:
Historically Pulp has written unit tests that actually are integration tests. We are trying to move more towards unit testing and allowing the integration test suite to test the end to end behavior of Pulp. Almost all of these tests are integration tests because they perform an action and then make assertions about the state of the database.

I believe writing unit tests can be a lot simpler than integration tests because using mock you only need to make assertions on what the unit under test actually does. You've written some really good integration tests here so I don't think it's worth it to rewrite anything, but I'm pointing it out for future test writing.

bmbouter:
The one additional test for `content_unit_uninstalled` needs to be added, but after that LGTM.

server/test/unit/server/managers/consumer/test_agent.py
ipanova: yes, you are right - i somehow missed that.

PR #1765, opened 2015-04-03
Removes dontaudit rules from pulp-server SELinux policy

bmbouter:
closes #564

### DIFF
--- a/server/selinux/server/pulp-server.te
+++ b/server/selinux/server/pulp-server.te
@@ -4,7 +4,6 @@ policy_module(pulp-server, 0.0.0)
 
 require {
     type httpd_t;
-    type rpm_var_lib_t;
 }
 
 type pulp_cert_t;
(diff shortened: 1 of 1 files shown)


beav:
@bmbouter  just to confirm my understanding of this issue, the `dontaudit` rule was from ~2012, and we are removing it since the selinux policy rewrite does not require it anymore?

bmbouter:
@beav Yes the restructuring is what allows us to remove it.

The `dontaudit` rule was probably unwise to add from the start. Separate from that, the SELinux restructuring we've done lets us remove it now. At that time, all Pulp operations ran inside of `httpd` which was constrained b the `pulp-server` SELinux policy which had the `dontaudit` rule added to it. Now sync runs inside of `pulp_workers` which is governed  by the `pulp-celery` SELinux policy. The right permissions have been granted there which allows the necessary operations. This is why removing the `dontaudit` doesn't cause the AVC denials we were seeing before.

beav:
LGTM :closed_lock_with_key: :penguin:

PR #1767, opened 2015-04-06
Allow setting the broker_login_method.

hgschmie:
When using RabbitMQ with SSL client certificates, it is necessary to use
the rabbitmq_auth_mechanism_ssl which uses the common_name (CN) of the
client certificate as user name and then use SASL EXTERNAL to have
RabbitMQ pick up the configuration.

Unfortunately, by default, pulp only supports the default (AMQPLAIN)
authentication and it can not be configured.

This change adds a new configuration parameter to the tasks section
which allows setting the requested authentication mechanism through the
celery BROKER_LOGIN_METHOD configuration setting.

This allows using RabbitMQ with strict SSL client certificates:

```erlang
[
  {rabbit, [
    {ssl_listeners, [5671]},
    {auth_mechanisms, ['EXTERNAL']},
    {ssl_options, [
      {cacertfile,           '.../cacertfile'},
      {certfile,             '.../cert.crt'},
      {keyfile,              '.../cert.key'},
      {verify,               verify_peer},
      {fail_if_no_peer_cert, true}
    ]},
    {ssl_cert_login_from,  common_name}]
  }
].
```

and adding the CN values from the client certificates as users to RabbitMQ.

### DIFF
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -295,6 +295,10 @@
 # certfile: The absolute path to the PEM encoded certificate used for authentication to the message
 #     bus. The default value is '/etc/pki/pulp/qpid/client.crt'.
 #
+# broker_login_method: Select the login method used to connect to the broker. This should be left
+#     unset (use the broker default) except in special cases (e.g. do SSL client certificate
+#     authentication. For RabbitMQ with SSL client certificates, this needs to be 'EXTERNAL'.
+#
 
 [tasks]
 # broker_url: qpid://guest@localhost/
--- a/server/pulp/server/async/celery_instance.py
+++ b/server/pulp/server/async/celery_instance.py
@@ -90,5 +90,7 @@ def configure_SSL():
         }
         celery.conf.update(BROKER_USE_SSL=BROKER_USE_SSL)
 
+        if config.has_option('tasks', 'broker_login_method'):
+            celery.conf.update(BROKER_LOGIN_METHOD=config.get('tasks', 'broker_login_method'))
 
 configure_SSL()
--- a/server/test/unit/server/async/test_celery_instance.py
+++ b/server/test/unit/server/async/test_celery_instance.py
@@ -140,6 +140,40 @@ class TestCeleryInstanceSSLConfig(unittest.TestCase):
             celery_instance.configure_SSL()
             mock_celery.conf.update.assert_called_once_with(BROKER_USE_SSL=EXPECTED_BROKER_USE_SSL)
 
+    @mock.patch('pulp.server.async.celery_instance.celery')
+    @mock.patch('pulp.server.async.celery_instance.config.getboolean')
+    def test_configure_SSL_when_broker_login_method_present(self, mock_getboolean, mock_celery):
+        """
+        Make sure that the Celery config has BROKER_LOGIN_METHOD set if given in the configuration.
+        """
+        mock_getboolean.return_value = True
+        mock_cacert = mock.Mock()
+        mock_keyfile = mock.Mock()
+        mock_certfile = mock.Mock()
+
+        CONFIG_OVERRIDE = {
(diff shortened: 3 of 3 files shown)


beav:
ok test

beav:
@hgschmie thanks for the patch! :koala:

It looks like the `test_configure_SSL_when_broker_login_method_present` test is failing:

```python
Expected call: update(BROKER_LOGIN_METHOD='EXTERNAL')
Actual call: update(BROKER_USE_SSL={'ca_certs': <Mock id='91207312'>, 'certfile': <Mock id='91085200'>, 'keyfile': <Mock id='91084624'>, 'cert_reqs': 2})
```

Can you take a look at this failure and resubmit the patch?

server/etc/pulp/server.conf
```diff
@@ -295,6 +295,10 @@
 # certfile: The absolute path to the PEM encoded certificate used for authentication to the message
 #     bus. The default value is '/etc/pki/pulp/qpid/client.crt'.
 #
+# broker_login_method: Select the login method used to connect to the broker. This should be left
+#     unset (use the broker default) except in special cases (e.g. do SSL client certificate
+#     authentication. For RabbitMQ with SSL client certificates, this needs to be 'EXTERNAL'.
+#
```
bmbouter: Please add a commented out entry below in the [tasks] section. By convention all server.conf entries use a commented entry to show its default.

bmbouter:
Please add an empty string as default entry in the defaults for config.py. I think you want to put it [here](https://github.com/bmbouter/pulp/blob/master/server/pulp/server/config.py#L89). At some point in the future, we hope to have empty config files so these defaults set us up for that.

server/etc/pulp/server.conf
```diff
@@ -295,6 +295,10 @@
 # certfile: The absolute path to the PEM encoded certificate used for authentication to the message
 #     bus. The default value is '/etc/pki/pulp/qpid/client.crt'.
 #
+# broker_login_method: Select the login method used to connect to the broker. This should be left
+#     unset (use the broker default) except in special cases (e.g. do SSL client certificate
+#     authentication. For RabbitMQ with SSL client certificates, this needs to be 'EXTERNAL'.
```
bmbouter: s/authentication/authentication)./

bmbouter: For clarity I'll suggest:  s/certificates/certificate verification/

server/pulp/server/async/celery_instance.py
```diff
@@ -90,5 +90,7 @@ def configure_SSL():
         }
         celery.conf.update(BROKER_USE_SSL=BROKER_USE_SSL)
 
+        if config.has_option('tasks', 'broker_login_method'):
```
bmbouter: Once the default gets added this check will need to check if the user has set it differently then the default.

server/etc/pulp/server.conf
bmbouter: The last sentence should probably be moved to a new section on RabbitMQ with SSL and Client Cert Verification at the bottom of this doc:  http://pulp.readthedocs.org/en/latest/user-guide/broker-settings.html#rabbitmq-with-ssl

bmbouter:
@hgschmie This is a good idea! Thanks for the PR.

I'm done reviewing this PR so I'm assigning it back to you. That is how we keep track of who has the next step. I've suggested several changes; they are all pretty easy I think. Leave a comment when you push your next commit so I can re-review. Feel free to leave questions here also.

Thanks again for the contribution!

bmbouter:
@hgschmie Thanks for the contribution. This was rebased against the 2.6-dev and 2.7-testing merge base and updated per the PR comments here. I've left your initial commit in tact and committed on top of it. I'm closing this and it will be merged with this PR:  https://github.com/pulp/pulp/pull/2002

PR #1768, opened 2015-04-07
clear the log handler on shutdown.

jortel:
The logging.shutdown() does not clear the handlers.  In normal operation this is not an issue but can cause issues for unit tests.  Also more complete and symmetrical with start_logging()


jortel:
Re-issue against 2.6-dev.

PR #1769, opened 2015-04-07
clear log handlers in stop_logging().

jortel:
More complete and symmetrical with start_logging().  Also, fixes issue with unit tests.


beav:
LGTM

PR #1771, opened 2015-04-08
Allow pulp CLI to print sub_errors from polled tasks

asmacdo:
When the API returns errors that the CLI is polling, only the top level errors were being printed. This prints sub_errors as well. One place that this was needed was related to https://pulp.plan.io/issues/291, where the information returned about why a validation failed was not being displayed. 

This PR also fixes some broken tests that were exposed by the changes here. Apparently, `mock.MagicMock.reset_mock()` does not reset `mock.MagicMock.side_effect`, so the state was not being properly cleaned up between tests.

[#828](https://pulp.plan.io/issues/828)


client_lib/pulp/client/commands/polling.py
```diff
@@ -296,6 +296,19 @@ def succeeded(self, task):
...
+    def _render_coded_error(self, e):
+        """
+        Recursively render a coded error and the errors contained by it.
+
+        :param e: error to display
+        :type  e: dictionary containing error information.
+        """
+        malformed_msg = _('Request error does not contain a description, '
+                          'please check client logs for more information.')
+        self.prompt.render_failure_message('%s' % (e.get('description', malformed_msg)))
```
mhrivnak: The string substitution doesn't seem to gain anything for you in this case.

client_lib/pulp/client/commands/polling.py
```diff
@@ -296,6 +296,19 @@ def succeeded(self, task):
...
         self.prompt.render_success_message(msg, tag='succeeded')
 
+    def _render_coded_error(self, e):
+        """
+        Recursively render a coded error and the errors contained by it.
+
+        :param e: error to display
+        :type  e: dictionary containing error information.
+        """
+        malformed_msg = _('Request error does not contain a description, '
```
mhrivnak: Replace the comma with a semicolon, or make it a period and start a new sentence on the next line.

server/test/unit/test_repo_distributor_manager.py
```diff
@@ -420,13 +420,16 @@ def test_update_validate_exception(self):
...
+            pass
+
+        mock_plugins.MOCK_DISTRIBUTOR.validate_config.side_effect = TestException()
 
         # Test
         try:
             self.distributor_manager.update_distributor_config('elf', dist_id, {})
             self.fail('Exception expected')
-        except exceptions.PulpDataException:
+        except TestException:
```
mhrivnak: If you feel like cleaning this up a touch, even though it's not directly related to your bug fix, it's a good use case for ``self.assertRaises``

asmacdo:
ok test

asmacdo:
@mhrivnak, ready for re-review.

client_lib/test/unit/client/commands/test_polling.py
```diff
@@ -311,6 +311,36 @@ def test_failed_task(self):
...
+        mock_render_coded.assert_called_once_with(mock_task.error)
+
+    def test__render_coded_error_called_recursively(self):
+        """
+        Pass an error with sub_errors, they should all be rendered.
+        """
+        mock_error = {'description': 'this is a test of the emergency broadcast system',
+                      'sub_errors': [{'description': 'beep1'}, {'description': 'beep2'}]}
+
+        # Though a context manager would be prefered, this pattern is necessary to run on python 2.4
```
asmacdo: @mhrivnak, could you double check that this is the correct pattern to replace `with mock.patch()` for 2.4 compatibility?

asmacdo:
@mhrivnak, this one is still floating out there. Mind having another look?

client_lib/test/unit/client/commands/test_polling.py
mhrivnak: the whole command instance gets re-created for each test (in ``setUp()``), so I wouldn't worry about un-doing the patch.

PR #1773, opened 2015-04-09
Removes docs on Reopened and Security tags since they are deleted

bmbouter:
These tags are no longer available on Redmine so they need to not be recommended in the docs.


beav:
LGTM :koala:

beav:
@bmbouter this looks like it's against `2.6-release` and not `2.6-testing`, can you confirm this is intended?

bmbouter:
@beav Yes this is intended because the change is already live on pulp.plan.io.

PR #1774, opened 2015-04-09
Ensure that if a publish_report is failed, the task is marked as a failur...

barnabycourt:
...e.

fixes #682

### DIFF
--- a/server/pulp/server/managers/repo/publish.py
+++ b/server/pulp/server/managers/repo/publish.py
@@ -26,15 +26,15 @@ from gettext import gettext as _
 
 from celery import task
 
-from pulp.common import dateutils, constants
+from pulp.common import dateutils, error_codes, constants
 from pulp.common.tags import resource_tag, RESOURCE_REPOSITORY_TYPE, action_tag
 from pulp.plugins.loader import api as plugin_api
-from pulp.plugins.loader import exceptions as plugin_exceptions
 from pulp.plugins.model import PublishReport
 from pulp.plugins.conduits.repo_publish import RepoPublishConduit
 from pulp.plugins.config import PluginCallConfiguration
 from pulp.server.db.model.repository import Repo, RepoDistributor, RepoPublishResult
-from pulp.server.exceptions import MissingResource, PulpExecutionException, InvalidValue
+from pulp.server.exceptions import MissingResource, PulpExecutionException, InvalidValue, \
+    PulpCodedException
 from pulp.server.managers import factory as manager_factory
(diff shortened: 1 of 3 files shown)


server/test/unit/server/managers/repo/test_publish.py
```diff
@@ -124,24 +124,26 @@ def test_publish_failure_report(self):
         mock_plugins.MOCK_DISTRIBUTOR.publish_repo.return_value = PublishReport(False, 'Summary of the publish', 'Details of the publish')
 
         # Test
-        report = self.publish_manager.publish('repo-1', 'dist-1', None)
+        try:
```
mhrivnak: with unittest2 (the PR for which hasn't been reviewed yet...), you can do basically this, which I think is a bit cleaner and more direct:

```
with self.assertRaises(PulpCodedException) as assertion:
  self.publish_manager.publish('repo-1', 'dist-1', None)
self.assertEqual(assertion.exception.error_code, error_codes.PLP0034)
```

No need to wait for unittest2, but it's worth knowing about.

server/test/unit/server/managers/repo/test_publish.py
```diff
@@ -124,24 +124,26 @@ def test_publish_failure_report(self):
...
-            self.assertEqual('mock-distributor', check_me['distributor_type_id'])
-            self.assertTrue(check_me['started'] is not None)
-            self.assertTrue(check_me['completed'] is not None)
-            self.assertEqual(RepoPublishResult.RESULT_FAILED, check_me['result'])
-            self.assertTrue(check_me['summary'] is not None)
-            self.assertTrue(check_me['details'] is not None)
-            self.assertTrue(check_me['error_message'] is None)
-            self.assertTrue(check_me['exception'] is None)
-            self.assertTrue(check_me['traceback'] is None)
+        check_me = entries[0]
```
mhrivnak: I like this change.

server/pulp/server/managers/repo/publish.py
```diff
@@ -34,7 +34,9 @@
 from pulp.plugins.conduits.repo_publish import RepoPublishConduit
 from pulp.plugins.config import PluginCallConfiguration
 from pulp.server.db.model.repository import Repo, RepoDistributor, RepoPublishResult
-from pulp.server.exceptions import MissingResource, PulpExecutionException, InvalidValue
+from pulp.common import error_codes
```
mhrivnak: It would be nice to group this with the other ``pulp.common`` imports above.

barnabycourt:
ok test

PR #1778, opened 2015-04-13
fix filters field so GET searches work

asmacdo:

closes #312


mhrivnak:
Unit test? If this was broken, then presumably it wasn't covered by an existing test.

asmacdo:
@mhrivnak, whoops, forgot to stage that file for my commit. Need more :coffee:

beav:
@asmacdo  can this go on `2.7-dev` instead?

PR #1780, opened 2015-04-13
Refactor repo auth to use entrypoints

beav:
Pulp's repo auth now allows for pluggable authentication methods. These can be
configured via a new entry point. The `repo_auth.wsgi` file was also refactored
to move code into a Python module to make testing easier.

More information is available in the 'contentauth' doc.

Additionally, the 'auth_handler_framework' was determined to be unused and has
been removed.

fixes #720


beav:
ok test

beav:
ok to test

beav:
ok test

previous test failed due to a github timeout

beav:
ok test

more timeouts during checkout

beav:
ok test

trying again to get all tests to succeed, one timed out again

beav:
ok test

looks like tests may be running now, trying again

beav:
ok test

trying again, will try once more in the afternoon if this times out

docs/dev-guide/contentauth.rst
```diff
@@ -0,0 +1,42 @@
...
+telling Pulp about this method via Python entry points.
+
+Note that **all** authenticators must return True in order to let the request
+through.
+
+For example, if you wanted to create a simple method that let everyone through
+but logged a message, you could do something like this:
+
+::
+    def authenticate(environ):
```
jortel: It would be helpful if the *environ* parameter was documented.

repoauth/pulp/repoauth/wsgi.py
```diff
@@ -0,0 +1,40 @@
...
+    :rtype:  Boolean
+    """
+
+    # If auth is disabled, then let the request continue. Note that this returns
+    # True if auth is disabled.
+    if auth_enabled_validation.authenticate(environ):
+        return True
+
+    # find all of the authenticator methods we need to try
+    authenticators = {}
```
jortel: Since I don't see the ep.name being used, I'm wondering if something like this might be simpler?

```
for method in [ep.load() for ep in iter_entry_points(group=AUTH_ENTRY_POINT)]:
    if not method(environ):
        return False
```

jortel:
LGTM

I know it's been convention in pulp to prefix all mocked object variable names with mock_ or sometimes fake_.  However, when looking a most tests (including yours here), this means that almost all variables begin with mock_ which, IMHO, negates the value.  Further, I'd suggest that it just clutters the code making it less readable.  -- just food for thought :)

rbarlow:
On 04/16/2015 02:02 PM, Jeff Ortel wrote:
> I know it's been convention in pulp to prefix all mocked object variable
> names with mock_ or sometimes fake_. However, when looking a most tests
> (including yours here), this means that almost all variables begin with
> mock_ which, IMHO, negates the value. Further, I'd suggest that it just
> clutters the code making it less readable. -- just food for thought :)

I agree. I prefer to just name them the same as what I'm replacing. The
code is more readable because the information density is higher.
However, I see it as a personal preference and not a thing we need to
standardize on as it sounds like Jeff does as well.

-- 
Randy Barlow

beav:
@jortel  thanks for the review!

PR #1782, opened 2015-04-14
convert "migration_tracker" collection to mongoengine.

ipanova:
https://pulp.plan.io/issues/728
closes #728


server/test/unit/server/db/test_manage.py
```diff
@@ -505,6 +506,25 @@ def test___init__(self):
         # By default, MigrationPackages should start at version 0
         self.assertEquals(mp._migration_tracker.version, 0)
 
+    @patch('pulp.server.db.migrate.models.MigrationTracker')
+    def test__init__2(self, mock_tracker):
+        """
+        Test package is not found.
```
bmbouter: s/is not found/does not exist/

server/test/unit/server/db/test_manage.py
```diff
@@ -505,6 +506,25 @@ def test___init__(self):
...
+        """
+        mock_tracker.objects.return_value.get.side_effect = DoesNotExist()
+        models.MigrationPackage(migration_packages.b)
+        self.assertEqual(mock_tracker.return_value.save.call_count, 1)
+
+    @patch('pulp.server.db.migrate.models.MigrationTracker')
+    def test__init__3(self, mock_tracker):
+        """
+        Test package is found.
```
bmbouter: s/is found/exists/

server/test/unit/server/db/test_manage.py
```diff
@@ -6,15 +6,16 @@
...
 
 from ... import base
 from pulp.common.compat import all, json
 from pulp.server.db import manage
 from pulp.server.db.migrate import models
 from pulp.server.db.model.migration_tracker import MigrationTracker
-from pulp.server.managers.migration_tracker import DoesNotExist, MigrationTrackerManager
 import pulp.plugins.types.database as types_db
 import migration_packages.a
+import migration_packages.b
```
bmbouter: Why can't all tests use `migration_packages.a`?

server/pulp/server/db/model/migration_tracker.py
```diff
@@ -24,29 +15,12 @@ class MigrationTracker(Model):
...
-        Save any changes made to this MigrationTracker to the database. If it doesn't exist in the
-        database already, insert a new record to represent it.
-        """
-        self.get_collection().update({'name': self.name},
-                                     {'name': self.name, 'version': self.version},
-                                     upsert=True, safe=True)
+    meta = {'collection': 'migration_trackers',
+            'indexes': [],  # small collection, does not need an index
+            'allow_inheritance': False,
+            'queryset_class': CriteriaQuerySet}
```
bmbouter: Does this actually need `CriteriaQuerySet`? I think all this code will work fine with the generic queryset manager. If that is the case this line should be deleted.

server/pulp/server/db/model/migration_tracker.py
```diff
@@ -1,18 +1,9 @@
...
-# 2 of the License (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+from mongoengine import Document, IntField, StringField
 
-from pulp.server.db.model.base import Model
+from pulp.server.db.model.base import CriteriaQuerySet
```
bmbouter: If this isn't used by the model definition any then it can be deleted.

bmbouter:
@ipanova I put a few questions on it, and I'm assigning back to you.

server/test/unit/server/db/test_manage.py
ipanova: We cannot use migration_packages.a because they are used in other testcases. I needed ne package that would be imported but not included here https://github.com/ipanova/pulp/blob/migration-tracker/server/test/unit/server/db/test_manage.py#L46

ipanova: I needed this for the situation when the package does not exist.

bmbouter: Makes sense, thanks.

bmbouter:
LGTM, great job

ipanova:
ok test

PR #1783, opened 2015-04-14
Remove owner type and and owner_id from repo_content_units

barnabycourt:
https://pulp.plan.io/issues/849

### DIFF
--- a/docs/user-guide/release-notes/2.7.x.rst
+++ b/docs/user-guide/release-notes/2.7.x.rst
@@ -57,6 +57,8 @@ Binding API Changes
 
 Plugin API Changes
 ------------------
+* The undocumented fields 'owner_type' and 'owner_id' have been removed from the
+  'repo_content_units' collection.
 
 Thank You
 ---------
--- a/nodes/test/unit/test_conduit.py
+++ b/nodes/test/unit/test_conduit.py
@@ -9,7 +9,6 @@
 # have received a copy of GPLv2 along with this software; if not, see
 # http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
 
-from mock import Mock
 from base import ServerTests
 from operator import itemgetter
 
@@ -67,9 +66,7 @@ def add_units(num_units=10):
             manager.associate_unit_by_id(
                 REPO_ID,
                 type_id,
-                unit_id,
-                RepoContentUnit.OWNER_TYPE_IMPORTER,
-                constants.HTTP_IMPORTER)
+                unit_id)
             units.append(unit)
             n += 1
     return units
--- a/server/pulp/server/db/migrations/0016_remove_repo_content_unit_owner_type_and_id.py
+++ b/server/pulp/server/db/migrations/0016_remove_repo_content_unit_owner_type_and_id.py
@@ -1,0 +1,75 @@
+"""
+This migration loads the content types into the database. Part of the process includes dropping
+of search indexes and their recreation.
+"""
+import logging
+
+from pulp.server.db import connection
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(*args, **kwargs):
+    """
+    Perform the migration as described in this module's docblock.
+
+    :param args:   unused
+    :type  args:   list
+    :param kwargs: unused
+    :type  kwargs: dict
+    """
+    db = connection.get_database()
+
+    # If 'repo_content_units' is not defined we don't need to do anything
+    if 'repo_content_units' not in db.collection_names():
+        return
+
+    collection = db['repo_content_units']
+    # Don't check whether we should run based on the index as that may have been cleared out
+    # by a different migration
+    if collection.find_one({'owner_type': {'$exists': True}}):
+        remove_duplicates(collection)
+
+        _logger.info("Removing unused fields (owner_type, owner_id) from repo_content_units")
+        collection.update({}, {'$unset': {'owner_type': "", 'owner_id': ''}}, multi=True)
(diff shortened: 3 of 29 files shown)


barnabycourt:
ok test

barnabycourt:
ok test

barnabycourt:
ok test

beav:
ok test

the builder appears to be doing better today :palm_tree:

ipanova:
@barnabycourt  was not it supposed to be this issue? https://pulp.plan.io/issues/849

barnabycourt:
@ipanova you are absolutely correct. I have updated the link. Thanks!

barnabycourt:
ok test

barnabycourt:
ok test

barnabycourt:
ok test

nodes/test/unit/test_conduit.py
```diff
@@ -9,7 +9,6 @@
 # have received a copy of GPLv2 along with this software; if not, see
 # http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
 
-from mock import Mock
```
beav: this is just for flake8?

server/pulp/server/db/migrations/0016_remove_repo_content_unit_owner_type_and_id.py
```diff
@@ -0,0 +1,68 @@
...
+    index_info = collection.index_information()
+    if "repo_id_-1_unit_type_id_-1_unit_id_-1_owner_type_-1_owner_id_-1" in index_info:
+        _logger.info("Dropping the uniqueness index that included the owner_type & owner_id")
+        collection.drop_index("repo_id_-1_unit_type_id_-1_unit_id_-1_owner_type_-1_owner_id_-1")
+
+
+def remove_duplicates(collection):
+    """
+    Remove entries from the repo_content_units collection that would be duplicates
+    after teh owner_type and owner_id fields are removed. Always remove older
```
beav: s/teh/the/

server/pulp/server/db/migrations/0016_remove_repo_content_unit_owner_type_and_id.py
```diff
@@ -0,0 +1,68 @@
...
+    entries so that incremental publishing will not be affected
+
+    :param collection: a reference to the repo_content_units collection
+    :type collection: pymongo.collection.collection
+    """
+    units_to_remove = []
+    last_unit = {"repo_id": None, "unit_type_id": None, "unit_id": None}
+    for unit in collection.find(sort=[('repo_id', 1), ('unit_type_id', 1),
+                                      ('unit_id', 1), ('updated', 1)]):
```
beav: Would it be possible to do an aggregate on all the units to find units with count > 1? That would avoid this loop

server/pulp/server/managers/repo/sync.py
```diff
@@ -79,8 +79,7 @@ def sync(repo_id, sync_config_override=None):
         repo_importer = importer_manager.get_importer(repo_id)
 
         # Assemble the data needed for the sync
-        conduit = RepoSyncConduit(repo_id, repo_importer['id'], RepoContentUnit.OWNER_TYPE_IMPORTER,
-                                  repo_importer['id'])
+        conduit = RepoSyncConduit(repo_id, repo_importer['id'])
 
         call_config = PluginCallConfiguration(importer_config, repo_importer['config'],
                                               sync_config_override)
```
beav: marking where I left off for lunch :smile:  will continue when I get back

:hamburger:

nodes/test/unit/test_conduit.py
barnabycourt: yup

server/test/unit/server/managers/repo/test_unit_association_query.py
```diff
@@ -673,385 +636,3 @@ def _assert_unit_integrity(self, unit):
         self.assertTrue(unit['metadata']['md_1'] is not None)
         self.assertTrue(unit['metadata']['md_2'] is not None)
         self.assertTrue(unit['metadata']['md_3'] is not None)
-
-
-class GetUnitsStressTest(base.PulpServerTests):
```
beav: I didn't realize this was here before, good catch.

The git log indicates this was updated in 2011-ish, aside from refactors and PEP8 updates.

beav:
The code change itself looks good. Two items:

* The migration might be able to be redone using an aggregation query instead of a loop.
* This change also needs a release note; the 2.7 notes are added as part of #1780 so you may need to add it in a new PR if this gets merged first.
* What happens if the migration is not run? Will the owner fields just be ignored on the mongo documents?

barnabycourt:
If the migration is not run two things occur. 1) The extra data would remain in the database, this would be messy and could result in errors when we stop filtering for duplicates in the unit association query. 2) The uniqueness constraint would not be updated for the collection so it would not be enforced properly in the future.

barnabycourt:
ok test

barnabycourt:
ok test

docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -57,6 +57,8 @@ Binding API Changes
 
 Plugin API Changes
 ------------------
+* The undocumented fields 'owner_type' and 'owner_id' have been removed from the
+  'repo_content_units' collection.
```
beav: cool, thanks :+1:

beav:
LGTM, nice work

PR #1785, opened 2015-04-14
Update installation.rst - Formatting fix

adam820:
Fix formatting of database line in server.conf description to be more in line with the other examples.


beav:
ok test

beav:
@adam820 Thanks again for this patch! Merging now.

PR #1787, opened 2015-04-14
  clear log handlers in stop_logging()

jortel:
Already reviewed on PR #1769 and just needed to be re-issued against 2.6-testing.

More complete and symmetrical with start_logging(). Also, fixes issue with unit tests.


bmbouter:
LGTM

PR #1788, opened 2015-04-15
Fix for issue 889.

kaos:
As described here: https://pulp.plan.io/issues/889

### DIFF
--- a/client_lib/pulp/client/commands/repo/status.py
+++ b/client_lib/pulp/client/commands/repo/status.py
@@ -83,7 +83,7 @@ class PublishStepStatusRenderer(StatusRenderer):
             return
 
         step.processed = step_details[reporting_constants.PROGRESS_NUM_PROCESSED_KEY]
-        step.details = step_details[reporting_constants.PROGRESS_DETAILS_KEY]
+        step.details = step_details.get(reporting_constants.PROGRESS_DETAILS_KEY, None)
         if not step.details:
             step.details = None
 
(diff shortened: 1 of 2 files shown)


beav:
@kaos can you give more info on how to reproduce this bug? I was not able to get it to happen when publishing a repo.

kaos:
I suspect it may be due to an older version of the pulp server, but that is
just a hunch.

But I think you should be able to see the data that caused this bug in the
server_calls.log.

I'm busy all day today, so I can't help further until tomorrow, at the
earliest.

2015-04-15 22:05 GMT+02:00 Chris Duryee <notifications@github.com>:

beav:
@kaos I wrote a small unit test that tests this patch:

```python
    def test_render_step_missing_details(self):
        """
        Test that no exception is raised when PROGRESS_DETAILS_KEY is missing.
        """
        self.step[reporting_constants.PROGRESS_STATE_KEY] = reporting_constants.STATE_RUNNING
        del self.step[reporting_constants.PROGRESS_DETAILS_KEY]
        self.renderer.render_step(self.step)
```

Can you add this test (or another if you want to write your own) into the `TestPublishStepStatusRenderer` test class and resubmit the PR?

beav:
ok test

beav:
ok test

beav:
@kaos This change looks good and tests pass for it. Many thanks for the patch!

kaos:
Thank you for accepting it :)
 Den 30 apr 2015 20:23 skrev "Chris Duryee" <notifications@github.com>:

PR #1789, opened 2015-04-15
Update to new contributor list for 2.6.1

beav:


beav:
failure was due to github timeout

beav:
ok test

mhrivnak:
I think we should find a way to not remove the 2.6.0 "thank-you"s. Perhaps add a new thank-you section under the 2.6.1 heading? Or just add this person to the original list?

PR #1790, opened 2015-04-15
fix filters field so GET searches work

asmacdo:

Moved from https://github.com/pulp/pulp/pull/1778 so this can be merged into 2.7-dev

closes #312


beav:
LGTM

PR #1791, opened 2015-04-15
Adjusts the upload delete API implementation

bmbouter:
The upload delete method used inode data to determine if a file
exists before deleting it. On filesystems with cached or slow-to-
respond inode data like NFS this was causing issues. The new
implementation always tries to remove a file and silences
only the error in the case where the file does not exist.

Also some small flake8 fixes.

closes #480

https://pulp.plan.io/issues/480

### DIFF
--- a/server/pulp/server/managers/content/upload.py
+++ b/server/pulp/server/managers/content/upload.py
@@ -1,21 +1,11 @@
-# -*- coding: utf-8 -*-
-#
-# Copyright © 2011 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
-# License as published by the Free Software Foundation; either version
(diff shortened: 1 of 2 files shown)


server/pulp/server/managers/content/upload.py
```diff
@@ -100,8 +90,11 @@ def delete_upload(self, upload_id):
         """
 
         file_path = ContentUploadManager._upload_file_path(upload_id)
-        if os.path.exists(file_path):
+        try:
             os.remove(file_path)
+        except OSError as e:
+            if e.errno != ENOENT:
+                raise
```
beav: hooray for more atomic operations

beav:
LGTM

PR #1793, opened 2015-04-15
set heartbeat interval to 30s

beav:
The heartbeat interval for celery was previously set to the default of 2
seconds. This caused the workers collection to be updated in mongo every two
seconds as well, even on an idle system.

Instead, set the heartbeat interval to 30 seconds. This should be sufficient
for worker status updates.

fixes #808


beav:
A `2.7.x.rst` release notes file is part of a different PR. I will add a release note about this change after that PR is merged.

PR #1796, opened 2015-04-16
Add initial model for repo_content_units

barnabycourt:
Depends on https://github.com/pulp/pulp/pull/1783
https://pulp.plan.io/issues/779

PR #1798, opened 2015-04-17
convert "reserved_resources" collection to mongoengine .

ipanova:
closes #727
https://pulp.plan.io/issues/727

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -113,7 +113,7 @@ def get_worker_for_reservation(resource_id):
                            `resource_id` associated with it.
     :rtype:                pulp.server.db.model.resources.Worker
     """
-    reservation = ReservedResource.get_collection().find_one({'resource_id': resource_id})
+    reservation = ReservedResource.objects(resource_id=resource_id).first()
     if reservation:
         return Worker.objects(name=reservation['worker_name']).first()
     else:
@@ -136,7 +136,7 @@ def _get_unreserved_worker():
     # Build a mapping of queue names to Worker objects
     workers_dict = dict((worker['name'], worker) for worker in Worker.objects())
     worker_names = workers_dict.keys()
-    reserved_names = [r['worker_name'] for r in ReservedResource.get_collection().find()]
+    reserved_names = [r['worker_name'] for r in ReservedResource.objects.all()]
 
     # Find an unreserved worker using set differences of the names, and filter
     # out workers that should not be assigned work.
@@ -176,7 +176,7 @@ def _delete_worker(name, normal_shutdown=False):
     Worker.objects(name=name).delete()
 
     # Delete all reserved_resource documents for the worker
-    ReservedResource.get_collection().remove({'worker_name': name})
+    ReservedResource.objects(worker_name=name).delete()
 
     # Cancel all of the tasks that were assigned to this worker's queue
     for task_status in TaskStatus.objects(worker_name=name,
@@ -199,7 +199,7 @@ def _release_resource(task_id):
     :param task_id: The UUID of the task that requested the reservation
     :type  task_id: basestring
     """
-    ReservedResource.get_collection().remove({'_id': task_id})
+    ReservedResource.objects(task_id=task_id).delete()
 
 
 class TaskResult(object):
--- a/server/test/unit/base.py
+++ b/server/test/unit/base.py
@@ -1,14 +1,3 @@
-# Copyright (c) 2011 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
-# License as published by the Free Software Foundation; either version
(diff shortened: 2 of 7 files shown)


ipanova:
ok test

server/pulp/server/async/tasks.py
```diff
@@ -136,7 +136,7 @@ def _get_unreserved_worker():
     # Build a mapping of queue names to Worker objects
     workers_dict = dict((worker['name'], worker) for worker in Worker.objects())
     worker_names = workers_dict.keys()
-    reserved_names = [r['worker_name'] for r in ReservedResource.get_collection().find()]
+    reserved_names = [r['worker_name'] for r in ReservedResource.objects()]
```
bmbouter: I think `ReservedResource.objects()` this would read more clearly as `ReservedResource.objects.all()`

bmbouter:
This object needs to have its ensure_indexes called on it. Like how these are added here:

https://github.com/pulp/pulp/blob/762ac5a9509e7aa27fb13227eb73186244896166/server/pulp/server/db/manage.py#L118

server/test/unit/server/db/model/test_resources.py
```diff
@@ -1,55 +1,21 @@
...
-from ....base import ResourceReservationTests
 from pulp.server.db.model import resources
 
 
-class TestReservedResource(ResourceReservationTests):
+class TestReservedResource(unittest.TestCase):
 
-    def test_unique_indices(self):
-        self.assertEqual(resources.ReservedResource.unique_indices, tuple())
+    def test_attributes(self):
```
bmbouter: Please add a test that asserts that `ReservedResource` is an instance of `Document`.

server/test/unit/server/db/model/test_resources.py
```diff
@@ -1,55 +1,21 @@
...
+        rr = resources.ReservedResource('my-task', 'some-worker', 'some-resource')
```
bmbouter: Since these are all class attributes, you shouldn't actually instantiate the class. Instead make assertions on the class itself. I think the tests @barnabycourt are a really good example. See:  https://github.com/pulp/pulp/pull/1796/files#diff-7b4c3232751ba4e87aa0dd34e5a81ba4R23

server/test/unit/server/db/model/test_resources.py
```diff
@@ -1,55 +1,21 @@
...
-        rr = resources.ReservedResource(task_id, 'some_worker', 'some_resource')
-
-        rr.save()
-
-        # Make sure the DB has the correct data
-        rrc = resources.ReservedResource.get_collection()
-        self.assertEqual(rrc.count(), 1)
-        self.assertEqual(rrc.find_one({'_id': task_id})['worker_name'], 'some_worker')
-        self.assertEqual(rrc.find_one({'_id': task_id})['resource_id'], 'some_resource')
+    def test_indixes(self):
```
bmbouter: s/test_indixes/test_indexes/

server/test/unit/server/db/model/test_resources.py
```diff
@@ -1,55 +1,21 @@
...
+        self.assertEqual(resources.ReservedResource._meta['indexes'],
+                         ['-worker_name', '-resource_id'])
```
bmbouter: Please add additional tests for the other aspects of the metadata. See this as a similar example:

https://github.com/pulp/pulp/pull/1796/files#diff-7b4c3232751ba4e87aa0dd34e5a81ba4R41

bmbouter:
I believe [this line](https://github.com/ipanova/pulp/blob/86957c5d71ede1746004eac7155c5ad684533c1a/server/pulp/server/async/tasks.py#L76) will cause a race condition because of how the save() works. Please rewrite it to use an atomic operation to issue the save()

ipanova:
ok test

server/test/unit/base.py
```diff
@@ -270,5 +260,5 @@ def test_assert_equal_ignoring_list_order(self):
 class ResourceReservationTests(PulpServerTests):
     def tearDown(self):
         Worker.objects().delete()
-        ReservedResource.get_collection().remove()
+        ReservedResource.objects().delete()
```
bmbouter: When chaining the `objects` call, calling it is not required and I think not calling `objects` makes it more readable. For example this line would become:  `ReservedResource.objects.delete()`. It's written using `objects()` in other places, but I would prefer those to be done differently too. Any rewrites of this in this PR should only deal with `ReservedResource.objects`.

bmbouter:
[One last comment](https://github.com/pulp/pulp/pull/1798#discussion_r28695049) which would be good to fix. Other than that, LGTM. The tests are really nice. I'm going to put the LGTM tag on, and assign it back to you.

ipanova:
ok test

PR #1799, opened 2015-04-17
Added persistence of /var/run/pulp

dkliban:
This patch provides a fix for both systemd and upstart. A configuration file for
systemd-tmpfiles helps ensure that /var/run/pulp is recreated at boot time. The
upstart script already created the directory if it was missing. The same script
now restores the context on the newly created directory. The relabel.sh script
now contains a statement to restore context of /var/run/pulp also.

Fixes #719

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -55,6 +55,8 @@ allow celery_t pulp_tmp_t:file manage_file_perms;
 files_tmp_filetrans(celery_t, pulp_tmp_t, file)
 
 allow celery_t pulp_var_run_t:file manage_file_perms;
+# dir access needed when celery searches for existing pid files on startup
+allow celery_t pulp_var_run_t:dir manage_dir_perms;
 files_pid_filetrans(celery_t, pulp_var_run_t, file)
 
 
--- a/pulp-dev.py
+++ b/pulp-dev.py
@@ -244,6 +244,9 @@ def get_paths_to_copy():
         paths.append({'source': 'server/usr/lib/systemd/system/pulp_workers.service',
                       'destination': '/etc/systemd/system/pulp_workers.service', 'owner': 'root',
                       'group': 'root', 'mode': '644', 'overwrite': True})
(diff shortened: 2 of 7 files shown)


server/selinux/server/pulp-celery.te
```diff
@@ -55,6 +55,7 @@ allow celery_t pulp_tmp_t:file manage_file_perms;
 files_tmp_filetrans(celery_t, pulp_tmp_t, file)
 
 allow celery_t pulp_var_run_t:file manage_file_perms;
+allow celery_t pulp_var_run_t:dir { write remove_name search add_name };
```
bmbouter: Consider replacing this with `manage_dir_perms`. I think that would be a good change. It's a little less fine grained, but I think it captures the intent of what you are going for.

dkliban:
ok test

dkliban:
ok test

server/selinux/server/pulp-celery.te
```diff
@@ -55,6 +55,8 @@ allow celery_t pulp_tmp_t:file manage_file_perms;
 files_tmp_filetrans(celery_t, pulp_tmp_t, file)
 
 allow celery_t pulp_var_run_t:file manage_file_perms;
+# celery searches the directory for any existing pid files when it starts
```
bmbouter: s/celery searches the directory for any existing pid files when it starts/dir access needed when celery searches for existing pid files on startup/

server/selinux/server/pulp-celery.te
```diff
@@ -55,6 +55,8 @@ allow celery_t pulp_tmp_t:file manage_file_perms;
...
+allow celery_t pulp_var_run_t:dir manage_dir_perms;
```
bmbouter: The more I think about this; I think reverting this to the more restrictive version would be better. Sorry for the back and forth.

bmbouter:
Also this needs changes to pulp-dev so that both folders will be created and survive a reboot

dkliban:
ok test

bmbouter:
pulpbot needs to still sign off on this, but LGTM!

PR #1801, opened 2015-04-20
Added a Vagrantfile to get us started with vagrant

rbarlow:


beav:
This does not run for me on F21:

```
==> default: generating RSA keys
==> default: generating certificates
==> default: ~/devel ~
==> default: Adjusting facls for apache
==> default: Starting services, this may take a few minutes due to mongodb journal allocation
==> default: Created symlink from /etc/systemd/system/multi-user.target.wants/qpidd.service to /usr/lib/systemd/system/qpidd.service.
==> default: Created symlink from /etc/systemd/system/multi-user.target.wants/goferd.service to /usr/lib/systemd/system/goferd.service.
==> default: Created symlink from /etc/systemd/system/multi-user.target.wants/mongod.service to /usr/lib/systemd/system/mongod.service.
==> default: Job for mongod.service failed. See "systemctl status mongod.service" and "journalctl -xe" for details.
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

rbarlow:
Hmmm, I wrote an e-mail to reply to this, but it didn't seem to make it here (at least, not yet). Here's what I wrote:

This can happen, and is outside of our control unfortunately. The
dev-setup.sh script actually suffers from this problem as well
occasionally. The issue is that the first time mongod starts, it
generates some very large files. If this takes too long, systemd will
give up on the unit and say that it failed to start.

When this happens, you can typically just reprovision the box and
everything will be just fine. In Vagrant you can just do this:

$ vagrant provision

Hopefully in the second try you will get past that issue.

beav:
I am re-testing now.

We could disable the journal on dev setups maybe, if you think that is OK to do.

beav:
This error appears on startup:

```
Apr 20 18:51:00 pulp-devel httpd[27191]: httpd: Syntax error on line 353 of /etc/httpd/conf/httpd.conf: Syntax error on line 54 of /etc/httpd/conf.d/pulp.conf: No matches for the wildcard '*.conf' in '/etc/pulp/vhosts80', failing (use IncludeOptional if required)
```

Do I need a pulp_rpm checkout next to my pulp checkout?

rbarlow:
On 04/20/2015 12:01 PM, Chris Duryee wrote:
> |==> default: Job for mongod.service failed. See "systemctl status mongod.service" and "journalctl -xe" for details.|

This can happen, and is outside of our control unfortunately. The
dev-setup.sh script actually suffers from this problem as well
occasionally. The issue is that the first time mongod starts, it
generates some very large files. If this takes too long, systemd will
give up on the unit and say that it failed to start.

When this happens, you can typically just reprovision the box and
everything will be just fine. In Vagrant you can just do this:

$ vagrant provision

Hopefully in the second try you will get past that issue.

-- 
Randy Barlow

rbarlow:
@beav ah, yes you do need pulp_rpm checked out next to pulp for this to work. I felt a little conflicted about that decision, but without any plugins it's not very useful to get Pulp installed and running. Also, as it is the setup script does make the example zoo repo and sync it.

We could refactor a bit and detect the repos that aren't checked out and check them out inside the VM for you. Or we could drop the zoo part. Or, if you don't mind it as it is for now, we could just leave it as is.

I'd be in favor of leaving it as is, with the option to change it later. What do you think?

rbarlow:
On 04/20/2015 02:51 PM, Chris Duryee wrote:
> We could disable the journal on dev setups maybe, if you think that is
> OK to do.

+1, though I'd prefer to do that in a separate PR if you don't mind.

-- 
Randy Barlow

rbarlow:
On 04/20/2015 02:53 PM, Chris Duryee wrote:
> Do I need a pulp_rpm checkout next to my pulp checkout?

I'm planning to document the above as well, in a separate PR, along with
the fact that we have Vagrant now! I'm thinking of adding a "for the
even lazier" section, or something like that. I want to keep the current
dev-setup.sh we have as well.

-- 
Randy Barlow

beav:
I was able to run this successfully. Can you create a redmine issue for the journal thing? LGTM otherwise :sunglasses:

rbarlow:
One other thing I'd like to do (as a separate PR) is to make our own Vagrant box that already has a lot of this stuff done (like the Mongo build). If we do that, then the vagrant up command will go much faster. I think that would be really nice, but this PR is just to get us going.

rbarlow:
@beav I made this task:

https://pulp.plan.io/issues/903

It's a little more general, but it covers the journalling and several more things.

PR #1802, opened 2015-04-20
refs:654 - remove rsyslog examples.

jortel:
Replace rsyslog.conf examples with reference to rsyslog.com and suggestion to use tail and grep.


beav:
LGTM

PR #1805, opened 2015-04-21
Removes orderreddict from Kombu Python deps. The rpm still includes it.

bmbouter:
This backports a patch that was built in a newer version of
Kombu so that multiple version of Pulp can all use the same
rebuilt dep.

closes #896
https://pulp.plan.io/issues/896


bmbouter:
Closing this because now it's supposed to be in 2.6.2 so it's against the wrong branch

PR #1806, opened 2015-04-21
As a user, I can see tasks in particular states.

ipanova:
closes #851
https://pulp.plan.io/issues/851

Also fixes https://pulp.plan.io/issues/906

### DIFF
--- a/client_admin/pulp/client/admin/tasks.py
+++ b/client_admin/pulp/client/admin/tasks.py
@@ -3,7 +3,8 @@ import sys
 
 import pulp.common.tags as tag_utils
 from pulp.bindings.exceptions import PulpServerException
-from pulp.client.extensions.extensions import PulpCliSection
+from pulp.client import parsers
+from pulp.client.extensions.extensions import PulpCliSection, PulpCliFlag, PulpCliOption
 
 
 # Guidance for render_document_list on how to display task info
@@ -31,6 +32,15 @@ class BaseTasksSection(PulpCliSection):
     to provide consistent functionality for a subset of tasks.
     """
 
+    all_flag = PulpCliFlag('--all',  _('if specified, all tasks in all states are shown'),
+                           aliases=['-a'])
+    state_option = PulpCliOption('--state',
+                                 _('comma-separated list of tasks states desired to be '
+                                   'shown. Example: "running,waiting,canceled,successful,failed". '
+                                   'Do not include spaces'), aliases=['-s'], required=False,
+                                 parse_func=parsers.csv)
+
+
     def __init__(self, context, name, description):
         PulpCliSection.__init__(self, name, description)
         self.context = context
@@ -39,7 +49,7 @@ class BaseTasksSection(PulpCliSection):
         # can manipulate them if necessary
 
         self.list_command = self.create_command(
-            'list', _('lists tasks queued or running in the server'), self.list
+            'list', _('lists tasks queued (waiting) or running on the server'), self.list
         )
 
         self.cancel_command = self.create_command('cancel', _('cancel one or more tasks'),
@@ -60,6 +70,11 @@ class BaseTasksSection(PulpCliSection):
         """
 
         self.context.prompt.render_title('Tasks')
+
+        if kwargs.get(self.all_flag.keyword) and kwargs.get(self.state_option.keyword):
+            msg = _('These arguments cannot be used together')
+            self.context.prompt.render_failure_message(msg)
+            return
 
         task_objects = self.retrieve_tasks(**kwargs)
 
@@ -231,12 +246,27 @@ class BaseTasksSection(PulpCliSection):
 class AllTasksSection(BaseTasksSection):
     FIELDS = ('tags', 'task_id', 'state', 'start_time', 'finish_time')
 
+    def __init__(self, context, name, description):
+        BaseTasksSection.__init__(self, context, name, description)
+
+        self.list_command.add_option(self.all_flag)
+
+        self.list_command.add_option(self.state_option)
+
     def retrieve_tasks(self, **kwargs):
         """
         :return:    list of pulp.bindings.responses.Task instances
         :rtype:     list
         """
-        tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+
+        if kwargs.get(self.all_flag.keyword):
+            tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+        elif kwargs.get(self.state_option.keyword):
+            tasks = self.context.server.tasks_search.search(
+                filters={'state': {'$in': kwargs[self.state_option.keyword]}}, fields=self.FIELDS)
+        else:
+            tasks = self.context.server.tasks_search.search(
+                filters={'state': {'$in': ['running', 'waiting']}}, fields=self.FIELDS)
         return tasks
 
 
--- a/client_admin/test/unit/test_pulp_tasks_extension.py
+++ b/client_admin/test/unit/test_pulp_tasks_extension.py
@@ -1,5 +1,7 @@
 import copy
 
+import mock
+
 import base_builtins
 
 from pulp.bindings.exceptions import NotFoundException, PulpServerException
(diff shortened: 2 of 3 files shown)


client_admin/pulp/client/admin/tasks.py
```diff
@@ -39,8 +39,13 @@ def __init__(self, context, name, description):
         # can manipulate them if necessary
 
         self.list_command = self.create_command(
-            'list', _('lists tasks queued or running in the server'), self.list
+            'list', _('lists tasks queued(waiting) or running on the server'), self.list
```
mhrivnak: I think there should be a space after "queued".

client_admin/pulp/client/admin/tasks.py
```diff
@@ -39,8 +39,13 @@ def __init__(self, context, name, description):
...
         )
+        d = _('if specified, all tasks in all states are shown.')
+        self.list_command.add_option(PulpCliFlag('--all', d))
```
mhrivnak: I suggest adding a single-character option ``-a``, which can be another argument to ``PulpCliFlag``.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -39,8 +39,13 @@ def __init__(self, context, name, description):
...
+
+        d = _('comma-separated list of tasks states desired to be shown.')
+        self.list_command.add_option(PulpCliOption('--state', d, required=False))
```
mhrivnak: I suggest adding a single-character option ``-s``, which can be another argument to ``PulpCliFlag``.

mhrivnak: See ``pulp.client.parsers.csv``, which can be passed to the ``PulpCliOption`` as the parser.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -39,8 +39,13 @@ def __init__(self, context, name, description):
         # can manipulate them if necessary
 
         self.list_command = self.create_command(
-            'list', _('lists tasks queued or running in the server'), self.list
+            'list', _('lists tasks queued(waiting) or running on the server'), self.list
         )
+        d = _('if specified, all tasks in all states are shown.')
```
mhrivnak: We tend to not put periods at the end of help text in the CLI. Same below.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -236,7 +245,15 @@ def retrieve_tasks(self, **kwargs):
         :return:    list of pulp.bindings.responses.Task instances
         :rtype:     list
         """
-        tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+
+        if kwargs.get('all'):
+            tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+        elif kwargs.get('state'):
+            tasks = self.context.server.tasks_search.search(
```
mhrivnak: Do you not need to specify ``fields=`` here and below, like you did above?

client_admin/pulp/client/admin/tasks.py
```diff
@@ -39,8 +39,13 @@ def __init__(self, context, name, description):
         # can manipulate them if necessary
 
         self.list_command = self.create_command(
-            'list', _('lists tasks queued or running in the server'), self.list
+            'list', _('lists tasks queued(waiting) or running on the server'), self.list
         )
+        d = _('if specified, all tasks in all states are shown.')
+        self.list_command.add_option(PulpCliFlag('--all', d))
+
+        d = _('comma-separated list of tasks states desired to be shown.')
```
mhrivnak: It would be helpful to list the options.

mhrivnak:
It looks like your changes to ``BaseTaskSection`` add new options to the ``RepoTasksSection`` class, which you may not have intended. The ``RepoTasksSection`` doesn't seem to use those options.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -60,6 +65,10 @@ def list(self, **kwargs):
         """
 
         self.context.prompt.render_title('Tasks')
+        if kwargs.get('all') and kwargs.get('state'):
```
mhrivnak: A nice pattern is to save the ``PulpCliFlag`` or ``PulpCliOption`` instance on your class, where you instantiate it, for example:

```
self.all_flag = PulpCliFlag(**stuff)
```

and then here (and in the tests) you can access the keyword directly like so:

```
if kwargs.get(self.all_flag.keyword) and kwargs.get(self.state_option.keyword):
```

This avoids having to track the keyword sting externally in some other constant, or by typing it in each time like you've done here.

client_admin/pulp/client/admin/tasks.py
ipanova: did not know about it, nice thanks :)

client_admin/pulp/client/admin/tasks.py
```diff
@@ -31,6 +32,15 @@ class BaseTasksSection(PulpCliSection):
     to provide consistent functionality for a subset of tasks.
     """
 
+    all_flag = PulpCliFlag('--all',  _('if specified, all tasks in all states are shown'),
+                           aliases=['-a'])
+    state_option = PulpCliOption('--state',
+                                 _('comma-separated list of tasks states desired to be '
```
mhrivnak: s/tasks/task

client_admin/pulp/client/admin/tasks.py
```diff
@@ -31,6 +32,15 @@ class BaseTasksSection(PulpCliSection):
...
+                                   'shown. Example: "runnining,waiting,canceled,successful,failed". '
```
mhrivnak: "running" is misspelled

mhrivnak: also, just confirm that we spell it "canceled". One "l" or two has been the source of much confusion in the past, and I can never remember which one we ended up choosing.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -61,8 +71,14 @@ def list(self, **kwargs):
...
         self.context.prompt.render_title('Tasks')
 
+        if kwargs.get(self.all_flag.keyword) and kwargs.get(self.state_option.keyword):
+            msg = _('These arguments cannot be used together')
+            self.context.prompt.render_failure_message(msg)
+            return
+
         task_objects = self.retrieve_tasks(**kwargs)
 
+
```
mhrivnak: This blank line should not be present.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -231,12 +247,27 @@ def retrieve_tasks(self):
 class AllTasksSection(BaseTasksSection):
     FIELDS = ('tags', 'task_id', 'state', 'start_time', 'finish_time')
 
+    def __init__(self, context, name, description):
+        BaseTasksSection.__init__(self, context, name, description)
+
+        self.list_command.add_option(self.all_flag)
```
mhrivnak: It would be better to add these options right after you create the list command, so everything is organized in one place. I don't see value in waiting to do it here.

client_admin/pulp/client/admin/tasks.py
ipanova: yeah, 'canceled' is right in this case
`$ pulp-admin -u admin -p admin tasks list --state canceled

                                 Tasks

Operations:  sync
Resources:   pup (repository)
State:       Cancelled
Start Time:  2015-04-22T13:16:32Z
Finish Time: 2015-04-22T13:16:50Z
Task Id:     d0ca6865-91dd-4fbe-ad17-f77db3c61a94

$ pulp-admin -u admin -p admin tasks list --state cancelled

                                 Tasks

No tasks found`

client_admin/pulp/client/admin/tasks.py
```diff
@@ -231,12 +247,27 @@ def retrieve_tasks(self):
...
         :return:    list of pulp.bindings.responses.Task instances
         :rtype:     list
         """
-        tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+
+        if kwargs.get(self.all_flag.keyword):
+            tasks = self.context.server.tasks_search.search(fields=self.FIELDS)
+        elif kwargs.get(self.state_option.keyword):
+            tasks = self.context.server.tasks_search.search(
+                filters={'state': {'$in': kwargs['state']}}, fields=self.FIELDS)
```
mhrivnak: Here's one more opportunity to use ``self.state_option.keyword``

client_admin/pulp/client/admin/tasks.py
ipanova: if i will add them right after the the list command is created(in BaseTasksSection) it will end up that RepoTasksSection will also have those options because it inherits the BaseTasksSection

client_admin/test/unit/test_pulp_tasks_extension.py
```diff
@@ -80,6 +89,79 @@ def test_list(self):
...
+
+        self.server_mock.request.return_value = (200, [])
+
+        # Test
+        self.all_tasks_section.list(**data)
+
+        # Verify
+        self.assertEqual(1, self.server_mock.request.call_count)
+        self.assertEqual('POST', self.server_mock.request.call_args[0][0])
+        search_crit = '{"criteria": {"fields": ["tags", "task_id", "state", "start_time", '\
```
mhrivnak: So, here's the thing... :) There aren't nearly as many examples of proper unit testing in our CLI code as in other parts of pulp, and you definitely didn't have any good examples to emulate in this module. It looks like this test, and the ones below, are testing a large stack of calls (into ``pulp.bindings`` and so on), and it looks like they may depend on the json serialization happening in a specific way with specific whitespace. That is actually rather fragile, so it would be better to mock ``pulp.bindings.tasks.TaskSearchAPI.search`` and make sure it got the arguments you expect. Let me know if it would be helpful to explain this more.

client_admin/pulp/client/admin/tasks.py
mhrivnak: Ah, I see.

PR #1807, opened 2015-04-21
Kombu SASL connection, connection closing, and python dep fixes

bmbouter:
Removes ordereddict from Kombu Python setuptools dependencies.
The spec file still includes it.

Also backports two fixes that were already included in
python-kombu-6.pulp

    Fixes a SASL negotiation issue when PLAIN is not available
    and the python-saslwrapper is installed

    Fixes a connection close issue that was already included
    in python-kombu-6.pulp. Included here so that multiple
    versions of Pulp can use the same version. They should be
    able to safely.

closes #896
https://pulp.plan.io/issues/896

### DIFF
--- a/deps/python-kombu/README
+++ b/deps/python-kombu/README
@@ -4,3 +4,26 @@ Celery 3.1.11 requires kombu>=3.0.15,<4.0, which is not in Fedora <= 20, or EPEL
 
 Pulp would like to use the Qpid transport which is available in upstream kombu>=3.0.24
 
+
+Permanent Patches:
+
+1212200.patch - This patch will need to continue to be with us forever as it works around a
+                permanent downstream issue whereby ordereddict is installed as part of the Python
+                system library, and not in site-packages. This is not contributed to upstream Kombu.
+
+
+Temporary Patches:
+
+This are contributed to upstream Kombu and will be included in a future release of Kombu, but until
+it is we include it in downstream.
+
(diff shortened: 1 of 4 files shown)


deps/python-kombu/README
```diff
@@ -4,3 +4,26 @@ Celery 3.1.11 requires kombu>=3.0.15,<4.0, which is not in Fedora <= 20, or EPEL
 
 Pulp would like to use the Qpid transport which is available in upstream kombu>=3.0.24
 
+
```
beav: thanks for documenting all of this

deps/python-kombu/python-kombu.spec
```diff
@@ -11,7 +11,7 @@ Name:           python-%{srcname}
 # The Fedora package is using epoch 1, so we need to also do that to make sure ours gets installed
 Epoch:          1
 Version:        3.0.24
-Release:        5.pulp%{?dist}
+Release:        7.pulp%{?dist}
```
beav: why does this go from 5 to 7? There isn't a shortage of integers, just curious :koala:

beav:
I had a question in the inline comments but LGTM.

deps/python-kombu/python-kombu.spec
bmbouter: `python-kombu-3.0.24-6.pulp` is already taken and koji already has it built and locked. The `7` release is the next available one for `3.0.24`. The `7` release also includes all the same stuff from `6`.

PR #1809, opened 2015-04-22
Added OSError 17 exception handling during race condition

dkliban:
Fixes #902

### DIFF
--- a/server/pulp/server/managers/content/query.py
+++ b/server/pulp/server/managers/content/query.py
@@ -289,8 +289,12 @@ class ContentQueryManager(object):
 
         unit_path = os.path.join(self.get_root_content_dir(content_type), relative_path)
         unit_dir = os.path.dirname(unit_path)
-        if not os.path.exists(unit_dir):
+        try:
             os.makedirs(unit_dir)
+        except OSError, e:
+            if e.errno != errno.EEXIST:
+                    raise
+
         return unit_path
 
 # utility methods --------------------------------------------------------------
--- a/server/test/unit/server/managers/content/test_query.py
+++ b/server/test/unit/server/managers/content/test_query.py
@@ -1,0 +1,35 @@
+import errno
+import inspect
+import os
+import unittest
+
+import mock
+
+from pulp.server.managers.content.query import ContentQueryManager
+
+
+class TestContentQueryManager(unittest.TestCase):
+
+    @mock.patch('pulp.server.managers.content.upload.os.makedirs')
+    @mock.patch.object(ContentQueryManager, 'get_root_content_dir')
+    def test_request_content_unit_file_path_exists(self, mock_root_dir, mock_makedirs):
+        mock_root_dir.return_value = '/var/lib/pulp/content/rpm/'
+        mock_makedirs.side_effect = OSError(errno.EEXIST, os.strerror(errno.EEXIST))
+        ContentQueryManager().request_content_unit_file_path('rpm', '/name/blah')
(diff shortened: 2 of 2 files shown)


server/pulp/server/managers/content/query.py
```diff
@@ -290,7 +290,14 @@ def request_content_unit_file_path(self, content_type, relative_path):
         unit_path = os.path.join(self.get_root_content_dir(content_type), relative_path)
         unit_dir = os.path.dirname(unit_path)
         if not os.path.exists(unit_dir):
-            os.makedirs(unit_dir)
+            try:
```
bmbouter: I propose that the conditional branch at 292 should go away altogether and that `os.makedirs` should be called in all cases. The try/except you've written will handle it correctly. The L292 check isn't working in all cases, and now we have two checks in a way for the same thing. Removing line 292 would reduce the complexity of this codepath. I just fixed a similar problem [in a recent PR](https://github.com/pulp/pulp/pull/1791/files#diff-6ceaf55d8c7a51c3a8243b1d577dcd11L103).

server/pulp/server/managers/content/query.py
```diff
@@ -290,7 +290,14 @@ def request_content_unit_file_path(self, content_type, relative_path):
...
+                os.makedirs(unit_dir)
+            except OSError, e:
+                # In case two workers are syncing exact same content simultaneously and the
+                # directory got created in between the above check and now. See issue #902.
+                if e.errno != 17:
```
bmbouter: The `17` should be replaced with a name provided by the python standard lib `errno`. Different systems may number these errors differently, but Python guarantees that the names provided by `errno` will map correctly on all systems Python runs on. See how I did this [in this other PR](https://github.com/pulp/pulp/pull/1791/files#diff-6ceaf55d8c7a51c3a8243b1d577dcd11R96).

server/pulp/server/managers/content/query.py
```diff
@@ -290,7 +290,14 @@ def request_content_unit_file_path(self, content_type, relative_path):
         unit_path = os.path.join(self.get_root_content_dir(content_type), relative_path)
         unit_dir = os.path.dirname(unit_path)
         if not os.path.exists(unit_dir):
-            os.makedirs(unit_dir)
+            try:
+                os.makedirs(unit_dir)
+            except OSError, e:
+                # In case two workers are syncing exact same content simultaneously and the
```
bmbouter: Consider removing this comment. Here are some reasons.

* It's linked in github to this PR and in Redmine using the ref, so it would be easy for a developer to find issue #902 and this PR for these lines if they ever have questions.
* More generally, we could fill the whole codebase with comments like these which need to be maintained over time and could potentially double the code size.
* If comments ever don't get maintained they are really unhelpful. Consider that this codepath may eventually not be called by two workers at the same time, and then this comment would suggest that they do.
* Comments should be reserved for really, really non-obvious programmer intent. I think this being race condition free is obvious.

bmbouter:
I think this needs to come with test code. If there are no tests to update/modify, I don't expect you to write test coverage for the whole function, but making assertions on the specific parts you changed/introduced would be good. I had to to do the same thing in my recent PR that is similar to this.

bmbouter:
Good job on resolving this issue. I'm done with the review so I'm assigning back to you. I left several comments that I think should be fixed before merging. I don't want to re-review so I'm adding the LGTM flag now.

bmbouter:
I've also taken to adding the link to the issue in the commit itself which will show up in the PR contents above automatically. That would be good for the future (or you could amend this commit)

PR #1815, opened 2015-04-22
Fixes accidental removal of an import in test_query.py

dkliban:


rbarlow:
LGTM, assuming @pulpbot is also cool with it.

PR #1817, opened 2015-04-22
add a mongoengine model for repositories

asmacdo:
closes: [#785](https://pulp.plan.io/issues/785)

### DIFF
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -115,6 +115,7 @@ def ensure_database_indexes():
     collection does not already exist.
     """
     model.RepositoryContentUnit.ensure_indexes()
+    model.Repository.ensure_indexes()
     model.ReservedResource.ensure_indexes()
     model.TaskStatus.ensure_indexes()
     model.Worker.ensure_indexes()
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -11,7 +11,59 @@ from pulp.server.db.model.base import CriteriaQuerySet
 from pulp.server.db.model.fields import ISO8601StringField
 from pulp.server.db.model.reaper_base import ReaperMixin
 
+
 _logger = logging.getLogger(__name__)
+
+
+class Repository(Document):
+    """
+    Defines schema for a pulp repository in the `repos` collection.
+
+    :ivar repo_id: unique across all repos
+    :type repo_id: mongoengine.StringField
+    :ivar display_name: user-readable name of the repository
+    :type display_name: mongoengine.StringField
+    :ivar description: free form text provided by the user to describe the repo
+    :type description: mongoengine.StringField
+    :ivar notes: arbitrary key-value pairs programmatically describing the repo;
+                 these are intended as a way to describe the repo usage or
+                 organizational purposes and should not vary depending on the
+                 actual content of the repo
+    :type notes: mongoengine.DictField
+    :ivar content_unit_counts: key-value pairs of number of units associated with this repo.
+                               This is different than the number of associations, since a
+                               unit may be associated multiple times.
+    :type content_unit_counts: mongoengine.DictField
+    :ivar scratchpad: Field used to persistently store arbitrary information from the plugins
+                      across multiple operations.
+    :type scratchpad: mongoengine.DictField
+    :ivar last_unit_added: Datetime of the most recent occurence of adding a unit to the repo
+    :type last_unit_added: mongoengine.DateTimeField
+    :ivar last_unit_removed: Datetime of the most recent occurence of removing a unit from the repo
+    :type last_unit_removed: mongoengine.DateTimeField
+    :ivar _ns: (Deprecated) Namespace of repo, included for backwards compatibility.
+    :type _is: mongoengine.StringField
+    """
+
+    # This field is named `repo_id` because `id`  cannot be accessed using the normal mongoengine
+    # idiom of `obj.id` because `obj.id` has been aliased (internally in mongoengine) to reference
+    # obj._id. So we name this field `repo_id`, and temporarily continue to store it in the db as
+    # `id` for backwards compatibility. This should be migrated to `repo_id` in a future X release.
+    repo_id = StringField(db_field='id', required=True)
+    display_name = StringField(required=True)
+    description = StringField()
+    notes = DictField()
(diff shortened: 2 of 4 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -51,3 +52,51 @@ class RepositoryContentUnit(Document):
...
+    :ivar notes: arbitrary key-value pairs programmatically describing the repo;
+                 these are intended as a way to describe the repo usage or
+                 organizational purposes and should not vary depending on the
+                 actual content of the repo
+    :type notes: mongoengine.DictField
+    :ivar content_unit_counts: key-value pairs of number of units associated with this repo.
+                               This is different than the number of associations, since a
+                               unit may be associated multiple times.
+    :type content_unit_counts: mongoengine.DictField
+    :ivar metadata: arbitrary data that describes the contents of the repo;
```
ipanova: why do you have the metadata described here? i do not see it among the attributes

server/pulp/server/db/model/__init__.py
```diff
@@ -51,3 +52,51 @@ class RepositoryContentUnit(Document):
...
+    # obj._id. So we name this field `repo_id`, and temporarily continue to store it in the db as
+    # `id` for backwards compatibility. This should be migrated to `repo_id` in a future X release.
+    repo_id = StringField(db_field='id', required=True)
+    display_name = StringField(required=True)
+    description = StringField()
+    notes = DictField()
+    scratchpad = DictField()
+    content_unit_counts = DictField()
+    last_unit_added = DateTimeField()
+    last_unit_removed = DateTimeField()
```
ipanova: please add `last_unit_added` and `last_unit_removed` description to the doc block

server/pulp/server/db/model/__init__.py
```diff
@@ -51,3 +52,51 @@ class RepositoryContentUnit(Document):
...
+    :ivar _ns: (Deprecated) Namespace of repo, included for backwards compatibility.
+    :type _is: mongoengine.StringField
+    """
+
+    # This field is named `repo_id` because `id`  cannot be accessed using the normal mongoengine
+    # idiom of `obj.id` because `obj.id` has been aliased (internally in mongoengine) to reference
+    # obj._id. So we name this field `repo_id`, and temporarily continue to store it in the db as
+    # `id` for backwards compatibility. This should be migrated to `repo_id` in a future X release.
+    repo_id = StringField(db_field='id', required=True)
+    display_name = StringField(required=True)
```
ipanova: can you explain why have you decided to make display_name as required field and, for example, description not required?

server/pulp/server/db/model/__init__.py
```diff
@@ -51,3 +52,51 @@ class RepositoryContentUnit(Document):
...
+    description = StringField()
+    notes = DictField()
+    scratchpad = DictField()
```
ipanova: please add `scratchpad` description to the doc block

server/pulp/server/db/model/__init__.py
```diff
@@ -51,3 +52,51 @@ class RepositoryContentUnit(Document):
...
+    content_unit_counts = DictField()
+    last_unit_added = DateTimeField()
```
ipanova: I am not sure, but should not we use the ISO8601StringField here?

server/pulp/server/db/manage.py
```diff
@@ -116,6 +116,7 @@ def ensure_database_indexes():
     collection does not already exist.
     """
     model.RepositoryContentUnit.ensure_indexes()
+    model.Repository.ensure_indexes()
```
ipanova: please add the testcase for this like here https://github.com/pulp/pulp/blob/master/server/test/unit/server/db/test_manage.py#L118

ipanova:
@asmacdo  assigning this back to you, ping me when you're ready

server/pulp/server/db/model/__init__.py
asmacdo: I copied this over from the Repo "model". I guess this was removed at some point. Good catch.

server/pulp/server/db/model/__init__.py
asmacdo: When creating a repo, if `display_name` was not specified, it is automatically set to `repo_id`, meaning there can never be a repo without a `display_name`. However, if `description` is not specified, it remains null.

server/pulp/server/db/model/__init__.py
asmacdo: the ISO8601StringField is actually not ideal, but is necessary for backwards compatibility. It stores a string equivalent of a real DateTime object. In this place, we are able to actually use a proper DateTime object, so DateTimeField is correct.

asmacdo:
ok test

asmacdo:
@ipanova, changes made, assigned back to you.

server/test/unit/server/db/test_manage.py
```diff
@@ -116,6 +116,7 @@ def test_ensure_index(self, mock_model):
         self.assertTrue(mock_model.ReservedResource.ensure_indexes.called)
         self.assertTrue(mock_model.TaskStatus.ensure_indexes.called)
         self.assertTrue(mock_model.Worker.ensure_indexes.called)
+        self.assertTrue(mock_model.Repository.ensure_indexes.called)
```
ipanova: I'd recommend make it in alphabetical order.

server/test/unit/server/db/test_model.py
```diff
@@ -152,3 +152,67 @@ def test_meta_collection(self):
...
+        Ensure that the Repository model is a subclass of Mongoengine's Document class.
+        """
+        sample_model = model.Repository()
+        self.assertTrue(isinstance(sample_model, Document))
+
+    def test_model_fields(self):
+        """
+        Assert that each field has the correct type and `required` status.
+        """
+        self.assertTrue(isinstance(model.Repository.repo_id, StringField))
```
ipanova: you could also check repo_id.db_field

ipanova:
@asmacdo There are only two small comments. LGTM after you'll make the changes. Well done ^_^

server/test/unit/server/db/test_model.py
asmacdo: good call.

asmacdo:
ok test

PR #1818, opened 2015-04-22
PEP-8 17 test modules in client_lib/.

rbarlow:


beav:
LGTM

If we are close to 100% PEP-8 compliance, maybe it's better to do this for `2.7-dev` instead of `master` to reduce merge conflicts when forward-merging 2.7 fixes in the future.

PR #1820, opened 2015-04-23
Moving converted models to mongoengine into __init__.py

ipanova:


server/pulp/server/db/manage.py
```diff
@@ -116,9 +115,9 @@ def ensure_database_indexes():
     collection does not already exist.
     """
     model.RepositoryContentUnit.ensure_indexes()
-    dispatch.TaskStatus.ensure_indexes()
-    workers.Worker.ensure_indexes()
-    resources.ReservedResource.ensure_indexes()
+    model.TaskStatus.ensure_indexes()
+    model.Worker.ensure_indexes()
+    model.ReservedResource.ensure_indexes()
```
barnabycourt: I would recommend putting these in alphabetical order

server/pulp/server/db/reaper.py
```diff
@@ -45,8 +46,8 @@ def reap_expired_documents():
     number of days as the argument.
     """
     _logger.info(_('The reaper task is cleaning out old documents from the database.'))
-    for model, config_name in _COLLECTION_TIMEDELTAS.items():
+    for _model, config_name in _COLLECTION_TIMEDELTAS.items():
```
barnabycourt: using _model makes me think it is a private module level variable.  How about model_class as an alternative?

barnabycourt:
Couple of comments but looks good on the whole

PR #1821, opened 2015-04-23
Allow basic auth when syncing from repos

beav:
Nectar already has support for basic auth when downloading files. However, Pulp
did not have an option to set this on the importer.

This commit adds a `basic_auth_username` and `basic_auth_password` to importer
configs to allow for basic auth. Options to `pulp-admin` have also been added.
Plugins will not automatically be able to use basic auth, and will have to set
`include_basic_auth=True` when using the *ImporterConfigMixin* mixin.

This commit contains an additional security-related change to obfuscate
`basic_auth_password` and `proxy_password` in importer configs when returned
via the REST api.

re #163

### DIFF
--- a/server/pulp/plugins/util/importer_config.py
+++ b/server/pulp/plugins/util/importer_config.py
@@ -264,6 +264,60 @@ def validate_proxy_password(config):
         raise ValueError(msg)
 
 
+def validate_basic_auth_username(config):
+    """
+    The basic_auth_username is optional. If it is set, this method will ensure
+    that it is a string, and it will also ensure that the basic_auth_password is set.
+
+    :param config: The configuration object that we are validating.
+    :type config: pulp.plugins.config.PluginCallConfiguration
+    """
+    basic_auth_username = config.get(importer_constants.KEY_BASIC_AUTH_USER)
+    # basic_auth username is not required unless the password is set
+    if basic_auth_username is None and config.get(importer_constants.KEY_BASIC_AUTH_PASS) is None:
+        return
+    elif basic_auth_username is None:
+        # If basic_auth_password is set, basic_auth_username must also be set
+        msg = _('The configuration parameter <%(password_name)s> requires the <%(username_name)s> '
+                'parameter to also be set.')
+        msg = msg % {'password_name': importer_constants.KEY_BASIC_AUTH_PASS,
+                     'username_name': importer_constants.KEY_BASIC_AUTH_USER}
+        raise ValueError(msg)
+
+    if not isinstance(basic_auth_username, basestring):
+        msg = _('The configuration parameter <%(name)s> should be a string, but it was %(type)s.')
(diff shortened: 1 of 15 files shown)


server/pulp/plugins/util/importer_config.py
```diff
@@ -264,6 +264,60 @@ def validate_proxy_password(config):
...
+                'parameter to also be set.')
+        msg = msg % {'password_name': importer_constants.KEY_BASIC_AUTH_PASS,
+                     'username_name': importer_constants.KEY_BASIC_AUTH_USER}
+        raise ValueError(msg)
+
+    if not isinstance(basic_auth_username, basestring):
+        msg = _('The configuration parameter <%(name)s> should be a string, but it was %(type)s.')
+        msg = msg % {'name': importer_constants.KEY_BASIC_AUTH_USER,
+                     'type': type(basic_auth_username)}
+        raise ValueError(msg)
```
mhrivnak: perhaps use ``TypeError`` in this case.

server/pulp/server/managers/repo/importer.py
```diff
@@ -346,6 +346,32 @@ def set_importer_scratchpad(self, repo_id, contents):
...
+        if not importer:
+            return
+
+        if 'config' in importer and importer['config']:
+            if 'proxy_password' in importer['config'] and \
+                    importer['config']['proxy_password'] is not None:
+                importer['config']['proxy_password'] = '*****'
+            if 'basic_auth_password' in importer['config'] and \
+                    importer['config']['basic_auth_password'] is not None:
+                importer['config']['basic_auth_password'] = '*****'
```
beav: I realized during breakfast that these should be using strings from `constants`. I will update the PR.

:egg: :coffee:

server/pulp/plugins/util/importer_config.py
```diff
@@ -264,6 +264,60 @@ def validate_proxy_password(config):
...
+        msg = msg % {'password_name': importer_constants.KEY_BASIC_AUTH_PASS,
+                     'username_name': importer_constants.KEY_BASIC_AUTH_USER}
+        raise ValueError(msg)
+
+    if not isinstance(basic_auth_password, basestring):
+        msg = _('The configuration parameter <%(basic_auth_password_name)s> should be a string, '
+                'but it was %(type)s.')
+        msg = msg % {'basic_auth_password_name': importer_constants.KEY_BASIC_AUTH_PASS,
+                     'type': type(basic_auth_password)}
+        raise ValueError(msg)
```
mhrivnak: same here. ``TypeError`` is the correct exception in python-land, but I'm not sure if the importer will handle that correctly. It's worth at least checking.

beav: This occurs in a number of places in this file. I opened https://pulp.plan.io/issues/921 so all of them can get fixed at once.

beav:
ok test

PR #1824, opened 2015-04-24
Turned off DEBUG in django settings.py

dkliban:
Fixes #910

### DIFF
--- a/docs/user-guide/installation.rst
+++ b/docs/user-guide/installation.rst
@@ -240,6 +240,9 @@ Server
 
         `SSLProtocol all -SSLv2 -SSLv3`
 
+  .. warning::
+     It is recommended that the web server only serves Pulp services.
+
 #. Start Apache httpd and set it to start on boot. For Upstart based systems::
 
     $ sudo service httpd start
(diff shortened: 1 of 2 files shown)


server/pulp/server/webservices/settings.py
```diff
@@ -3,15 +3,14 @@
...
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = False
 
-TEMPLATE_DEBUG = True
+TEMPLATE_DEBUG = False
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['*']
```
bmbouter: Can this be set to the a more secure value.  Perhaps using the server hostname from `server.conf` would work.

docs/user-guide/installation.rst
```diff
@@ -240,6 +240,9 @@ Server
 
         `SSLProtocol all -SSLv2 -SSLv3`
 
+  .. warning::
+     It is recommended that the web server only serves Pulp related services.
```
bmbouter: s/servers/serve/
s/related//

bmbouter:
Small wording suggestion. LGTM

PR #1825, opened 2015-04-27
The status API return body example in docs is not formatted correctly.

ipanova:
closes #852


beav:
This renders correctly now, tested in firefox :cat2: 

LGTM

PR #1826, opened 2015-04-27
API call to Update a scheduled publish returns 400 with invalid schedule_id

ipanova:
closes #768

This issue also was present in update/delete methods in schedule sync/publish and in
consumer schedule content install/update/uninstall.

Manual testing for:
scheduled sync http://fpaste.org/215722/43012388/
scheduled publish http://fpaste.org/215721/43012374/
scheduled consumer content install http://fpaste.org/215723/24126143/ (skipped update/uninstall - they are the same)

### DIFF
--- a/server/pulp/server/managers/schedule/utils.py
+++ b/server/pulp/server/managers/schedule/utils.py
@@ -98,7 +98,9 @@ def delete(schedule_id):
     try:
         spec = {'_id': ObjectId(schedule_id)}
     except InvalidId:
-        raise exceptions.InvalidValue(['schedule_id'])
+        # During schedule deletion, MissingResource should be raised even if
+        # schedule_id is invalid object_id.
+        raise exceptions.MissingResource(schedule_id=schedule_id)
 
     schedule = ScheduledCall.get_collection().find_and_modify(
         query=spec, remove=True, safe=True)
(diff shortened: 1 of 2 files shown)


barnabycourt:
Hi, I'ma  little confused by the fix. The issue is not so much a missing resource but that the value specified was not a valid UUID if I read the code properly. Shouldn't this be returning a validation error 400 and not a 404? I could see where we should be wrapping an appropriate error message to go with the 400 but 400 does appear to be more specific than 404 for this error

server/pulp/server/managers/schedule/utils.py
```diff
@@ -98,7 +98,7 @@ def delete(schedule_id):
     try:
         spec = {'_id': ObjectId(schedule_id)}
     except InvalidId:
-        raise exceptions.InvalidValue(['schedule_id'])
+        raise exceptions.MissingResource(schedule_id=schedule_id)
```
barnabycourt: Add a comment here on why this is being converted to a MissingResource rexception

server/pulp/server/managers/schedule/utils.py
```diff
@@ -147,7 +147,7 @@ def update(schedule_id, delta):
...
```
barnabycourt: Add a comment here on why this is being converted to a MissingResource rexception

PR #1829, opened 2015-04-28
Initial content unit model

barnabycourt:
https://pulp.plan.io/issues/847

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -84,6 +84,11 @@ PLP0033 = Error("PLP0033", _("Working Directory requested outside of asynchronou
 PLP0034 = Error("PLP0034", _("The distributor %(distributor_id)s indicated a failed response when "
                              "publishing repository %(repository_id)s."),
                 ['distributor_id', 'repository_id'])
+PLP0035 = Error("PLP0035", _("The ContentUnit model class %(class_name)s failed to define the "
+                             "attribute unit_key_fields"), ['class_name'])
+PLP0036 = Error("PLP0036", _("The source_location: %(source_location) specified for the content unit is invalid."),
+                ['source_location'])
+PLP0037 = Error("PLP0037", _("The relative_path specified for the content unit is empty."), [])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -1,11 +1,16 @@
 import copy
+import errno
 import logging
+import os
+import shutil
+import uuid
 
 from mongoengine import (DateTimeField, DictField, Document, DynamicField, IntField,
                          ListField, StringField)
 from mongoengine import signals
 
-from pulp.common import constants
+from pulp.common import constants, dateutils, error_codes
+from pulp.server import config, exceptions
 from pulp.server.async.emit import send as send_taskstatus_message
 from pulp.server.db.model.base import CriteriaQuerySet
 from pulp.server.db.model.fields import ISO8601StringField
@@ -307,3 +312,129 @@ class TaskStatus(Document, ReaperMixin):
 
 
 signals.post_save.connect(TaskStatus.post_save, sender=TaskStatus)
+
+
+class ContentUnit(Document):
+    """
+    The base class for all content units.
+
+    All classes inheriting from this class must override the unit_type_id and _ns to ensure
+    they are populated properly.
+
+    :ivar id: content unit id
+    :type id: mongoengine.StringField
+    :ivar last_updated: last time this unit was updated (since epoch, zulu time)
+    :type last_updated: mongoengine.IntField
+    :ivar user_metadata: Bag of User supplied data to go along with this unit
+    :type user_metadata: mongoengine.DictField
+    :ivar storage_path: Location on disk where the content associated with this unit lives
+    :type storage_path: mongoengine.StringField
+
+    :ivar _ns: (Deprecated), Contains the name of the collection this model represents
+    :type _ns: mongoengine.StringField
+    :ivar unit_type_id: content unit type
+    :type unit_type_id: mongoengine.StringField
+    """
+
+    id = StringField(primary_key=True)
+    last_updated = IntField(db_field='_last_updated', required=True)
+    user_metadata = DictField(db_field='pulp_user_metadata')
+    storage_path = StringField(db_field='_storage_path')
+
+    # For backward compatibility
+    _ns = StringField(required=True)
+    unit_type_id = StringField(db_field='_content_type_id', required=True)
+
+    meta = {
+        'abstract': True,
+    }
+
+    def __init__(self, *args, **kwargs):
+        super(ContentUnit, self).__init__(*args, **kwargs)
+        self._source_location = None
+        self._relative_path = None
+
+    @classmethod
+    def post_init_signal(cls, sender, document):
+        """
+        The signal that is triggered before a unit is initialized
+
+        This is used to validate that the unit_key_fields attribute is set properly
+
+        :param sender: sender class
+        :type sender: object
+        :param document: Document that sent the signal
+        :type document: ContentUnit
+        :raises: PLP0035 if the unit_key_fields attribute has not been defined
+        """
+        if not hasattr(document, 'unit_key_fields'):
+            class_name = type(document).__name__
+            raise exceptions.PulpCodedException(error_codes.PLP0035, class_name=class_name)
+
+    @classmethod
+    def pre_save_signal(cls, sender, document, **kwargs):
+        """
+        The signal that is triggered before a unit is saved, this is used to
+        support the legacy behavior of generating the unit id and setting
+        the last_updated timestamp
+
+        :param sender: sender class
+        :type sender: object
+        :param document: Document that sent the signal
+        :type document: ContentUnit
+        """
+        if not document.id:
+            document.id = str(uuid.uuid4())
+        document.last_updated = dateutils.now_utc_timestamp()
+
+        # If content was set on this unit, copy the content into place
+        if document._source_location and document._relative_path:
+            server_storage_dir = config.config.get('server', 'storage_dir')
+            platform_storage_dir = os.path.join(server_storage_dir, 'content',
+                                                document.unit_type_id)
+            target_location = os.path.join(platform_storage_dir, document._relative_path)
+            # Make if source is a directory, recursively copy it, otherwise copy the file
+            if os.path.isdir(document._source_location):
+                shutil.copytree(document._source_location, target_location)
+            else:
+                check_dir = os.path.dirname(target_location)
+                try:
+                    os.makedirs(check_dir)
+                except OSError as e:
+                    if e.errno != errno.EEXIST:
+                        raise
+                shutil.copy(document._source_location, target_location)
+
+    def set_content(self, source_location, relative_path):
+        """
+        Store the source of the content for the unit and the relative path
+        where it should be stored within the plugin content directory.
+
+        :param source_location: The absolute path to the content in the plugin working directory.
+        :type source_location: str
+        :param relative_path: The relative path where the content should be stored inside the
+                              content type specific content folder
+        :type relative_path: str
+
+        :raises PulpCodedException: PLP0036 if the source_location doesn't exist.
+                                    PLP0037 if the relative_path is not included
+        """
+        if not os.path.exists(source_location):
+            raise exceptions.PulpCodedException(error_code=error_codes.PLP0036,
+                                                source_location=source_location)
+        if relative_path is None or relative_path.strip() == '':
+            raise exceptions.PulpCodedException(error_code=error_codes.PLP0037)
+
+        self._source_location = source_location
+        self._relative_path = relative_path
+
+    def get_repositories(self):
+        """
+        Get an iterable of Repository models for all the repositories that contain this unit
+
+        :return: Repositories that contain this content unit
+        :rtype: iterable of Repository
+        """
+        content_list = RepositoryContentUnit.objects(unit_id=self.id)
+        id_list = [item.repo_id for item in content_list]
+        return Repository.objects(repo_id__in=id_list)
--- a/server/test/unit/server/db/test_model.py
+++ b/server/test/unit/server/db/test_model.py
@@ -1,17 +1,223 @@
 """
 Tests for the pulp.server.db.model module.
 """
+import os
+import shutil
+from tempfile import mkdtemp
+
+from mock import patch
+
 try:
     import unittest2 as unittest
 except ImportError:
     import unittest
 
-
 from mongoengine import DateTimeField, DictField, Document, IntField, StringField
 
+from pulp.common import error_codes
+from pulp.devel.unit.util import touch
+from pulp.server.exceptions import PulpCodedException
 from pulp.server.db import model
 from pulp.server.db.model.base import CriteriaQuerySet
 from pulp.server.db.model.fields import ISO8601StringField
+
+
+class TestContentUnit(unittest.TestCase):
+    """
+    Test ContentUnit model
+    """
+
+    def setUp(self):
+        self.working_dir = mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.working_dir, ignore_errors=True)
+
+    def test_model_fields(self):
+        self.assertTrue(isinstance(model.ContentUnit.id, StringField))
+        self.assertTrue(model.ContentUnit.id.primary_key)
+
+        self.assertTrue(isinstance(model.ContentUnit.last_updated, IntField))
+        self.assertTrue(model.ContentUnit.last_updated.required)
+        self.assertEquals(model.ContentUnit.last_updated.db_field, '_last_updated')
+
+        self.assertTrue(isinstance(model.ContentUnit.user_metadata, DictField))
+        self.assertEquals(model.ContentUnit.user_metadata.db_field, 'pulp_user_metadata')
+
+        self.assertTrue(isinstance(model.ContentUnit.storage_path, StringField))
+        self.assertEquals(model.ContentUnit.storage_path.db_field, '_storage_path')
+
+        self.assertTrue(isinstance(model.ContentUnit._ns, StringField))
+        self.assertTrue(model.ContentUnit._ns)
+        self.assertTrue(isinstance(model.ContentUnit.unit_type_id, StringField))
+        self.assertTrue(model.ContentUnit.unit_type_id)
+
+    def test_meta_abstract(self):
+        self.assertEquals(model.ContentUnit._meta['abstract'], True)
+
+    def test_post_init_signal_with_unit_key_fields_defined(self):
+        """
+        Test the init signal handler that validates the existence of the
+        unit_key_fields list
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            unit_key_fields = ['id']
+
+        model.ContentUnit.post_init_signal({}, ContentUnitHelper())
+
+    def test_post_init_signal_without_unit_key_fields_defined(self):
+        """
+        Test the init signal handler that raises an exception if the unit_key_fields list
+        has not been defined.
+        """
+        try:
+            model.ContentUnit.post_init_signal({}, {})
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0035)
+
+    @patch('pulp.server.db.model.dateutils.now_utc_timestamp')
+    def test_pre_save_signal(self, mock_now_utc):
+        """
+        Test the pre_save signal handler
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            id = None
+            last_updated = None
+
+        mock_now_utc.return_value = 'foo'
+        helper = ContentUnitHelper()
+        helper.last_updated = 50
+
+        model.ContentUnit.pre_save_signal({}, helper)
+
+        self.assertIsNotNone(helper.id)
+
+        # make sure the last updated time has been updated
+        self.assertEquals(helper.last_updated, 'foo')
+
+    def test_pre_save_signal_leaves_existing_id(self):
+        """
+        Test the pre_save signal handler leaves an existing id on an object in place
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            id = None
+            last_updated = None
+
+        helper = ContentUnitHelper()
+        helper.id = "foo"
+
+        model.ContentUnit.pre_save_signal({}, helper)
+
+        # make sure the id wasn't replaced
+        self.assertEquals(helper.id, 'foo')
+
+    def test_set_content(self):
+        class ContentUnitHelper(model.ContentUnit):
+            pass
+        helper = ContentUnitHelper()
+        helper.set_content(self.working_dir, 'apples')
+        self.assertEquals(helper._source_location, self.working_dir)
+        self.assertEquals(helper._relative_path, 'apples')
+
+    def test_set_content_bad_source_location(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with a non existent source_location
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            pass
+        helper = ContentUnitHelper()
+        try:
+            helper.set_content(os.path.join(self.working_dir, 'foo'), 'apples')
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException as raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0036)
+
+    def test_set_content_no_relative_path(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with None for the relative_path
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            pass
+        helper = ContentUnitHelper()
+        try:
+            helper.set_content(self.working_dir, None)
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0037)
+
+    def test_set_content_empty_relative_path(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with an empty relative_path
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            pass
+        helper = ContentUnitHelper()
+        try:
+            helper.set_content(self.working_dir, '  ')
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException as raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0037)
+
+    @patch('pulp.server.db.model.config')
+    def test_pre_save_signal_directory_content(self, mock_config):
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
+        os.mkdir(target_dir)
+        mock_config.config.get.return_value = target_dir
+        # create something in the source directory to test the copy
+        touch(os.path.join(source_dir, 'foo', 'bar'))
+
+        class ContentUnitHelper(model.ContentUnit):
+            unit_type_id = 'foo_unit'
+        foo = ContentUnitHelper()
+        foo.set_content(source_dir, 'apples')
+        model.ContentUnit.pre_save_signal(object(), foo)
+        full_path = os.path.join(target_dir, 'content', 'foo_unit', 'apples', 'foo', 'bar')
+        self.assertTrue(os.path.exists(full_path))
+
+    @patch('pulp.server.db.model.config')
+    def test_pre_save_signal_file_content(self, mock_config):
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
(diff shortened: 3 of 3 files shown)


bmbouter:
Please amend the commit to have "re #847" so that it will show up as a related ref

common/pulp/common/error_codes.py
```diff
@@ -84,6 +84,11 @@
 PLP0034 = Error("PLP0034", _("The distributor %(distributor_id)s indicated a failed response when "
                              "publishing repository %(repository_id)s."),
                 ['distributor_id', 'repository_id'])
+PLP0035 = Error("PLP0035", _("The ContentUnit model class %(class_name)s failed to define the "
```
bmbouter: Is there a way to make this less technical. I don't think a use would know what a ContentUnit model class is or familiarity with the unit_key_fields attribute. I don't have a specific suggestion for how to improve it.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
 
 
 signals.post_save.connect(TaskStatus.post_save, sender=TaskStatus)
+
+
+class ContentUnit(Document):
+    """
+    The base class for all content units.
+
+    All unit inheriting from this class should override the unit_type_id and _ns to ensure
```
bmbouter: s/unit/unit classes/
s/this class/this object/

bmbouter: Do all subclasses need to override unit_type_id and _ns? If they all must then s/should/must/

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    that they are populated properly
+
```
bmbouter: Suggest removing the extra line of whitespace

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
 
 signals.post_save.connect(TaskStatus.post_save, sender=TaskStatus)
+
+
+class ContentUnit(Document):
+    """
+    The base class for all content units.
+
+    All unit inheriting from this class should override the unit_type_id and _ns to ensure
+    that they are populated properly
```
bmbouter: s/properly/properly./

bmbouter: s/that they/they/

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+
+    :ivar id: string representation of content unit id
+    :type id: mongoengine.StringField
+    :ivar last_updated: integer representation of last time this unit was updated (mer
+    :type last_updated: mongoengine.IntField
+    :ivar user_metadata: Bag of user supplied data to go along with this unit
```
bmbouter: Is this a bag of dicts?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+class ContentUnit(Document):
+    """
+    The base class for all content units.
+
+    All unit inheriting from this class should override the unit_type_id and _ns to ensure
+    that they are populated properly
+
+
+    :ivar id: string representation of content unit id
```
bmbouter: the type should be clear from the next line. `string` can be removed.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    :type id: mongoengine.StringField
+    :ivar last_updated: integer representation of last time this unit was updated (mer
```
bmbouter: the type should be clear from the next line. `integer` can be removed. Suggest rewritten to:  "last time this unit was updated" This also gets rid of the `(mer` at the end. I don't know what this is.

server/pulp/server/db/model/__init__.py
bmbouter: s/Bag of user/User/

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    :type last_updated: mongoengine.IntField
+    :ivar user_metadata: Bag of user supplied data to go along with this unit
+    :type user_metadata: mongoengine.DictField
+    :ivar storage_path: Location on disk where the content associated with this unit lives
+    :type storage_path: mongoengine.StringField
+
+    :ivar _ns: The namespace field (Deprecated), reading
+    :type _ns: mongoengine.StringField
+    :ivar unit_type_id: string representation of content unit type
```
bmbouter: the type should be clear from the next line. `string` can be removed.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    :ivar id: string representation of content unit id
+    :type id: mongoengine.StringField
+    :ivar last_updated: integer representation of last time this unit was updated (mer
+    :type last_updated: mongoengine.IntField
+    :ivar user_metadata: Bag of user supplied data to go along with this unit
+    :type user_metadata: mongoengine.DictField
+    :ivar storage_path: Location on disk where the content associated with this unit lives
+    :type storage_path: mongoengine.StringField
+
+    :ivar _ns: The namespace field (Deprecated), reading
```
bmbouter: s/, reading/Contains the name of the collection this model represents./

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+    id = StringField(primary_key=True)
+    last_updated = IntField(db_field='_last_updated', required=True)
+    user_metadata = DictField(db_field='pulp_user_metadata')
+    storage_path = StringField(db_field='_storage_path')
+
+    # For backward compatibility
+    _ns = StringField(required=True)
+    unit_type_id = StringField(db_field='_content_type_id', required=True)
+    _source_location = None
```
bmbouter: This isn't in the docblock above, what does it do?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    _relative_path = None
```
bmbouter: This isn't in the docblock above, what does it do?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    def post_init_signal(cls, sender, document):
+        """
+        The signal that is triggered before a unit is initialized
+
+        This is used to validate that the unit_key_fields attribute is set properly
+
+        :param sender: sender class
+        :type sender: object
+        :param document: Document that sent the signal
+        :type document: ContentUnit
```
bmbouter: This needs a `:raises:` section documenting what is raised and when.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+        """
+
+        if not hasattr(document, 'unit_key_fields'):
+            class_name = type(document).__name__
+            raise exceptions.PulpCodedException(error_codes.PLP0035, class_name=class_name)
```
bmbouter: Can't `cls` be used instead of class_name? If so the above line can also be removed.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+    @classmethod
+    def pre_save_signal(cls, sender, document, **kwargs):
+        """
+        The signal that is triggered before a unit is saved
```
bmbouter: Can an additional descriptive sentence be added here that describes what this code does?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+        :type document: ContentUnit
+        """
+        if not document.id:
+            document.id = str(uuid.uuid4())
+        document.last_updated = dateutils.now_utc_timestamp()
+
+    @classmethod
+    def post_save_signal(cls, sender, document, **kwargs):
+        """
+        The signal that is triggered after a unit is saved
```
bmbouter: Can an additional descriptive sentence be added here that describes what this code does?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+        :param sender: sender class
+        :type sender: object
+        :param document: Document that sent the signal
+        :type document: ContentUnit
+        """
+        # If content was set on this unit, move the content into place
```
bmbouter: Why do we move content after the save and not before? If we do it after there is a small window of time where the record is in place but the content is not. Also if an error occurs where content couldn't be moved then the record is already in place. What do you think?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+            server_storage_dir = config.config.get('server', 'storage_dir')
+            platform_storage_dir = os.path.join(server_storage_dir, 'content',
+                                                document.unit_type_id)
+            target_location = os.path.join(platform_storage_dir, document._relative_path)
+            # Make sure the directory for the target exists
+            if os.path.isdir(document._source_location):
+                shutil.copytree(document._source_location, target_location)
+            else:
+                check_dir = os.path.dirname(target_location)
+                misc.mkdir(check_dir)
```
bmbouter: Can we not use misc.mkdir() and use just the standard library stuff?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+        """
+        # If content was set on this unit, move the content into place
+        if document._source_location and document._relative_path:
+            server_storage_dir = config.config.get('server', 'storage_dir')
+            platform_storage_dir = os.path.join(server_storage_dir, 'content',
+                                                document.unit_type_id)
+            target_location = os.path.join(platform_storage_dir, document._relative_path)
+            # Make sure the directory for the target exists
+            if os.path.isdir(document._source_location):
+                shutil.copytree(document._source_location, target_location)
```
bmbouter: The note above says that this moves content, but this and line 409 copy things. Can one of them be updated so they agree on the operation performed?

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+        :param document: Document that sent the signal
+        :type document: ContentUnit
+        """
+        # If content was set on this unit, move the content into place
+        if document._source_location and document._relative_path:
+            server_storage_dir = config.config.get('server', 'storage_dir')
+            platform_storage_dir = os.path.join(server_storage_dir, 'content',
+                                                document.unit_type_id)
+            target_location = os.path.join(platform_storage_dir, document._relative_path)
+            # Make sure the directory for the target exists
```
bmbouter: This comment doesn't seem to sync up with the check on L#403 which checks if the source location exists (not the target).

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+            if os.path.isdir(document._source_location):
+                shutil.copytree(document._source_location, target_location)
+            else:
+                check_dir = os.path.dirname(target_location)
+                misc.mkdir(check_dir)
+                # Copy the content into place
```
bmbouter: This can probably be removed L#409 is very readable.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+                shutil.copy(document._source_location, target_location)
+
+    def set_content(self, source_location, relative_path):
+        """
+        Store the source of the content for the unit and the relative path
+        where it should be stored within the plugin content directory
```
bmbouter: s/directory/directory./

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+
+        :param source_location: The absolute path to the content in the plugin working directory
+        :type source_location: str
```
bmbouter: I'm not sure if this is a good suggestion, but maybe `basestr` instead for these since unicode should work for these.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
 """
 Tests for the pulp.server.db.model module.
 """
+from tempfile import mkdtemp
```
bmbouter: These three imports are in reverse alphabetical order.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+    :ivar storage_path: Location on disk where the content associated with this unit lives
+    :type storage_path: mongoengine.StringField
+
+    :ivar _ns: The namespace field (Deprecated), reading
+    :type _ns: mongoengine.StringField
+    :ivar unit_type_id: string representation of content unit type
+    :type unit_type_id: mongoengine.StringField
+    """
+
+    id = StringField(primary_key=True)
```
bmbouter: Will this cause problems like the ones we were discussing last week where the 'id' attribute is aliased to _id internally in mongoengine?

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+    Test ContentUnit model
+    """
+
+    def setUp(self):
+        self.working_dir = mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.working_dir, ignore_errors=True)
+
+    def test_model_fields(self):
```
bmbouter: Is there any way to add the db_field kwargs to these assertions?

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+
+    def test_post_init_signal_with_unit_key_fields_defined(self):
+        """
+        Test the init signal handler that validates the existence of the
+        unit_key_fields list
+        """
+        class ContentUnitHelper(model.ContentUnit):
+            unit_key_fields = ['id']
+
+        model.ContentUnit.post_init_signal({}, ContentUnitHelper())
```
bmbouter: I don't see any assertions in this. I assume the test is so that if the validation fails an exception will be raised and the test will Error. In that case it would be good to catch the expected exception type and then call `self.fail()`. Something should be done to either add an assertion or convert an unexpected error into a graceful failure.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+
+    def test_post_init_signal_without_unit_key_fields_defined(self):
+        """
+        Test the init signal handler that raises an exception if the unit_key_fields list
+        has not been defined.
+        """
+        try:
+            model.ContentUnit.post_init_signal({}, {})
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
```
bmbouter: Do you mean `except PulpCodedException as raised_error:`. I thought the comma syntax was for multiple Exception types on one except block which is different then what I think this is attempting to do.

server/pulp/server/db/model/__init__.py
```diff
@@ -307,3 +312,132 @@ def post_save(cls, sender, document, **kwargs):
...
+        The signal that is triggered before a unit is saved
+
+        :param sender: sender class
+        :type sender: object
+        :param document: Document that sent the signal
+        :type document: ContentUnit
+        """
+        if not document.id:
+            document.id = str(uuid.uuid4())
+        document.last_updated = dateutils.now_utc_timestamp()
```
bmbouter: Does it make sense for this to only use standard library stuff? I'm trying to get us away from our litany of helper functions.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            self.assertEquals(raised_error.error_code, error_codes.PLP0035)
+
+    def test_pre_save_signal(self):
```
bmbouter: Can this test mock the utc time generation so that the assertion can be more precise.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
 
+from pulp.common import error_codes
+from pulp.devel.unit.util import touch
+from pulp.server.exceptions import PulpCodedException
 from pulp.server.db import model
 from pulp.server.db.model.base import CriteriaQuerySet
 from pulp.server.db.model.fields import ISO8601StringField
 
 
+class TestContentUnit(unittest.TestCase):
```
bmbouter: unittest.TestCase is the right this to use; good deal. :+1:

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            model.ContentUnit.post_init_signal({}, {})
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0035)
+
+    def test_pre_save_signal(self):
+        """
+        Test the pre_save signal handler
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Since these tests are on the Contentnit itself, can all of these in-test subclasses be removed, and instead instantiate a ContentUnit() and then set attributes on the instantiated object? I'm trying to separate the test code from the subclassing behavior/concern.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+        self.assertIsNotNone(helper.id)
+
+        # make sure the last updated time has been updated
+        self.assertTrue(helper.last_updated > 50)
+
+    def test_pre_save_signal_leaves_existing_id(self):
+        """
+        Test the pre_save signal handler leaves an existing id on an object in place
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar comment here as L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+        helper = ContentUnitHelper()
+        helper.id = "foo"
+
+        model.ContentUnit.pre_save_signal({}, helper)
+
+        # make sure the id wasn't replaced
+        self.assertEquals(helper.id, 'foo')
+
+    def test_set_content(self):
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            pass
+        helper = ContentUnitHelper()
+        helper.set_content(self.working_dir, 'apples')
```
bmbouter: I expected this test to either have some kind of assertion.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+
+    def test_set_content_bad_source_location(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with a non existent source_location
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0036)
+
+    def test_set_content_no_relative_path(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with None for the relative_path
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0037)
+
+    def test_set_content_empty_relative_path(self):
+        """
+        Test that the appropriate exception is raised when set_content
+        is called with an empty relative_path
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            self.fail("Previous call should have raised a PulpCodedException")
+        except PulpCodedException, raised_error:
+            self.assertEquals(raised_error.error_code, error_codes.PLP0037)
+
+    def test_post_save_signal_no_content(self):
+        """
+        Check to make sure that things pass through with no work or errors if no content
+        is set
+        """
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+
+    @patch('pulp.server.db.model.config')
+    def test_post_save_signal_directory_content(self, mock_config):
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
+        os.mkdir(target_dir)
+        mock_config.config.get.return_value = target_dir
+        # create something in the source directory to test the copy
+        touch(os.path.join(source_dir, 'foo', 'bar'))
```
bmbouter: Can we avoid using the pulp utility helpers. There must be a standard library way to do this easily.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+    @patch('pulp.server.db.model.config')
+    def test_post_save_signal_file_content(self, mock_config):
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
+        os.mkdir(target_dir)
+        mock_config.config.get.return_value = target_dir
+        # create something in the source directory to test the copy
+        source_file = os.path.join(source_dir, 'bar')
+        touch(source_file)
```
bmbouter: Can we avoid using the pulp utility helpers. There must be a standard library way to do this easily.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+    def test_post_save_signal_directory_content(self, mock_config):
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
+        os.mkdir(target_dir)
+        mock_config.config.get.return_value = target_dir
+        # create something in the source directory to test the copy
+        touch(os.path.join(source_dir, 'foo', 'bar'))
+
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            unit_type_id = 'foo_unit'
+        foo = ContentUnitHelper()
+        foo.set_content(source_dir, 'apples')
+        model.ContentUnit.post_save_signal(object(), foo)
+
+        self.assertTrue(os.path.exists(os.path.join(target_dir, 'content', 'foo_unit',
```
bmbouter: This could be broken into two lines for readability.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+        source_dir = os.path.join(self.working_dir, 'src')
+        target_dir = os.path.join(self.working_dir, 'target')
+        os.mkdir(source_dir)
+        os.mkdir(target_dir)
+        mock_config.config.get.return_value = target_dir
+        # create something in the source directory to test the copy
+        source_file = os.path.join(source_dir, 'bar')
+        touch(source_file)
+
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+            unit_type_id = 'foo_unit'
+            pass
+        foo = ContentUnitHelper()
+        foo.set_content(source_file, 'foo/bar')
+        model.ContentUnit.post_save_signal(object(), foo)
+
+        self.assertTrue(os.path.exists(os.path.join(target_dir, 'content', 'foo_unit',
```
bmbouter: This could be broken into two lines for readability.

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
+                                                    'foo', 'bar')))
+
+    @patch('pulp.server.db.model.Repository.objects')
+    @patch('pulp.server.db.model.RepositoryContentUnit.objects')
+    def test_get_repositories(self, mock_rcu_query, mock_repository_query):
+        class ContentUnitHelper(model.ContentUnit):
```
bmbouter: Similar here as in L#80

server/test/unit/server/db/test_model.py
```diff
@@ -1,19 +1,232 @@
...
 from pulp.server.db.model.fields import ISO8601StringField
 
 
+class TestContentUnit(unittest.TestCase):
+    """
+    Test ContentUnit model
+    """
+
+    def setUp(self):
+        self.working_dir = mkdtemp()
```
bmbouter: I expected this setUp and tearDown to use a lot of mocking and never actually touch the filesystem.

bmbouter:
This PR looks really good, despite the egregious amount of comments I left on it. I also ask several questions for clarification. @barnabycourt I'm sending back to you since my review is done.

server/pulp/server/db/model/__init__.py
barnabycourt: good point

server/pulp/server/db/model/__init__.py
barnabycourt: msc.mkdir() doesn't fail if the directory already exists

server/pulp/server/db/model/__init__.py
barnabycourt: nope, in this case that is the behavior we want to have.

server/test/unit/server/db/test_model.py
barnabycourt: nope, The comma syntax is the old style of how you specify the exception and is supported for backwards compatibility, I'm not sure what is supported on all our platforms so I revert to the old style by default.  https://docs.python.org/2/tutorial/errors.html

server/pulp/server/db/model/__init__.py
barnabycourt: I kept this the same to ensure that the behavior did not change during the conversion

server/test/unit/server/db/test_model.py
barnabycourt: Mongoengine throws an error if the class is instantiated on it's own as it is an abstract class

server/test/unit/server/db/test_model.py
barnabycourt: This checks that it passes if the unit_key_fields have been defined properly

barnabycourt:
ok test

server/pulp/server/db/model/__init__.py
bmbouter: This makes sense for compatibility reasons.

server/pulp/server/db/model/__init__.py
bmbouter: I think having a try/except here would be better than using the helper function.

server/pulp/server/db/model/__init__.py
bmbouter: I think having a try/except here would be better than using the `misc.mkdir()`. Something [just like this](https://github.com/pulp/pulp/pull/1809/files#diff-83c54e49104a094800b69d01f7385232R292) would make the behavior of this code clearer and more intentional.

The reason I don't want us to use `misc.mkdir` is because it is very similarly named to the standard library function, but with a completely different behavior. This causes confusion and the line savings aren't worth it.

server/test/unit/server/db/test_model.py
bmbouter: Consider wrapping this line in a `try/except` and if an exception is caught it should call something like `self.fail("An unexpected Exception was raised even with unit_key_fields correctly defined")`.

server/test/unit/server/db/test_model.py
bmbouter: I think the new style is clearer, and these tests are have a 2.6 compatibility requirement. 2.6 support the `except PulpCodedException as raised_error` statement.

https://docs.python.org/2.6/tutorial/errors.html#user-defined-exceptions

bmbouter:
@barnabycourt thanks for making so many of those changes. I left some more comments. Moving those two class variables to be inside of an __init__ would be good, and doing away with the helper functions as much as possible would be good. Adding the LGTM now; I don't need to re-review.

barnabycourt:
ok test

barnabycourt:
ok test

PR #1832, opened 2015-04-30
Create entry point to load unit models

barnabycourt:
refs: #853

https://pulp.plan.io/issues/853

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -89,6 +89,12 @@ PLP0035 = Error("PLP0035", _("The ContentUnit model class %(class_name)s failed 
 PLP0036 = Error("PLP0036", _("The source_location: %(source_location) specified for the content unit is invalid."),
                 ['source_location'])
 PLP0037 = Error("PLP0037", _("The relative_path specified for the content unit is empty."), [])
+PLP0038 = Error("PLP0038", _("The unit model with id %(model_id)s and class "
+                             "%(model_class)s failed to register. Another model has already "
+                             "been registered with the same id."), ['model_id', 'model_class'])
+PLP0039 = Error("PLP0039", _("The unit model with the id %(model_id)s failed to register. The "
+                             "class %(model_class)s is not a subclass of ContentUnit."),
+                ['model_id', 'model_class'])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
--- a/server/pulp/plugins/loader/api.py
+++ b/server/pulp/plugins/loader/api.py
@@ -40,6 +40,7 @@ def initialize(validate=True):
     if _is_initialized():
         return
 
+    # Initialize the plugin manager, this includes initialization of the unit_model entry point
     _create_manager()
 
     plugin_entry_points = (
@@ -81,7 +82,21 @@ def list_content_types():
     :rtype: list of str
     """
     assert _is_initialized()
-    return database.all_type_ids()
+    types_list = _MANAGER.unit_models.keys()
+    legacy_types = database.all_type_ids()
+    types_list.extend(legacy_types)
+    return types_list
+
+
+def list_unit_models():
+    """
+    Get the id's of the supported unit_models.
+
+    :return: list of unit model content type IDs
+    :rtype: list of str
+    """
+    assert _is_initialized()
+    return _MANAGER.unit_models.keys()
 
 
 def list_group_distributors():
@@ -281,6 +296,20 @@ def is_valid_cataloger(cataloger_id):
 
 
 # plugin api -------------------------------------------------------------------
+
+def get_unit_model_by_id(model_id):
+    """
+    Get the ContentUnit model class that corresponds to the given id.
+
+    :param model_id: id of the model
+    :type model_id: str
+
+    :return: the Model class or None
+    :rtype: pulp.server.db.model.ContentUnit
+    """
+    assert _is_initialized()
+    return _MANAGER.unit_models.get(model_id)
+
 
 def get_distributor_by_id(distributor_id):
     """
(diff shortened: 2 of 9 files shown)


bmbouter:
I'm not 100% sure redmine will recognize `refs: #853`. I think the commit message should adhere to these:  https://pulp.plan.io/issues/103

common/pulp/common/error_codes.py
```diff
@@ -89,6 +89,12 @@
 PLP0036 = Error("PLP0036", _("The source_location: %(source_location) specified for the content unit is invalid."),
                 ['source_location'])
 PLP0037 = Error("PLP0037", _("The relative_path specified for the content unit is empty."), [])
+PLP0038 = Error("PLP0038", _("The unit model with the id %(model_id)s and the class "
```
bmbouter: s/the id/id/
s/the class/class/

common/pulp/common/error_codes.py
```diff
@@ -89,6 +89,12 @@
...
+                             "%(model_class)s failed to register as another model has already "
```
bmbouter: s/ as another/\. Another/

common/pulp/common/error_codes.py
```diff
@@ -89,6 +89,12 @@
...
+                             "been registered with the same id."), ['model_id', 'model_class'])
+PLP0039 = Error("PLP0039", _("The unit model with the id %(model_id)s failed to register as the "
```
bmbouter: s/as the/\. The/

server/pulp/plugins/loader/api.py
```diff
@@ -81,7 +82,20 @@ def list_content_types():
...
-    return database.all_type_ids()
+    types_list = _MANAGER.unit_models.keys()
+    legacy_types = database.all_type_ids()
+    types_list.extend(legacy_types)
+    return types_list
+
+
+def list_unit_models():
+    """
+    List the supported unit_models.
```
bmbouter: List makes it seem like this will print. I'll suggest:  s/List/Get/  The return args explain the typing details clearly.

bmbouter: Consider adding a newline under this to separate the description from the params in the docstring.

server/pulp/plugins/loader/api.py
```diff
@@ -282,6 +296,19 @@ def is_valid_cataloger(cataloger_id):
 
 # plugin api -------------------------------------------------------------------
 
+def get_unit_model_by_id(model_id):
+    """
+    Get the ContentUnit model class that corresponds to the given id.
+
+    :param model_id: id of the model
+    :type model_id: str
```
bmbouter: Consider a newline underneath this separating the input parameters from output parameters.

server/pulp/plugins/loader/manager.py
```diff
@@ -22,6 +30,43 @@ def __init__(self):
...
         self.profilers = _PluginMap()
         self.catalogers = _PluginMap()
+        self.unit_models = dict()
+
+        # Load the unit models
+        self._load_unit_models()
+
+    def _load_unit_models(self):
+        """"
+        Load all of the Unit Models from the ENTRY_POINT_UNIT_MODELS entry point
```
bmbouter: Consider renaming ENTRY_POINT_UNIT_MODELS to PULP_UNIT_MODELS or just UNIT_MODELS as the entry point. It is an entry point so having that in the name makes it longer without adding anything imho.

server/pulp/plugins/loader/manager.py
```diff
@@ -22,6 +30,43 @@ def __init__(self):
...
+                                         model_id=model_id,
+                                         model_class=class_name)
+
+            if model_id in self.unit_models:
+                raise PulpCodedException(error_code=error_codes.PLP0038,
+                                         model_id=model_id,
+                                         model_class=class_name)
+            self.unit_models[model_id] = model_class
+            # Attach all the signals
+            signals.post_init.connect(model_class.post_init_signal, sender=model_class)
```
bmbouter: Can these be moved to their own function. The function name and docstring make doing this operation inside of here not seem like it belongs. Having a staticmethod like `def connect_signals_to_unit_model` would be good. Also it would be more clear in the future where signals get wired.

server/test/unit/plugins/loader/test_api.py
```diff
@@ -252,3 +252,56 @@ def test_catalogers(self):
     def test_finalize(self):
         api.finalize()
         self.assertEqual(api._MANAGER, None)
+
+
+class TestAPIModels(unittest.TestCase):
+
+    @mock.patch('pulp.plugins.loader.api._is_initialized')
+    @mock.patch('pulp.plugins.loader.api._MANAGER')
+    def test_get_unit_model(self, mock_manager, mock_is_initialized):
```
bmbouter: Consider renaming to match literal function name being tested:
`def test_get_unit_model` = `def test_get_unit_model_by_id`

server/test/unit/plugins/loader/test_manager.py
```diff
@@ -0,0 +1,86 @@
...
+        mock_entry_points_iter.return_value = [entry]
+
+        plugin_manager = manager.PluginManager()
+
+        self.assertTrue('foo' in plugin_manager.unit_models)
+        test_model = ContentUnitHelper
+        found_model = plugin_manager.unit_models.get('foo')
+
+        self.assertTrue(test_model == found_model)
+        mock_signals.post_init.connect.assert_called_once_with(ContentUnitHelper.post_init_signal,
```
bmbouter: These would be moved to their own tests when the signal wiring moves to its own class.

bmbouter:
@barnabycourt The mocking looks especially good on this. I left some comments, but I don't need to re-review. Putting the LGTM on it and assigning back to you.

PR #1834, opened 2015-04-30
Ensure index fails on mongoengine 0.7.10

barnabycourt:
https://pulp.plan.io/issues/916

### DIFF
--- a/server/pulp/plugins/loader/manager.py
+++ b/server/pulp/plugins/loader/manager.py
@@ -4,8 +4,6 @@ import copy
 import logging
 import pkg_resources
 
-from mongoengine import signals
-
 from pulp.common import error_codes
 from pulp.plugins.loader import exceptions as loader_exceptions
 from pulp.server.db.model import ContentUnit
@@ -65,8 +63,6 @@ class PluginManager(object):
 
             # Attach all the signals
             model_class.attach_signals()
-            signals.post_init.connect(model_class.post_init_signal, sender=model_class)
-            signals.pre_save.connect(model_class.pre_save_signal, sender=model_class)
 
         _logger.debug(_("Unit Model Loading Completed"))
 
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -115,16 +115,34 @@ def ensure_database_indexes():
     MongoEngine based models require a manual check as they will only create indexes if the
     collection does not already exist.
     """
-    model.RepositoryContentUnit.ensure_indexes()
-    model.Repository.ensure_indexes()
-    model.ReservedResource.ensure_indexes()
-    model.TaskStatus.ensure_indexes()
-    model.Worker.ensure_indexes()
+
+    _ensure_indexes(model.RepositoryContentUnit)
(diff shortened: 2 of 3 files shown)


barnabycourt:
ok test

server/pulp/server/db/manage.py
```diff
@@ -115,16 +115,34 @@ def ensure_database_indexes():
...
+
+    :param model_class: document model
+    :type model_class: mongoenigne.Document
+    """
+    if hasattr(model_class, 'ensure_indexes'):
         model_class.ensure_indexes()
+    else:
+        # Using a private method as there is no public method to access
+        # the index list in mongoengine 0.7.10, as soon as we can update
+        # the mongoengine dependency on all platforms this should be removed.
```
beav: I have some similar code at https://github.com/pulp/pulp/blob/master/server/pulp/server/db/model/base.py#L179-201. Is there a good way to track the items we can get rid of after we are off of 0.7.10?

server/pulp/server/db/manage.py
```diff
@@ -115,16 +115,34 @@ def ensure_database_indexes():
...
+    _ensure_index(model.TaskStatus)
+    _ensure_index(model.Worker)
 
     # Load all the model classes that the server knows about and ensure their inexes as well
     plugin_manager = PluginManager()
     for model_class in plugin_manager.unit_models.itervalues():
+        _ensure_index(model_class)
+
+
+def _ensure_index(model_class):
```
beav: Later versions of mongoengine have both `ensure_index` and `ensure_indexes` methods on Documents. It may be better to rename this to `_ensure_indexes` just to make it consistent with the other methods it calls.

server/pulp/server/db/manage.py
barnabycourt: I added a tracking story at https://pulp.plan.io/issues/934

server/pulp/server/db/manage.py
barnabycourt: good call

server/pulp/plugins/loader/manager.py
```diff
@@ -65,8 +63,6 @@ def _load_unit_models(self):
 
             # Attach all the signals
             model_class.attach_signals()
-            signals.post_init.connect(model_class.post_init_signal, sender=model_class)
```
barnabycourt: I cleaned these out since the attach_signals() method above performs these actions

server/pulp/server/db/manage.py
beav: cool thanks :koala:

server/pulp/plugins/loader/manager.py
beav: Does `TaskStatus` need to implement `attach_signals()`?  Here is how it currently does it:

https://github.com/pulp/pulp/blob/master/server/pulp/server/db/model/__init__.py#L314

beav: oh n/m, this is just for unit models

beav:
LGTM

PR #1835, opened 2015-04-30
ref #214, support distributor search.

jortel:
Contained in the PR:
- Change storage of *last_publish* on distributors from iso8601 string to native date object in mongo.  This is needed to support date range searches.  The API will continue to return this field as an ios8601 string so backwards compatibility is maintained.
- The distributor  *last_publish* only updated on successful publish.
- Added /v2/distributors/seach/ endpoint to support search.

### DIFF
--- a/server/test/unit/plugins/conduits/test_repo_publish.py
+++ b/server/test/unit/plugins/conduits/test_repo_publish.py
@@ -57,15 +57,15 @@ class RepoPublishConduitTests(base.PulpServerTests):
         self.assertTrue(unpublished is None)
 
         # Setup - Previous publish
-        last_publish = datetime.datetime.now()
+        last_publish = datetime.datetime(2015, 4, 29, 20, 23, 56, 0)
         repo_dist = RepoDistributor.get_collection().find_one({'repo_id': 'repo-1'})
-        repo_dist['last_publish'] = dateutils.format_iso8601_datetime(last_publish)
+        repo_dist['last_publish'] = last_publish
         RepoDistributor.get_collection().save(repo_dist, safe=True)
 
         # Test - Last publish
         found = self.conduit.last_publish()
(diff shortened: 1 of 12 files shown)


beav:
This works OK in that it returns a nice error telling me what I'm missing:

```
curl -k https://admin:admin@localhost/pulp/api/v2/distributors/search/ -d '{}'
```

However this returns a 500 (same thing with no POST data):
```
 curl -k https://admin:admin@localhost/pulp/api/v2/distributors/search/
```

server/test/unit/server/db/migrations/test_0017_distributor_list_published.py
```diff
@@ -0,0 +1,51 @@
...
+MIGRATION = 'pulp.server.db.migrations.0017_distributor_last_published'
+
+
+class TestMigration(TestCase):
+    """
+    Test the migration.
+    """
+
+    @patch('.'.join((MIGRATION, 'parse_iso8601_datetime')))
+    @patch('.'.join((MIGRATION, 'RepoDistributor')))
```
beav: that is a cool way to reduce the line length :panda_face:

server/test/unit/plugins/conduits/test_repo_publish.py
```diff
@@ -57,15 +57,15 @@ def test_last_publish(self):
         self.assertTrue(unpublished is None)
 
         # Setup - Previous publish
-        last_publish = datetime.datetime.now()
+        last_publish = datetime.datetime(2015, 4, 29, 20, 23, 56, 0)
```
beav: :clock830:

beav:
LGTM overall. The `500` when calling the search with GET instead of POST may be OK, I hit it when testing because I forgot to add `-X POST` to the call.

PR #1836, opened 2015-05-01
Updates the docs that Pulp Server requires CentOS or RHEL Server.

bmbouter:


bmbouter:
Closing so I can open against 2.6-testing.

PR #1841, opened 2015-05-04
Fix error code PLP0008

barnabycourt:
https://pulp.plan.io/issues/947


jortel:
LGTM

jortel:
Shouldn't this be causing a unit test to fail somewhere?

PR #1844, opened 2015-05-05
Ensure that tests share a single DB conenction.

rbarlow:
Many of our tests independently connected to the database, and they did
not always ensure that they were connecting to the test database. This
commit fixes the tests so they all use the test database.

Additionally, this commit refactors all of the entry points such that
the connection is no longer started as a side effect of importing a
module. Most of our code already worked this way, but this commit
refactors the Celery workers to start the database connection using a
signal instead of through an import.

This commit also makes it an error for the connection to be initialized
more than once.

https://pulp.plan.io/issues/131
https://pulp.plan.io/issues/460

fixes #131
fixes #460

This pull request must be merged at the same time as https://github.com/pulp/pulp_rpm/pull/682. If this is merged first, pulp_rpm master will fail. If the RPM PR is merged first, it will fail until this is merged.


rbarlow:
ok test

rbarlow:
ok test

rbarlow:
ok test

PR #1848, opened 2015-05-06
Update .repo files for 2.7

beav:
This commit updates the .repo files that are on fedorahosted and removes some
older repos that are no longer used.

fixes #955

### DIFF
--- a/rel-eng/fedora-pulp.repo
+++ b/rel-eng/fedora-pulp.repo
@@ -16,38 +16,6 @@ enabled=0
 skip_if_unavailable=1
 gpgcheck=0
 
-# Version 2.4 Beta Builds
-[pulp-2.4-beta]
-name=Pulp 2.4 Beta Builds
-baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/beta/2.4/fedora-$releasever/$basearch/
-enabled=0
-skip_if_unavailable=1
-gpgcheck=0
-
-# Version 2.4 Testing Builds
-[pulp-2.4-testing]
-name=Pulp 2.4 Testing Builds
-baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/2.4/fedora-$releasever/$basearch/
-enabled=0
-skip_if_unavailable=1
-gpgcheck=0
-
-# Version 2.5 Beta Builds
-[pulp-2.5-beta]
-name=Pulp 2.5 Beta Builds
-baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/beta/2.5/fedora-$releasever/$basearch/
-enabled=0
-skip_if_unavailable=1
-gpgcheck=0
-
-# Version 2.5 Testing Builds
-[pulp-2.5-testing]
-name=Pulp 2.5 Testing Builds
-baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/2.5/fedora-$releasever/$basearch/
-enabled=0
-skip_if_unavailable=1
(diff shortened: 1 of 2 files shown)


beav:
this worked ok for me on el6

rel-eng/fedora-pulp.repo
```diff
@@ -63,3 +31,11 @@ baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/2.6/fedora-$relea
...
 skip_if_unavailable=1
 gpgcheck=0
+
+# Version 2.7 Testing Builds
+[pulp-2.7-testing]
+name=Pulp 2.7 Testing Builds
+baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/2.7/dev/fedora-$releasever/$basearch/
+enabled=0
+skip_if_unavailable=1
+gpgcheck=0
```
bmbouter: Should we `gpgcheck=1`? I think so.

rel-eng/rhel-pulp.repo
```diff
@@ -63,3 +32,11 @@ baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/2.6/$releasever/$
...
 skip_if_unavailable=1
 gpgcheck=0
+
+# Version 2.7 Testing Builds
+[pulp-2.7-testing]
+name=Pulp 2.7 Testing Builds
+baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/2.7/dev/$releasever/$basearch/
+enabled=0
+skip_if_unavailable=1
+gpgcheck=0
```
bmbouter: Same here

bmbouter:
@beav When will this be pushed to fedorapeople?

bmbouter:
It would be good if the commit message contained the text `https://pulp.plan.io/issues/955`. I usually put it above or below the `fixes #955`

bmbouter:
@beav I'm done reviewing for now so I'm going to assign back to you. I'm putting the LGTM flag on it.

rel-eng/fedora-pulp.repo
beav: (adding @barnabycourt to cc)

This url points to an automation repo which isn't signed. We'd need to start publishing and signing alphas if we wanted to do this.

Would calling the repo something like `pulp-2.7-nightly` be more accurate? I just need the repo definition available somewhere so I can install w/ Vagrant from rpm. A nightly would work better for my purposes than a hand-tagged alpha since it would be fresher.

beav:
@bmbouter PR has been updated for new message. I was planning on updating it on fedorapeople immediately after its merged.

Would it make more sense for this to live in pulp_packaging?

bmbouter:
@beav I do think it would be better in pulp_packaging, but we should ask @barnabycourt

rel-eng/fedora-pulp.repo
bmbouter: @beav Oh I didn't realize they weren't signed. If we want to sign them, OK, but I'm also ok publishing the .repo file with `gpgcheck=0`. @barnabycourt what do you recommend?

barnabycourt:
I'm not sure that linking directly to the automation builds is a good idea. If we sign & release a beta that's one thing but automation builds are not signed. In addition we should not make any guarantee's as to the structure of the automation builds. If there is a need to change how those directories should look, nothing external should be depending on them.

beav:
It sounds like we should not point to the automation dir in this file.

withdrawing PR per github and IRC convo

PR #1849, opened 2015-05-06
Set server_name to FQDN for default server config

dkliban:
The server_name and the CN in certificates generated by pulp-gen-ca-certificate is
the fully qualified domain name (FQDN).

https://pulp.plan.io/issues/926

fixes #926

### DIFF
--- a/docs/user-guide/release-notes/2.7.x.rst
+++ b/docs/user-guide/release-notes/2.7.x.rst
@@ -60,6 +60,15 @@ Deprecation
 Upgrade Instructions for 2.6.x --> 2.7.0
 -----------------------------------------
 
+.. warning::
+
+   This release adjusts the default 'server_name' in `/etc/pulp/server.conf` to be the fully
+   qualified domain name (FQDN), but previous releases on some platforms may have used the short
+   hostname. Similarly, this release adjusts the pulp-gen-ca-certificate tool to use the FQDN as
+   the `CN` in a certificate it generates. As a result of this change, you may experience a
+   hostname mismatch when you run Pulp. If you experience that issue, you can either set the value
+   of 'server_name' in /etc/pulp/server.conf or regenerate your certificate using pulp-gen-ca-certificate.
+
 All services should be stopped. At that point you can issue an upgrade via:
 
 ::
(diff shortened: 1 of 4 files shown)


bmbouter:
This needs a release note added to it similar to comments 7 and 8 in issue 926.

bmbouter:
Also update server.conf comments so its clear that the FQDN is used.

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

server/etc/pulp/server.conf
```diff
@@ -52,7 +52,8 @@
 # Controls general Pulp web server behavior.
 #
 # server_name:      hostname the admin client and consumers should use when accessing
-#                   the server; if not specified, this is defaulted to the server's hostname
+#                   the server; if not specified, this is defaulted to the server's fully qualified
```
bmbouter: s/is defaulted/defaults/

docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -60,6 +60,13 @@ Deprecation
 Upgrade Instructions for 2.6.x --> 2.7.0
 -----------------------------------------
 
+.. warning::
+
+   The default 'server_name' is the fully qualified domain name (FQDN). The CN in the certificate
```
bmbouter: I think this should be rewritten to be clearer about what is changing with this release. I'm suggesting the following rewrite:

```
This release adjusts the default 'server_name' in `/etc/pulp/server.conf` to be the fully qualified domain name (FQDN), but previous releases on some platforms may have used the short hostname. Similarly, this release adjusts the pulp-gen-ca-certificate tool to use the FQDN as the `CN` in a certificate it generates.

As a result of this change, you may experience a hostname mismatch when you run Pulp. If you experience that issue, you can either set the value of 'server_name' in /etc/pulp/server.conf or regenerate your certificate using pulp-gen-ca-certificate.
```

bmbouter:
@dkliban back to you

dkliban:
ok test

dkliban:
ok test

PR #1851, opened 2015-05-08
Remove web.py reference from setup.py

beav:
web.py is not available on "fresh" 2.7 installs from RPM, which causes
pulp-manage-db to not run.

fixes #959

https://pulp.plan.io/issues/959


ipanova:
LGTM

PR #1854, opened 2015-05-11
Fixes a file descriptor leak in python-kombu

bmbouter:
I did a tito test build of the RPM and the patch applied cleanly. I did not bump the release because release 7 has not yet been built on koji.

This is a fix for upstream issue 476.
https://github.com/celery/kombu/issues/476

fixes #470
https://pulp.plan.io/issues/470


jortel:
LGTM

PR #1855, opened 2015-05-12
742 - Reload systemctl daemon after install

dkliban:
fixes #742

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -426,6 +426,13 @@ then
   pulp-gen-ca-certificate
 fi
 
+%if %{pulp_systemd} == 1
+  if [ $1 -eq 2 ]; # an upgrade
+  then
+    /sbin/systemctl daemon-reload > /dev/null 2>&1
+  fi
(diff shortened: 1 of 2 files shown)


pulp.spec
```diff
@@ -426,6 +426,13 @@ then
   pulp-gen-ca-certificate
 fi
 
+if [ $1 -eq 2 ]; # an upgrade
+then
+  %if %{pulp_systemd} == 1
+    /sbin/systemctl daemon-reload > /dev/null 2>&1
+  %endif
+fi
+
```
beav: can you confirm this works ok on el6?

```bash
if [ 1 -eq 1 ];
then
  echo "HELLO"
fi
```

the above works, but if I comment out the echo statement, it fails. You may need to wrap the if statement in the %if macro.

beav:
LGTM but the reload should be using a macro. Can you put in an issue for this?

new spec passes rpmlint with no additional warnings/errors

PR #1856, opened 2015-05-12
Add  criteria search operator.

jortel:
https://pulp.plan.io/issues/214

To completely support the use case in #214, pulp *criteria* needs to support filters on fields stored in the DB as mongo *ISODate*.  Since filters are passed to pymongo, the value of filters on fields stored as ISODate need to be parsed and converted into python *datetime* before being passed to mongo.

This approach attempts to follow the *operator* pattern already established by mongo for expressing queries.  It introduces an operator extension named ``$date`` which coverts the iso-8601 string value into a python *datetime* object before it is handed to pymongo.

### DIFF
--- a/docs/dev-guide/conventions/criteria.rst
+++ b/docs/dev-guide/conventions/criteria.rst
@@ -24,6 +24,15 @@ find specification syntax, the resource fields and values to match. For more
 information on the syntax, see:
 http://www.mongodb.org/display/DOCS/Querying
 
+Pulp supports an operator extension for matching date fields. The ``$date``
+operator may be used to specify ISO-8601 date strings for fields that are
+stored as a *Date* or *ISODate* (instead of string) in the database.
+
+For example::
+
+ {"created": {"$gt": {"$date": "2015-01-01T00:00:00Z"}}}
+
+
 The **sort** field is an array of arrays. Each specifying a field and a
 direction.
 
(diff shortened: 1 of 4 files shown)


docs/dev-guide/conventions/criteria.rst
```diff
@@ -24,6 +24,15 @@ find specification syntax, the resource fields and values to match. For more
 information on the syntax, see:
 http://www.mongodb.org/display/DOCS/Querying
 
+Pulp supports a operator extension for matching date fields. The ``$date``
```
ipanova: s/Pulp supports a operator extension/ Pulp supports an operator extension

ipanova:
@jortel  shall we somehow handle this error, instead of receiving internal server error? http://fpaste.org/221476/43151971/ how do you think?

jortel:
I considered this but decided to let this fail with a 500 to be consistent with how other invalid queries fail.  For example, if a query contains an invalid operator, it results in a 500.  My intention is for this 500 to be handled as a 400 (bad request) along with all of the other cases.  I opened a separate issue for this: https://pulp.plan.io/issues/981

ipanova:
yeah, that makes sense, LGTM

jortel:
ok test

ipanova:
ok test

jortel:
ok test

ipanova:
ok test

barnabycourt:
ok test

PR #1858, opened 2015-05-14
better support for installing standalone repoauth

beav:
Previously, the pulp repoauth RPM was not installable without also installing
the rest of pulp-server. This was not ideal for instances that served protected
content to client systems, but did not run full Pulp installations.

This patch does the following:

 * moves an openssl lib from `pulp.server.common` to `pulp.repoauth`.  Note that we couldn't move it to `pulp.common` since our openssl code has some calls to `subprocess` that do not work on Python 2.4.

 * adds a few deps to python-pulp-repoauth that previously were provided by
   pulp-server's deps

 * fixes an unrelated issue where a "real" conf file was used in a few unit
   tests

I confirmed that no plugins are using `pulp.server.common.*`.

Fixes #975

https://pulp.plan.io/issues/975


barnabycourt:
Hi, this still doesn't address the fact that the repo-auth depends on the default values being in the config file in etc during runtime. A lot of work was done to get away from that pattern for the general server conf (most, if not all the values are commented out by default). That pattern should be followed here as well. 

Also, how do you ensure that this is installed every time pulp-rpm is installed?

barnabycourt:
One other thought, @rbarlow just added code that will fail if the config is modified in tests instead of patching, will this run into problems with that?

beav:
The config test changes from @rbarlow appear to only be in master, but this patch is for 2.7-dev. I'm not sure right now if it's better to backport the test changes or alter the tests in this PR when they are forward-merged to master. I don't believe it will interfere with Randy's changes based on a prior convo but I can merge this PR to master locally and try it out.

re: pulp_rpm, the pulp-rpm spec file already has a `Requires:` for python-pulp-repoauth as of Pulp 2.7.

rbarlow:
On 05/14/2015 06:09 PM, Chris Duryee wrote:
> I don't believe it will interfere with Randy's changes based on a prior
> convo but I can merge this PR to master locally and try it out.

The change I made only forbids changing pulp.server.config.config during
test runs that use pulp.devel.unit.server.base to start the database
connection. I suspect that the repoauth changes do not use
pulp.server.config.config. Is that true? If so, it will not be affected!

-- 
Randy Barlow

run-tests.py
```diff
@@ -38,7 +38,6 @@
     'server/pulp/server/agent/',
     'server/pulp/server/async/',
     'server/pulp/server/auth/',
-    'server/pulp/server/common/',
```
barnabycourt: Why was pulp/server/common/ removed from the pep8 checks?

barnabycourt:
This does indeed fix the dependence on the /etc/pulp/repoauth.conf file for running the unit tests.

run-tests.py
beav: good catch :fishing_pole_and_fish: 

I meant to comment on that in the PR and forgot. `server/pulp/server/common` only had one file, `openssl.py`. It got moved to `repoauth/pulp/repoauth` which is already covered, so the dir was removed.

PR #1861, opened 2015-05-15
Updates to schedules are now respected by celerybeat

bmbouter:
The schedule field was being updated, but next_run and
first_run were not being updated. These values are used
to determine if a schedued task should be run so updating
them is necessary.

fixes #922
https://pulp.plan.io/issues/922


barnabycourt:
ok test

beav:
LGTM

PR #1864, opened 2015-05-19
Adds a confirmation from DB on saving TaskStatus

dkliban:
fixes #974

### DIFF
--- a/server/pulp/server/db/model/dispatch.py
+++ b/server/pulp/server/db/model/dispatch.py
@@ -22,6 +22,8 @@ from celery import beat
 from celery.schedules import schedule as CelerySchedule
 from celery.utils.timeutils import timedelta_seconds
 import isodate
+import mongoengine
+
 
 from pulp.common import dateutils
 from pulp.server.async.celery_instance import celery as app
@@ -614,7 +616,8 @@ class TaskStatus(Model, ReaperMixin):
         update = {'$set': stuff_to_update}
         if set_on_insert:
             update['$setOnInsert'] = set_on_insert
-        self.get_collection().update(
-            {'task_id': task_id},
-            update,
-            upsert=True)
+
+        wc_param = 'write_option' if mongoengine.get_version()  < "0.9.0" else 'write_concern'
+        update_kwargs = {wc_param: {'w':1, 'fsync':True}}
+
+        self.get_collection().update({'task_id': task_id}, update, upsert=True, **update_kwargs)


server/pulp/server/async/tasks.py
```diff
@@ -255,7 +255,7 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         # To avoid the race condition where __call__ method below is called before
         # this change is propagated to all db nodes, using an 'upsert' here and setting
         # the task state to 'waiting' only on an insert.
-        task_status.save(fields_to_set_on_insert=['state', 'start_time'])
+        task_status.save(fields_to_set_on_insert=['state', 'start_time'], safe=True)
```
bmbouter: as @mhrivnak alluded to, safe=True will be completely ignored in newer versions of mongoengine. :disappointed: I see safe=True as an option on the 0.7.10 docs I've built locally, but in the newer docs I think you'll have to use write_concern. See the [docs on save()](http://docs.mongoengine.org/en/latest/apireference.html#mongoengine.Document.save)

bmbouter: Pulp isn't really ready to work with a clustered MongoDB. It used to be with seeds, but then MongoDB approach to replication evolved and moved away from that. My understanding of these things is weak, so maybe check with @rbarlow who may have a better understanding of this stuff.

Focusing on just the goal of this PR, I'll suggest that instead of using `safe=True` you should pass the option `write_concern={fsync: True}` or `write_concern={w:1, fsync: True}`. The fsync option will perform a synchronous flush of all mongoDB writes which should guarantee that the write is committed to disk before returning. See the [fsync docs](http://docs.mongodb.org/v2.4/reference/command/fsync/) for more details. Note, in the 0.7.10 version it was not called `write_concern`, it was called `write_options`. You'll need this code to work correctly with all the versions that could be found on our platforms. See [this table](https://apps.fedoraproject.org/packages/python-mongoengine) for the list of versions on platforms.

server/pulp/server/db/model/dispatch.py
```diff
@@ -588,7 +588,7 @@ def __init__(
         self.exception = None
         self.traceback = None
 
-    def save(self, fields_to_set_on_insert=None):
+    def save(self, fields_to_set_on_insert=None, safe=False):
```
bmbouter: This additional kwarg needs to be documented in the docstring.

bmbouter:
The code changes here also need test updates. Even though this is a change that won't be in Pulp long in needs test coverage.

bmbouter:
Also pulpbot needs to be happy.

bmbouter:
Please also put a link to the pulp.plan.io issue explicitly in the commit message and PR description. It should be:  `https://pulp.plan.io/issues/974`

server/pulp/server/async/tasks.py
rbarlow: We do actually almost work with replica sets, except for the Celery results backend bug. I suspect we'll have that bug fixed soon since it's high priority.

I agree that we should stop using the safe kwarg since it is deprecated and start using write concern.

Unfortunately, write concern set to 1 is not safe either so I recommend against the w:1 setting. This will only ensure that the write makes it to a single node and not to the other nodes. If there is a conflict, there is no guarantee that your write will "win", but your code will believe it did. With MongoDB 3.0, there is a w:majority setting, but we unfortunately cannot do that.

What we need is some code that calculates the majority for us that all of our Mongo queries can use. Something similar to floor(len(db_seeds)/2)+1. If we put something like that in pulp.server.db somewhere, then everyone can benefit. It could be calculated as a constant, or a function. Then we can set w to that value and be sure that we are getting our writes to a majority of nodes.

bmbouter: @rbarlow I think Pulp has more than just the Celery results backend bug to work properly with mongoDB replicate sets. We have the auto-retry issue and that failover has been untested. You could say it's our [number 1 issue](https://pulp.plan.io/issues/1).

+1 to not using `safe` anywhere in the codebase.

I agree we likely will need to compute write_concern, but do we have to do it now for this specific bugfix? To avoid getting into all that now, couldn't we set write_concern as `write_concern={fsync: True}` and leave off the `w` option? That would ensure the write occurred to the mongodb Pulp is connected to and leaves the rest to be resolved as [issue 1](https://pulp.plan.io/issues/1).

rbarlow: @bmbouter I believe we had been using the deprecated safe=True before. Though deprecated, I believe it was safer than w=1 would be, and guaranteed that the write hit more than one node. In other words, I think not calculating the write concern here and doing just the fsync will be a regression from what we had before.

I agree that we should perform the work described in issue number 1, but IMO we should always write our code with Mongo replica sets in mind regardless of whether we have or have not tested that feature yet. Since we do use safe now, we must calculate the appropriate write concern as we replace it or we will lose ground on our HA story.

bmbouter: @rbarlow In the case of this specific PR, it introduces `safe=True` so it was not being used before here.

Also the `safe` parameter is not deprecated, it's literally missing from the most current version of mongoengine. That means in other areas of code when we set that option, we would expect it to be committed synchronously but the option will be ignored entirely and won't provide the expected behavior. I'm suggesting `write_concern={fsync: True}` as an alternative because it seems to provide that expected behavior which is "don't return until the write has been committed".

I propose we introduce the setting of the `w` option throughout the entire codebase as part of issue 1, or maybe a separate issue itself. It's such a broad problem that I don't think there is much value in setting `w` for this case when its still such a problem elsewhere.

rbarlow: When I and @skarmark originally wrote all the task system DB code, we used safe=True on all of our pymongo queries. That is what I meant when I said we originally used safe=True. I think it's imperative that we continue doing safe queries as we transition to Mongoengine, as we had such safe queries when we used pymongo directly.

The reason that ```write_concern={fsync: True}``` is not equivalent to the old safe=True is that users can configure default write_concern on their replica sets. I confirmed that this was possible at least as early as mongo 2.2:

http://docs.mongodb.org/v2.2/applications/replication/#modify-default-write-concern

So if a user sets the default write concern to the majority of their nodes, this code will override that default and we will not have a safe write. The safe=True that we used to use respected the default write concern of the replica set.

rbarlow: One more note - it is possible to set the write_concern in one place, so that we do not have to configure it on every single query. For example, we could do it on the MongoClient when we make the connection:

http://api.mongodb.org/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient

I believe you can also set the write_concern when you create the Collection instance.

rbarlow: Mongoengine may have similar options, though I haven't looked into it.

bmbouter: @rbarlow Since everything is converting to mongoengine, anything we do needs to work with mongoengine. How do we set write_concern in one place with mongoengine?

I figured `write_concern={fsync: True}` would not override the configured write concern as set by the user. I figured telling mongoDB to perform a `fsync` and specifying the value for `w` would be independent options. It is possible it happens the way you describe, but I don't think either of us know for sure. We both want it safe though! Maybe @dkliban can come up with more info.

rbarlow: It appears that mongoengine just passes any kwargs you give it on to the pymongo connection:

http://docs.mongoengine.org/apireference.html#connecting

If that is true, we can employ a proper write_concern in one place that will be just as safe as what we had before. The reason I believe that passing your own write concern will override the replica set one I linked to above is that they call that setting the *default* write concern. The language means to me that it is the write concern used when no other write concern is specified.

There may be a way to test to find out for sure - if we were to make a replica set with 3 nodes and set the default write concern to 3, we could kill -9 one of them and see if the above call returns or not.

Another point I forgot to make earlier is that Pulp used to work well with replica sets. Issue 1 isn't going to be the first time someone ever tries this - it used to work and was tested by @jdob. We regressed later on when we started using MongoDB for Celery's result backend, which was unfortunate. It was also unfortunate that the issue was not reported for so long and that there was even a comment in the result backend code to that effect.

bmbouter: @rbarlow  I agree with the way you've read the language, but I just want to be sure. +1 to testing it out for ourselves!

I don't want to rely on the assumption that Pulp was doing it right before. Pulp probably did work well with replica sets, but I've got some doubts. For example, for a long time now, Pulp has a blind auto-retry behavior. That is a bad behavior which has been expressly told by both the MongoDB developers AND the PyMongo community not to be done and is not safe. Pulp does it anyway. This is the motivation that caused me to write issue 1 in the first place.

rbarlow: +1 to not assuming ☺

Let's make a simple lab and try it out!

rbarlow: I made a lab to demonstrate what we've been discussing here, and it is conclusive that we should not set the write concern like this on this query:

https://www.redhat.com/archives/pulp-list/2015-May/msg00070.html

bmbouter: Great job setting up the testbed! You've proven that specifying `{'fsync': True}` allows a write request to return before hitting all the nodes, even though the replica set is configured for a safer write concern. IMHO, this is a bug in mongoDB because the replica set is configured for one level of safety, but a client can override it. That doesn't seem right. If client software needs to force a flush of all writes to disk, it's forced to "mirror" the specification of the safety level which an administrator has already configured mongoDB. That is redundant and potentially unsafe if done incorrectly (as you've proven). I remember you mentioning that they brought back the `w:majority` setting in 3.0, which keeps the client out of the business of specifying the replication level, but we still use mongoDB 2.x. Again, great investigation work. 

To work around this bug, the only thing we can do is have Pulp [compute the write concern as you suggest](https://github.com/pulp/pulp/pull/1864/files#r30699492). Let's do that.

rbarlow: On 05/22/2015 09:03 AM, bmbouter wrote:
> IMHO, this is a bug in mongoDB because the replica set is configured for
> one level of safety, but a client can override it. That doesn't seem
> right.

I think it is useful that we can specify a write-concern on any
particular query. There are times that data isn't particularly important
to record so safely, and the write concern makes the query take longer
to return. For example, we record all of the heartbeats from the workers
in the database. I don't think it's critical that those writes fsync or
propagate, because it'll be ok if the next one is the one that makes it.
I think we would rather just hand it to mongo and hope for the best in
that situation. Does that make sense? It's up to the application writer
to know when we can override the default write concern in either
direction (safer or less safe).

In the Mongo work I've done in the past with large databases, I took
great advantage of being able to specify which queries were critically
important, and which ones were OK to miss when eventual consistency is OK.

> I remember
> you mentioning that they brought back the |w:majority| setting in 3.0,
> which keeps the client out of the business of specifying the replication
> level, but we still use mongoDB 2.x.

We should read more about this. It does say that something about this
changed in 3.0, and maybe also in 2.6? The wording confused me, but I
think I might have been wrong that majority was new in 3.0. I think
w:majority might actually exist in some 2.x versions, but maybe it means
different things or something? We could use the lab to test this by
switching out the Vagrant box OS to get differing versions of Mongo.

> Again, great investigation work.

Thanks!

-- 
Randy Barlow

rbarlow: I don't think we should have a safe= parameter here. A TaskStatus should always be updated safely. Eliminating this parameter will also make the if statement later simpler.

server/pulp/server/db/model/dispatch.py
rbarlow: I highly recommend dropping the extra kwarg and just always being safe.

server/pulp/server/db/model/dispatch.py
```diff
@@ -614,7 +616,9 @@ def save(self, fields_to_set_on_insert=None):
...
-        self.get_collection().update(
-            {'task_id': task_id},
-            update,
-            upsert=True)
+        if mongoengine.get_version()  < "0.9.0":
+            self.get_collection().update({'task_id': task_id}, update, upsert=True,
+                                         write_option={'w':1, 'fsync':True})
+        else:
+            self.get_collection().update({'task_id': task_id}, update, upsert=True,
+                                         write_concern={'w':1, 'fsync':True})
```
rbarlow: If we drop safe, the above becomes:

```
if mongoengine.get_version()  < "0.9.0":
    self.get_collection().update({'task_id': task_id}, update, upsert=True, write_option={'w':1, 'fsync':True})
else:
    self.get_collection().update({'task_id': task_id}, update, upsert=True, write_concern={'w':1, 'fsync':True})
```

There is repetition in there to achieve write_option vs. write_concern. What if you did this instead:

```
wc_param = 'write_option' if mongoengine.get_version()  < "0.9.0" else 'write_concern'
update_kwargs = {wc_param: {'w':1, 'fsync':True}}

self.get_collection().update({'task_id': task_id}, update, upsert=True, **update_kwargs)
```

rbarlow: Ugh, I can't remember if you can do thing = x if y else z in Python 2.6, so I guess we should make sure that's allowed, but the rest of the above pattern I think is still a little cleaner to avoid the repetition, so I think we should go with something like it at least.

dkliban:
This is getting fixed in a complete overhaul of how we do replica sets and write concerns.

PR #1866, opened 2015-05-20
Don't use full path to systemctl in pulp-dev.py.

rbarlow:
Not all distributions put the systemctl executable in the same location.
pulp-dev.py had been calling it as /sbin/systemctl, but Fedora 21 keeps
that executable in /usr/bin/systemctl. This commit calls the executable
as just systemctl and allows the PATH variable to be used to locate it.

https://pulp.plan.io/issues/991

fixes #991


dkliban:
LGTM

PR #1869, opened 2015-05-21
Break out OID validation into its own RPM

beav:
Previously, OID validation was bundled in with the repoauth RPM. This causes
problems for users who want to enable cert-based repo authentication, but do
not want this particular authenticator installed.

Instead, break out OID validation into its own RPM. This lets users install it
or leave it out depending on if they want this authenticator.

fixes #975

https://pulp.plan.io/issues/975


beav:
ok test

test failure was due to a known issue in test timings

dkliban:
LGTM

PR #1870, opened 2015-05-21
Removes duplicate calling of systemctl daemon reload

dkliban:
This was already being performed as a macro.
fixes #980
https://pulp.plan.io/issues/980


beav:
LGTM after jenkins returns

PR #1872, opened 2015-05-22
Speed up the vagrant provisioning.

rbarlow:
This commit speeds up vagrant provision in two ways:
* Turn off the mongo journal.
* Use dnf instead of yum.


beav:
I tested the mongo journalling change the other day; LGTM once jenkins returns

PR #1876, opened 2015-05-22
Adding a publish step that creates a PULP_MANIFEST

mhrivnak:
https://pulp.plan.io/issues/304

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -14,7 +14,7 @@ from pulp.common import error_codes
 from pulp.common.plugins import reporting_constants, importer_constants
 from pulp.common.util import encode_unicode
 from pulp.plugins.model import Unit
-from pulp.plugins.util import misc
+from pulp.plugins.util import manifest_writer, misc
 from pulp.plugins.util.nectar_config import importer_config_to_nectar_config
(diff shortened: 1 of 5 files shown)


server/pulp/plugins/util/publish_step.py
```diff
@@ -824,6 +824,34 @@ def process_main(self):
...
+class CreatePulpManifestStep(Step):
+    """
+    This will create a PULP_MANIFEST file in the specified directory. This step should be used when
+    the checksums of the files are not already known, because it will read and calculate new
+    checksums for each one.
+
+    If you already know the SHA256 checksums of the files going in the manifest, see an example
+    in the FileDistributor that creates this file in a different way.
+    """
+    def __init__(self, target_dir):
```
barnabycourt: What if this took an optional argument of a generator of (file_path, sha256_checksum, size) tuples that would be used instead of the make_manifest_for_dir method if provided? That way all places that need a  manifest could use the same code path to generate it.

mhrivnak: I think that would be a great enhancement once we have something that could use it. The only other place we have that does this is in the FileDistributor, where writing the manifest is already built in to the step. It's also just 4 lines of code: open file, create writer, write, close file.

I think only the ISO distributor subclasses the FileDistributor. The use cases and context are sufficiently different that it's not worth refactoring to pull those 4 lines out into this separate step.

PR #1879, opened 2015-05-26
Add prestart and friends to Vagrant provisioning.

rbarlow:
This commit adds some useful bash functions to the Vagrant provisioned
.bashrc file that will start, stop, restart, and display the status of
all the Pulp services. The functions are called pstart, pstop, prestart,
and pstatus, respectively.


rbarlow:
Test failure was due to a test that times how long something takes ☹ We need to get rid of tests like that!

rbarlow:
ok test

PR #1882, opened 2015-05-27
Perform all Worker setup in the same signal.

rbarlow:
We had two different Celery startup signal handlers registered, and it
seemed that the second was no longer being sent. This commit ensures
that all worker startup code happens in the same signal handler, and
that it occurs in the correct order.

https://pulp.plan.io/issues/988

fixed #988

### DIFF
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -3,27 +3,43 @@ This module is the Pulp Celery App. It is passed to the workers from the command
 will see the "celery" module attribute and use it. This module also initializes the Pulp app after
 Celery setup finishes.
 """
-from celery.signals import celeryd_after_setup
+from celery.signals import worker_init
 
 # This import will load our configs
 from pulp.server import config  # noqa
 from pulp.server import initialization
 # We need this import so that the Celery setup_logging signal gets registered
 from pulp.server import logs  # noqa
+from pulp.server.async import tasks
 # This import is here so that Celery will find our application instance
 from pulp.server.async.celery_instance import celery  # noqa
+from pulp.server.managers.repo import _common as common_utils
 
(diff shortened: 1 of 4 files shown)


server/pulp/server/async/app.py
```diff
@@ -3,27 +3,43 @@
...
-    signal[0] so that it gets called by Celery after logging is initialized, but before Celery
-    starts to run tasks.
+    This function performs all the necessary initialization of the Celery worker.
+
+    It starts by cleaning up old state if this worker was previously running, but died unexpectedly.
+    In such cases, any Pulp tasks that were running or waiting on this worker will show incorrect
+    state. Any reserved_resource reservations associated with the previous worker will also be
+    removed along with the worker entry in the database itself. The working directory specified in
+    /etc/pulp/server.conf (/var/cache/pulp/<worker_name>) by default is removed and recreated. This
+    is called early in the worker start process, and later when its fully online pulp_celerybeat
```
mhrivnak: s/its/it's/
s/online/online,/

server/pulp/server/async/app.py
```diff
@@ -3,27 +3,43 @@
...
+    will discover the worker as usual to allow new work to arrive at this worker. If there is no
+    previous work to cleanup this method still runs, but has not effect on the database.
```
mhrivnak: s/cleanup/cleanup,/
s/not/no/

server/pulp/server/async/app.py
rbarlow: Fixed.

server/pulp/server/async/app.py
rbarlow: Fixed.

PR #1884, opened 2015-05-28
Adds batch size to cursor used to iterate repo profile applicabilities

dkliban:
According to mongo documentation [0] the cursor will initially return about 101
documents or slightly more than 1 megabyte of data. The subsequent fetches return
4 times as much data. The default timeout of mongo cursor (version 2.4) is 600
seconds. This timeout cannot be adjusted until mongodb 2.6. If 400 applicability
calculations need to be performed in 600 seconds, each calculation cannot take
any longer than 1.5 seconds. In my testing I found the calculations to take 12 to
13 seconds. Limiting the batch size to 25 ensures that calculations can take up
to 24 seconds each before the cursor times out.

[0] http://docs.mongodb.org/v2.4/core/cursors/#cursor-batches

https://pulp.plan.io/issues/998
fixes #998

### DIFF
--- a/server/pulp/server/managers/consumer/applicability.py
+++ b/server/pulp/server/managers/consumer/applicability.py
@@ -132,9 +132,11 @@ class ApplicabilityRegenerationManager(object):
         repo_ids = [r['id'] for r in repo_query_manager.find_by_criteria(repo_criteria)]
 
         for repo_id in repo_ids:
-            # Find all existing applicabilities for given repo_id
+            # Find all existing applicabilities for given repo_id. Setting batch size of 25 ensures
+            # the MongoDB cursor does not time out. See https://pulp.plan.io/issues/998#note-6 for
+            # more details.
             existing_applicabilities = RepoProfileApplicability.get_collection().find(
-                {'repo_id': repo_id})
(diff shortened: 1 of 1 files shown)


bmbouter:
Can a test be added for this asserting that batch_size is called with 25?

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -132,9 +132,12 @@ def regenerate_applicability_for_repos(repo_criteria):
         repo_ids = [r['id'] for r in repo_query_manager.find_by_criteria(repo_criteria)]
 
         for repo_id in repo_ids:
-            # Find all existing applicabilities for given repo_id
+            # Find all existing applicabilities for given repo_id. If batch size is not specified,
```
bmbouter: This should contain a link to the issue.

bmbouter: Not a required change but consider moving these details to the issue and shortening this to something more direct for the developer. Maybe link directly to your comment with all this good detail. Something like:

`setting batch_size to 25 ensures the MongoDB cursor doesn't timeout. See https:/pulp.plan.... for more details.`

PR #1887, opened 2015-06-01
Don't print out the file contents when using the uploads api

barnabycourt:
fixes #705

https://pulp.plan.io/issues/705

Ugly to not have unittests on this but the log messages are not currently being confirmed in unit tests


jortel:
Changes look good but what about unit tests?  Looks like the unit tests for PulpConnection are pretty sparse but might be good to add .. something?

jortel:
LGTM

PR #1891, opened 2015-06-02
Fixing discrepancy "canceled" vs. "cancelled

ipanova:
I also changed all the docs and messages not to confuse the users. I did not touch any variables' or methods' name.
closes #966
https://pulp.plan.io/issues/966

these PRs should be merged together:
https://github.com/pulp/pulp_rpm/pull/694
https://github.com/pulp/pulp_docker/pull/71

### DIFF
--- a/client_admin/pulp/client/admin/tasks.py
+++ b/client_admin/pulp/client/admin/tasks.py
@@ -202,7 +202,7 @@ class BaseTasksSection(PulpCliSection):
                 finish_time = _('N/A')
                 result = _('N/A')
             elif task.was_cancelled():
-                state = _('Cancelled')
+                state = _('Canceled')
                 result = _('N/A')
 
         return state, start_time, finish_time, result
--- a/common/pulp/common/plugins/progress.py
+++ b/common/pulp/common/plugins/progress.py
@@ -1,15 +1,3 @@
-# -*- coding: utf-8 -*-
-#
-# Copyright © 2013 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
-# License as published by the Free Software Foundation; either version
-# 2 of the License (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
 """
 Contains classes and functions related to tracking the progress of the
 importer and distributor.
--- a/nodes/common/pulp_node/reports.py
+++ b/nodes/common/pulp_node/reports.py
@@ -1,17 +1,3 @@
-# Copyright (c) 2013 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
-# License as published by the Free Software Foundation; either version
-# 2 of the License (GPLv2) or (at your option) any later version.
-# There is NO WARRANTY for this software, express or implied,
-# including the implied warranties of MERCHANTABILITY,
-# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
-# have received a copy of GPLv2 along with this software; if not, see
-# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
-
-# --- summary reporting  -----------------------------------------------------
-
-
 class RepositoryReport(object):
     """
     Repository merge report.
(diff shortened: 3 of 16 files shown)


rbarlow:
I think this fix is still problematic - it turns all the spellings "cancelled" into "canceled", but the reported bug is about how the official task state (i.e., the API) uses "cancelled" and the -s flag uses "canceled". Further, this PR will change our states from "cancelled" to "canceled", which is not a backwards-compatible API change.

I suggest one of two things:

0) Reduce the scope of this change to only change the help text on the reported -s flag to use the "cancelled" spelling, since we cannot change the state to "canceled" without raising Pulp's major version. If you do this, I suggest filing another bug about the overall discrepancy in our code against Pulp 3.0.0 with the backwards incompatible tag, and we can fix the issue once and for all at that time.

1) If you want to resolve all the discrepancies, I suggest going with the "cancelled" spelling, since that is the spelling used by the API in released versions of Pulp. If you go this route, be careful that it is true that we don't use "canceled" in any other part of the API because that would also be backwards incompatible.

ipanova:
@rbarlow The official state state( in API) is 'canceled' http://fpaste.org/228525/35037214/
that's why i decided to remove in all parts of API the 'cancelled' word

rbarlow:
It seems you are correct! I don't know how I got it confused, but I did. I apologize for the confusion! I see that in pulp.common.constants that ```CALL_CANCELED_STATE = 'canceled'``` and this is indeed the state used in the ```pulp.server.async.cancel()``` function.

client_admin/pulp/client/admin/tasks.py
```diff
@@ -202,7 +202,7 @@ def parse_state(task):
                 finish_time = _('N/A')
                 result = _('N/A')
             elif task.was_cancelled():
-                state = _('Cancelled')
+                state = _('Canceled')
```
rbarlow: Ahhhhh, perhaps this is why I thought the API was "cancelled" - the CLI was printing "Cancelled" as the task status and I assumed it was that stupid CLI thing where it capitalizes words from the API when printing them. It turns out that we had code that detected "canceled" and then printed "cancelled"! lol.

common/pulp/common/plugins/progress.py
```diff
@@ -33,7 +21,7 @@ class ProgressReport(object):
     # If an error occurs outside in progress states, this general failed state can be set
     STATE_FAILED = 'failed'
     # When the user has cancelled a sync
-    STATE_CANCELED = 'cancelled'
+    STATE_CANCELED = 'canceled'
```
rbarlow: I think this change might not be safe. The progress reports are part of our API, and I think this might change the progress report API. I recommend leaving this as-is.

common/pulp/common/plugins/reporting_constants.py
```diff
@@ -19,7 +19,7 @@
 STATE_COMPLETE = u'FINISHED'
 STATE_FAILED = u'FAILED'
 STATE_SKIPPED = u'SKIPPED'
-STATE_CANCELLED = u'CANCELLED'
+STATE_CANCELLED = u'CANCELED'
```
rbarlow: I think this change might not be safe - it may change the progress report API.

nodes/common/pulp_node/reports.py
```diff
@@ -27,7 +13,7 @@ class RepositoryReport(object):
 
     # actions
     PENDING = 'pending'
-    CANCELLED = 'cancelled'
+    CANCELLED = 'canceled'
```
rbarlow: This change may also be unsafe, as this sounds like it is used in the Node report API.

nodes/extensions/admin/pulp_node/extensions/admin/rendering.py
```diff
@@ -36,7 +25,7 @@
 
 ACTIONS = {
     RepositoryReport.PENDING: _('Pending'),
-    RepositoryReport.CANCELLED: _('Cancelled'),
+    RepositoryReport.CANCELLED: _('Canceled'),
```
rbarlow: Note that the variable is still spelled "CANCELLED" - it might be best to keep that consistent.

rbarlow:
I think there might still be a few state names that we can't change in here, as they are used in the progress reports which are considered as part of our API.

In my opinion it would be best to reduce the scope of this PR to just changing the CLI code that prints "Cancelled" to print "Canceled" instead so that it is consistent with the API, as per the ticket filed. Then, I think we should file a second ticket that is backwards incompatible to "search and destroy" all discrepancies with Pulp 3.0.0. This way we don't have to worry about making any backwards incompatible changes and it will be an easy sed operation. It also makes fixing this bug very simple in the meantime. What do you think?

ipanova:
ok test

ipanova:
hi @rbarlow, I tend to agree with you. I created a task for this https://pulp.plan.io/issues/1028

PR #1892, opened 2015-06-02
Use the documented celeryd_after_setup signal.

rbarlow:
We had used the undocumented worker_init signal, which turns out to send
unknown arguments to initialize_worker(). This commit switches us to the
well documented celeryd_after_setup signal instead:

http://celery.readthedocs.org/en/latest/userguide/signals.html#celeryd-after-setup


jortel:
LGTM

PR #1898, opened 2015-06-04
sort with _id in every unit association query (prevents mongo from freeze)

midnightercz:
This prevented us from long delays in association units

### DIFF
--- a/server/pulp/server/managers/repo/unit_association_query.py
+++ b/server/pulp/server/managers/repo/unit_association_query.py
@@ -411,13 +411,11 @@ class RepoUnitAssociationQueryManager(object):
         sort = criteria.unit_sort
 
         if sort is None:
-            unit_key = types_db.type_units_unit_key(unit_type_id)
+            # sort by _id by default (better performance)
+            sort = [('_id', SORT_ASCENDING)]
 
-            if unit_key is not None:
-                sort = [(u, SORT_ASCENDING) for u in unit_key]
 
-        if sort is not None:
-            cursor.sort(sort)
+        cursor.sort(sort)
 
         return cursor
 


ipanova:
Seems like that this patch is related to the issue https://pulp.plan.io/issues/970
shall we change the status of the issue?

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -411,10 +411,12 @@ def _associated_units_by_type_cursor(unit_type_id, criteria, associated_unit_ids
...
         if sort is None:
-            unit_key = types_db.type_units_unit_key(unit_type_id)
+            # dgregor - HACK - sort by _id so we use a single index
+            sort = [('_id', SORT_ASCENDING)]
 
-            if unit_key is not None:
-                sort = [(u, SORT_ASCENDING) for u in unit_key]
+            #unit_key = types_db.type_units_unit_key(unit_type_id)
+            #if unit_key is not None:
+            #    sort = [(u, SORT_ASCENDING) for u in unit_key]
```
rbarlow: Can you remove the above lines?

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -411,10 +411,12 @@ def _associated_units_by_type_cursor(unit_type_id, criteria, associated_unit_ids
         sort = criteria.unit_sort
 
         if sort is None:
-            unit_key = types_db.type_units_unit_key(unit_type_id)
+            # dgregor - HACK - sort by _id so we use a single index
```
rbarlow: Can you make the comment just say "sort by _id by default"?

rbarlow:
ok test

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -411,10 +411,12 @@ def _associated_units_by_type_cursor(unit_type_id, criteria, associated_unit_ids
...
+            sort = [('_id', SORT_ASCENDING)]
 
-            if unit_key is not None:
-                sort = [(u, SORT_ASCENDING) for u in unit_key]
+            #unit_key = types_db.type_units_unit_key(unit_type_id)
+            #if unit_key is not None:
+            #    sort = [(u, SORT_ASCENDING) for u in unit_key]
 
         if sort is not None:
             cursor.sort(sort)
```
rbarlow: The above if statement can never be False. I know this wasn't your change, but it does seem like this code block could use a refactor to me.

rbarlow:
Is it important that this method have a sort at all? I'm not 100% familiar with this difficult-to-mentally-parse module, but it seems to me that perhaps the best option is to only sort if the criteria requested a sort. Does that make sense?

I may be missing something so if having a default sort makes sense, I agree that _id is a good choice. However, I suggest reworking the code here to read like this:

```
sort = criteria.unit_sort or [('_id', SORT_ASCENDING)]
cursor.sort(sort)
```

This way we have two lines of code instead of five, a little bit simpler.

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -411,10 +411,12 @@ def _associated_units_by_type_cursor(unit_type_id, criteria, associated_unit_ids
...
 
         if sort is None:
-            unit_key = types_db.type_units_unit_key(unit_type_id)
+            # dgregor - HACK - sort by _id so we use a single index
+            sort = [('_id', SORT_ASCENDING)]
 
-            if unit_key is not None:
-                sort = [(u, SORT_ASCENDING) for u in unit_key]
+            #unit_key = types_db.type_units_unit_key(unit_type_id)
+            #if unit_key is not None:
```
rbarlow: These two lines are causing the Jenkins failure, since they violate PEP-8. We should just remove them, because we have a policy of not having commented-out code.

rbarlow:
Having a default sort may help with pagination. Perhaps that is the reason we should pick something? According to MongoDB's docs, there is not a guarantee about the order of documents in a query when there is no sort:

http://docs.mongodb.org/manual/reference/method/cursor.sort/#result-ordering

If pagination is what we need this for, I'm on board with having _id be the default sort. It's much simpler than using the unit key as was being done before and also faster, while still accomplishing a pageable view of the data.

I'm good with merging this if you can do the following:
0) Remove all commented code
1) Let's get rid of the unnecessary if statements, and go with the simple two-line approach as I suggested in another comment.
2) Let's not use the word "hack" in comments ☺
3) Let me know when to take another look!

rbarlow:
I think @ipanova is correct - this appears to be the fix for https://pulp.plan.io/issues/970. Could you also rewrite your commit message to reference that issue? We have a special syntax that Redmine looks for to attach commits to issues where you write "fixes #970" on its own line in the commit message (I usually put it last). I personally like to put links to the issue in my commit messages too, though I think that part is optional. I suggest something like this:

```
sort with _id in every unit association query

This commit changes the unit association query's default sort to use the _id field instead of the unit key as it did before. Since the default sort is arbitrary, this improves performance while maintaining API compatibility.

https://pulp.plan.io/issues/970

fixes #970
```

Thanks!

rbarlow:
This looks good, thanks for the contribution!

PR #1899, opened 2015-06-04
Distributor config added to publish history

midnightercz:
It could be handy to have distributor config included in each entry of publish history

### DIFF
--- a/server/pulp/server/db/model/repository.py
+++ b/server/pulp/server/db/model/repository.py
@@ -427,8 +427,9 @@ class RepoPublishResult(Model, ReaperMixin):
         return r
 
     @classmethod
-    def expected_result(cls, repo_id, distributor_id, distributor_type_id, started,
-                        completed, summary, details, result_code):
+    def expected_result(cls, repo_id, distributor_id, distributor_type_id,
+                        distributor_config, started, completed, summary,
+                        details, result_code):
         """
         Creates a new history entry for a successful publish.
 
@@ -457,7 +458,8 @@ class RepoPublishResult(Model, ReaperMixin):
         @type  result_code: str
         """
 
-        r = cls(repo_id, distributor_id, distributor_type_id, started, completed, result_code)
+        r = cls(repo_id, distributor_id, distributor_type_id,
+                distributor_config, started, completed, result_code)
         r.summary = summary
         r.details = details
 
@@ -497,7 +499,8 @@ class RepoPublishResult(Model, ReaperMixin):
 
         return r
 
-    def __init__(self, repo_id, distributor_id, distributor_type_id, started, completed, result):
+    def __init__(self, repo_id, distributor_id, distributor_type_id,
+                 distributor_config, started, completed, result):
         """
         Describes the results of a single completed (potentially errored) publish.
         Rather than directory instantiating instances, use one of the above
@@ -508,6 +511,7 @@ class RepoPublishResult(Model, ReaperMixin):
         self.repo_id = repo_id
         self.distributor_id = distributor_id
         self.distributor_type_id = distributor_type_id
+        self.distributor_config = distributor_config
         self.started = started
         self.completed = completed
         self.result = result
--- a/server/pulp/server/managers/repo/publish.py
+++ b/server/pulp/server/managers/repo/publish.py
@@ -165,7 +165,9 @@ class RepoPublishManager(object):
             result_code = RepoPublishResult.RESULT_SUCCESS
 
         result = RepoPublishResult.expected_result(
-            repo_id, repo_distributor['id'], repo_distributor['distributor_type_id'],
+            repo_id, repo_distributor['id'],
+            call_config.flatten(),
+            repo_distributor['distributor_type_id'], call_config.flatten(),
             publish_start_timestamp, publish_end_timestamp, summary, details, result_code)
         publish_result_coll.save(result, safe=True)
         return result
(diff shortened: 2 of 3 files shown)


ipanova:
this PR is related to https://pulp.plan.io/issues/973

server/pulp/server/managers/repo/publish.py
```diff
@@ -165,7 +165,9 @@ def _do_publish(repo, distributor_id, distributor_instance, transfer_repo, condu
             result_code = RepoPublishResult.RESULT_SUCCESS
 
         result = RepoPublishResult.expected_result(
-            repo_id, repo_distributor['id'], repo_distributor['distributor_type_id'],
+            repo_id, repo_distributor['id'],
+            call_config.flatten(),
+            repo_distributor['distributor_type_id'], call_config.flatten(),
             publish_start_timestamp, publish_end_timestamp, summary, details, result_code)
```
rbarlow: The above doesn't seem to match the call arguments:

```
def expected_result(cls, repo_id, distributor_id, distributor_type_id,
                        distributor_config, started, completed, summary,
                        details, result_code):
```

server/test/unit/server/managers/repo/test_publish.py
```diff
@@ -606,6 +606,30 @@ def test_publish_history_missing_repo(self):
...
+        """
+        Test there's distributor config included in publish history
+        """
+        # Setup
+        self.repo_manager.create_repo('test_date')
+        self.distributor_manager.add_distributor('test_date', 'mock-distributor', {}, True,
+                                                 distributor_id='test_dist')
+        add_result('test_date', 'test_dist', 1)
+        entry = self.publish_manager.publish_history('test_date', 'test_dist')[0]
+        self.assertTrue(hasattr(entry, "distributor_config"))
```
rbarlow: This test should assert more than just the existence of an attribute called distributor_config - it should verify that the data in the distributor_config is corect.

server/test/unit/server/managers/repo/test_publish.py
```diff
@@ -606,6 +606,30 @@ def test_publish_history_missing_repo(self):
...
+        """
+        Test there's distributor config included in publish history
+        """
+        # Setup
+        self.repo_manager.create_repo('test_date')
+        self.distributor_manager.add_distributor('test_date', 'mock-distributor', {}, True,
+                                                 distributor_id='test_dist')
+        add_result('test_date', 'test_dist', 1)
+        entry = self.publish_manager.publish_history('test_date', 'test_dist')[0]
+        self.assertTrue("distributor_config" in entry, entry)
```
rbarlow: This test has the same name as the one above it, which is an error. Also, the code is almost identical except the assertion is slightly different.

server/pulp/server/db/model/repository.py
```diff
@@ -427,8 +427,9 @@ def error_result(cls, repo_id, distributor_id, distributor_type_id, started, com
         return r
 
     @classmethod
-    def expected_result(cls, repo_id, distributor_id, distributor_type_id, started,
-                        completed, summary, details, result_code):
+    def expected_result(cls, repo_id, distributor_id, distributor_type_id,
+                        distributor_config, started, completed, summary,
+                        details, result_code):
```
rbarlow: There doesn't seem to be a test that explicitly tests this change.

server/test/unit/server/managers/repo/test_publish.py
```diff
@@ -482,7 +482,7 @@ def test_publish_history_ascending_sort(self):
         date_string = '2013-06-01T12:00:0%sZ'
         for i in range(0, 10, 2):
             r = RepoPublishResult.expected_result(
-                'test_sort', 'test_dist', 'bar', date_string % str(i), date_string % str(i + 1),
+                'test_sort', 'test_dist', 'bar', {}, date_string % str(i), date_string % str(i + 1),
```
rbarlow: This and many other tests have been updated to pass an empty dictionary here. Please make a new test that passes some data to this method and asserts that expected_result() did the right thing with the information.

server/pulp/server/db/model/repository.py
```diff
@@ -508,6 +511,7 @@ def __init__(self, repo_id, distributor_id, distributor_type_id, started, comple
         self.repo_id = repo_id
         self.distributor_id = distributor_id
         self.distributor_type_id = distributor_type_id
+        self.distributor_config = distributor_config
```
rbarlow: Please create a test that specifically asserts that this __init__() method sets this attribute.

server/pulp/server/managers/repo/publish.py
rbarlow: Please make sure that the above is explicitly tested so we can be sure that it works properly.

rbarlow:
By policy, we need these changes to be made in a single commit. We can't merge this with 2.7-dev, because it is a behavioral change. I'm also concerned that the tests are not making strong assertions about the data being stored, and that the call to that function does not match the method signature.

I'm going to close this since it needs to be changed in the master branch instead of 2.7-dev. Here are the things we need fixed about this change before we can merge to master:

0) Make the change in a single commit.
1) Have the commit message reference the Redmine issue with "fixes #973" on its own line.
2) Have the tests make strong assertions about the distributor_config, not just the attribute existing.
3) Make sure there are no duplicate tests.
4) Make sure the call parameters are used correctly.
5) Make sure there is a test for each item that you've changed, asserting that the new behavior is working correctly.

PR #1900, opened 2015-06-04
- use cp instead of shutil.copytree due file attributes preservation

midnightercz:
cp can preserver file attributes

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -746,7 +746,7 @@ class AtomicDirectoryPublishStep(PublishStep):
         # for items where http & https are published to a separate directory
 
         _logger.debug('Copying tree from %s to %s' % (self.source_dir, timestamp_master_dir))
-        shutil.copytree(self.source_dir, timestamp_master_dir, symlinks=True)
+        os.system("cp -a %s %s" % (self.source_dir, timestamp_master_dir))
 
         for source_relative_location, publish_location in self.publish_locations:
             if source_relative_location.startswith('/'):
@@ -858,7 +858,7 @@ class CopyDirectoryStep(PublishStep):
         """
         if self.delete_before_copy:
             shutil.rmtree(self.target_dir, ignore_errors=True)
-        shutil.copytree(self.source_dir, self.target_dir, symlinks=self.preserve_symlinks)
+        os.system("cp -a %s %s" % (self.source_dir, timestamp_master_dir))
 
 
 class PluginStepIterativeProcessingMixin(object):


ipanova:
this fixes https://pulp.plan.io/issues/971

bmbouter:
For cherry picking purposes we, please add two things to each of your commit messages.

(1) a link in the commit message to the issue associated with it
(2) a bit of redmine syntax so that redmine shows the commit that fixes an issue. See this issue with demos of the syntax:  https://pulp.plan.io/issues/103

For this one for instance, please redo the commit message as:

```
Use cp instead of shutil.copytree due file attributes preservation

https://pulp.plan.io/issues/971
fixes #971
```

bmbouter:
ok test

bmbouter:
My thought is that this change creates a correctness problem with Pulp. The [motivation for the change](https://pulp.plan.io/issues/971#note-3) is to have the published timestamps reflect an older publish/sync time and not have the symlinks of a new publish carry new created times. When Pulp publishes symlinks as part of a publish, I expect Pulp to "make new files" and those symlinks should carry created timestamps that reflect the new publish date, not the previous publish/sync created date. I can see the benefit in some specific use cases but for general Pulp use it seems incorrect.

The comment on the ticket mentions that created timestamps could have an additional step done during the rsync. This is probably the best thing to do since doing nothing would cause caching issues.

I'm going to leave some comments on the inline issues, but the above topic is the most important point to discuss; the other problems are implementation level things that could be resolved.

server/pulp/plugins/util/publish_step.py
```diff
@@ -746,7 +746,7 @@ def process_main(self):
         # for items where http & https are published to a separate directory
 
         _logger.debug('Copying tree from %s to %s' % (self.source_dir, timestamp_master_dir))
-        shutil.copytree(self.source_dir, timestamp_master_dir, symlinks=True)
+        os.system("cp -a %s %s" % (self.source_dir, timestamp_master_dir))
```
bmbouter: This should pass an additional option to `cp ` that will ensure symlinks are preserved. Other codepaths that haven't been tested could be relying on that.

server/pulp/plugins/util/publish_step.py
```diff
@@ -858,7 +858,7 @@ def process_main(self):
         """
         if self.delete_before_copy:
             shutil.rmtree(self.target_dir, ignore_errors=True)
-        shutil.copytree(self.source_dir, self.target_dir, symlinks=self.preserve_symlinks)
+        os.system("cp -a %s %s" % (self.source_dir, timestamp_master_dir))
```
bmbouter: This will break export and iso export codepaths because the resulting publish would by all symlinks which will break when they are disconnected from the base filesystem with the actual content. To fix this have the `self.preserve_symlinks` option pass the right options to `cp` to either preserve or dereference symlinks.

bmbouter:
This change [breaks existing tests.](https://github.com/pulp/pulp/pull/1900#issuecomment-114647475). Also any new codepaths that would be introduced would need to have full test coverage with assertions. Also any calls to `os.system` would need to be mocked in new or existing tests.

rbarlow:
On 06/24/2015 05:09 PM, bmbouter wrote:
> My thought is that this change creates a correctness problem with Pulp.
> The motivation for the change <https://pulp.plan.io/issues/971#note-3>
> is to have the published timestamps reflect an older publish/sync time
> and not have the symlinks of a new publish carry new created times. When
> Pulp publishes symlinks as part of a publish, I expect Pulp to "make new
> files" and those symlinks should carry created timestamps that reflect
> the new publish date, not the previous publish/sync created date. I can
> see the benefit in some specific use cases but for general Pulp use it
> seems incorrect.

FWIW, I agree.

bmbouter:
@midnightercz 

In the last month we've learned that copying files into `/var/lib/pulp` in a way that preserves attributes will break SELinux. Because of this we carry an [internal copytree](https://github.com/pulp/pulp/blob/3a1fe3cf0a95684ed16a4068a463afbad8d52102/server/pulp/server/util.py#L175) which is the same as the standard copytree except that it doesn't copy extended attributes. See the docstring for a more detailed descrpition as to why.

Also, the use of `mv` is prohibited, we have to use copy instead. Finally, we cannot use `shutil.copy2()`, and instead we need to use [`shutil.copy()`](https://docs.python.org/2/library/shutil.html#shutil.copy)

midnightercz:
I consulted with Dennis and we don't need preserve file attributes anymore (due better rsync strategy)

PR #1901, opened 2015-06-04
Move password masking & serialization of the importer out of the cont…

barnabycourt:
…roller and into the views

fixes #962

https://pulp.plan.io/issues/962

### DIFF
--- a/server/pulp/server/webservices/views/repositories.py
+++ b/server/pulp/server/webservices/views/repositories.py
@@ -14,7 +14,7 @@ from pulp.server.managers.repo import importer as repo_importer_manager
 from pulp.server.managers.repo.distributor import RepoDistributorManager
 from pulp.server.managers.repo.unit_association import associate_from_repo, unassociate_by_criteria
 from pulp.server.tasks import repository as repo_tasks
-from pulp.server.webservices.views import search
+from pulp.server.webservices.views import search, serializers
 from pulp.server.webservices.views.decorators import auth_required
 from pulp.server.webservices.views.schedule import ScheduleResource
 from pulp.server.webservices.views.util import (generate_json_response,
@@ -54,7 +54,11 @@ def _merge_related_objects(name, manager, repos):
         repo[name] = []
 
     for item in manager.find_by_repo_list(repo_ids):
-        repo_dict[item['repo_id']][name].append(item)
+        if name == 'importers':
+            serializer = serializers.ImporterSerializer(item)
+            repo_dict[item['repo_id']][name].append(serializer.data)
+        else:
+            repo_dict[item['repo_id']][name].append(item)
 
     return repos
 
@@ -375,7 +379,9 @@ class RepoImportersView(View):
 
         importer_manager = manager_factory.repo_importer_manager()
         importers = importer_manager.get_importers(repo_id)
-        return generate_json_response_with_pulp_encoder(importers)
+        serializer = serializers.ImporterSerializer(importers, multiple=True)
+
+        return generate_json_response_with_pulp_encoder(serializer.data)
 
     @auth_required(authorization.CREATE)
     @json_body_required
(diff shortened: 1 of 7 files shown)


server/pulp/server/webservices/views/serializers/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+
+from django.core.urlresolvers import reverse
+
+
+class BaseSerializer(object):
+    """
+    Base class to be used for creating serializers
+    """
+
+    def __init__(self, instance=None, context=None, multiple=False):
```
jortel: I *instance* really an optional parameter?

Would it be reasonable to detect if *instance* is iterable instead of relying to *multiple* being passed correctly?

server/pulp/server/webservices/views/serializers/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        root_key = accessor[0]
+        if root_key in representation:
+            if len(accessor) > 1:
+                self._mask_field(accessor[1:], representation[root_key])
+            else:
+                representation[root_key] = '*****'
+
+    @property
+    def data(self):
```
jortel: Thanks for not having a *serialize()* method.  The redundancy of serializer.serialize() just kills me :)
That said, I wonder if a command pattern would be a better semantic.  Using __call__() instead.  Might look like this:
```
serializer = MySerializer(thing)
data = serializer()
```

Just offered as an alternative suggestion.

barnabycourt: So, the reason this uses serializer.data is so that we could think about adding a create() or update() method similar to http://www.django-rest-framework.org/api-guide/serializers/ to support taking a raw data dictionary from the request and turning it into a mongoengine model for use in other calls.

server/pulp/server/webservices/views/serializers/__init__.py
barnabycourt: At this point instance is not really optional, though it becomes optional with the create/update pattern referenced below.

Unfortunately I know of no absolute way to detect iterability that doesn't leave the possibility of false positives. For example, a string is iterable even though you probably wouldn't consider it to be iterable from a serialization standpoint. The cleanest thing is to ask the user of the class to state their intent.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -54,7 +54,11 @@ def _merge_related_objects(name, manager, repos):
         repo[name] = []
 
     for item in manager.find_by_repo_list(repo_ids):
-        repo_dict[item['repo_id']][name].append(item)
+        if name == 'importers':
+            serializer = serializers.ImporterSerializer(item)
+            repo_dict[item['repo_id']][name].append(serializer.data)
+        else:
+            repo_dict[item['repo_id']][name].append(item)
```
jortel: Line 61 missing unit test coverage.

jortel:
Really glad you fixed this.  I ran into this when testing ostree proxy support.  Really like moving this to the view layer.  Also, like the *Serializer* classes approach.

server/pulp/server/webservices/views/serializers/__init__.py
rbarlow: On 06/04/2015 03:52 PM, Barnaby Court wrote:
> Unfortunately I know of no absolute way to detect iterability that
> doesn't leave the possibility of false positives. For example, a string
> is iterable even though you probably wouldn't consider it to be iterable
> from a serialization standpoint. The cleanest thing is to ask the user
> of the class to state their intent.

The collections module has an abstract base class called Iterable that
matches a list and a string:

```
>>> import collections
>>> isinstance([1, 2, 3], collections.Iterable)
True
>>> isinstance('hello', collections.Iterable)
True
>>> isinstance(5, collections.Iterable)
False
```

Or was it that you don't want a string to count as an iterable? If
that's the case, you can test this way:

```
>>> hasattr('hello', '__iter__')
False
>>> hasattr([1, 2, 3], '__iter__')
True
```

-- 
Randy Barlow

barnabycourt:
ok test

PR #1903, opened 2015-06-04
Fixed 952: Delete unused models from server.db.models.dispatch

shubham90:
https://pulp.plan.io/issues/952

Fixes #952

### DIFF
--- a/docs/user-guide/release-notes/2.7.x.rst
+++ b/docs/user-guide/release-notes/2.7.x.rst
@@ -52,6 +52,11 @@ Deprecation
 
 .. _2.6.x_upgrade_to_2.7.0:
 
+* The ``archived_calls`` setting in ``server.conf`` is removed. It has not been used in several releases.
+
+* The ``archived_calls`` collection is dropped from the database with a migration.
+
+
 Upgrade Instructions for 2.6.x --> 2.7.0
 -----------------------------------------
 
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -147,8 +147,6 @@
 # reaper_interval: float; time in days between checks for old data in
 #     the database
 #
-# archived_calls: float; time in days to store archived calls
-#
 # consumer_history: float; time in days to store consumer history events
 #
 # repo_sync_history: float; time in days to store repository sync history events
@@ -164,7 +162,6 @@
 
 [data_reaping]
 # reaper_interval: 0.25
-# archived_calls: 0.5
 # consumer_history: 60
 # repo_sync_history: 60
 # repo_publish_history: 60
--- a/server/pulp/server/db/migrations/0018_remove_archived_calls.py
+++ b/server/pulp/server/db/migrations/0018_remove_archived_calls.py
@@ -1,0 +1,30 @@
+"""
+This migration deletes the archived_calls collection if it is present.
+"""
+from gettext import gettext as _
+import logging
+
+from pulp.server.db import connection
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(*args, **kwargs):
+    """
+    Perform the migration as described in this module's docblock.
+
+    :param args:   unused
+    :type  args:   list
+    :param kwargs: unused
+    :type  kwargs: dict
+    """
+    db = connection.get_database()
+    # If 'archived_calls' is not present, do nothing.
+    if 'archived_calls' not in db.collection_names():
+        return
+
+    # If it's present, drop it
+    collection = db['archived_calls']
+    collection.drop()
+    msg = _("Deleted the archived_calls collection.")
+    _logger.info(msg)
--- a/server/pulp/server/db/model/dispatch.py
+++ b/server/pulp/server/db/model/dispatch.py
@@ -13,58 +13,10 @@ from celery.utils.timeutils import timedelta_seconds
 from pulp.common import dateutils
 from pulp.server.async.celery_instance import celery as app
 from pulp.server.db.model.base import Model
-from pulp.server.db.model.reaper_base import ReaperMixin
 from pulp.server.managers import factory
 
 
 _logger = logging.getLogger(__name__)
-
-
-class CallResource(Model):
-    """
-    Information for an individual resource used by a call request.
-    """
-
-    collection_name = 'call_resources'
-    search_indices = ('call_request_id', 'resource_type', 'resource_id')
-
-    def __init__(self, call_request_id, resource_type, resource_id, operation):
-        super(CallResource, self).__init__()
-        self.call_request_id = call_request_id
-        self.resource_type = resource_type
-        self.resource_id = resource_id
-        self.operation = operation
-
-
-class QueuedCall(Model):
-    """
-    Serialized queued call request
-    """
-
-    collection_name = 'queued_calls'
-    unique_indices = ()
-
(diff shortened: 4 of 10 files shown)


ipanova:
ok test

server/test/unit/server/db/test_reaper.py
```diff
@@ -23,8 +23,7 @@ def test_reaper_collections(self):
         """
         Test that the expected key-value pairs exist in the reaper collection to timedelta mapping.
         """
-        collections_to_reap = [dispatch.ArchivedCall,
```
ipanova: flake8 says: server/test/unit/server/db/test_reaper.py:13:1: F401 'dispatch' imported but unused

server/pulp/server/db/model/dispatch.py
```diff
@@ -13,60 +13,11 @@
 from pulp.common import dateutils
 from pulp.server.async.celery_instance import celery as app
 from pulp.server.db.model.base import Model
-from pulp.server.db.model.reaper_base import ReaperMixin
 from pulp.server.managers import factory
 
 
 _logger = logging.getLogger(__name__)
```
ipanova: add one more blank line here

ipanova:
@shubham90  there are some flake8 errors, please fix them. You can see them in the logs of jenkins build. Overwise LGTM

shubham90:
@ipanova ok will do. Thanks.

ipanova:
ok test

ipanova:
ok test

ipanova:
LGTM

ipanova:
@shubham90  for the next time, it is better to write  in the commit message "closes #123" instead of "done", i do not think that with 'done' it will automatically change the status of the issue

shubham90:
@ipanova Thanks for the suggestion. Will keep that in mind from next time.

rbarlow:
On 06/08/2015 10:41 AM, Shubham Bhawsinka wrote:
> @ipanova <https://github.com/ipanova> Thanks for the suggestion. Will
> keep that in mind from next time.

You can still amend your commit message if you haven't merged the commit.

-- 
Randy Barlow

shubham90:
Thanks @rbarlow. Did it the way you suggested last time! It was really helpful.

nodes/test/nodes_tests/base.py
```diff
@@ -50,7 +49,7 @@ def setUpClass(cls):
         managers.initialize()
 
     def setUp(self):
```
bmbouter: this method can be removed entirely since it only contains pass

server/etc/pulp/server.conf
```diff
@@ -164,7 +163,6 @@
 
 [data_reaping]
 # reaper_interval: 0.25
-# archived_calls: 0.5
```
bmbouter: This can't be removed due to semantic versioning applied to the config files. It needs a deprecated comment added after it. See this line example:  https://github.com/shubham90/pulp/blob/i952/server/etc/pulp/server.conf#L120

bmbouter:
This needs a release note also since it deprecates a setting in server.conf.

bmbouter:
@shubham90 Please also write a migration in platform that will delete archived_calls collection entirely if it is present. If it is not present the migration should pass without error. Ask one of the devs about how to do this in IRC.

bmbouter:
@ipanova I originally wrote this bug so I looked at the PR. There are more things that it needed so I removed the LGTM flag.

shubham90:
Thanks @bmbouter for the update. I'll make the changes as you suggested. They seems to be reasonable.

ipanova:
@bmbouter  thanks for looking into this, i also learned some things

server/etc/pulp/server.conf
rbarlow: I disagree - I think we can remove this setting since it is getting removed elsewhere. It doesn't make sense to mark it deprecated when it actually doesn't do anything. Also, users who have this setting present in their config file will not experience any problems - it just doesn't get used.

I do think we should mention in the release notes that this setting is not used anymore.

bmbouter: You're saying the other ones marked as deprecated are still used while this one is not. Removing this line from the `server.conf` and adding the release note only sounds fine to me. Also the release note needs to mention that both the setting is removed from server.conf and that the `archived_calls` collection will be removed with a migration.

rbarlow: On 06/09/2015 04:24 PM, bmbouter wrote:
> You're saying the other ones marked as deprecated are still used while
> this one is not.

I don't know if the above is true or not. I was just saying that for
this particular setting, there's not a practical reason to mark it
deprecated since it's actually not used (and thus, removed).

-- 
Randy Barlow

shubham90:
@bmbouter Please review the recent changes. Thanks a lot for helping me out with this.

shubham90:
ok test

shubham90:
ok test

shubham90:
@bmbouter please review once you are free. Thanks.

docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -52,6 +52,10 @@ Deprecation
 
 .. _2.6.x_upgrade_to_2.7.0:
 
+* The setting for ``archived_calls`` is removed from ``server.conf`` as it was not used in the previous releases.
```
bmbouter: Suggested rewrite.

```
The ``archived_calls`` setting in ``server.conf`` is removed. It has not been used in several releases.
```

docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -52,6 +52,10 @@ Deprecation
...
+  Also, the ``archived_calls`` collection is dropped from the database with a migration.
```
bmbouter: suggest making this its own bullet point and slightly reword:
```
* The ``archived_calls`` collection is dropped from the database with a migration.
```

server/etc/pulp/server.conf
```diff
@@ -147,7 +147,6 @@
 # reaper_interval: float; time in days between checks for old data in
 #     the database
 #
-# archived_calls: float; time in days to store archived calls
 #
```
bmbouter: This line should also be removed because it has inconsistent spacing now.

server/pulp/server/db/migrations/0018_remove_archived_calls.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    db = connection.get_database()
+    # If 'archived_calls' is not defined we don't need to do anything
+    if 'archived_calls' not in db.collection_names():
+        return
+
+    # If it is present, delete it
+    collection = db['archived_calls']
+    collection.drop()
+    msg = _("Deleted archived_calls collection.")
```
bmbouter: s/Deleted/Deleted the/

bmbouter:
@shubham90 some small language changes and a line removal. Please ping me after that to look again.

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
Closing this PR and creating a new PR https://github.com/pulp/pulp/pull/1916 under 2.7-dev.

PR #1911, opened 2015-06-08
Fixing behaviour of individual orphan removal and orphan removal by type

ipanova:
closes #667
closes #1035
https://pulp.plan.io/issues/667
https://pulp.plan.io/issues/1035

### DIFF
--- a/docs/dev-guide/integration/rest-api/content/orphan.rst
+++ b/docs/dev-guide/integration/rest-api/content/orphan.rst
@@ -167,6 +167,7 @@ Remove all the orphaned content of a particular content type.
 | :response_list:`_`
 
 * :response_code:`202,even if no content is to be deleted`
+* :response_code:`404,if content type does not exist`
 
 | :return:`a` :ref:`call_report`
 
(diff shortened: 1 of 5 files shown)


server/pulp/server/managers/content/orphan.py
```diff
@@ -17,6 +17,13 @@
 
 class OrphanManager(object):
 
+    @staticmethod
+    def validate_type(content_type_id):
```
mhrivnak: needs a doc block

docs/dev-guide/integration/rest-api/content/orphan.rst
```diff
@@ -183,7 +184,7 @@ Remove and individual orphaned content unit by content type and content id.
 | :response_list:`_`
 
 * :response_code:`202,if the content unit is to be deleted`
-* :response_code:`404,if the content does not exist`
+* :response_code:`404,if the content unit or content type do not exist`
```
mhrivnak: "does" is still the correct word here, because the subject is a single unit or single type.

mhrivnak:
I think this is missing a couple of important test assertions. For each view, we need to verify that the MissingResource exception gets raised when appropriate. If you want to isolate testing that in the ``validate_type`` function, then you could just verify that the view calls it.

ipanova:
ok test

ipanova:
ok test

PR #1913, opened 2015-06-09
This reduces the batch size to 5

dkliban:
25 was still causing timeouts for some users. 10 worked, so 5 seems
like a safe bet.

https://pulp.plan.io/issues/998
fixes #998


dkliban:
ok test

beav:
LGTM after jenkins passes

PR #1915, opened 2015-06-11
Fixes 917 and 1006 by stopping an thread that should have exited

bmbouter:
https://pulp.plan.io/issues/917
https://pulp.plan.io/issues/1006

fixes #917
fixes #1006
re #711

I also did a `tito build --test --rpm` where all unit tests in python-kombu were run and passed, and it successfully built python-kombu-3.0.24-8.pulp.


beav:
an elegant fix :wine_glass: 

LGTM after jenkins passes

bmbouter:
ok test

PR #1917, opened 2015-06-11
Use paths property instead of url property in the example.

jortel:
https://pulp.plan.io/issues/1047


docs/user-guide/content-sources.rst
```diff
@@ -38,25 +38,25 @@ are supported:
  - **paths** <str>
      An *optional* list of URL relative paths. Delimited by space or newline.
  - **max_concurrent** <int>
-     Limit the number of concurrent downloads.
+     An *optional* limit the number of concurrent downloads.
```
mhrivnak: stick "to", "for" or similar after "limit". Same 3 lines below.

PR #1918, opened 2015-06-12
unit search works again with the "include_repos" argument

mhrivnak:
fixes #1038 

https://pulp.plan.io/issues/1038

### DIFF
--- a/server/pulp/server/webservices/views/content.py
+++ b/server/pulp/server/webservices/views/content.py
@@ -238,7 +238,7 @@ class ContentUnitSearch(search.SearchView):
     """
     Adds GET and POST searching for content units.
     """
-    optional_fields = ['include_repos']
+    optional_bool_fields = ('include_repos',)
     manager = content_query.ContentQueryManager()
 
     @staticmethod
@@ -283,7 +283,7 @@ class ContentUnitSearch(search.SearchView):
         type_id = kwargs['type_id']
         units = list(search_method(type_id, query))
         units = [_process_content_unit(unit, type_id) for unit in units]
-        if options.get('include_repos', 'false').lower() == "true":
+        if options.get('include_repos') is True:
             cls._add_repo_memberships(units, type_id)
         return units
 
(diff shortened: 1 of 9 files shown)


server/pulp/server/webservices/views/content.py
```diff
@@ -275,6 +275,28 @@ def _add_repo_memberships(units, type_id):
...
+                     removed and returned separately.
+        :type  args: django.http.QueryDict
+
+        :return: filtered search parameters and options
+        :rtype:  tuple containing a 2 dicts
+        """
+        search_params, options = super(ContentUnitSearch, cls)._parse_args(args)
+
+        # if this was a GET request, the argument will be a string. We convert it to a bool here.
+        if isinstance(options.get('include_repos'), basestring):
```
asmacdo: I think we could this logic into the base search class and not need to override _parse_args. The bool issue will be present for all of the search endpoints that have the optional params separated in _parse_args.

server/pulp/server/webservices/views/search.py
```diff
@@ -64,7 +64,8 @@ def _parse_args(cls, args):
         options = {}
         for field in cls.optional_fields:
             if search_params.get(field):
-                options[field] = search_params.pop(field)
+                options[field] = args.get(field)
```
asmacdo: perhaps:

    if isinstance(args.get(field), basestring):
        options[field] = args.get(field).lower == 'true'
    else:
         options[field] = args.get(field)
    search_params.pop(field)


This will require some changes in the subclasses to make sure that they expect bools, but it should fix the problem more generally.

mhrivnak:
I wanted to keep this fix isolated because the other two views that use extra options do things completely differently, and I didn't want this simple bug fix to scope-creep into a big refactor. However, investigating your suggestions has revealed that the other two endpoints have the same bug. One of them is repository search, which is pretty important. I'll consolidate to the base class and fix those as well.

mhrivnak:
ok test

PR #1919, opened 2015-06-12
parse filters json only if filters are present

asmacdo:
…ches

fixes: 1040

https://pulp.plan.io/issues/1040

### DIFF
--- a/server/test/unit/server/webservices/views/test_search.py
+++ b/server/test/unit/server/webservices/views/test_search.py
@@ -78,6 +78,19 @@ class TestSearchView(unittest.TestCase):
 
     @mock.patch('pulp.server.webservices.views.decorators._verify_auth',
                 new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.search.SearchView._parse_args')
+    def test_get_with_invalid_filters(self, mock_parse):
+        """
+        InvalidValue should be raised if param 'filters' is not json.
+        """
+        mock_parse.return_value = ({'mock': 'query'}, 'tuple')
+        search_view = search.SearchView()
+        mock_request = mock.MagicMock()
+        mock_request.GET = {'filters': 'invalid json'}
+        self.assertRaises(exceptions.InvalidValue, search_view.get, mock_request)
+
+    @mock.patch('pulp.server.webservices.views.decorators._verify_auth',
+                new=assert_auth_READ())
     def test_post(self):
         """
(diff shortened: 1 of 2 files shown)


asmacdo:
ok test

asmacdo:
ok test

server/test/unit/server/webservices/views/test_search.py
```diff
@@ -78,6 +78,19 @@ class FakeSearchView(search.SearchView):
...
                 new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.search.SearchView._parse_args')
+    def test_get_without_filters(self, mock_parse):
+        """
+        MissingValue should be raised if required param 'filters' is not passed.
+        """
+        mock_parse.return_value = ('mock', 'tuple')
+        search_view = search.SearchView()
+        mock_request = mock.MagicMock()
+        mock_request.GET = {}
```
mhrivnak: subtle difference, but I'd use a ``QueryDict`` here, because why not?

mhrivnak:
I'm not sure this is the right solution. I can't find that we document anywhere that this is required, but I can think of ways this API is useful without using filters (pagination for example). If you do think it's a good idea to make this a requirement, please elaborate on why.

asmacdo:
@mhrivnak, I agree with you. updated.

asmacdo:
ok test

asmacdo:
ok test

server/test/unit/server/webservices/views/test_search.py
```diff
@@ -78,6 +78,19 @@ class FakeSearchView(search.SearchView):
...
                 new=assert_auth_READ())
+    @mock.patch('pulp.server.webservices.views.search.SearchView._parse_args')
+    def test_get_with_invalid_filters(self, mock_parse):
+        """
+        InvalidValue should be raised if param 'filters' is not json.
+        """
+        mock_parse.return_value = ({'mock': 'query'}, 'tuple')
+        search_view = search.SearchView()
+        mock_request = mock.MagicMock()
+        mock_request.GET = {'filters': 'invalid json'}
```
mhrivnak: I'd use a ``QueryDict`` here for authenticity. It might not make much difference, but we may as well keep tests as close to real circumstances as we can.

asmacdo:
moving to 2.7-testing https://github.com/pulp/pulp/pull/1923

PR #1920, opened 2015-06-12
Removed F22 from dist_list

dkliban:


beav:
LGTM after jenkins passes

PR #1922, opened 2015-06-15
Last associated

midnightercz:
new attriubte required by rcm.


rbarlow:
Can you explain what the use case is for this attribute? It seems strange to keep a record of the last datetime that the unit was associated with a repo, but keep none of the other relevant information about the association. I'm trying to think of a use case that would benefit upstream Pulp users, if that makes sense.

midnightercz:
This is statement of Sebastian which is maintainer of Unified Downloads:
" soooo, the long story short is that we're using this attribute to sort RPMs in order to deliver so-called XML feeds for the internal search engine and that feed needs to be paginated and ordered chronologically. If you ask why this attribute is needed in addition to _last_modified
the answer is                     
in our understanding, when a package gets associated/disassociated from a repo 
it does not change its _last_modified. The XML feeds include the information on which product/versions the RPM is mapped to so when we generate a feed that displays eveyrthing that has been changed in last 24 hours we want to include not only new RPMs, but also the existing RPMs for which the repo mappings have been changed "

mhrivnak:
Could you instead accomplish the same thing by using this new feature? https://pulp.plan.io/issues/86

That lets you use the API to put any custom metadata you want on a unit.

rbarlow:
I believe that @mhrivnak's suggestion will accomplish your purposes, so I am closing this PR for now. Feel free to chat with us if you think that doesn't help you!

PR #1924, opened 2015-06-15
Allow disabling specific repo auth plugins

beav:
Previously in Pulp 2.7, repo authenticator plugins were controlled only via the
plugin's entry points. This meant that if a plugin was installed by default, we
were not able to disable it.

A new config value has been added to repo_auth.conf to disable specific
plugins. This lets us turn plugins on and off at will, without having to
uninstall any default plugins.

fixes #1055

https://pulp.plan.io/issues/1055

### DIFF
--- a/server/etc/pulp/repo_auth.conf
+++ b/server/etc/pulp/repo_auth.conf
@@ -11,6 +11,10 @@ max_num_certs_in_chain: 100
 # maintain backwards compatibility.
 # verify_ssl: true
 
+# If set, this disables specific repo auth plugins. More than one plugin can be
+# specified in the form of "plugin1,plugin2,plugin3".
+# disabled_authenticators = oid_validation
+
 [repos]
 cert_location: /etc/pki/pulp/content
 global_cert_location: /etc/pki/pulp/content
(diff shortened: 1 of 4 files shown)


beav:
I attempted to make a more comprehensive patch that cleaned up a few other areas as well, but ran afoul of making changes that would not be supported in a Y stream release. This patch does what I need in downstream without breaking Y-stream compatibility.

beav:
ok test

beav:
ok test

beav:
ok test

server/etc/pulp/repo_auth.conf
```diff
@@ -11,6 +11,9 @@ max_num_certs_in_chain: 100
 # maintain backwards compatibility.
 # verify_ssl: true
 
+# If set, this disables specific repo auth plugins
```
dkliban: Should probably add comment about comma separated list being allowed.

dkliban:
You also need to update the spec file and pulp-dev.py to include the new config file.

beav:
The spec change should already be in there from an earlier commit: https://github.com/pulp/pulp/blob/2.7-testing/pulp.spec#L886

and for pulp-dev.py: https://github.com/pulp/pulp/blob/2.7-testing/pulp-dev.py#L197-198

I will make the update to the conf file's comment to mention that commas are allowed

dkliban:
@beav in that case LGTM after you update the comment in the conf file.

PR #1925, opened 2015-06-15
Added config for dependencies

dkliban:
python-requests now builds for F20 and F21. mod_wsgi now builds for F20.


deps/python-requests/dist_list.txt
```diff
@@ -1 +1 @@
-el6 el7
```
barnabycourt: Do we need f21 support here? What version of python requests ships with f21?

dkliban: @barnabycourt it looks like 2.3.0 ships with F21

PR #1926, opened 2015-06-15
Added config for dependencies

dkliban:
python-requests now builds for F20 and F21. mod_wsgi now builds for F20.


beav:
This changed was LGTM'd in https://github.com/pulp/pulp/pull/1925. LGTM here as well.

PR #1930, opened 2015-06-16
Install koji on the dev box.

rbarlow:


beav:
is tito already on there?

rbarlow:
On 06/16/2015 01:08 PM, Chris Duryee wrote:
> is tito already on there?

I'm not sure, but I would guess not. Does koji pull in tito? If not, it
would be easy to add it to the list. Is that your suggestion?

-- 
Randy Barlow

beav:
yeah, if you don't mind, adding tito would be good :space_invader: 

For my use case at least, I'm usually using tito to create an srpm to test spec file changes, and then scratch-building in koji. Having tito already installed would save me one step :smile:

PR #1931, opened 2015-06-16
Use mongoengine to access the repos collection

asmacdo:
https://pulp.plan.io/issues/860

This replaces the direct access of the `repos` collection with the use of the Repository Mongoengine model. In addition, this began an [alteration of our MVC pattern](https://gist.github.com/asmacdo/fb194f2709b8bb62e302) by removing managers. The manager code is mostly moved into a controller, but many uses of the manager are also replaced by using the model directly. 

Additionally, there is task logic that belongs in the controllers (which I was forced to move at the same time due to circular imports). Tasks are named by default after their location, so I renamed them to maintain backwards compatibility for the task scheduler. This means that their names at the moment are misleading, so once all tasks have been moved, it would be a good idea to change them all at once. This is being tracked here: https://pulp.plan.io/issues/1052


client_lib/pulp/client/commands/repo/query.py
```diff
@@ -34,7 +34,7 @@ def run(self, **kwargs):
         # Limit to only repositories of a specific type
         if kwargs.get('str-eq', None) is None:
             kwargs['str-eq'] = []
-        kwargs['str-eq'].append(['notes.%s' % constants.REPO_NOTE_TYPE_KEY, self.repo_type])
+        kwargs['str-eq'].append(['notes__%s' % constants.REPO_NOTE_TYPE_KEY, self.repo_type])
```
barnabycourt: This should be fixed on the server side using the field mapping stuff we worked on earlier today.

nodes/parent/pulp_node/distributors/http/distributor.py
```diff
@@ -214,8 +215,11 @@ def _add_repository(self, repo_id, payload):
         :param payload: The repository payload
         :type payload: dict
         """
-        manager = factory.repo_query_manager()
-        payload['repository'] = manager.get_repository(repo_id)
+        repo_obj = model.Repository.objects.get_repo_or_missing_resource(repo_id)
+        # Pseudo serialize the repository object so that it can be used by a node.
+        payload['repository'] = {'id': repo_obj.repo_id, 'display_name': repo_obj.display_name,
```
barnabycourt: Couldn't this be fixed using the serializer at this point?

nodes/test/nodes_tests/test_plugins.py
```diff
@@ -186,7 +190,7 @@ def tearDown(self):
         shutil.rmtree(self.childfs)
         Consumer.get_collection().remove()
         Bind.get_collection().remove()
-        Repo.get_collection().remove()
+        model.Repository.drop_collection()
```
barnabycourt: Why the drop_collection() instead of remove?

nodes/parent/pulp_node/distributors/http/distributor.py
asmacdo: The scratchpad is necessary so the repository serializer won't work as is. Also, @jortel and I didn't like the idea of using a view serializer to change what is being sent to pulp_nodes.

nodes/test/nodes_tests/test_plugins.py
asmacdo: No real reason. Why `model.Repository.objects.delete()`?

barnabycourt: not sure it makes a big difference, remove clears the contents of a collection but the collection remains. drop_collection deletes the collection as well as the contents. I noticed the change in behavior and thus was curious as to why.

rbarlow: On 06/18/2015 09:29 AM, Barnaby Court wrote:
> not sure it makes a big difference, remove clears the contents of a
> collection but the collection remains. drop_collection deletes the
> collection as well as the contents. I noticed the change in behavior and
> thus was curious as to why.

There is one way in which it makes an important difference. As said,
remove() removes the documents in the collection and drop_collection()
drops the collection itself. The significant bit is that the indices get
dropped as well with drop_collection() but they remain with remove().

Since we hopefully aren't writing large amounts of data in our unit
tests, it probably doesn't make a noticeable difference in test
performance if the indices are present or not. But technically it is
probably slightly advantageous for write-once read-many tests (if we
have such things).

This actually could affect correctness too. We do have some code that
relies on uniqueness constraints. If the tests drop the database
enforced uniqueness constraints, the code being tested isn't going to
benefit from the more realistic environment that the code would
experience in production. That said, I think our tests should move away
from using the database entirely so perhaps this point is moot.

-- 
Randy Barlow

asmacdo:
skip ci

server/pulp/plugins/conduits/repo_publish.py
```diff
@@ -52,17 +54,20 @@ def __str__(self):
...
+        :return: timestamp instance describing the last publish
+        :rtype:  datetime.datetime or None
+
+        :raises DistributorConduitException: if any errors occur
         """
         try:
-            repo_publish_manager = manager_factory.repo_publish_manager()
-            last = repo_publish_manager.last_publish(self.repo_id, self.distributor_id)
-            return last
+            collection = RepoDistributor.get_collection()
```
barnabycourt: Could this be condensed into a controller method?

server/pulp/server/controllers/repository.py
```diff
@@ -125,3 +145,708 @@ def associate_single_unit(repository, unit):
...
+    :rtype:  TaskResult
+
+    :raises pulp_exceptions.InvalidValue: if repo_delta is not a dictionary
+    """
+    if repo_delta:
+        if isinstance(repo_delta, dict):
+            repo.update_from_delta(repo_delta)
+            repo.save()
+        else:
+            raise pulp_exceptions.InvalidValue(['delta'])
```
barnabycourt: This is a horribly imprecise error when all that is being checked is if delta is a Dict

server/pulp/server/controllers/repository.py
```diff
@@ -125,3 +145,708 @@ def associate_single_unit(repository, unit):
...
+    :param repo_id: identifies the repo
+    :type  repo_id: str
+    :param unit_type_id: identifies the unit type to update
+    :type  unit_type_id: str
+    :param delta: amount by which to increment the total count
+    :type  delta: int
+
+    :raises pulp_exceptions.PulpCodedException: if there is an error in the update
+    """
+    atomic_inc_key = 'inc__content_unit_counts__{unit_type_id}'.format(unit_type_id=unit_type_id)
```
barnabycourt: nice!

server/pulp/plugins/conduits/repo_publish.py
asmacdo: Yes, but I think that it should be in the distributor controller when that gets written.

nodes/parent/pulp_node/distributors/http/distributor.py
asmacdo: https://pulp.plan.io/issues/1067

server/pulp/server/db/migrations/0004_content_unit_counts.py
```diff
@@ -22,6 +22,6 @@ def migrate(*args, **kwargs):
     will be that the 'content_unit_counts' attribute does not yet exist, but
     this migration is idempotent just in case.
     """
-    RepoManager().rebuild_content_unit_counts()
-    repo_collection = Repo.get_collection()
+    managers.RepoManager().rebuild_content_unit_counts()
+    repo_collection = get_collection('repos')
```
barnabycourt: import migrations_lib and then migrations_lib.get_collection()

server/pulp/server/db/querysets.py
```diff
@@ -84,3 +90,26 @@ def update_one(self, *args, **kwargs):
...
         kwargs['multi'] = False
         self.update(*args, **kwargs)
+
+
+class RepoQuerySet(CriteriaQuerySet):
+    """
+    Custom queryset for repositories.
+    """
+
+    def get_repo_or_missing_resource(self, repo_id):
```
barnabycourt: Seems odd to add a whole method like this  as opposed to teaching the pulp middleware about the DoesNotExist exception.

server/pulp/server/managers/repo/group/cud.py
```diff
@@ -35,9 +34,8 @@ def create_repo_group(group_id, display_name=None, description=None, repo_ids=No
         """
         if repo_ids:
             # Check if ids in repo_ids belong to existing repositories
-            repo_query_manager = manager_factory.repo_query_manager()
             for repo_id in repo_ids:
-                repo_query_manager.get_repository(repo_id)
+                model.Repository.objects.get_repo_or_missing_resource(repo_id)
```
barnabycourt: You could get all repos by id using one query and count the number returned.

server/pulp/server/webservices/views/consumer_groups.py
```diff
@@ -382,13 +383,12 @@ def verify_group_resources(group_id, repo_id, distributor_id):
...
     missing_resources = {}
     group_manager = factory.consumer_group_query_manager()
-    repo_manager = factory.repo_query_manager()
     distributor_manager = factory.repo_distributor_manager()
     try:
         group_manager.get_group(group_id)
     except pulp_exceptions.MissingResource:
         missing_resources['group_id'] = group_id
-    repo = repo_manager.find_by_id(repo_id)
+    repo = model.Repository.objects(repo_id).first()
```
barnabycourt: repo = model.Repository.objects.get(repo_id)?

server/pulp/server/webservices/views/repositories.py
```diff
@@ -60,37 +56,17 @@ def _merge_related_objects(name, manager, repos):
...
-
-
-def _process_repos(repos, details, importers, distributors):
-    """
     Apply standard processing to a collection of repositories being returned to a client. Adds
     the object link and optionally adds related importers and distributors.
 
-    :param repos: collection of repositories
+    :param repo_objs: collection of repository objects
+    :type  repo_objs: list or tuple of model.Repository objects
```
barnabycourt: fully qualify pulp.server.db.model.Repository

server/pulp/server/webservices/views/repositories.py
```diff
@@ -165,27 +126,20 @@ def post(self, request):
...
+            display_name=repo_data.get('display_name', repo_id),
+            description=repo_data.get('description'),
+            notes=repo_data.get('notes'),
+            importer_type_id=repo_data.get('importer_type_id'),
+            importer_repo_plugin_config=repo_data.get('importer_config'),
+            distributor_list=repo_data.get('distributors')
+        )
+
+        repo = serializers.Repository(repo_obj).data
         response = generate_json_response_with_pulp_encoder(repo)
```
barnabycourt: Whatever this is doing should be encompassed in the Repository Serializer

server/pulp/server/webservices/views/repositories.py
```diff
@@ -200,7 +154,7 @@ def get(self, request, repo_id):
         """
         Looks for query parameters 'importers' and 'distributors', and will add
         the corresponding fields to the repository returned. Query parameter
-        'details' is equivalent to passing both 'importers' and 'distributors'.
+        'details' is equivalent to passing both 'importers.' and 'distributors'.
```
barnabycourt: 'importers'

server/pulp/server/webservices/views/repositories.py
```diff
@@ -1036,82 +973,110 @@ def post(self, request):
...
+                dateutils.parse_iso8601_datetime(end_date)
+            except (ValueError, isodate.ISO8601Error):
+                invalid_values.append('end_date')
 
-class RepoPublishHistory(View):
-    """
-    View for publish history of a specified repository.
-    """
+        if invalid_values:
+            raise pulp_exceptions.InvalidValue(invalid_values)
```
barnabycourt: How about more specific Coded Validation Exceptions?

server/pulp/server/webservices/views/serializers/__init__.py
```diff
@@ -149,6 +154,86 @@ def to_representation(self, instance):
...
+class ModelSerializer(BaseSerializer):
+    """
+    Base serializer for Mongoengine Documents.
+    """
+
+    def to_representation(self, instance):
+        """
+        Transforms a Mongoengine Document into a serialized dictionary.
+        """
+        document_dict = {}
```
barnabycourt: Should this translate the iso8601 stringfield from datetime to the string representation?

asmacdo: I just filed https://pulp.plan.io/issues/1068 in response to this.

The date time conversion is currently being done in `generate_json_respone_with_pulp_encoder`, which makes use of a custom json encoder. If we were to add this functionality to the serializer, we could not stop using our custom json encoder because at that point, the repo dict may have had importers and distributors injected into them (and their non-serialized date times).

I think it would make more sense to first switch to the use a single type of date time (see 1068 for more details), add this conversion to the serializer, and remove the use of pulp's custom json encoder all at once. Otherwise, we are implementing a new way of doing the same thing (pulp will attempt to do it twice) and hope that we get all the uses organically, then remove it. I suspect that in that case, the other date serializers would never be removed and we would just end up with one extra bit of complication.

server/pulp/server/controllers/repository.py
asmacdo: This is a new behavior (non dict would have been a 500), so I could raise a different error. `InvalidValue` is currently what the rest of the API and manger/controller code raises in this situation, what would you suggest?

server/pulp/server/db/querysets.py
asmacdo: The use of `MissingResource` is very inconsistent over the code base, but it does take args and kwargs which means that we can use it this way: `MissingResource(repository=repo_id)` 

If we were to catch `DoesNotExist` in the middleware, we would be notified that the Repository doesn't exist, but without reference to the data that the user expected to work. I am not necessarily opposed to doing it that way since it seems like the user *should* know what data they just sent. Also, it would be simpler and more consistent around the API, but it is in the gray area of backwards compatibility. 

If we were to do it that way, this is how our use would change:

``` python
>>> # Current/my implementation
>>> model.Repository.objects.get_repo_or_missing_resource('fake_repo')
Missing resource(s): repository=fake_repo

>>> # If the middleware were to catch DoesNotExist
>>> model.Repository.objects.get(repo_id='fake_repo')
DoesNotExist: Repository matching query does not exist.
```
To me, the main problem with backwards compatibility is that the MissingResource exception has a field `MissingResource.resources` which is directly returned via the API:

``` json
{
    "_href": "/pulp/api/v2/repositories/fake/", 
    "error": {
        "code": "PLP0009", 
        "data": {
            "resources": {
                "repository": "fake"
            }
        }, 
        "description": "Missing resource(s): repository=fake", 
        "sub_errors": []
    }, 
    "error_message": "Missing resource(s): repository=fake", 
    "exception": null, 
    "http_request_method": "GET", 
    "http_status": 404, 
    "resources": {
        "repository": "fake"
    }, 
    "traceback": null
}
```

server/pulp/server/managers/repo/group/cud.py
asmacdo: Nice catch.

Rather than push the code and hope that you happen to notice it, I'll post my proposed change here and get feedback first.

``` python
# Check if ids in repo_ids belong to existing repositories
existing_repos = model.Repository.objects(repo_id__in=repo_ids)
if len(existing_repos) != len(repo_ids):
    existing_repo_ids = set([repo.repo_id for repo in existing_repos])
    non_existing_repo_ids = list(set(repo_ids) - existing_repo_ids)
    raise pulp_exceptions.MissingResource(repositories=non_existing_repo_ids)
```

server/pulp/server/webservices/views/consumer_groups.py
asmacdo: This is one of a few places I have seen recently that use this pattern; rather than raising the first exception, it continues on so that a list of all invalid params can be raised together. I know this isn't your favorite pattern, but I don't think it would be appropriate to change it as a part of this story.

In order to preserve this, I thought it would be easier to read if I used `.first()` to allow `repo` to be `None` rather than using a try-except block.

server/pulp/server/webservices/views/repositories.py
asmacdo: I may have a misunderstanding here. I thought it would be better to use `model.Repository` since this module imports `model`. This means that if we were to move where `model` lives, we would not have to update all the doc blocks. I believe that sphinx should know how to create this link. 

Either way, lets add this to our style guide!

asmacdo: Got it, you are correct. I'll make the change throughout and add it to our style guide.

server/pulp/server/db/migrations/0004_content_unit_counts.py
asmacdo: I am not sure what you mean

server/pulp/server/webservices/views/repositories.py
asmacdo: Please see  https://github.com/pulp/pulp/pull/1931#discussion_r32776564

barnabycourt:
After setting up the latest version and running the migrations I get an error from pulp_rpm:
File ".../pulp_rpm/plugins/pulp_rpm/plugins/migrations/0004_pkg_group_category_repoid.py", line 9, in <module>
    from pulp.server.managers.repo.cud import RepoContentUnit

Is there an additional PR for pulp_rpm to fix this issue?

barnabycourt:
ok test

barnabycourt:
Repo group creation fails:
$ pulp-admin repo group create --group-id parks
An internal error occurred on the Pulp server:

RequestException: POST request
on /pulp/api/v2/repo_groups/ failed with 500 - 'NoneType' object is not iterable

barnabycourt:
Should repo_id be included in the new repo object along with a deprecation notice for id field in the documentation?

barnabycourt:
I don't know if this is new or not but a repo group export fails if one of the repos has no content

asmacdo:
ok test

asmacdo:
@barnabycourt, I have rebased to include the upgraded deps from https://github.com/pulp/pulp/pull/1956

asmacdo:
ok test

asmacdo:
ok test

barnabycourt:
ok test

barnabycourt:
ok test

asmacdo:
Go pulpbot go!

PR #1934, opened 2015-06-18
Set storage path in ContentUnit.pre_save handler correctly.

bmbouter:
The pre-save handler did not save the full path to the
file_path set with set_content in the case where
file_path is a file.

re #862


barnabycourt:
This really didn't require a unit test update? How did I miss that?

PR #1936, opened 2015-06-18
Provision a virtualenv for Crane if present.

rbarlow:
Our Vagrant provisioning didn't do anything for Crane. This commit
creates basic support for creating a virtualenv and installing Crane's
dependencies.

In order to avoid copying the RPM dependency information, I determined
how to query Crane's spec file for the RPMs that Crane depends on.
Since this line was also useful for our other projects, I converted our
dependencies to all be installed this way. By
doing this, I also solved #990 and so this commit also fixes that issue.

https://pulp.plan.io/issues/990

fixes #990

### DIFF
--- a/playpen/vagrant-setup.sh
+++ b/playpen/vagrant-setup.sh
@@ -64,10 +64,7 @@ fi
 
 # install rpms, then remove pulp*
 echo "installing RPMs"
-sudo dnf install -y @pulp-server-qpid @pulp-admin @pulp-consumer
-sudo dnf remove --setopt=clean_requirements_on_remove=false -y pulp-\* python-pulp\*
-sudo dnf install -y git mongodb mongodb-server python-debian python-django python-glanceclient \
-                    python-keystoneclient python-mongoengine python-paste python-qpid-qmf \
+sudo dnf install -y git mongodb mongodb-server python-gofer-qpid python-qpid python-qpid-qmf \
                     python-setuptools python-sphinx qpid-cpp-server qpid-cpp-server-store
 
 # disable mongo journaling since this is a dev setup
@@ -88,6 +85,7 @@ for r in {pulp,pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_
     ! mkvirtualenv --system-site-packages $r
     workon $r
     setvirtualenvproject
+    sudo dnf install -y $(rpmspec -q --queryformat '[%{REQUIRENAME}\n]' *.spec | grep -v "/.*" | grep -v "python-pulp.* " | grep -v "pulp.*" | uniq)
     # Install dependencies for automated tests
     pip install -r test_requirements.txt
     sudo python ./pulp-dev.py -I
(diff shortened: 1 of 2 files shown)


playpen/vagrant-setup.sh
```diff
@@ -64,10 +64,7 @@ fi
 
 # install rpms, then remove pulp*
 echo "installing RPMs"
-sudo dnf install -y @pulp-server-qpid @pulp-admin @pulp-consumer
-sudo dnf remove --setopt=clean_requirements_on_remove=false -y pulp-\* python-pulp\*
-sudo dnf install -y git mongodb mongodb-server python-debian python-django python-glanceclient \
-                    python-keystoneclient python-mongoengine python-paste python-qpid-qmf \
+sudo dnf install -y git mongodb mongodb-server python-gofer-qpid python-qpid python-qpid-qmf \
```
jeremycline: Does this need to include `qpid-tools` or is that part of a spec? Or do we not need that?

rbarlow: On 06/19/2015 12:11 PM, Jeremy Cline wrote:
> Does this need to include |qpid-tools| or is that part of a spec? Or do
> we not need that?

It seems to get pulled in automatically by one of the above (not sure
which). qpid-tools is present after vagrant up.

-- 
Randy Barlow

playpen/vagrant-setup.sh
```diff
@@ -96,8 +94,38 @@ for r in {pulp,pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_
...
+if [ -d crane ]; then
+    echo "installing crane's environment"
+    pushd crane
+    ! mkvirtualenv --system-site-packages crane
+    workon crane
+    setvirtualenvproject
+    # Install dependencies
+    sudo dnf install -y $(rpmspec -q --queryformat '[%{REQUIRENAME}\n]' python-crane.spec | grep -v "/.*" | uniq)
+    pip install -r test-requirements.txt
+    python setup.py test
```
jeremycline: The tests seem to expect that `/var/lib/crane/metadata/` exists:
```
==> dev: ERROR:root:The data directory specified does not exist: /var/lib/crane/metadata/
```

jeremycline:
Looks good to me after the setup creates the  `/var/lib/crane/metadata/` directory.

rbarlow:
On 06/19/2015 03:42 PM, Jeremy Cline wrote:
> Looks good to me after the setup creates the |/var/lib/crane/metadata/|
> directory.

Because I have just opened another pull request to stop vagrant from
running tests, I think I will just stop running the tests here for now
instead.

IMO, the tests themselves shouldn't mess with real paths, so if the
tests are assuming /var/lib/crane exists and fail if not, I think we
need to use mock to fix that.

-- 
Randy Barlow

jeremycline:
Agreed about the tests (I think there's a task to fix the tests), and your solution seems reasonable to me. Merge away!

rbarlow:
Thanks @jeremycline!

PR #1938, opened 2015-06-19
Remove resolve_dependecies from pulp.

ipanova:
closes #1022
https://pulp.plan.io/issues/1022


ipanova:
this also needs to be merged https://github.com/pulp/pulp_rpm/pull/699

docs/dev-guide/integration/rest-api/content/associate.rst
```diff
@@ -40,7 +40,6 @@ with an importer that supports the type of units being copied.
       }
     },
    'override_config': {
-     'resolve_dependencies: true,
```
rbarlow: Does this mean we had documented the code we are removing? If so, how can we be sure that removing this API is OK without waiting until 3.0?

rbarlow:
@ipanova that one documentation thing concerns me a little bit. Are we sure that we had never exposed this code for our API users?

docs/dev-guide/integration/rest-api/content/associate.rst
ipanova: The API endpoint resolve_dependencies was already removed in some other PR. It was unused, undocumented and how this line got into the doc example i have no idea. You can check this issue for more info https://pulp.plan.io/issues/812 
In this PR i remove the remainings of resolve_dependencies like manager and tests.

ipanova: I also checked in 2.6release branch were this endpoint was still present and as you can see from the paste it did not solve dependencies for package lion http://fpaste.org/236485/52269011/

Anyway i think we are safe to remove this dead and unused code.

rbarlow:
We had an IRC conversation to discuss this a bit, so that I could verify that the API really wasn't used. This API required the plugins to implement a method to resolve the dependencies, and none of our plugins had implemented that method (they all raised NotImplemented). Therefore we can safely know that this API was never used.

LGTM!

PR #1939, opened 2015-06-19
Add a ptests command to the dev environment.

rbarlow:
This commit adds a new ptests bash function to the development
environment. This function loops over all the Pulp projects that you
have installed and runs each test suite.

This commit also removes the tests from the initial provisioning so that
we can save some time on vagrant up.


jeremycline:
This is also nice because if the tests fail, Vagrant gets upset and assumes the provisioning failed. Looks good to me!

PR #1940, opened 2015-06-19
Pulp with multiple celeryBeat instances

shubham90:
https://pulp/plan.io/issues/1060

closes #1060

### DIFF
--- a/common/pulp/common/constants.py
+++ b/common/pulp/common/constants.py
@@ -63,4 +63,7 @@ PRIMARY_ID = '___/primary/___'
 # this is used by both platform and plugins to find the default CA path
 DEFAULT_CA_PATH = '/etc/pki/tls/certs/ca-bundle.crt'
 
+# CeleryBeat constants
 SCHEDULER_WORKER_NAME = "scheduler"
+CELERYBEAT_WAIT_SECONDS = 200
+TICK_SECONDS = 90
--- a/docs/user-guide/installation.rst
+++ b/docs/user-guide/installation.rst
@@ -279,17 +279,12 @@ Server
       ``PULP_CONCURRENCY - 1``. For example, you can use ``sudo systemctl status pulp_worker-1`` to
       see how the second worker is doing.
 
-#. There are two more services that need to be running, but it is important that these two only run
-   once each (i.e., do not enable either of these on any more than one Pulp server).
-
-   .. warning::
-
-      ``pulp_celerybeat`` and ``pulp_resource_manager`` must both be singletons, so be sure that you
-      only enable each of these on one host if you are Pulp's clustered deployment.
+#. There are two more services that need to be running.
 
    On some Pulp system, configure, start and enable the Celerybeat process. This process performs a
    job similar to a cron daemon for Pulp. Edit ``/etc/default/pulp_celerybeat`` to your liking, and
-   then enable and start it. Again, do not enable this on more than one host. For Upstart::
+   then enable and start it. Multiple instances of ``pulp_celerybeat`` may run concurrently, which
+   will make the Pulp installation more failure tolerant. For Upstart::
 
       $ sudo chkconfig pulp_celerybeat on
       $ sudo service pulp_celerybeat start
@@ -298,6 +293,11 @@ Server
 
       $ sudo systemctl enable pulp_celerybeat
       $ sudo systemctl start pulp_celerybeat
+
+   .. warning::
+
+      ``pulp_resource_manager`` must be singleton, so be sure that you
+      only enable this on one host if you are Pulp's clustered deployment.
 
    Lastly, one ``pulp_resource_manager`` process must be running in the installation. This process
    acts as a task router, deciding which worker should perform certain types of tasks. Apologies
--- a/docs/user-guide/release-notes/master.rst
+++ b/docs/user-guide/release-notes/master.rst
@@ -8,6 +8,8 @@ Pulp master
 New Features
 ------------
 
+* Multiple instances of ``pulp_celerybeat`` can now run simultaneously.
+  If one of them goes down, another instance will dispatch scheduled tasks as usual.
 
 Deprecation
 -----------
--- a/docs/user-guide/scaling.rst
+++ b/docs/user-guide/scaling.rst
@@ -31,8 +31,9 @@ Pulp consists of several components:
   Additionally, it performs task cancellations in the event of a worker
   shutdown or failure. The celerybeat process also initiates scheduled tasks,
   and automatically cancels tasks that have failed more than *X* times. This
-  process also initiates periodic jobs that Pulp runs internally. In a Pulp
-  cluster, exactly one of these should be running!
+  process also initiates periodic jobs that Pulp runs internally. Multiple
+  instances of the ``pulp_celerybeat`` process can run at the same time to
+  ensure high availability of the Pulp.
 
 * ``pulp_resource_manager`` - The resource manager assigns tasks to workers,
   and ensures multiple conflicting tasks on a repo are not executed at the same
@@ -45,9 +46,8 @@ Additionally, Pulp relies on other components:
 * `Apache Qpid`_ or `RabbitMQ`_ - the queuing system that Pulp uses to assign
   work to workers. Pulp can operate equally well with either Qpid or RabbitMQ.
 
-.. warning:: It is critical to note that ``pulp_celerybeat`` and
-    ``pulp_resource_manager`` should *never* have more than a single instance
-    running under any circumstance!
+.. warning:: It is critical to note that ``pulp_resource_manager`` should
+   *never* have more than a single instance running under any circumstance!
 
 The diagram below shows an example default deployment.
 
@@ -132,8 +132,7 @@ httpd workers are.
 
 To add additional httpd server capacity, configure the desired number of
 `Pulp clustered servers` and start ``httpd`` on them. Remember only one
-instance of ``pulp_celerybeat`` and ``pulp_resource_manager`` should be
-running across all `Pulp clustered servers`.
+instance of ``pulp_resource_manager`` should be running across all `Pulp clustered servers`.
 
 
 Scaling workers
@@ -144,9 +143,8 @@ and redundancy.
 
 To add additional Pulp worker capacity, configure the desired number of `Pulp
 clustered servers` according to the the `clustering`_ docs and start
-``pulp_workers`` on each of them. Remember only one instance of
-``pulp_celerybeat`` and ``pulp_resource_manager`` should be running across
-all `Pulp clustered servers`.
+``pulp_workers`` on each of them. Remember only one instance of ``pulp_resource_manager``
+should be running across all `Pulp clustered servers`.
 
 
 .. _clustering:
--- a/docs/user-guide/server.rst
+++ b/docs/user-guide/server.rst
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
   available workers. If a worker with the same name is started again after being missing, it is
   added into the pool of workers as any worker starting up normally would.
 
-* If ``pulp_celerybeat`` dies, and in case then new workers start, they won't be given work. If
-  existing workers stop, Pulp will continue assigning them work. Once restarted, pulp_celerybeat
-  will synchronize with the current state of all workers. Scheduled tasks will not run while
-  pulp_celerybeat is down, but they will instead run when celerybeat is restarted.
+* If all instances of ``pulp_celerybeat`` die and new workers start, they won't
+  be given work or if existing workers stop, Pulp will continue assigning them work incorrectly.
+  Once restarted, ``pulp_celerybeat`` will synchronize with the current state of all workers.
+  Scheduled tasks will not run if there are no ``pulp_celerybeat`` processes running, but
+  they will run when the first ``pulp_celerybeat`` process is restarted.
 
 * If ``pulp_resource_manager`` dies, the Pulp tasking system will halt. Once restarted it will
   resume.
@@ -97,9 +98,9 @@ The service name is ``pulp_workers``.
 Celery Beat
 ^^^^^^^^^^^
 
-This is a singleton (there must only be one celery beat process per pulp deployment)
-that is responsible for queueing scheduled tasks. It also plays a role in
-monitoring the availability of workers.
+This process is responsible for queueing scheduled tasks and is responsible for
+monitoring the availability of workers. For fault tolerance, there can be multiple
+instances of this process running. If one of them fails, another will take over.
 
 The service name is ``pulp_celerybeat``.
 
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -9,15 +9,16 @@ import time
 
 from celery import beat
 from celery.result import AsyncResult
+import mongoengine
 
-from pulp.common.constants import SCHEDULER_WORKER_NAME
+from pulp.common.constants import SCHEDULER_WORKER_NAME, CELERYBEAT_WAIT_SECONDS, TICK_SECONDS
 from pulp.server.async import worker_watcher
 from pulp.server.async.celery_instance import celery as app
 from pulp.server.async.tasks import _delete_worker
 from pulp.server.db import connection as db_connection
 from pulp.server.db.connection import retry_decorator
 from pulp.server.db.model.dispatch import ScheduledCall, ScheduleEntry
-from pulp.server.db.model import Worker
+from pulp.server.db.model import Worker, CeleryBeatLock
 from pulp.server.managers.schedule import utils
 
 # The import below is not used in this module, but it needs to be kept here. This module is the
@@ -305,6 +306,13 @@ class Scheduler(beat.Scheduler):
         worker_timeout_monitor.daemon = True
         worker_timeout_monitor.start()
 
+    @staticmethod
+    def call_tick(self, celerybeat_name):
+        ret = super(Scheduler, self).tick()
+        _logger.debug(_("%(celerybeat_name)s will tick again in %(ret)s secs")
+                      % {'ret': ret, 'celerybeat_name': celerybeat_name})
+        return ret
+
     def tick(self):
         """
         Superclass runs a tick, that is one iteration of the scheduler. Executes
@@ -317,15 +325,50 @@ class Scheduler(beat.Scheduler):
         :return:    number of seconds before the next tick should run
         :rtype:     float
         """
-        ret = super(Scheduler, self).tick()
         self._failure_watcher.trim()
+
+        # Setting the celerybeat name
+        celerybeat_name = SCHEDULER_WORKER_NAME + "@" + platform.node()
 
         # this is not an event that gets sent anywhere. We process it
         # immediately.
         scheduler_event = {'timestamp': time.time(),
                            'type': 'scheduler-event',
-                           'hostname': ("%s@%s" % (SCHEDULER_WORKER_NAME, platform.node()))}
+                           'hostname': celerybeat_name}
         worker_watcher.handle_worker_heartbeat(scheduler_event)
+
+        old_timestamp = datetime.utcnow() - timedelta(seconds=CELERYBEAT_WAIT_SECONDS)
+
+        # Updating the current lock if lock is on this instance of celerybeat
+        result = CeleryBeatLock.objects(celerybeat_name=celerybeat_name).\
+            update(set__timestamp=datetime.utcnow())
+
+        # If current instance has lock and updated lock_timestamp, call super
+        if result == 1:
+            _logger.debug(_('Lock updated by %(celerybeat_name)s')
+                          % {'celerybeat_name': celerybeat_name})
+            ret = self.call_tick(self, celerybeat_name)
+        else:
+            # check for old enough time_stamp and remove if such lock is present
+            CeleryBeatLock.objects(timestamp__lte=old_timestamp).delete()
+            try:
+                lock_timestamp = datetime.utcnow()
+
+                # Insert new lock entry
+                new_lock = CeleryBeatLock(celerybeat_name=celerybeat_name,
+                                          timestamp=lock_timestamp)
+                new_lock.save()
+                _logger.info(_("New lock acquired by %(celerybeat_name)s") %
+                             {'celerybeat_name': celerybeat_name})
+                # After acquiring new lock call super to dispatch tasks
+                ret = self.call_tick(self, celerybeat_name)
+
+            except mongoengine.NotUniqueError:
+                # Setting a default wait time for celerybeat instances with no lock
+                ret = TICK_SECONDS
+                _logger.info(_("Duplicate or new celerybeat Instance, "
+                               "ticking again in %(ret)s seconds.")
+                             % {'ret': ret})
         return ret
 
     def setup_schedule(self):
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -122,6 +122,7 @@ def ensure_database_indexes():
     model.ReservedResource.ensure_indexes()
     model.TaskStatus.ensure_indexes()
     model.Worker.ensure_indexes()
+    model.CeleryBeatLock.ensure_indexes()
 
     # Load all the model classes that the server knows about and ensure their inexes as well
     plugin_manager = PluginManager()
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -534,3 +534,24 @@ class ContentUnit(Document):
         type and the same unit key will get the same hash value.
         """
         return hash(self.unit_type_id + self.unit_key_str)
+
+
+class CeleryBeatLock(Document):
+    """
+    Single document collection which gives information about the current celerybeat lock.
+
+    :ivar celerybeat_name: string representing the celerybeat instance name
+    :type celerybeat_name: basestring
+    :ivar timestamp: The timestamp(UTC) at which lock is acquired
+    :type timestamp: datetime.datetime
+    :ivar lock: A unique key set to "locked" when lock is acquired.
+    :type lock: basestring
+    :ivar _ns: (Deprecated), Contains the name of the collection this model represents
+    :type _ns: mongoengine.StringField
+    """
+    celerybeat_name = StringField(required=True)
+    timestamp = DateTimeField(required=True)
+    lock = StringField(required=True, default="locked", unique=True)
+
+    # For backward compatibility
+    _ns = StringField(default='celery_beat_lock')
--- a/server/test/unit/server/async/test_scheduler.py
+++ b/server/test/unit/server/async/test_scheduler.py
@@ -1,9 +1,11 @@
 from datetime import datetime
 import time
 import unittest
+import platform
 
 from celery.beat import ScheduleEntry
 import mock
+from mongoengine import NotUniqueError
 
 from pulp.server.async import scheduler
 from pulp.server.async.celery_instance import celery as app
@@ -255,7 +257,9 @@ class TestSchedulerSpawnPulpMonitorThreads(unittest.TestCase):
 class TestSchedulerTick(unittest.TestCase):
     @mock.patch('celery.beat.Scheduler.__init__', new=mock.Mock())
     @mock.patch('celery.beat.Scheduler.tick')
-    def test_calls_superclass(self, mock_tick):
+    @mock.patch('pulp.server.async.scheduler.worker_watcher')
+    @mock.patch('pulp.server.async.scheduler.CeleryBeatLock')
+    def test_calls_superclass(self, mock_celerybeatlock, mock_worker_watcher, mock_tick):
         sched_instance = scheduler.Scheduler()
 
         sched_instance.tick()
@@ -265,7 +269,9 @@ class TestSchedulerTick(unittest.TestCase):
     @mock.patch('celery.beat.Scheduler.__init__', new=mock.Mock())
     @mock.patch.object(scheduler.FailureWatcher, 'trim')
     @mock.patch('celery.beat.Scheduler.tick')
-    def test_calls_trim(self, mock_tick, mock_trim):
+    @mock.patch('pulp.server.async.scheduler.worker_watcher')
+    @mock.patch('pulp.server.async.scheduler.CeleryBeatLock')
+    def test_calls_trim(self, mock_celerybeatlock, mock_worker_watcher, mock_tick, mock_trim):
         sched_instance = scheduler.Scheduler()
 
         sched_instance.tick()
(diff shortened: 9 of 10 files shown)


docs/user-guide/installation.rst
```diff
@@ -279,17 +279,12 @@ Server
...
-   .. warning::
-
-      ``pulp_celerybeat`` and ``pulp_resource_manager`` must both be singletons, so be sure that you
-      only enable each of these on one host if you are Pulp's clustered deployment.
+#. There are two more services that need to be running.
 
    On some Pulp system, configure, start and enable the Celerybeat process. This process performs a
    job similar to a cron daemon for Pulp. Edit ``/etc/default/pulp_celerybeat`` to your liking, and
-   then enable and start it. Again, do not enable this on more than one host. For Upstart::
+   then enable and start it. Multiple instances of this may be enabled, which will make the system
```
bmbouter: s/this/pulp_celerybeat/
s/enabled/run concurrently/
s/the system/the Pulp installation/

docs/user-guide/release-notes/master.rst
```diff
@@ -8,6 +8,8 @@ Pulp master
 New Features
 ------------
 
+* Multiple instances of celery beat can now run simultaneously in pulp. If one of them goes down,
```
bmbouter: s/celery beat/pulp_celerybeat/
s/ in pulp//

docs/user-guide/scaling.rst
```diff
@@ -31,8 +31,9 @@ Pulp consists of several components:
   Additionally, it performs task cancellations in the event of a worker
   shutdown or failure. The celerybeat process also initiates scheduled tasks,
   and automatically cancels tasks that have failed more than *X* times. This
-  process also initiates periodic jobs that Pulp runs internally. In a Pulp
-  cluster, exactly one of these should be running!
+  process also initiates periodic jobs that Pulp runs internally. Multiple
+  instances of pulp_celerybeat process can be started together to ensure high
```
bmbouter: s/of pulp_celerybeat/of the pulp_celerybeat/

docs/user-guide/scaling.rst
```diff
@@ -31,8 +31,9 @@ Pulp consists of several components:
...
+  availability of the system.
```
bmbouter: s/the system/Pulp/

docs/user-guide/scaling.rst
```diff
@@ -45,8 +46,7 @@ Additionally, Pulp relies on other components:
 * `Apache Qpid`_ or `RabbitMQ`_ - the queuing system that Pulp uses to assign
   work to workers. Pulp can operate equally well with either Qpid or RabbitMQ.
 
-.. warning:: It is critical to note that ``pulp_celerybeat`` and
-    ``pulp_resource_manager`` should *never* have more than a single instance
+.. warning:: It is critical to note that ``pulp_resource_manager`` should *never* have more than a single instance
```
bmbouter: This line needs to be shorter. I usually limit docs to 70ish lines or less.

bmbouter:
Everywhere that you refer in the docs to pulp_celerybeat, ensure to wrap it in two backticks to rst escape the term.

```
``pulp_celerybeat``
```

docs/user-guide/server.rst
```diff
@@ -30,7 +30,7 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
   available workers. If a worker with the same name is started again after being missing, it is
   added into the pool of workers as any worker starting up normally would.
 
-* If ``pulp_celerybeat`` dies, and in case then new workers start, they won't be given work. If
+* If all instances of ``pulp_celerybeat`` dies, and in case then new workers start, they won't be given work. If
```
bmbouter: shorten this line.

bmbouter: s/dies/die/
s/ in case then//

docs/user-guide/server.rst
```diff
@@ -97,9 +97,9 @@ The service name is ``pulp_workers``.
 Celery Beat
 ^^^^^^^^^^^
 
-This is a singleton (there must only be one celery beat process per pulp deployment)
-that is responsible for queueing scheduled tasks. It also plays a role in
-monitoring the availability of workers.
+This process is responsible for queueing scheduled tasks. It also plays a role in
+monitoring the availability of workers. There can be multiple instances of this process running.
+If one of them fails, the other will take over the work.
```
bmbouter: s/the other will take over the work/another will take over/

docs/user-guide/server.rst
```diff
@@ -97,9 +97,9 @@ The service name is ``pulp_workers``.
 Celery Beat
 ^^^^^^^^^^^
 
-This is a singleton (there must only be one celery beat process per pulp deployment)
-that is responsible for queueing scheduled tasks. It also plays a role in
-monitoring the availability of workers.
+This process is responsible for queueing scheduled tasks. It also plays a role in
+monitoring the availability of workers. There can be multiple instances of this process running.
```
bmbouter: s/There/For fault tolerance, there/

server/pulp/server/async/scheduler.py
```diff
@@ -27,6 +29,7 @@
 
 
 _logger = logging.getLogger(__name__)
+celery_name = uuid.uuid4()
```
bmbouter: :+1:

server/pulp/server/async/scheduler.py
```diff
@@ -425,4 +473,5 @@ def apply_async(self, entry, publisher=None, **kwargs):
             has_failure = bool(entry._scheduled_call.consecutive_failures)
             self._failure_watcher.add(result.id, entry.name, has_failure)
             _logger.debug(_('watching task %s') % {'id': result.id})
+            _logger.info(_('watching task %s') % {'id': result.id})
```
bmbouter: This should be logged at either `debug` or `info` but not both. I'll suggest doing it at `info` so keep this line and delete the one above it.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+        # Updating the current lock if lock is on this instance of celerybeat
+        result = collection.find_and_modify(
+            query={'celerybeat_name': celerybeat_name},
+            update={'$set': {'lock_timestamp': datetime.utcnow()}}
+        )
+
+        # If current instance has lock and updated lock_timestamp, call super
+        if result is not None:
+            ret = super(Scheduler, self).tick()
+            _logger.info("Lock updated, Ticking again in %s secs "
```
bmbouter: This should be internationalized with _() and you need to switch the logging to dictionary based (not position based).

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+            # check for old enough time_stamp and remove if such lock is present
+            collection.remove({'lock_timestamp': {'$lte': old_timestamp}})
+            try:
+                # Insert new lock entry
+                collection.insert({
+                    'celerybeat_name': celerybeat_name,
+                    'lock_timestamp': datetime.utcnow(),
+                    'lock': "locked"
+                })
+                _logger.info("New lock acquired by %s ", celerybeat_name)
```
bmbouter: This should be internationalized with _() and you need to switch the logging to dictionary based (not position based).

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+                # After acquiring new lock call super to dispatch tasks
+                ret = super(Scheduler, self).tick()
+                _logger.info("Ticking again in %s secs and "
```
bmbouter: This should be internationalized with _() and you need to switch the logging to dictionary based (not position based).

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+                             "celerybeat_name is %s ", ret, celerybeat_name)
+
+            except pymongo.errors.DuplicateKeyError:
+                # Setting a default wait time for celerybeat instances with no lock
+                ret = DEFAULT_TICK_WAIT
+                _logger.info("Duplicate or new celerybeat Instance, ticking again in 90 seconds.")
```
bmbouter: This should be internationalized with _() and you need to switch the logging to dictionary based (not position based).

bmbouter:
A lot of these files have had their permissions changed and show `100644 → 100755` at the top. This needs to not change their file permissions.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+        if result is not None:
+            ret = super(Scheduler, self).tick()
+            _logger.info("Lock updated, Ticking again in %s secs "
+                         "and celerybeat_name is %s ", ret, celerybeat_name)
+        else:
+            # check for old enough time_stamp and remove if such lock is present
+            collection.remove({'lock_timestamp': {'$lte': old_timestamp}})
+            try:
+                # Insert new lock entry
+                collection.insert({
```
bmbouter: This should use the model definition and use `save()` instead.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +489,26 @@ def __init__(self, call_request, call_report):
         self.timestamp = dateutils.now_utc_timestamp()
         self.call_request_string = str(call_request)
         self.serialized_call_report = call_report.serialize()
+
+
+class CeleryBeatLock(Model):
```
bmbouter: Convert this to inherit from `Document`. Document should come from `mongoengine`. Grep the codebase for existing objects that inherit from `Document` as examples.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
         worker_watcher.handle_worker_heartbeat(scheduler_event)
+
+        celerybeat_name = SCHEDULER_WORKER_NAME + "@" + str(celery_name) + "@" + platform.node()
+
+        # Establishing connection with DB
+        db = db_connection.get_database()
+        collection = db['celerybeat_lock']
+
+        # Creating a unique constraint on the collection
+        collection.create_index('lock', unique=True)
```
bmbouter: Don't do this here. Instead you should add your CeleryBeatLock object to the list of things that pulp-manage-db will call. Do that here:  https://github.com/pulp/pulp/blob/7a1703ceb1913c757c1f91caabd3ed44d458a7c5/server/pulp/server/db/manage.py#L123

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
         # this is not an event that gets sent anywhere. We process it
         # immediately.
         scheduler_event = {'timestamp': time.time(),
                            'type': 'scheduler-event',
-                           'hostname': ("%s@%s" % (SCHEDULER_WORKER_NAME, platform.node()))}
+                           'hostname': ("%s@%s@%s" % (SCHEDULER_WORKER_NAME,
+                                                      str(celery_name), platform.node()))}
         worker_watcher.handle_worker_heartbeat(scheduler_event)
+
+        celerybeat_name = SCHEDULER_WORKER_NAME + "@" + str(celery_name) + "@" + platform.node()
```
bmbouter: This is being calculated twice. once here in line 332 and once in lines 328-239. Compute it once above and use it in both places.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,60 @@ def tick(self):
...
+
+        # Establishing connection with DB
+        db = db_connection.get_database()
+        collection = db['celerybeat_lock']
+
+        # Creating a unique constraint on the collection
+        collection.create_index('lock', unique=True)
+
+        _logger.debug('Updating the lock with celerybeat_name %s ', celerybeat_name)
```
bmbouter: This should be internationalized with _() and you need to switch the logging to dictionary based (not position based).

server/pulp/server/async/scheduler.py
bmbouter: Have the first part of the sentence read with "Lock updated by <name>..." This will be shorter and more readable.

server/pulp/server/async/scheduler.py
bmbouter: Have the first part of the sentence read with "<name> will tick again in ..." This will be shorter and more readable.

server/test/unit/server/async/test_scheduler.py
```diff
@@ -282,6 +284,17 @@ def test_calls_handle_heartbeat(self, mock_heartbeat, mock_tick):
 
         mock_heartbeat.assert_called_once()
 
+    @mock.patch('celery.beat.Scheduler.__init__', new=mock.Mock())
+    @mock.patch('celery.beat.Scheduler.tick')
+    def test_heartbeat_lock(self, mock_tick):
```
bmbouter: There needs to be a test for each codepath you have added. There are at least 4 because there is an if/else and try/except so used together that is 4 codepaths. I recommend putting them all on the TestSchedulerTick class.

bmbouter:
@shubham90 I'm done for now I'll re-review after the changes are made.

server/pulp/server/async/scheduler.py
rbarlow: On 06/19/2015 03:40 PM, bmbouter wrote:
> This should be logged at either |debug| or |info| but not both. I'll
> suggest doing it at |info| so keep this line and delete the one above it.

IMO this one makes sense to stay at debug. I don't think it will be
meaningful to users, and it adds noise to the log for most users.

-- 
Randy Barlow

barnabycourt:
ok test

docs/user-guide/release-notes/master.rst
```diff
@@ -8,6 +8,8 @@ Pulp master
 New Features
 ------------
 
+* Multiple instances of ``pulp_celerybeat`` can now run simultaneously.
+  If one of them goes down, another instance will re-dispatch tasks as usual.
```
bmbouter: s/re-dispatch tasks/dispatch scheduled tasks/

common/pulp/common/constants.py
```diff
@@ -64,3 +64,6 @@
 DEFAULT_CA_PATH = '/etc/pki/tls/certs/ca-bundle.crt'
 
 SCHEDULER_WORKER_NAME = "scheduler"
+# CeleryBeat HA constants
+CELERYBEAT_OLD_TIMESTAMP = 200
+DEFAULT_TICK_WAIT = 90
```
bmbouter: consider changing this name to be more meaningful. I'll suggest `TICK_SECONDS`

common/pulp/common/constants.py
```diff
@@ -64,3 +64,6 @@
 DEFAULT_CA_PATH = '/etc/pki/tls/certs/ca-bundle.crt'
 
 SCHEDULER_WORKER_NAME = "scheduler"
+# CeleryBeat HA constants
```
bmbouter: This comment could be generalized some. Consider moving this comment above the line above and rewording it to be `# celerybeat constants`

docs/user-guide/scaling.rst
```diff
@@ -31,8 +31,9 @@ Pulp consists of several components:
   Additionally, it performs task cancellations in the event of a worker
   shutdown or failure. The celerybeat process also initiates scheduled tasks,
   and automatically cancels tasks that have failed more than *X* times. This
-  process also initiates periodic jobs that Pulp runs internally. In a Pulp
-  cluster, exactly one of these should be running!
+  process also initiates periodic jobs that Pulp runs internally. Multiple
+  instances of the ``pulp_celerybeat`` process can be started together to ensure
```
bmbouter: s/started together/run at the same time/

It's not important when they start, but that they can run together.

docs/user-guide/server.rst
```diff
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
   available workers. If a worker with the same name is started again after being missing, it is
   added into the pool of workers as any worker starting up normally would.
 
-* If ``pulp_celerybeat`` dies, and in case then new workers start, they won't be given work. If
-  existing workers stop, Pulp will continue assigning them work. Once restarted, pulp_celerybeat
-  will synchronize with the current state of all workers. Scheduled tasks will not run while
-  pulp_celerybeat is down, but they will instead run when celerybeat is restarted.
+* If all instances of ``pulp_celerybeat`` die, and new workers start, they won't
```
bmbouter: I don't think you need the comma after `die`

docs/user-guide/server.rst
```diff
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
...
+  be given work. If existing workers stop, Pulp will continue assigning them work. Once
```
bmbouter: consider rewriting this line to be:

`be given work, or if existing workers stop, Pulp will incorrectly continue assigning them work`

docs/user-guide/server.rst
```diff
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
...
+  restarted, ``pulp_celerybeat`` will synchronize with the current state of all workers.
+  Scheduled tasks will not run while ``pulp_celerybeat`` is down, but they will instead run
```
bmbouter:
```
while ``pulp_celerybeat`` is down
```
=>
```
if there are no ``pulp_celerybeat`` processes running
```

bmbouter: also
s/instead //

docs/user-guide/server.rst
```diff
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
...
+  when ``pulp_celerybeat`` is restarted.
```
bmbouter: s/when/when the first/
s/is restarted/process is restarted/

docs/user-guide/server.rst
```diff
@@ -97,9 +98,9 @@ The service name is ``pulp_workers``.
 Celery Beat
 ^^^^^^^^^^^
 
-This is a singleton (there must only be one celery beat process per pulp deployment)
-that is responsible for queueing scheduled tasks. It also plays a role in
-monitoring the availability of workers.
+This process is responsible for queueing scheduled tasks. It also plays a role in
```
bmbouter:
```
tasks. It also plays a role in
```
=>
```
tasks and is responsible for
```

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,58 @@ def tick(self):
...
+
         # this is not an event that gets sent anywhere. We process it
         # immediately.
         scheduler_event = {'timestamp': time.time(),
                            'type': 'scheduler-event',
-                           'hostname': ("%s@%s" % (SCHEDULER_WORKER_NAME, platform.node()))}
+                           'hostname': celerybeat_name}
         worker_watcher.handle_worker_heartbeat(scheduler_event)
+
+        _logger.debug(_('Lock updated by celerybeat_name %(celerybeat_name)s')
```
bmbouter: debug is the right level. :+1:

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,58 @@ def tick(self):
...
+            ret = super(Scheduler, self).tick()
+            _logger.info(_("Will tick again in %(ret)s secs "
+                         "and celerybeat_name is %(celerybeat_name)s") %
+                         {'ret': ret, 'celerybeat_name': celerybeat_name})
+        else:
+            # check for old enough time_stamp and remove if such lock is present
+            CeleryBeatLock.objects(lock_timestamp__lte=old_timestamp).delete()
+            try:
+                lock_timestamp = datetime.utcnow()
+                lock = "locked"
```
bmbouter: I think this should be moved to be the default value of the lock field on CeleryBeatLock. Then your code never has to set it.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,58 @@ def tick(self):
...
+
+        old_timestamp = datetime.utcnow() - timedelta(seconds=CELERYBEAT_OLD_TIMESTAMP)
+
+        # Updating the current lock if lock is on this instance of celerybeat
+        result = CeleryBeatLock.objects(celerybeat_name=celerybeat_name).\
+            update(set__lock_timestamp=datetime.utcnow())
+
+        # If current instance has lock and updated lock_timestamp, call super
+        if result == 1:
+            ret = super(Scheduler, self).tick()
```
bmbouter: L#436 - L#349 are the same as L#364 - L#367. Make this a `staticmethod` and having the same code called from two places.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,58 @@ def tick(self):
...
+            _logger.info(_("Will tick again in %(ret)s secs "
```
bmbouter: I think this should read more actively.    `"%(celerybeat_name)s will tick again in %(ret)s secs"`

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +320,58 @@ def tick(self):
...
+            CeleryBeatLock.objects(lock_timestamp__lte=old_timestamp).delete()
+            try:
+                lock_timestamp = datetime.utcnow()
+                lock = "locked"
+
+                # Insert new lock entry
+                new_lock = CeleryBeatLock(celerybeat_name=celerybeat_name,
+                                          lock_timestamp=lock_timestamp, lock=lock)
+                new_lock.save()
+                _logger.info(_("New lock acquired by %(celerybeat_name)s") %
```
bmbouter: I think this is good at the info level :+1:

server/pulp/server/async/scheduler.py
bmbouter: I think this should be at `debug` level.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,29 @@ def __init__(self, call_request, call_report):
...
+    """
+        :param celerybeat_name:        string representing the celerybeat instance name
+        :type  celerybeat_name:        basestring
+        :param lock_timestamp:         The timestamp(UTC) at which lock is acquired
+        :type  lock_timestamp:         datetime
+        :param lock:                   A unique key set to "locked" when lock is acquired.
+        :type  lock:                   basestring
+    """
+    celerybeat_name = StringField(required=True)
+    lock_timestamp = DateTimeField(required=True)
```
bmbouter: consider changing this field name to `timestamp`. Since the object is named CeleryBeatLock I think it's clear it's a lock_timestamp.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,29 @@ def __init__(self, call_request, call_report):
...
         self.call_request_string = str(call_request)
         self.serialized_call_report = call_report.serialize()
+
+
+class CeleryBeatLock(Document):
+    """
+    Single document collection which gives information about the current celerybeat lock.
+    """
+
+    """
```
bmbouter: this should be combined with the above docblock so that there is only 1 docblock for this function.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,29 @@ def __init__(self, call_request, call_report):
...
+        :param celerybeat_name:        string representing the celerybeat instance name
```
bmbouter: This is a style thing, but I prefer the other way. Consider making these not to be vertically aligned. Each ending colon should have a single space and then the value.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,29 @@ def __init__(self, call_request, call_report):
...
+        :type  celerybeat_name:        basestring
+        :param lock_timestamp:         The timestamp(UTC) at which lock is acquired
+        :type  lock_timestamp:         datetime
```
bmbouter: `datetime` should use a fully class path to identify which object type it is. Ask me if this comment doesn't make sense.

server/pulp/server/db/manage.py
```diff
@@ -121,6 +121,7 @@ def ensure_database_indexes():
     _ensure_indexes(model.ReservedResource)
     _ensure_indexes(model.TaskStatus)
     _ensure_indexes(model.Worker)
+    _ensure_indexes(model.CeleryBeatLock)
```
bmbouter: :+1: good job

shubham90:
ok test

shubham90:
@bmbouter  I have made the changes you suggested. They all seemed reasonable. Please do a re-review on this. Thanks a lot!

shubham90:
ok test

docs/user-guide/server.rst
```diff
@@ -30,10 +30,11 @@ For a recap of Pulp components and the work they are responsible for, read :ref:
   available workers. If a worker with the same name is started again after being missing, it is
   added into the pool of workers as any worker starting up normally would.
 
-* If ``pulp_celerybeat`` dies, and in case then new workers start, they won't be given work. If
-  existing workers stop, Pulp will continue assigning them work. Once restarted, pulp_celerybeat
-  will synchronize with the current state of all workers. Scheduled tasks will not run while
-  pulp_celerybeat is down, but they will instead run when celerybeat is restarted.
+* If all instances of ``pulp_celerybeat`` die and new workers start, they won't
+  be given work or if existing workers stop, Pulp will continue assigning them work. Once
```
bmbouter: s/work/work incorrectly/

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
...
         self.call_request_string = str(call_request)
         self.serialized_call_report = call_report.serialize()
+
+
+class CeleryBeatLock(Document):
+    """
+    Single document collection which gives information about the current celerybeat lock.
+
+        :param celerybeat_name: string representing the celerybeat instance name
+        :type  celerybeat_name: basestring
```
bmbouter: `type` should have a single space after it.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
...
+        :param timestamp: The timestamp(UTC) at which lock is acquired
+        :type  timestamp: datetime.datetime
```
bmbouter: `type` should have a single space after it.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
...
+        :param lock: A unique key set to "locked" when lock is acquired.
+        :type  lock: basestring
```
bmbouter: `type` should have a single space after it.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
...
+    """
+    celerybeat_name = StringField(required=True)
+    timestamp = DateTimeField(required=True)
+    lock = StringField(required=True, default="locked")
+
+    meta = {'collection': 'celery_beat_lock',
+            'indexes': [
```
bmbouter: L#511 - L#515 can be removed entirely, and the keyword argument `unique=True` can be added to lock.

http://docs.mongoengine.org/apireference.html#fields

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
         self.timestamp = dateutils.now_utc_timestamp()
         self.call_request_string = str(call_request)
         self.serialized_call_report = call_report.serialize()
+
+
+class CeleryBeatLock(Document):
+    """
+    Single document collection which gives information about the current celerybeat lock.
+
+        :param celerybeat_name: string representing the celerybeat instance name
```
bmbouter: This entire parameter block should be at the same indention level (4 spaces) as the word Single above.

server/pulp/server/db/model/dispatch.py
```diff
@@ -489,3 +490,27 @@ def __init__(self, call_request, call_report):
...
+        :type  celerybeat_name: basestring
+        :param timestamp: The timestamp(UTC) at which lock is acquired
+        :type  timestamp: datetime.datetime
+        :param lock: A unique key set to "locked" when lock is acquired.
+        :type  lock: basestring
+    """
+    celerybeat_name = StringField(required=True)
+    timestamp = DateTimeField(required=True)
+    lock = StringField(required=True, default="locked")
```
bmbouter: You should add a _ns option here for backwards compatibility. Add these two lines with a space above and below:

```
    # For backward compatibility
    _ns = StringField(default='celery_beat_lock')
```

server/test/unit/server/async/test_scheduler.py
```diff
@@ -282,6 +285,55 @@ def test_calls_handle_heartbeat(self, mock_heartbeat, mock_tick):
...
+
+        sched_instance = scheduler.Scheduler()
```
bmbouter: This should be a unit test, but it's written as an integration test.

server/test/unit/server/async/test_scheduler.py
```diff
@@ -282,6 +285,55 @@ def test_calls_handle_heartbeat(self, mock_heartbeat, mock_tick):
...
+        test = CeleryBeatLock.objects.get(lock="locked")
+        self.assertFalse(test is None)
+        self.assertTrue(ret <= TICK_SECONDS)
+        self.assertEqual(CeleryBeatLock.objects().count(), 1)
+        mock_tick.assert_called_once_with()
+
+    @mock.patch('celery.beat.Scheduler.__init__', new=mock.Mock())
+    @mock.patch('celery.beat.Scheduler.tick')
+    def test_heartbeat_lock_old_timestamp(self, mock_tick):
+        self.test_heartbeat_lock()
```
bmbouter: This should be a unit test, but it's written as an integration test.

server/test/unit/server/async/test_scheduler.py
```diff
@@ -282,6 +285,55 @@ def test_calls_handle_heartbeat(self, mock_heartbeat, mock_tick):
...
+        self.assertEqual(CeleryBeatLock.objects().count(), 1)
+
+#       Check if the delete is successful
+        test2.delete()
+        self.assertEqual(CeleryBeatLock.objects().count(), 0)
+
+    @mock.patch('celery.beat.Scheduler.__init__', new=mock.Mock())
+    @mock.patch('celery.beat.Scheduler.tick')
+    def test_heartbeat_lock_new(self, mock_tick):
+        self.test_heartbeat_lock()
```
bmbouter: This should be a unit test, but it's written as an integration test.

server/pulp/server/db/model/dispatch.py
rbarlow: On 06/26/2015 04:40 PM, bmbouter wrote:
> |type| should have a single space after it.

I think this might be a personal preference thing. I usually do it the
way Shubham did.

bmbouter:
I was hand testing this, and I got the following traceback when I run pulp-manage-db:

```
pulp.server.db.manage:CRITICAL: 'module' object has no attribute 'CeleryBeatLock'
pulp.server.db.manage:CRITICAL: (18666-81760) Traceback (most recent call last):
pulp.server.db.manage:CRITICAL: (18666-81760)   File "/home/bmbouter/Documents/pulp/server/pulp/server/db/manage.py", line 160, in main
pulp.server.db.manage:CRITICAL: (18666-81760)     return _auto_manage_db(options)
pulp.server.db.manage:CRITICAL: (18666-81760)   File "/home/bmbouter/Documents/pulp/server/pulp/server/db/manage.py", line 221, in _auto_manage_db
pulp.server.db.manage:CRITICAL: (18666-81760)     migrate_database(options)
pulp.server.db.manage:CRITICAL: (18666-81760)   File "/home/bmbouter/Documents/pulp/server/pulp/server/db/manage.py", line 104, in migrate_database
pulp.server.db.manage:CRITICAL: (18666-81760)     ensure_database_indexes()
pulp.server.db.manage:CRITICAL: (18666-81760)   File "/home/bmbouter/Documents/pulp/server/pulp/server/db/manage.py", line 124, in ensure_database_indexes
pulp.server.db.manage:CRITICAL: (18666-81760)     _ensure_indexes(model.CeleryBeatLock) 
pulp.server.db.manage:CRITICAL: (18666-81760) AttributeError: 'module' object has no attribute 'CeleryBeatLock'
pulp.server.db.manage:CRITICAL: (18666-81760)
```

The issue is that the CeleryBeatLock model needs to be moved out of dispatch and into pulp.server.db with the other subclasses of Document.

dkliban:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

server/pulp/server/db/model/__init__.py
```diff
@@ -534,3 +534,24 @@ def __hash__(self):
         type and the same unit key will get the same hash value.
         """
         return hash(self.unit_type_id + self.unit_key_str)
+
+
+class CeleryBeatLock(Document):
+    """
+    Single document collection which gives information about the current celerybeat lock.
+
+    :param celerybeat_name: string representing the celerybeat instance name
```
asmacdo: Very minor, but these are all instance variables, not params.

`:ivar celerybeat_name: string representing the celerybeat instance name`

server/test/unit/server/db/model/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+import mock
+
+
+class TestCeleryBeatLock(unittest.TestCase):
+    """
+    Test the CeleryBeatLock class.
+    """
+    @mock.patch('pulp.server.db.model.__init__', new=mock.Mock())
+    @mock.patch('pulp.server.db.model.CeleryBeatLock')
```
asmacdo: Generally not a good idea to mock the class you are testing.

The long term plan for the files in `pulp/srver/db/model/*` was that they will all be converted to mongoengine and removed. We are changing that plan slightly, but it doesn't really matter. The point is that the `model/__init__.py` tests are in `pulp/server/db/test_model.py`. There is not much to test with these since we are letting Mongoengine do the work, but take a look at the other model tests to see what is asserted.

asmacdo: Also please move the CeleryBeatLock tests to that same module.

asmacdo:
@shubham90, you don't need to, but feel free to delete the old pulpbot test comments and triggers to clean up.

shubham90:
@asmacdo Thanks for reviewing. I have made the changes you suggested. Those were really helpful.

shubham90:
ok test

server/pulp/server/async/scheduler.py
```diff
@@ -6,18 +6,20 @@
 import platform
 import threading
 import time
+import uuid
+import mongoengine
```
mhrivnak: for pep8, this should be in the second group of imports. The first group is reserved for standard library imports.

mhrivnak:
This looks good. Well done!

server/pulp/server/async/scheduler.py
bmbouter: This should be removed.

server/pulp/server/async/scheduler.py
```diff
@@ -317,15 +327,50 @@ def tick(self):
         :return:    number of seconds before the next tick should run
         :rtype:     float
         """
-        ret = super(Scheduler, self).tick()
         self._failure_watcher.trim()
 
+        # Setting the celerybeat name
+        celerybeat_name = SCHEDULER_WORKER_NAME + "@" + str(celery_name) + "@" + platform.node()
```
bmbouter: This should be changed back to `SCHEDULER_WORKER_NAME + "@" + platform.node()`.

The normal init scripts will only allow 1 scheduler per hostname and so we no longer need this. Also with this implementation every time `pulp_celerybeat` is restarted the new one can't recognize that it already had the lock because the name has changed with each restart.

bmbouter: Please update the tests accordingly.

server/pulp/server/async/scheduler.py
bmbouter: Also within each section imports should be alphabetized so `mongoengine` would come before `uuid`.

shubham90:
ok test

bmbouter:
I did several rounds of hand testing. I tested the following scenarios:

* when two pulp_celerybeat processes are running only one scheduled task gets sent out
* when the pulp_celerybeat who aquired the lock was killed the other failed over in the expected amount of time
* starting a the first pulp_celerybeat behaves as expected
* starting a pulp_celerybeat when old lock records are in place works as expected

@shubham90 great job, I'm putting the LGTM on it!

PR #1943, opened 2015-06-22
Removed pulp-manage-workers entry point from setup.py

dkliban:
The lastest version of setuptools does not support '/' character in name of
install scripts. This patch removes the entry point which lived in
/usr/libexec/pulp-manage-workers. The unit file now uses the
'pulp.server.asynch.manage_workers' as a python module.

### DIFF
--- a/server/pulp/server/async/manage_workers.py
+++ b/server/pulp/server/async/manage_workers.py
@@ -115,7 +115,7 @@ def _stop_workers():
         sys.exit(exit_code)
 
 
-def main():
+if __name__ == "__main__":
     """
     This is the entry point method that becomes pulp-manage-workers.
     """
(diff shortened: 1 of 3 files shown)


server/pulp/server/async/manage_workers.py
```diff
@@ -115,7 +115,7 @@ def _stop_workers():
         sys.exit(exit_code)
 
 
-def main():
+if __name__ == "__main__":
     """
     This is the entry point method that becomes pulp-manage-workers.
```
rbarlow: You should probably update this comment since pulp-manage-workers doesn't exist anymore. Also, it's not really a method anymore. Perhaps something like "This code block is executed by the systemd unit file to manage the worker units"?

rbarlow:
Nice work!

PR #1945, opened 2015-06-22
Added back the main() method so we can unit test it

dkliban:


server/pulp/server/async/manage_workers.py
```diff
@@ -128,3 +128,6 @@ def _stop_workers():
         _start_workers()
     elif _action == 'stop':
         _stop_workers()
+
```
rbarlow: One more space here for PEP-8.

server/pulp/server/async/manage_workers.py
```diff
@@ -115,9 +115,9 @@ def _stop_workers():
         sys.exit(exit_code)
 
 
-if __name__ == "__main__":
+def main():
     """
-    This is the entry point method that becomes pulp-manage-workers.
+    This method is executed by the systemd unit file to manage the worker units.
```
rbarlow: I know I originally used the word "method", but technically I think this is a function…

rbarlow:
Is Jenkins going to test this? It doesn't look like it's started yet. If not, LGTM. If so, make sure that the Jenkins tests pass before merging (and still LGTM).

PR #1946, opened 2015-06-22
Improve custom pulp exception incorrect syntax error

asmacdo:
Incorrect syntax in a new PLP error resulted in `str(e) = None` and was not obvious to fix.

### DIFF
--- a/server/pulp/server/webservices/middleware/exception.py
+++ b/server/pulp/server/webservices/middleware/exception.py
@@ -31,28 +31,34 @@ class ExceptionHandlerMiddleware(object):
         :return: Response containing processed exception
         :rtype: django.http.HttpResponseServerError
         """
-        if isinstance(exception, PulpException):
-            status = exception.http_status_code
-            response = error.http_error_obj(status, str(exception))
-            response.update(exception.data_dict())
-            response['error'] = exception.to_dict()
-            logger.info(str(exception))
-            response_obj = HttpResponse(json.dumps(response), status=status,
-                                        content_type="application/json; charset=utf-8")
-        else:
-            status = httplib.INTERNAL_SERVER_ERROR
-            response = error.http_error_obj(status, str(exception))
-            msg = _('Unhandled Exception')
-            logger.error(msg)
 
-        if status == httplib.INTERNAL_SERVER_ERROR:
-            logger.exception(str(exception))
-            e_type, e_value, trace = sys.exc_info()
-            response['exception'] = traceback.format_exception_only(e_type, e_value)
-            response['traceback'] = traceback.format_tb(trace)
-            response_obj = HttpResponseServerError(json.dumps(response),
-                                                   content_type="application/json; charset=utf-8")
-        else:
-            logger.info(str(exception))
+        try:
+            if isinstance(exception, PulpException):
+                status = exception.http_status_code
+                response = error.http_error_obj(status, str(exception))
+                response.update(exception.data_dict())
+                response['error'] = exception.to_dict()
+                logger.info(str(exception))
+                response_obj = HttpResponse(json.dumps(response), status=status,
+                                            content_type="application/json; charset=utf-8")
+            else:
+                status = httplib.INTERNAL_SERVER_ERROR
+                response = error.http_error_obj(status, str(exception))
+                msg = _('Unhandled Exception')
+                logger.error(msg)
 
-        return response_obj
+            if status == httplib.INTERNAL_SERVER_ERROR:
+                logger.exception(str(exception))
+                e_type, e_value, trace = sys.exc_info()
+                response['exception'] = traceback.format_exception_only(e_type, e_value)
+                response['traceback'] = traceback.format_tb(trace)
+                response_obj = HttpResponseServerError(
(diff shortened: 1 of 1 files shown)


server/pulp/server/exceptions.py
```diff
@@ -92,7 +92,10 @@ def __init__(self, error_code=error_codes.PLP0001, **kwargs):
                                          field=key)
 
     def __str__(self):
-        msg = self.error_code.message % self.error_data
+        try:
+            msg = self.error_code.message % self.error_data
+        except ValueError:
+            msg = "Incorrectly formatted error string for {code}".format(code=self.error_code)
```
rbarlow: This seems like it would obscure the real error message for the developer. I see that the intent is to try to help a developer out a bit, but I think this actually would be more difficult to understand. Compare:

```
>>> 'This is a %(test)' % {'test': 'test'}
  File "<stdin>", line 1, in <module>
ValueError: incomplete format
```

to:

```
>>> try:
...  'This is a %(test)' % {'test': 'test'}
... except ValueError:
...  print 'Incorrectly formatted error string for PLP-1234'
... 
Incorrectly formatted error string for PLP-1234
```

I think that ValueError: incomplete format is more informative, and it avoids dropping the stack trace which I think would be helpful during debugging. Also, I wonder if there's a way to trigger a ValueError by passing the wrong argument on the other side of the ```%``` operator. I tried to find something, but I was unsuccessful. If that is possible, this error message could be very misleading.

IMO, it's best to let developer facing errors happen without trying to intervene. For outside users, it is good to try to catch errors and give more human understandable meaning to them, but I think we can reasonably expect developers to be able to read ```ValueError: incomplete format``` and know what to do.

rbarlow:
Also, I'm not seeing how this would result in str(e) = None, unless something above this in the stack is catching and eating the Exception. IMO, we should find and fix that instead of doing this.

asmacdo:
@rbarlow, I agree. Rather than playing with the error message, I solved a more general problem: if there are errors in the middleware exception handing, they are logged and reraised.

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
@rbarlow could you have another look at this?

server/pulp/server/webservices/middleware/exception.py
```diff
@@ -31,28 +31,36 @@ def process_exception(self, request, exception):
...
+                response['traceback'] = traceback.format_tb(trace)
+                response_obj = HttpResponseServerError(
+                    json.dumps(response), content_type="application/json; charset=utf-8")
+            else:
+                logger.info(str(exception))
+
+            response_obj['Content-Encoding'] = 'utf-8'
+            return response_obj
+        # Because this defines our own exception handling, if something goes wrong, we need to make
+        # sure that it gets logged.
```
rbarlow: s/wrong,/wrong/

server/pulp/server/webservices/middleware/exception.py
```diff
@@ -31,28 +31,36 @@ def process_exception(self, request, exception):
...
+
+            if status == httplib.INTERNAL_SERVER_ERROR:
+                logger.exception(str(exception))
+                e_type, e_value, trace = sys.exc_info()
+                response['exception'] = traceback.format_exception_only(e_type, e_value)
+                response['traceback'] = traceback.format_tb(trace)
+                response_obj = HttpResponseServerError(
+                    json.dumps(response), content_type="application/json; charset=utf-8")
+            else:
+                logger.info(str(exception))
```
rbarlow: I know this line was like this before your change, but it seems like this line is going to re-log the str() of the Exception in non INTERNAL_SERVER_ERROR cases. I think we can drop the else and this line and not lose any info.

server/pulp/server/webservices/middleware/exception.py
```diff
@@ -31,28 +31,36 @@ def process_exception(self, request, exception):
...
+                msg = _('Unhandled Exception')
+                logger.error(msg)
+
+            if status == httplib.INTERNAL_SERVER_ERROR:
+                logger.exception(str(exception))
+                e_type, e_value, trace = sys.exc_info()
+                response['exception'] = traceback.format_exception_only(e_type, e_value)
+                response['traceback'] = traceback.format_tb(trace)
+                response_obj = HttpResponseServerError(
+                    json.dumps(response), content_type="application/json; charset=utf-8")
```
rbarlow: This entire block could just be moved into the else block above it, and we could drop this if check. If we do that combined with dropping the extra logger.info() below this then I think this code gets simpler and more readable.

rbarlow:
LGTM. Feel free to consider the "refactoring" suggestions optional (or you could make a separate PR just about those if you want to do that, or not… whatevs ☺)

server/pulp/server/webservices/middleware/exception.py
asmacdo: I wanted to do that also, but it is possible that the http_status_code is a 500 and the exception is an instance of a `PulpException`. In that case this block still needs to run.

asmacdo:
ok test

PR #1954, opened 2015-06-24
Fixed failure during pulp-admin user update.

ipanova:
closes #1058
https://pulp.plan.io/issues/1058


rbarlow:
As I noted in the Redmine issue, I lol'd at the fact that we have a command in pulp-admin that uses -p, when -p is already used to specify the password on all pulp-admin commands. Oooof! I'll file another issue to change that.

ipanova:
@rbarlow  oh yeah, we need to fix that, good catch ^_^

PR #1955, opened 2015-06-24
Adding fc22 to dist list of celery and semantic versioning

dkliban:


deps/python-celery/dist_list.txt
```diff
@@ -1 +1 @@
-el6 el7 fc20 fc21
+el6 el7 fc20 fc21 fc22
```
rbarlow: I think we can drop fc20.

rbarlow:
I think we can drop fc20 in all of the packages. It is officially not maintained anymore as of yesterday. LGTM!

PR #1956, opened 2015-06-25
Adds pymongo 2.7.2 and mongoengine 0.9.0

bmbouter:
Both dependencies have been test built in koji against fc21, 
fc22, el6, and el7.

pymongo builds:
http://koji.katello.org/koji/taskinfo?taskID=287868
http://koji.katello.org/koji/taskinfo?taskID=287866
http://koji.katello.org/koji/taskinfo?taskID=287874
http://koji.katello.org/koji/taskinfo?taskID=287871

mongoengine builds:
http://koji.katello.org/koji/taskinfo?taskID=287881
http://koji.katello.org/koji/taskinfo?taskID=287883
http://koji.katello.org/koji/taskinfo?taskID=287877
http://koji.katello.org/koji/taskinfo?taskID=287879

The pymongo spec file skips tests that require a running
instance of mongodb. I manually ran the tests against a
running instance once and they all passed for me. If we
ever configure the builder with a running instance of mongodb
we can delete the --exclude list in the pymongo spec file.

Updates pulp.server.db.manage:ensure_database_indexes()
to only use the Document.ensure_indexes() method.

Removes pulp.server.db.model.queryset:update_one() which was
here for 0.7.10 compatability.

This also updates all tests to run correctly when the test
runner has mongoengine==0.9.0 and pymongo==2.7.2 installed.

Test server.managers.repo.test_cud:test_update_repo_and_plugins
is removed altogether. I don't know why it was failing after
upgrading the deps, but when I downgrade the deps back to
mongoengine 0.7.10 and pymongo 2.5.3 it passed. It's safe to
remove altogether here because it has completely been removed
and rewritten with #1931.

https://pulp.plan.io/issues/934
fixes #934


dkliban:
ok test

pulp.spec
```diff
@@ -304,8 +304,8 @@ Requires: python-%{name}-common = %{pulp_version}
 Requires: python-blinker
 Requires: python-celery >= 3.1.0
 Requires: python-celery < 3.2.0
-Requires: python-pymongo >= 2.5.2
-Requires: python-mongoengine >= 0.7.10
+Requires: python-pymongo >= 2.7.1
```
dkliban: 2.7.1 or 2.7.2 ?

deps/python-mongoengine/python-mongoengine.spec
```diff
@@ -1,110 +1,88 @@
...
-BuildRequires: python-pillow
-Requires:      python-pillow
-%else
-Requires: python-imaging
-BuildRequires: python-imaging
-%endif
-BuildRequires: python-coverage
-BuildRequires: python-nose
-Requires: python-blinker
+Requires:       python-pymongo >= 2.7.1
```
dkliban: same question here ... do you mean 2.7.2 ?

asmacdo:
@bmbouter, why the patch to remove pillow? I assume that it is for older python support?

pulp.spec
bmbouter: 2.7.1 b/c that is the actual version that mongoengine requires.

deps/python-mongoengine/python-mongoengine.spec
bmbouter: same, here it's 2.7.1. Good eye though!

bmbouter:
@asmacdo Pillow isn't a dep that is available on EL6. See more details in the README notes I wrote:

https://github.com/pulp/pulp/pull/1956/files#diff-17deb32a98804a9357b319e08adc2677R3

PR #1957, opened 2015-06-26
index task_status.task_id only once

asmacdo:
The task status model incorrectly generated 2 indexes for task_id. I believe that this will fix https://pulp.plan.io/issues/1092 but I have been unable to replicate the failure of the migration.

Regardless, this extra index must be dropped in 2.7.0 (where it first appears) so that we do not need to write a migration to remove it.

@dkliban, I would appreciate if you could try this on your system to see if it solves your issue.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -234,7 +234,7 @@ class TaskStatus(Document, ReaperMixin):
     :type traceback:   None
     """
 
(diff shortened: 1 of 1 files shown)


dkliban:
I was able to build RPMs and run pulp-manage-db on top of a 2.6.3 database. All services are able to start and perform basic repo create, sync, publish and delete.

PR #1961, opened 2015-06-30
Fixing get, update, delete call when non existent importer is provided.

ipanova:
closes #750
https://pulp.plan.io/issues/750

All this three calls behave the same:
if a non_exitent_importer provided in the call, but the repo itself exists a 404 missing_resource *repo_id* will be raised, though the repo exists.
Now with the fix it will return 404 missing_resource *importer_id*

### DIFF
--- a/server/pulp/server/webservices/views/repositories.py
+++ b/server/pulp/server/webservices/views/repositories.py
@@ -121,6 +121,35 @@ def _process_repos(repos, details, importers, distributors):
     return repos
 
 
+def _get_valid_importer(repo_id, importer_id):
+    """
+    Validates if the specified repo_id and importer_id are valid.
+    If not MissingResource exception is raised.
+
+    :param repo_id: id of the repo
+    :type repo_id: str
+    :param importer_id: id of the importer
+    :type importer_id: str
+
+    :return: key-value pairs describing the importer in use
+    :rtype:  dict
+    :raises pulp_exceptions.MissingResource: if repo or importer cannot be found
+    """
+
+    repo_query_manager = manager_factory.repo_query_manager()
+    repo = repo_query_manager.find_by_id(repo_id)
+    if repo is None:
+        raise pulp_exceptions.MissingResource(repo_id=repo_id)
+    importer_manager = manager_factory.repo_importer_manager()
+    try:
+        importer = importer_manager.get_importer(repo_id)
+        if importer['id'] != importer_id:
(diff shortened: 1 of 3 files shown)


server/pulp/server/webservices/views/repositories.py
```diff
@@ -121,6 +121,35 @@ def _process_repos(repos, details, importers, distributors):
...
+
+    :return: key-value pairs describing the importer in use
+    :rtype:  dict
+    :raises pulp_exceptions.MissingResource: if repo or importer cannot be found
+    """
+
+    repo_query_manager = manager_factory.repo_query_manager()
+    repo = repo_query_manager.find_by_id(repo_id)
+    if repo is None:
+        raise pulp_exceptions.MissingResource(repo_id=repo_id)
```
asmacdo: This works here but will fail when merging to master because of my recent mongoengine change. Here is what you will need to do in master (replacing the previous 4 lines):

``` python
from pulp.server.db import model

model.Repository.objects.get_repo_or_missing_resource(repo_id)
```

server/pulp/server/webservices/views/repositories.py
```diff
@@ -496,11 +515,7 @@ def put(self, request, repo_id, importer_id):
         :raises pulp_exceptions.OperationPostponed: dispatch a task
         """
 
-        importer_manager = manager_factory.repo_importer_manager()
-        importer = importer_manager.get_importer(repo_id)
-        if importer['id'] != importer_id:
-            raise pulp_exceptions.MissingResource(importer_id=importer_id)
-
+        _validate_repo_importer_existance(repo_id, importer_id)
```
asmacdo: nice!

server/pulp/server/webservices/views/repositories.py
```diff
@@ -121,6 +121,35 @@ def _process_repos(repos, details, importers, distributors):
     return repos
 
 
+def _validate_repo_importer_existance(repo_id, importer_id):
```
asmacdo: With a name like this, I do not expect it to return the importer. Perhaps this should be `_get_importer` or something similar. I think it is reasonable that this function could both retrieve and validate.

asmacdo:
@ipanova, one minor thing, but I think it is ready. The merge forward will be more involved than normal due to some of my recent changes, let me know if I can help.

ipanova:
ok test

PR #1963, opened 2015-06-30
import unittest2 in compat module

asmacdo:
In order to use new things like `self.assertDictEqual`, we are increasingly using unittest2 in older versions of python. By moving this import to the compat module, we can avoid doing this same import magic in every test module.

### DIFF
--- a/common/pulp/common/compat.py
+++ b/common/pulp/common/compat.py
@@ -1,16 +1,4 @@
 # -*- coding: utf-8 -*-
-#
-# Copyright © 2012 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public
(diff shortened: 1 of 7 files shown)


common/pulp/common/compat.py
```diff
@@ -77,9 +60,13 @@ def any(iterable):
...
 
 
-################################################
 # This provides pkgutil.iter_modules.
-################################################
 @check_builtin(pkgutil)
 def iter_modules(*args, **kwargs):
     return backports.pkgutil.iter_modules(*args, **kwargs)
+
+
```
ipanova: maybe right a comment like previous imports have? this provides unittest

PR #1964, opened 2015-06-30
Fixes mongoengine model usage that did not have kwargs

bmbouter:
The upgrade to mongoengine 0.9.0 revealed that parts of the
Pulp codebase which used mongoengine models did not specify
kwargs. This raises an exception (as it should) in mongoengine
0.9.0.

I audited the existing models on 2.7-testing, fixed those, and
I updated the tests to match.

re #934
https://pulp.plan.io/issues/934


dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

PR #1965, opened 2015-06-30
sort sync and publish history by an indexed key

asmacdo:
History results were sorted by the `started` key which was not indexed. Sorting with `_id` which is an ObjectId will result in similar sorting using an indexed key. 

From the mongo [docs](http://docs.mongodb.org/v2.4/reference/object-id/#objectid):
"sorting on an _id field that stores ObjectId values is roughly equivalent to sorting by creation time."

https://pulp.plan.io/issues/1043

This was branched from the merge-base of 2.6-testing and 2.7-testing. When merged, this feature branch will be manually merged into 2.7-testing as well as 2.7-dev. There will be some modification (removing the repetition) on the merge forward to master to account for a refactor: https://github.com/pulp/pulp/pull/1931.

### DIFF
--- a/server/pulp/server/managers/repo/publish.py
+++ b/server/pulp/server/managers/repo/publish.py
@@ -340,8 +340,10 @@ class RepoPublishManager(object):
 
         # Retrieve the entries
         cursor = RepoPublishResult.get_collection().find(search_params)
-        # Sort the results on the 'started' field. By default, descending order is used
-        cursor.sort('started', direction=constants.SORT_DIRECTION[sort])
+        # Sort the results by the ObjectId field, which will effectively sort based on the start
+        # time of the event. This way of sorting is preferred because for a sufficiently large data
+        # set, the sort field must be an indexed field.
+        cursor.sort('_id', direction=constants.SORT_DIRECTION[sort])
         if limit is not None:
             cursor.limit(limit)
 
--- a/server/pulp/server/managers/repo/sync.py
+++ b/server/pulp/server/managers/repo/sync.py
@@ -294,8 +294,10 @@ class RepoSyncManager(object):
 
         # Retrieve the entries
         cursor = RepoSyncResult.get_collection().find(search_params)
-        # Sort the results on the 'started' field. By default, descending order is used
-        cursor.sort('started', direction=constants.SORT_DIRECTION[sort])
+        # Sort the results by the ObjectId field, which will effectively sort based on the start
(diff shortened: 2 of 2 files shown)


server/pulp/server/managers/repo/publish.py
```diff
@@ -340,8 +340,10 @@ def publish_history(self, repo_id, distributor_id, limit=None, sort=constants.SO
 
         # Retrieve the entries
         cursor = RepoPublishResult.get_collection().find(search_params)
-        # Sort the results on the 'started' field. By default, descending order is used
-        cursor.sort('started', direction=constants.SORT_DIRECTION[sort])
+        # Sort the results on the ObjectId field. This will effectively sort by when the item was
+        # created. This is preferred because 'started' is not indexed. By default, descending order
```
ipanova: "This is preferred because 'started' is not indexed" I don't think this comment will be effective for person who is looking at this code for the first time unless he will look at the history of this file and find out that previously it was sorted by 'started' field.
I think more efficient would be to write something like this instead: This way of sorting is preferred to non indexed fields, as those can lead to the database error: too much data for sort() with no index.

server/pulp/server/managers/repo/sync.py
```diff
@@ -294,8 +294,10 @@ def sync_history(self, repo_id, limit=None, sort=constants.SORT_DESCENDING, star
 
         # Retrieve the entries
         cursor = RepoSyncResult.get_collection().find(search_params)
-        # Sort the results on the 'started' field. By default, descending order is used
-        cursor.sort('started', direction=constants.SORT_DIRECTION[sort])
+        # Sort the results on the ObjectId field. This will effectively sort by when the item was
+        # created. This is preferred because 'started' is not indexed. By default, descending order
+        # is used
```
ipanova: same here

ipanova:
@asmacdo seems like added code lines are missing coverage and you also would need to update the tests https://github.com/asmacdo/pulp/blob/1043/server/test/unit/server/managers/repo/test_publish.py#L453 as they are still using the 'started' field

Assigning this back to you.

asmacdo:
@ipanova, I should have mentioned this in the original comment. I have opted not to write a unit test for this because it is extremely simple and this code will need to be significantly changed on the merge forward. I will add a unit test for the master code when I merge it forward, but to do the same here would not be efficient.

ipanova:
LGTM, but don't forget to add/update needed testcases when you'll merge forward

asmacdo:
ok test

PR #1969, opened 2015-07-01
Fixed: Unable to remove a user from super-user role

shubham90:
https://pulp.plan.io/issues/1078

closes #1078


ipanova:
@shubham90 Great finding! good job
will you also that the test for ensure_super_user_role() with assertions of proper keys and values, so we can be sure that the super user role is correctly created.

ipanova:
Also, please, put the target release on the issue

shubham90:
@ipanova I will add the test case and the target release. Thanks for reviewing!

PR #1973, opened 2015-07-07
schedulers should use queue_syncs and publishs rather than run

asmacdo:
https://pulp.plan.io/issues/1116


asmacdo:
ok test

bmbouter:
I just tested master, and the issue went away for me. Good job @asmacdo

PR #1975, opened 2015-07-08
load data structures from GET json

asmacdo:

The BZ originally led me to believe that the problem was with the `filters` field. In actuality, all fields will be passed as strings and need to be loaded from json. The `limit` and `skip` fields do not need to be included because they are cast to integers elsewhere.


asmacdo:
ok test

dkliban:
ok test

asmacdo:
ok test

asmacdo:
ok test

PR #1976, opened 2015-07-09
Adds logging if Pulp is missing critical Celery services

bmbouter:
https://pulp.plan.io/issues/1115
closes 1115

### DIFF
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -10,9 +10,9 @@ import time
 from celery import beat
 from celery.result import AsyncResult
 
-from pulp.common.constants import SCHEDULER_WORKER_NAME
+from pulp.common.constants import RESOURCE_MANAGER_WORKER_NAME, SCHEDULER_WORKER_NAME
 
-from pulp.server.async.celery_instance import celery as app, RESOURCE_MANAGER_QUEUE
+from pulp.server.async.celery_instance import celery as app
 from pulp.server.async.tasks import _delete_worker
 from pulp.server.async import worker_watcher
 from pulp.server.db import connection as db_connection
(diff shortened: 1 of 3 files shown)


server/pulp/server/async/scheduler.py
```diff
@@ -186,57 +186,72 @@ def monitor_events(self):
...
         for worker in worker_list:
-            msg = _("Workers '%s' has gone missing, removing from list of workers") % worker.name
+            if worker['last_heartbeat'] < oldest_heartbeat_time:
+                msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
+                _logger.error(msg)
+                _delete_worker(worker.name)
+            elif worker['name'].startswith(SCHEDULER_WORKER_NAME):
+                scheduler_count = scheduler_count + 1
+            elif worker['name'].startswith(RESOURCE_MANAGER_WORKER_NAME):
+                resource_manager_count = resource_manager_count + 1
```
beav: It may be worthwhile to add a debug-level logging statement here for total number of schedulers and resource managers.

beav:
LGTM but I would not mind a second review since it's for 2.6.x :koala:

server/pulp/server/async/scheduler.py
```diff
@@ -186,57 +186,72 @@ def monitor_events(self):
...
+        for cleanup.
 
-        This method logs and the debug and error levels.
+        This method also checks that at least one resource_manager and one scheduler process is
+        present. If there are zero of either, log at the error level that Pulp will not operate
+        correctly.
         """
-        msg = _('Looking for workers missing for more than %s seconds') % self.WORKER_TIMEOUT_SECONDS
+        msg = _('Checking if pulp_workers, pulp_celerybeat, or pulp_resource_manager '
+                'processes are missing for more than %s seconds') % self.CELERY_TIMEOUT_SECONDS
```
mhrivnak: python is very forgiving, but in theory this should be ``%d``

server/pulp/server/async/scheduler.py
bmbouter: This is a good idea; I'm doing it and an associated test.

server/pulp/server/async/scheduler.py
bmbouter: good catch; fixed

PR #1978, opened 2015-07-12
Moving model/__init__.py to models.py

pcreech:
https://pulp.plan.io/issues/1066

Refactoring model/__init__.py to models.py


barnabycourt:
ok test

barnabycourt:
@pcreech This looks good on a first pass, the primary hangup is that it will require a corresponding change to pulp_docker's use of these models. It shouldn't be a big thing to fix.  @asmacdo, what do you think?

asmacdo:
@pcreech, is it possible to use `git mv` to preserve the git history? If you do that, make sure to recreate the `model/__init__.py` so that package works.

asmacdo:
@barnabycourt, we will also need a corresponding change in pulp_rpm.

pcreech:
@barnabycourt @asmacdo , will update to use git mv instead, as well as make corresponding changes to pulp_rpm and pulp_docker (and any others that would need it, just list them here)

asmacdo:
@pcreech could you also move `server/test/unit/server/db/test_model.py` to `server/test/unit/server/db/test_models.py`

pcreech:
Updated!

asmacdo:
@pcreech, looking over this, I realized that we may have a problem. 

I have posted about it here:
http://stackoverflow.com/questions/32769663/have-git-track-changes-on-a-renamed-file-with-new-file-to-complicate-matters

There may be no way around it (not due to your changes), but this will probably cause some pretty irritating merge conflicts going forward if we continue our current branching strategy. I want to wait until next week to merge it to make sure that everyone is on board.

pcreech:
@asmacdo 

Agreed.  It does appear to have the potential to cause some headaches in the short term.

asmacdo:
@pcreech, we have discussed this and have determined that it will cause too much headache to be worth the move now. As we complete the mongoengine conversion, all of the modules in the `model/` directory will me moved out of there. At that point we can move the file and git will be smart enough to track the changes because we will not need a new `__init__.py`. We really appreciate the work you did on this, especially since we missed the crucial detail (that model changes could not be merged forward) until after you had done the work.

pcreech:
@asmacdo No problem, glad I could help!

PR #1981, opened 2015-07-13
ref #865 - mongoengine modeling for shared storage.

jortel:
The ContentUnit remodeled to have (2) derived classes: FileContentUnit and SharedContentUnit.  All plugins currently subclassing ContentUnit will need to be changed to subclass FileContentUnit instead.

Content (bits) storage is delegated to ContentStorage classes.  These classes are focused on how content is stored.  They are not meant to address subjects like *Object Storage* but ARE modeled in such a way as to not be orthogonal.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -1,8 +1,6 @@
 import copy
-import errno
 import logging
 import os
-import shutil
 import uuid
 from collections import namedtuple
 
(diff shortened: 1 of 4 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -536,6 +490,86 @@ def __hash__(self):
...
+
+
+class SharedContentUnit(ContentUnit):
+    """
+    A content unit representing content that is stored in a
+    shared storage facility.
+    """
+
+    @property
+    def storage_id(self):
```
barnabycourt: Are there any requirements on what this string can be?

jortel: Good question.  Perhaps not requirements or limitations here but the SharedStorage should probably hash it so that it is of deterministic makeup and length.

jortel:
ok test

jortel:
ok test

PR #1982, opened 2015-07-13
Story 1111: Purge completed task status records

shubham90:
https://pulp.plan.io/issues/1111

closes #1111

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -94,6 +94,7 @@ PLP0038 = Error("PLP0038", _("The unit model with id %(model_id)s and class "
 PLP0039 = Error("PLP0039", _("The unit model with the id %(model_id)s failed to register. The "
                              "class %(model_class)s is not a subclass of ContentUnit."),
                 ['model_id', 'model_class'])
+PLP0040 = Error("PLP0040", _("This request is forbidden."), [])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
@@ -113,3 +114,6 @@ PLP1008 = Error("PLP1008", _("The importer type %(importer_type_id)s does not ex
 PLP1009 = Error("PLP1009", _("The request body does not contain valid JSON"), [])
 PLP1010 = Error("PLP1010", _("Provided value %(value)s for field %(field)s must be of type "
                              "%(field_type)s."), ["value", "field", "field_type"])
+PLP1011 = Error("PLP1011", _("Invalid task state passed to parameters list: %(state)s. "
+                             "Valid states are finished, error, timed out and skipped"), ["state"])
+PLP1012 = Error("PLP1012", _("No task state given to parameters list for delete."), [])
--- a/docs/dev-guide/integration/rest-api/tasks.rst
+++ b/docs/dev-guide/integration/rest-api/tasks.rst
@@ -119,6 +119,30 @@ All currently running and waiting tasks may be listed. This returns an array of
 
 | :return:`array of` :ref:`task_report`
 
+
+Deleting Completed Tasks
+------------------------
+
+All completed tasks with states *finished, error, timed out, skipped* may be deleted.
+This call returns response code 204 if successful or code 403 if the request is forbidden.
+
+| :method:`delete`
+| :path:`/v2/tasks/`
+| :permission:`delete`
+
+* :param:`?state,str,only states *finished, error, timed out, skipped* can be deleted.
+
+For example::
+
+  /pulp/api/v2/tasks/?state=finished&state=skipped
+
+| :response_list:`_`
+
+* :response_code:`204,if the tasks were successfully deleted`
+* :response_code:`403, if there was a forbidden request`
+
+| :return:`HttpResponse or pulp Exception`
+
 Searching for Tasks
 -------------------
 
--- a/docs/user-guide/release-notes/master.rst
+++ b/docs/user-guide/release-notes/master.rst
@@ -51,6 +51,8 @@ After migrating the database, restart `httpd`, `pulp_workers`, `pulp_celerybeat`
 Rest API Changes
 ----------------
 
+* Tasks with complete states (except `canceled` state) can now be deleted.
+
 Binding API Changes
 -------------------
 
--- a/server/pulp/server/webservices/views/tasks.py
+++ b/server/pulp/server/webservices/views/tasks.py
@@ -4,8 +4,12 @@ This module contains views related to Pulp's task system models.
 from datetime import datetime
 
 from django.views.generic import View
+from django.http import HttpResponse
 from mongoengine.queryset import DoesNotExist
 
+from pulp.common import error_codes
+from pulp.common.constants import CALL_CANCELED_STATE, CALL_COMPLETE_STATES
+from pulp.server import exceptions as pulp_exceptions
 from pulp.server.async import tasks
 from pulp.server.auth import authorization
 from pulp.server.db.model import Worker, TaskStatus
@@ -15,6 +19,10 @@ from pulp.server.webservices.views.decorators import auth_required
 from pulp.server.webservices.views.serializers import dispatch as serial_dispatch
 from pulp.server.webservices.views.util import (generate_json_response,
                                                 generate_json_response_with_pulp_encoder)
+
+
+# This constant set is used for deleting the completed tasks from the collection.
+VALID_STATES = set(filter(lambda state: state != CALL_CANCELED_STATE, CALL_COMPLETE_STATES))
 
 
 def task_serializer(task):
@@ -46,12 +54,11 @@ class TaskCollectionView(View):
     """
     View for all tasks.
     """
-
     @auth_required(authorization.READ)
     def get(self, request):
         """
-        Return a response containing a list of all tasks or a response containg a list of all tasks
-        that can be filtered by the optional GET parameter 'tags'.
+        Return a response containing a list of all tasks or a response containing
+        a list of tasks filtered by the optional GET parameter 'tags'.
 
         :param request: WSGI request object
         :type  request: django.core.handlers.wsgi.WSGIRequest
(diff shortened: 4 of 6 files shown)


server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +67,32 @@ def get(self, request):
...
+        :return: Status of delete operation or pulp Exception
+        :rtype:  String or pulp.Exception
+        """
+        task_state = request.GET.getlist('state')
+        if not task_state:
+            raise pulp_exceptions.PulpCodedForbiddenException()
+        complete_states = ("skipped", "error", "finished", "timed out")
+        for state in task_state:
+            if state not in complete_states:
+                raise pulp_exceptions.PulpCodedForbiddenException()
```
mhrivnak: Validation should happen before any deletion can happen. In this implementation, if three states are provided, and only the last is invalid, deletion will occur for the first two before the exception is raised.

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +67,32 @@ def get(self, request):
...
+            try:
+                TaskStatus.objects(state=state).delete()
+                result = "Tasks Deleted Successfully"
```
mhrivnak: REST API conventions are that DELETE requests are idempotent. There is no need to tell the client whether you found the resources or not, and in fact doing so is undesirable.

I suggest not returning a message at all, and use response code "204" with an empty body.

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +67,32 @@ def get(self, request):
...
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+
+        :return: Status of delete operation or pulp Exception
+        :rtype:  String or pulp.Exception
+        """
+        task_state = request.GET.getlist('state')
+        if not task_state:
+            raise pulp_exceptions.PulpCodedForbiddenException()
```
mhrivnak: If the request fails validation, we should return an error message about why.

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +67,32 @@ def get(self, request):
...
+        complete_states = ("skipped", "error", "finished", "timed out")
```
mhrivnak: You should reference these states from the constants file: https://github.com/pulp/pulp/blob/master/common/pulp/common/constants.py#L51

Perhaps make a copy of the tuple, but exclude the canceled state for now. Maybe ``VALID_STATES = set(filter(lambda state: state != constants.CALL_CANCELED_STATE, constants.CALL_COMPLETE_STATES)``

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +67,32 @@ def get(self, request):
...
+        for state in task_state:
+            if state not in complete_states:
+                raise pulp_exceptions.PulpCodedForbiddenException()
+            try:
+                TaskStatus.objects(state=state).delete()
+                result = "Tasks Deleted Successfully"
+            except:
```
mhrivnak: This could mask all kinds of other problems. What if there was an error connecting to the database, for example? Or what if the database experiences some internal error, like it timed out because there were too many records to delete? In general, you should almost never have a naked ``except`` statement in python.

I'd specifically catch the exception you expect if no resources are found. Although would mongoengine actually raise such an exception in this case?

Otherwise, let any other exception bubble up and get caught by the middleware. That should turn uncaught exceptions into 5xx responses.

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
@mhrivnak Thanks for the review. I have made the changes as you suggested. If possible, please have have a look at it.

common/pulp/common/constants.py
```diff
@@ -53,6 +53,7 @@
 CALL_STATES = (CALL_WAITING_STATE, CALL_SKIPPED_STATE, CALL_ACCEPTED_STATE, CALL_RUNNING_STATE,
                CALL_SUSPENDED_STATE, CALL_FINISHED_STATE, CALL_ERROR_STATE, CALL_CANCELED_STATE,
                CALL_TIMED_OUT_STATE)
+VALID_STATES = set(filter(lambda state: state != CALL_CANCELED_STATE, CALL_COMPLETE_STATES))
```
mhrivnak: this belongs in ``tasks.py``, because this collection of states is "valid" in the context of that code only.

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +69,32 @@ def get(self, request):
...
+
+        for state in task_state:
+            if state not in VALID_STATES:
+                raise pulp_exceptions.PulpCodedForbiddenException(
+                    error_message=_("Invalid task state given as parameter for delete"))
+
+        for state in task_state:
+                TaskStatus.objects(state=state).delete()
+
+        return generate_json_response(None)
```
mhrivnak: I think this will produce a json document with the value ``null`` and put that in the response body, with a code of 200. That's a behavior seen in other parts of our API, but I don't think it's a good one. As noted in a previous comment, it would be better to return a 204 with an empty body.

shubham90:
ok test

shubham90:
ok test

shubham90:
@mhrivnak I thought its better to be consistent with the current display of successful response. But, your comments made more sense and I have updated the code accordingly. Thanks a lot for the re-review.

shubham90:
ok test

docs/dev-guide/integration/rest-api/tasks.rst
```diff
@@ -119,6 +119,30 @@ All currently running and waiting tasks may be listed. This returns an array of
...
+------------------------
+
+All completed tasks with states *finished, error, timed out, skipped* may be deleted.
+This call returns response code 204 if successful or code 403 if the request is forbidden.
+
+| :method:`delete`
+| :path:`/v2/tasks/`
+| :permission:`delete`
+
+* :param:`?state,str,only delete tasks with state as state parameters`
```
barnabycourt: Can the valid values for this param be listed here? I see them above but it might be clearer as part of the param listing

shubham90: Sure. will add it here.

shubham90:
ok test

server/pulp/server/webservices/views/tasks.py
```diff
@@ -46,12 +54,11 @@ class TaskCollectionView(View):
...
     View for all tasks.
     """
-
     @auth_required(authorization.READ)
     def get(self, request):
         """
-        Return a response containing a list of all tasks or a response containg a list of all tasks
-        that can be filtered by the optional GET parameter 'tags'.
+        Return a response containing a list of all tasks or a response containing
+        a list of all tasks that can be filtered by the optional GET parameter 'tags'.
```
barnabycourt: s/all tasks that can be filtered by/tasks filtered by/

server/pulp/server/webservices/views/tasks.py
```diff
@@ -67,6 +74,32 @@ def get(self, request):
...
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+
+        :return: Response containing None or pulp Exception
+        :rtype:  django.http.HttpResponse or pulp.Exception
+        """
+        task_state = request.GET.getlist('state')
+        if not task_state:
+            raise pulp_exceptions.PulpCodedForbiddenException(
```
barnabycourt: Should this be a validation error or a 400 instead of the 403?

shubham90: I thought it to be 403 because giving no arguments is forbidden. Client can't keep the argument list empty in case of delete. However, http 400 also makes sense because server won't understand the client's request.

shubham90: So, do you suggest this to be 400 instead of 403?

shubham90:
ok test

shubham90:
ok test

shubham90:
@barnabycourt I changed the exceptions as per our discussion. Please review them once you are free. Thanks!

server/pulp/server/webservices/views/tasks.py
```diff
@@ -17,6 +21,10 @@
                                                 generate_json_response_with_pulp_encoder)
 
 
+# This constant set is used for deleting the completed tasks from the collection.
+VALID_STATES = set(filter(lambda state: state != CALL_CANCELED_STATE, CALL_COMPLETE_STATES))
```
barnabycourt: This seems somewhat risky if additional states are added in the future. 50/50 whether the additional state could be deleted or not. Is there some what this could be defined on the Model?

common/pulp/common/error_codes.py
```diff
@@ -113,3 +114,5 @@
 PLP1009 = Error("PLP1009", _("The request body does not contain valid JSON"), [])
 PLP1010 = Error("PLP1010", _("Provided value %(value)s for field %(field)s must be of type "
                              "%(field_type)s."), ["value", "field", "field_type"])
+PLP1011 = Error("PLP1011", _("Invalid task state passed to parameters list: %(state)s"), ["state"])
```
barnabycourt: Would it make sense to include the list of valid states as well?

docs/user-guide/release-notes/master.rst
```diff
@@ -51,6 +51,8 @@ After migrating the database, restart `httpd`, `pulp_workers`, `pulp_celerybeat`
 Rest API Changes
 ----------------
 
+* Tasks with complete states can now be deleted.
```
barnabycourt: This is not strictly accurate as canceled is considered a completed state.

server/pulp/server/webservices/views/tasks.py
shubham90: As an alternative, the states can be defined here instead of importing from constants.

barnabycourt: We really can't support removing canceled tasks?

common/pulp/common/error_codes.py
shubham90: That makes sense. Will add them

server/pulp/server/webservices/views/tasks.py
shubham90: Not at this point of time. We will have to change the way we tag a task as canceled.

shubham90:
ok test

docs/user-guide/release-notes/master.rst
```diff
@@ -51,6 +51,8 @@ After migrating the database, restart `httpd`, `pulp_workers`, `pulp_celerybeat`
 Rest API Changes
 ----------------
 
+* Tasks with complete states(except `canceled` state) can now be deleted.
```
barnabycourt: Missing a space between states and (

shubham90:
ok test

shubham90:
ok test

PR #1984, opened 2015-07-16
ref #1136 - support agent task cancellation.

jortel:
https://pulp.plan.io/issues/1136

During the migration to celery, the agent task cancellation did not get wired up.

PR #1985, opened 2015-07-17
gofer 2.6.3

jortel:
https://pulp.plan.io/issues/1136

Cancellation updated in 2.6.3 to fix an issue with canceling requests to plugins configured with a thread pool capacity=1.  Fix back-ported from gofer master (2.7).


jortel:
Related: https://github.com/pulp/pulp/pull/1984

PR #1986, opened 2015-07-20
remove some code that was never used

asmacdo:
Poking around in the auth stuff as part of a story, I saw this scary looking code and realized that it hasn't been used in any pulp release.


rbarlow:
I didn't verify that it wasn't used, but I believe you. LGTM once Jenkins passes.

PR #1991, opened 2015-07-24
Fix in Unit test for CeleryBeatLock

shubham90:
@asmacdo Here is the patch for the issue. Thanks for pointing this out.


asmacdo:
Feel free to merge when pulpbot is done.

PR #1992, opened 2015-07-27
Users mongoengine

asmacdo:
Replace the Users model with a model that inherits from Mongoengine's Document, implement a serializer, move the user managers to a controller, and convert everything to use the new interface. I will squash commits after review.


shubham90:
@asmacdo  I am having the same issue in Jenkins for rhel6. Let me know if you figure a work around for that.

devel/pulp/devel/unit/server/base.py
```diff
@@ -57,58 +52,3 @@ def _load_test_config():
...
-        """
-        self.mock_user_query_manager.return_value.is_authorized.assert_called_once_with(
-            mock.ANY, mock.ANY, operation)
-
-    def get_mock_uri_path(self, *args):
-        """
-        :param object_id: the id of the object to get the uri for
-        :type object_id: str
-        """
-        return os.path.join('/mock', *args) + '/'
```
jortel: Not sure how removing all of this is related to mongoengine conversion of users but hoping I will later.

server/pulp/server/db/migrations/lib/managers.py
```diff
@@ -77,3 +86,141 @@ def find_with_importer_type(importer_type_id):
...
+        If no super users are found, the default admin user (from the pulp config)
+        is looked up or created and added to the super users role.
+        """
+        role_manager = factory.role_manager()
+        if self.get_admins():
+            return
+
+        default_login = config.config.get('server', 'default_login')
+
+        admin = get_collection('users').find_one({'login': default_login})
```
jortel: Why not use
```
model.User.objects(login=default_login).first()
```

server/pulp/server/db/migrations/lib/managers.py
```diff
@@ -77,3 +86,141 @@ def find_with_importer_type(importer_type_id):
...
+
+    @staticmethod
+    def find_all():
+        """
+        Returns serialized versions of all users in the database.
+
+        @return: list of serialized users
+        @rtype:  list of dict
+        """
+        all_users = list(get_collection('users').find())
```
jortel: Why not use?:
```
model.User.objects()
```

server/pulp/server/db/migrations/lib/managers.py
```diff
@@ -77,3 +86,141 @@ def find_with_importer_type(importer_type_id):
...
+        @type  name: str
+
+        @param roles: list of roles user will belong to
+        @type  roles: list
+
+        @raise DuplicateResource: if there is already a user with the requested login
+        @raise InvalidValue: if any of the fields are unacceptable
+        """
+
+        existing_user = get_collection('users').find_one({'login': login})
```
jortel: Why not use?
```
model.User.objects(login=login).first()
```

server/pulp/server/db/migrations/lib/managers.py
```diff
@@ -77,3 +86,141 @@ def find_with_importer_type(importer_type_id):
...
+        create_me = model.User(login=login, name=name, roles=roles)
+        create_me.set_password(password)
+        create_me.save()
+
+        # Grant permissions
+        permission_manager = factory.permission_manager()
+        permission_manager.grant_automatic_permissions_for_user(create_me.login)
+
+        # Retrieve the user to return the SON object
+        created = get_collection('users').find_one({'login': login})
```
jortel: Why not use?
```
model.User.objects(login=login).first()

File: server/pulp/server/db/migrations/lib/managers.py [hunk 6]
```diff
@@ -77,3 +86,141 @@ def find_with_importer_type(importer_type_id):
...
+        created.pop('password')
+
+        return created
+
+
+def invalid_type(input_value, valid_type):
+    """
```
jortel: Description in docstring.

server/pulp/server/db/model/__init__.py
```diff
@@ -597,3 +607,147 @@ class CeleryBeatLock(Document):
...
+
+        :param plain_password: plaintext password to be hashed
+        :type  plain_password: str
+
+        :return: salt concatenated with the hashed password
+        :rtype:  str
+        """
+        salt = self._random_bytes(8)  # 64 bits
+        hashed_password = self._pbkdf_sha256(str(plain_password), salt, PASSWORD_ITERATIONS)
+        # return the salt and hashed password, encoded in base64 and split with ","
```
jortel: Be careful adding comments that simply state in English what the code is doing.  It just adds code-noise.
Eg:
```
# add two numbers and assign to a variable
a = b + c
```
As the code reader, I have to stop processing code in my mind, context switch to processing English, then back to processing code.  I *can* make the code more effort to read.

I suspect you've never heard this and surprised I am suggesting you actually remove a comment.  Just suggesting that more comments are not always better.  There is a cost to the reader.

server/pulp/server/db/model/__init__.py
```diff
@@ -597,3 +607,147 @@ class CeleryBeatLock(Document):
...
+    """
+    Singleton user class that represents the "system" user (i.e. no user).
+
+    The entire point of this singleton class is that you can generate any number of new ones, and
+    if it is initialized with the same args, then SystemUser(same_args) is SystemUser(same_args)
+    returns True.
+
+    This class cannot inherrit from the Users mongoengine document because the metaclasss
+    does not play well with the mongoengine metaclasses. Fortunately, the only functionality
+    required is access to user instance variables.
```
jortel: Please add instance variables to docstring.

server/pulp/server/db/querysets.py
```diff
@@ -67,6 +67,19 @@ def update(self, *args, **kwargs):
                 # if post_save() is not defined for this particular document, that's ok
                 pass
 
+    def get_or_404(self, **kwargs):
```
jortel: Is the _or_404 redundant to the:
```
:raise pulp_exceptions.MissingResource
```
Not opposed but method names don't typically contain information about raised exceptions.

server/pulp/server/db/querysets.py
```diff
@@ -67,6 +67,19 @@ def update(self, *args, **kwargs):
...
+        """
+        Generally allow querysets to raise a MissingResource exception if getting an object that
+        does not exist.
+
+        :param kwargs: keyword arguments that specify field of the document and value it should be
```
jortel: Line length.

server/pulp/server/managers/auth/authentication.py
```diff
@@ -31,8 +32,7 @@ def _check_username_password_local(self, username, password=None):
         :rtype: L{pulp.server.db.model.auth.User} instance or None
         :return: user corresponding to the credentials
         """
-        user_query_manager = factory.user_query_manager()
-        user = user_query_manager.find_by_login(username)
+        user = model.User.objects(login=username).first()
```
jortel: This really makes the code so much cleaner :)

server/test/unit/server/controllers/test_user.py
```diff
@@ -0,0 +1,282 @@
...
+    Tests for finding a list of users that belong to a role.
+    """
+
+    def test_role_does_not_exist(self, mock_role):
+        """
+        Ensure that a MissingResource is raised if the role does not exist.
+        """
+        mock_role.return_value.find_one.return_value = None
+        self.assertRaises(
+            pulp_exceptions.MissingResource, user_controller.find_users_belonging_to_role, 'role')
```
jortel: line length.

server/test/unit/server/controllers/test_user.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def test_as_expected(self, mock_f, mock_model, mock_last_su):
+        """
+        Test delete that works as expected.
+        """
+        mock_last_su.return_value = False
+        m_permission_manager = mock_f.permission_manager.return_value
+        user_controller.delete_user('curiosity')
+
+        m_permission_manager.revoke_all_permissions_from_user.assert_called_once_with('curiosity')
```
jortel: line length.

jortel:
Very nice work, Austin! 

The code cleaned up nicely and I especially like the tests you wrote from scratch.

Just a few nit-pick comments to consider.  The only blocker would be for comments related to using User.objects() instead of the collection.fetch_one().  There is probably a good explanation or perhaps I just failed to understand something.

server/pulp/server/db/migrations/lib/managers.py
asmacdo: Since this is in the migrations library, I want it to be as independent of the production code as possible and as similar to the old code as possible. This is *almost* identical to the manager code that was removed, so I believe that doing it this way will have the least risk. If we use the mongoengine model code and we change the model, we may inadvertently also change how the database is initialized. If we initialize in the way that we used to, then any changes should be reflected in the migrations. The idea is that once in the migrations lib, the code should not need to change and core code changes shouldn't affect it either.

asmacdo: @bmbouter added a task to document this. https://pulp.plan.io/issues/1220

jortel: Makes sense.

devel/pulp/devel/unit/server/base.py
asmacdo: It doesn't actually, I will remove this dead code in a separate PR.

server/pulp/server/db/model/__init__.py
asmacdo: Good point.

server/pulp/server/db/querysets.py
asmacdo: This is based on a Django shortcut function. 
https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404

Since we handle 404's by raising an exception, this replaces code that is repeated all over the code base:

    thing = model.Collection.objects().first()
    if thing is None:
        raise MissingResource

This does more than simply reduce lines of code. Everywhere that this is used, the function has one fewer code path that needs to be tested. I agree that it is a little weird, but I think that the benefits outweigh the awkwardness. Thoughts?

server/test/unit/server/controllers/test_user.py
asmacdo: All of the line length comments are showing 98 characters for me?

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

dkliban:
ok test

asmacdo:
ok test

dkliban:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
@jortel, it has been a long and glorious battle, but pulpbot and I have resolved our differences and are now friends. That said, Pulpbot will probably fail since I just rebased and did a force push.

Comments taken or responded to, would you mind having another look?

asmacdo:
ok test

dkliban:
ok test

PR #1993, opened 2015-07-28
Adjusts Qpid broker string and includes Kombu SASL fixes

bmbouter:
This includes all the patches from before consolidated into one
patch and also includes the fix for
https://github.com/celery/kombu/issues/498

The pulp broker string for the [tasks] section is adjusted to
not specify 'guest' as the default username. This will force
Kombu to use ANONYMOUS auth by default.

https://pulp.plan.io/issues/1152
closes #1152


bmbouter:
I built this locally with tito and it built successfully for me. I then installed it and tested it with Pulp with a sync+publish and everything worked as expected.

beav:
changes LGTM for myself and @dkliban

PR #1996, opened 2015-07-31
Remove the -e from the bash provisioner.

rbarlow:
For some reason, the -e started causing provisioning failures recently.
I suspect that a Fedora update might have changed the behavior of that
flag, but I have not confirmed this. Since we are slowly switching from
the shell provisioner to Ansible I believe it is best to disable this
flag for now to get around the immediate issue.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -31,6 +31,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         domain.cpus   = 4
     end
 
-    dev.vm.provision "shell", inline: "sudo -u vagrant bash -e /home/vagrant/devel/pulp/playpen/vagrant-setup.sh"
+    dev.vm.provision "shell", inline: "sudo -u vagrant bash /home/vagrant/devel/pulp/playpen/vagrant-setup.sh"
  end
 end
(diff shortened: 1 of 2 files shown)


mhrivnak:
It's very unlikely that the behavior of that flag has changed. Here is its description:

"-e  Exit immediately if a command exits with a non-zero status."

So that indicates some command is exiting non-zero. We should probably chase down why that's happening.

rbarlow:
@mhrivnak yeah I'm aware of what it does (I set it originally on purpose). When the failure occurs, there is no output at all on stdin or stdout. This might mean that it's the very first thing that is failing (building the .bashrc). However, without -e the .bashrc is built correctly.

My thinking is that we are leaving the bash provisioner anyway, and so this particular commit is a "quick fix" for the meantime without investing any time trying to figure out what is happening. If you would rather, I can work on converting the rest of the bash provisioner to ansible until it is either done, or until this error goes away. However, I do not think it is worth the time to try to fix the bash provisioner itself since it is not long for this world.

What do you think given the above?

mhrivnak:
Ok. It sounds like this will leave us better-off than what we currently have, so I'm ok with this incremental change. Hopefully we'll also discover what's causing the error to begin with.

PR #2000, opened 2015-08-03
Convert all the dev bashrc building to Ansible.

rbarlow:
This commit also removes the hard dependency on pulp_rpm and updates the documentation to be more current.


rbarlow:
ok test

PR #2001, opened 2015-08-04
Adds login_method options to [tasks] section of server.conf

bmbouter:
Adds login_method options to [tasks] section of server.conf
    
If login_method is set, the value is used to configure the BROKER_LOGIN_METHOD Celery option. login_method defaults to '' which allows Celery to use its default values.

This change includes docs updates on practical usage of the new setting and a release note.

closes #1168
https://pulp.plan.io/issues/1168


bmbouter:
ok test

bmbouter:
ok test

bmbouter:
This work was rebased against the 2.6-dev and 2.7-testing merge base and will be merged with this PR instead:   https://github.com/pulp/pulp/pull/2002

PR #2006, opened 2015-08-12
Ensures the unit key contains only UTF-8 characters

bmbouter:


barnabycourt:
Should something like this be done for the other unit key generators on the ContentUnit?

barnabycourt:
Closing after discussion with Brian

PR #2007, opened 2015-08-12
Add controller method to disassociate units

barnabycourt:


server/pulp/server/controllers/repository.py
```diff
@@ -146,6 +147,22 @@ def associate_single_unit(repository, unit):
         upsert=True)
 
 
+def disassociate_units(repository, unit_iterable):
+    """
+    Disassociate all units in the iterable from the repository
+
+    :param repository: The repository to update.
+    :type repository: pulp.server.db.model.Repository
+    :param unit_iterable: The unit to associate to the repository.
```
ipanova: s/unit to associate/units to disassociate

PR #2008, opened 2015-08-14
Rework the dev-setup.sh script to use Ansible.

rbarlow:
This commit reworks the dev-setup.sh script to use the same code that
Vagrant uses to provision the dev box.

https://pulp.plan.io/issues/1137


dkliban:
I had to install yum-utils to get this to work. Here is what prompted me to install the package: http://fpaste.org/255940/43983287/

dkliban:
After installing yum-utils I was able to finish the setup process. Only thing I didn't like is that this was done without Vagrant but the last line of output said "Done. You can ssh into your development environment with vagrant ssh."

rbarlow:
On 08/17/2015 01:43 PM, Dennis Kliban wrote:
> Only thing I didn't like is that this was done without Vagrant but the
> last line of output said "Done. You can ssh into your development
> environment with vagrant ssh."

Ah yes, good note. I think I might be able to tell which is happening by
seeing if $USER == "vagrant", and then only printing that then so that
might be easy enough to fix.

-- 
Randy Barlow

rbarlow:
I've added yum-utils to dev-setup.sh, and I've partially resolved the vagrant ssh issue. I made a todo to fully resolve the vagrant ssh messaging issue for later, as I would like to merge this today.

PR #2011, opened 2015-08-25
Removes MongoDB as celery result backend 

dkliban:
This patch also removes the FailureHandler which relied on checking the
results backend to determine if a scheduled task needs to have it's
schedule disabled after reaching a failure threshold. The logic is moved
to the on_success and on_failure methods for Task defined in Pulp.


dkliban:
ok test

PR #2012, opened 2015-08-26
add mongoengine orphan-cleanup support

barnabycourt:


server/pulp/server/managers/content/orphan.py
```diff
@@ -231,6 +239,42 @@ def delete_orphans_by_type(content_type_id, content_unit_ids=None):
...
+        This method only applies to new style content units that are loaded via entry points
+
+        NOTE: this method deletes the content unit's bits from disk, if applicable.
+
+        :param type_id: id of the content type
+        :type type_id: basestring
+        """
+        # get the model matching the type
+        content_model = plugin_api.get_unit_model_by_id(type_id)
+        content_units = content_model.objects().only('id', 'storage_path')
```
bmbouter: FYI, you don't need the `()` on `objects()`. This could be written as `content_model.objects.only('id', 'storage_path')`.

bmbouter:
I tested both list oprhans and delete with `-all` and they both worked for me with this fix. LGTM!

PR #2014, opened 2015-08-27
Adds release note about pulp-admin tasks list --all

dkliban:


docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -107,6 +107,9 @@ Client Changes
 * The `pulp-admin` command now supports obtaining detailed information for just
   one repo via the :ref:`details <admin-client-list-repos>` flag.
 
+* `pulp-admin tasks list` command now returns only tasks in 'waiting' or 'running' state. `--all`
```
bmbouter: Suggested rewrite:

```
* `pulp-admin tasks list` command is adjusted to only display tasks in 'waiting' or 'running' state. Previously all tasks were shown regardless of state. The `--all` flag can be passed in to retrieve a list of tasks in all states.
```

PR #2015, opened 2015-08-27
Drops celery_taskmeta collection

dkliban:


server/test/unit/server/db/migrations/test_0020_drop_celery_taskmeta.py
```diff
@@ -0,0 +1,23 @@
+"""
+This module contains tests for pulp.server.db.migrations.0019_repo_collection_id.py
```
bmbouter: s/0019/0020/

bmbouter:
I think this change needs a release note

bmbouter:
LGTM after release note added and copy/paste error fixed.

bmbouter:
Also the commit should have this in it:

```
Adds migration 0020 that removes the celery_taskmeta collection.

re #1084
https://pulp.plan.io/issues/1084
```

PR #2020, opened 2015-09-01
Adds django settings environment variable to all celery processes

dkliban:
re #1216
https://pulp.plan.io/issues/1216

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 
 from pulp.devel.test_runner import run_tests
+from pulp.server.constants import PULP_DJANGO_SETTINGS_MODULE
 
 # Find and eradicate any existing .pyc files, so they do not eradicate us!
 PROJECT_DIR = os.path.dirname(__file__)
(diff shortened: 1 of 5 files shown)


dkliban:
ok test

server/test/unit/server/async/test_app.py
```diff
@@ -0,0 +1,18 @@
+"""
```
bmbouter: nice module docblock

server/test/unit/server/async/test_app.py
```diff
@@ -0,0 +1,18 @@
...
+from pulp.server.constants import PULP_DJANGO_SETTINGS_MODULE
+
+
+class TestCeleryApp(unittest.TestCase):
+    """
+    Assert that the DJANGO_SETTINGS_MODULE environment variable is set.
+    """
+    @mock.patch('pulp.server.initialization.initialize')
+    def test_django_env_variable(self, mock_initialize):
+        from pulp.server.async import app  # noqa
```
bmbouter: This should be moved up to the module itself. Having it here doesn't really guard against the problem that other test code could have imported a line of code that set the environment variable. If it was already modified having the import statement here versus at the top won't solve that problem. As such, moving it to the top would be good.

bmbouter:
For the same reasons as [this comment](https://github.com/pulp/pulp/pull/2020#discussion_r38424907), the changes made to `server/test/unit/server/async/test_scheduler.py` can probably be reverted. I don't think this is actually asserting much that the test in `server/test/unit/server/async/test_app.py` isn't already.

bmbouter:
I suggested moving the import statement and also reverting the changes in `server/test/unit/server/async/test_scheduler.py`. I don't need to re-review, LGTM.

PR #2021, opened 2015-09-01
Moves post_init functionality to attach_signals.

bmbouter:
This platform changes causes validation to happen earlier
before mongoengine begins its init. This causes a coded
exception to be raised earlier in the case that the model
definition is invalid.


bmbouter:
@barnabycourt thanks for the review.

PR #2023, opened 2015-09-03
Adjust the example Vagrantfile for the new F22 box location.

rbarlow:
I learned in a bug report I filed against vagrant-libvirt that the Fedora 22 Vagrant box we use has moved
locations[0]. This commit updates the example Vagrantfile to use the new location. There are 301 redirects in place
for the old location, so there is no immediate urgency for developers to adopt this change.

[0] https://github.com/pradels/vagrant-libvirt/issues/452#issuecomment-137332077


rbarlow:
ok test

PR #2024, opened 2015-09-03
update pulp-selinux for ostree gpg-related perms

beav:
This patch adds an additional SELinux allow rule so pulp-ostree can invoke the
gpg binary.

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -101,6 +101,7 @@ corecmd_exec_shell(celery_t)
 corecmd_read_bin_symlinks(celery_t)
 dev_read_urand(celery_t)
 files_rw_pid_dirs(celery_t)
+gpg_exec(celery_t) # needed for ostree
 kernel_read_system_state(celery_t)
 libs_exec_ldconfig(celery_t)
 logging_send_syslog_msg(celery_t)


server/selinux/server/pulp-celery.te
```diff
@@ -59,6 +60,8 @@ allow celery_t pulp_var_run_t:file manage_file_perms;
 allow celery_t pulp_var_run_t:dir manage_dir_perms;
 files_pid_filetrans(celery_t, pulp_var_run_t, file)
 
+# needed for ostree
+allow celery_t gpg_exec_t:file { read execute open execute_no_trans };
```
bmbouter: Is there a refpol statement that is equivalent to this? Usually I ask in #selinux on freenode or on tech-list. It would make it look more [line #55](https://github.com/beav/pulp/blob/e677845b69189309f570b6a3a0650536c6e91d92/server/selinux/server/pulp-celery.te#L55).

It would be more like allow:

```
celery_t gpg_exec_t:file exec;
```

Or maybe a macro to have celery_t receive the execute like [line #98](https://github.com/beav/pulp/blob/e677845b69189309f570b6a3a0650536c6e91d92/server/selinux/server/pulp-celery.te#L98)

bmbouter: Can this be rewritten and moved into the lower section. After asking in #selinux it was suggested to use:

```
gpg_exec(celery_t) ;
```

Also can it be moved into the [lower section](https://github.com/beav/pulp/blob/ostree-gpg-selinux/server/selinux/server/pulp-celery.te#L98-L112) with a comment to the right of the statement.

beav: Thanks for investigating this. I will make the changes now.

bmbouter:
@beav LGTM, do you want to merge it and merge forward?

beav:
yup will do. thanks for the review :koala:

PR #2026, opened 2015-09-03
Add provider to shared storage location.

jortel:
https://pulp.plan.io/issues/1233

Add provider to shared storage and include it in the path.  This needed to be propagated to the shared content unit class.

### DIFF
--- a/server/pulp/server/content/storage.py
+++ b/server/pulp/server/content/storage.py
@@ -116,17 +116,24 @@ class SharedStorage(ContentStorage):
     """
     Direct shared storage.
 
+    :ivar provider: A storage provider.
+        This defines the storage mechanism and qualifies the storage_id.
+    :type provider: str
     :ivar storage_id: A shared storage identifier.
     :ivar storage_id: str
     """
 
-    def __init__(self, storage_id):
+    def __init__(self, provider, storage_id):
         """
+        :param provider: A storage provider.
+            This defines the storage mechanism and qualifies the storage_id.
+        :type provider: str
         :param storage_id: A shared storage identifier.
         :ivar storage_id: str
         """
         super(SharedStorage, self).__init__()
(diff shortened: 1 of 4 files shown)


server/test/unit/server/db/test_model.py
```diff
@@ -237,6 +237,11 @@ class TestUnit(model.SharedContentUnit):
     def test_abstract(self):
         unit = TestSharedContentUnit.TestUnit()
         try:
+            unit.storage_provider
```
ipanova: would not it be better to split this test into two tescases- test_storage_provider and test_storage_id? to avoid 2 try/except that are testing 2 different things.

server/pulp/server/content/storage.py
```diff
@@ -116,17 +116,24 @@ class SharedStorage(ContentStorage):
...
 
+    :ivar provider: A storage provider.
+        This defines the storage mechanism and qualifies the storage_id.
+    :type provider: str
     :ivar storage_id: A shared storage identifier.
     :ivar storage_id: str
     """
 
-    def __init__(self, storage_id):
+    def __init__(self, provider, storage_id):
```
ipanova: i would name it 'storage_provider' to avoid confusion in the future

ipanova:
@jortel i proposed some suggestions, no need to re-review,LGTM

server/test/unit/server/db/test_model.py
```diff
@@ -237,6 +237,11 @@ class TestUnit(model.SharedContentUnit):
     def test_abstract(self):
         unit = TestSharedContentUnit.TestUnit()
         try:
```
mhrivnak: you could still use ```assertRaises``` here by using ```getattr``` as the callable.

```
self.assertRaises(NotImplementedError, getattr, unit, 'storage_provider')
```

PR #2029, opened 2015-09-04
Fixed the unit unassociate workflow to work with mongoengine models and filters

mhrivnak:
fixes #1212 

https://pulp.plan.io/issues/1212

### DIFF
--- a/server/pulp/server/managers/repo/unit_association.py
+++ b/server/pulp/server/managers/repo/unit_association.py
@@ -414,15 +414,21 @@ def calculate_associated_type_ids(source_repo_id, associated_units):
 
 
 def create_transfer_units(associate_units, associated_unit_type_ids):
-    type_defs = {}
-    for def_id in associated_unit_type_ids:
-        type_def = types_db.type_definition(def_id)
-        type_defs[def_id] = type_def
+    unit_key_fields = {}
+    for unit_type_id in associated_unit_type_ids:
+        type_def = types_db.type_definition(unit_type_id)
+        if type_def is not None:
+            # this is an "old style" model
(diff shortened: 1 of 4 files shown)


server/pulp/server/managers/repo/unit_association.py
```diff
@@ -414,15 +414,21 @@ def calculate_associated_type_ids(source_repo_id, associated_units):
...
 
 def create_transfer_units(associate_units, associated_unit_type_ids):
-    type_defs = {}
-    for def_id in associated_unit_type_ids:
-        type_def = types_db.type_definition(def_id)
-        type_defs[def_id] = type_def
+    unit_key_fields = {}
+    for unit_type_id in associated_unit_type_ids:
+        type_def = types_db.type_definition(unit_type_id)
+        if type_def is not None:
```
bmbouter: Consider simplifying the `if` statement to not have a negation. The new-style could become the `if` codepath with

```
if type_def is None:
```

bmbouter:
I tested this against the puppet mongoengine reproducer described in the issue and it resolved the problem. Great job, LGTM.

server/pulp/server/managers/repo/unit_association.py
mhrivnak: I considered that and initially went that way, but here's why I changed my mind. As-is, the code that gets the type_def is near the code that uses it. If that doesn't work, we move on to the code that gets the model and then uses it. Ignoring control statements, it can be reduced to:

get A
use A
get B
use B

I think this a more friendly and readable layout than the alternative, but it's definitely a subtlety.

bmbouter: cool

PR #2030, opened 2015-09-09
fixed orphan removal by type when the type uses mongoengine

mhrivnak:
https://pulp.plan.io/issues/1215

Several places in the code, only two of which I touched here, need to know the
unit key fields for a given type ID. I created a new controller function that
takes care of finding that information regardless of whether the type is
defined the new way (mongonengine) or old way. The next step will be to use
that controller function in additional places, but that is beyond the scope of
this bug fix.

fixes #1215

### DIFF
--- a/server/pulp/server/controllers/units.py
+++ b/server/pulp/server/controllers/units.py
@@ -1,5 +1,7 @@
 import mongoengine
 
+from pulp.plugins.loader import api as plugin_api
+from pulp.plugins.types import database as types_db
 from pulp.plugins.util import misc
 
 
@@ -37,3 +39,30 @@ def find_units(units, pagination_size=50):
 
         for found_unit in query:
             yield found_unit
+
+
+def get_unit_key_fields_for_type(type_id):
+    """
+    Based on a unit type ID, determine the fields that compose that type's unit key.
+
+    This supports both the new mongoengine models and the old "types_def" collection, so the caller
+    need not worry about whether a type has been converted to use mongoengine or not.
+
+    :param type_id: unique ID for a unit type
+    :type  type_id: str
+
+    :return:    tuple containing the name of each field in the unit key
+    :rtype:     tuple
+
+    :raises ValueError: if the type ID is not found
+    """
+    model_class = plugin_api.get_unit_model_by_id(type_id)
+    if model_class is not None:
+        return model_class.unit_key_fields
+
+    type_def = types_db.type_definition(type_id)
+    if type_def is not None:
+        # this is an "old style" model
+        return tuple(type_def['unit_key'])
+
+    raise ValueError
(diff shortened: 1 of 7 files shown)


mhrivnak:
FYI I have a fix for https://pulp.plan.io/issues/1239 that is ready to submit, but it depends on this being merged.

server/pulp/server/controllers/units.py
```diff
@@ -37,3 +39,30 @@ def find_units(units, pagination_size=50):
 
         for found_unit in query:
             yield found_unit
+
+
+def get_unit_key_fields_for_type(type_id):
+    """
+    Based on a unit type ID, determine the fields that compose that type's unit key.
+
+    This uses both the new mongoengine models and the old "types_def" collection, so the caller
```
bmbouter: s/uses/supports/

server/test/unit/server/controllers/test_units.py
```diff
@@ -57,3 +57,38 @@ def test_results(self):
...
+@patch('pulp.plugins.types.database.type_definition', spec_set=True)
+class TestGetUnitKeyFieldsForType(unittest.TestCase):
+    def test_returns_from_model(self, mock_type_def, mock_get_model):
+        """
+        test when the requested type is a mongoengine model
+        """
+        mock_get_model.return_value = DemoModel
+        mock_type_def.return_value = None
+
+        ret = units_controller.get_unit_key_fields_for_type('demo_model')
```
bmbouter: maybe deduplicate the string by making this:

```
ret = units_controller.get_unit_key_fields_for_type(DemoModel.unit_type_id)
```

server/test/unit/server/controllers/test_units.py
```diff
@@ -57,3 +57,38 @@ def test_results(self):
...
+
+        self.assertEqual(ret, ('key_field',))
```
bmbouter: maybe deduplicate the definition by using `self.assertEqual(ret, DemoModel.unit_key_fields)`

server/test/unit/server/webservices/views/test_content.py
```diff
@@ -137,13 +137,14 @@ def test_get_orphan_type_subcollection_with_empty_list(self, mock_factory, mock_
...
-    @mock.patch('pulp.server.webservices.views.content.factory.content_orphan_manager')
+    @mock.patch('pulp.server.controllers.units.get_unit_key_fields_for_type', spec_set=True)
     @mock.patch('pulp.server.webservices.views.content.content_orphan')
-    def test_delete_orphan_type_subcollection(self, mock_orphan_manager, mock_type):
+    def test_delete_orphan_type_subcollection(self, mock_orphan_manager, mock_get_unit_key_fields):
         """
         Delete orphans should be called with the correct arguments and OperationPostponed is raised.
         """
         request = mock.MagicMock()
+        mock_get_unit_key_fields.return_value = tuple('id',)
```
bmbouter: I like that you used `tuple` explicitly here

bmbouter:
I tested the fix and it resolved the problem. Great job. I left some optional feedback depending on what you wanted to do. LGTM.

PR #2031, opened 2015-09-09
convert consumer register to bindings

asmacdo:


asmacdo:
ok test

bindings/pulp/bindings/static.py
```diff
@@ -0,0 +1,18 @@
+from pulp.bindings.base import PulpAPI
+
+
+class StaticRequest(PulpAPI):
+    """
+    Connection class to access static calls
+    """
+    def __init__(self, pulp_connection):
+        super(StaticRequest, self).__init__(pulp_connection)
```
rbarlow: This method doesn't seem to do anything specific to this subclass. I recommend removing it.

bindings/pulp/bindings/server.py
```diff
@@ -223,7 +225,7 @@ def _build_url(self, path, queries=()):
                     of self.path_prefix
         """
         # build the request url from the path and queries dict or tuple
-        if not path.startswith(self.path_prefix):
+        if not path.startswith(self.path_prefix) and not ignore_prefix:
```
rbarlow: You've added this ignore_prefix throughout this file without documenting it in any docblocks. Every method that this appears in needs to have it documented.

bindings/pulp/bindings/static.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def get_server_key(self):
+        """
+        Retrieve the server's public key.
+        """
```
rbarlow: Add returns and rtype documentation to this.

bindings/pulp/bindings/static.py
```diff
@@ -0,0 +1,18 @@
...
+        return self.server.GET('/pulp/static/rsa_pub.key', ignore_prefix=True)
+
+
+
```
rbarlow: Remove these empty lines at the end of the file - it will cause a flake8 error.

client_consumer/pulp/client/consumer/cli.py
```diff
@@ -91,43 +91,39 @@ def initialize(context):
...
-        fp = open(location, 'w+')
-        try:
-            fp.write(content)
-        finally:
-            fp.close()
+        os.makedirs(os.path.dirname(location))
+    except OSError, e:
+        if e.errno != errno.EEXIST:
+            raise
+    fp = open(location, 'w+')
```
rbarlow: This open needs to be guarded in the try/finally close below so you can be sure that it gets closed.

client_consumer/pulp/client/consumer/cli.py
```diff
@@ -91,43 +91,39 @@ def initialize(context):
...
+    try:
+        fp.write(content)
     finally:
-        request.close()
+        fp.close()
 
 
-def update_server_key(conf):
+def update_server_key(self):
```
rbarlow: This appears to be a module-level function. What is self? I recommend not using the name self here because of the convention for using it to reference a class instance on methods.

client_consumer/pulp/client/consumer/cli.py
```diff
@@ -91,43 +91,39 @@ def initialize(context):
...
     """
-    Download the server's RSA key and store in the location
-    specified in the configuration.
-    :param conf: The consumer configuration object.
-    :type conf: dict
+    Ensure that the server's public key stored on the consumer is up to date.
     """
```
rbarlow: You need to document the parameter that this function needs.

client_consumer/test/unit/test_cli.py
```diff
@@ -0,0 +1,113 @@
...
+            pass
+
+        mock_e = MockException()
+        mock_e.errno = 17
+        mock_os.makedirs.side_effect = mock_e
+        mock_fp = open.return_value
+
+        cli.write_to_location('test/loc', 'content')
+        mock_fp.write.assert_called_once_with('content')
+        mock_fp.close.assert_called_once_with()
```
rbarlow: You haven't asserted that the correct call to open() was made, only that the correct data was passed to it's write(). I recommend adding a test that the correct calls to open() and mkdir() were made.

client_consumer/test/unit/test_cli.py
```diff
@@ -0,0 +1,113 @@
...
+
+        class MockException(Exception):
+            pass
+
+        mock_fp = open.return_value
+        mock_fp.write.side_effect = MockException
+        try:
+            cli.write_to_location('test/loc', 'content')
+        except:
+            pass
```
rbarlow: If you use self.assertRaises here, it still accomplishes what you need and also eliminates three lines of code.

client_consumer/test/unit/test_cli.py
```diff
@@ -0,0 +1,113 @@
...
+    @mock.patch('pulp.client.consumer.cli.os')
+    @mock.patch('__builtin__.open', new_callable=mock.MagicMock())
+    def test_as_expected(self, *_):
+        """
+        When everything works as expected, ensure that the file is closed.
+        """
+        mock_fp = open.return_value
+        cli.write_to_location('test/loc', 'content')
+        mock_fp.write.assert_called_once_with('content')
+        mock_fp.close.assert_called_once_with()
```
rbarlow: You should also assert here that the correct path was opened, and that mkdir was also called with the correct path.

client_consumer/test/unit/test_cli.py
```diff
@@ -0,0 +1,113 @@
...
+        Everything is as expected, content is written to the location in the config file.
+        """
+        mock_cmd = mock.MagicMock()
+        key_response = mock_cmd.context.server.static.get_server_key.return_value
+        key_loc = mock_cmd.context.config['getter']['getter']
+        cli.update_server_key(mock_cmd)
+        mock_write.assert_called_once_with(key_loc, key_response.response_body)
+
+    @mock.patch('pulp.client.consumer.cli.write_to_location')
+    def test_binding_excpeption(self, mock_write):
```
rbarlow: s/excpeption/exception/

rbarlow:
Let me know when you have addressed the comments I've made so far, and I'll take another look.

bindings/pulp/bindings/server.py
asmacdo: Since these have no docblocks, I did not want to add an incomplete/inaccurate docblock, and I also didn't want to spend the time to investigate each of these params just to add it.

rbarlow:
@asmacdo Feel free to merge after squashing your commits. Nice work!

PR #2033, opened 2015-09-10
sort sync and publish history by an indexed key

asmacdo:
fixes #1043 

Original PR: https://github.com/pulp/pulp/pull/1965, reopened against 2.6-dev so it won't hold up the release.


asmacdo:
ok test

PR #2036, opened 2015-09-11
Adds error logging to platform base FileDistributor exceptions

bmbouter:
Write the exception and its traceback to the log at error level
when FileDistributor publish has an error.

https://pulp.plan.io/issues/1248
closes #1248

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -107,6 +107,7 @@ class FileDistributor(Distributor):
             progress_report.state = progress_report.STATE_COMPLETE
             return progress_report.build_final_report()
         except Exception, e:
+            _logger.exception(e)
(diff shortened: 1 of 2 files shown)


server/pulp/plugins/file/distributor.py
```diff
@@ -107,6 +108,8 @@ def publish_repo(self, repo, publish_conduit, config):
             progress_report.state = progress_report.STATE_COMPLETE
             return progress_report.build_final_report()
         except Exception, e:
+            _logger.error(str(e))
```
mhrivnak: I think ```_logger.exception(e)``` may be able to replace these two lines.

bmbouter: This is exactly what I was looking for; great!

PR #2037, opened 2015-09-11
fixing more places that used the old-style types db

mhrivnak:


server/pulp/server/managers/content/query.py
```diff
@@ -22,11 +22,6 @@ def list_content_types(self):
         """
         return content_types_db.all_type_ids()
 
-    def get_content_type(self, type_id):
```
mhrivnak: this isn't used anywhere, isn't documented, and obviously has no value based on the code you see below.

mhrivnak: oh, and of course it also wasn't tested.

bmbouter: :+1:

bmbouter:
@mhrivnak LGTM! Good fix.

PR #2040, opened 2015-09-14
Added saslwrapper to the external deps list

dkliban:


bmbouter:
The json is valid, and it looks like this points at "the right SRPM in Koji":http://koji.katello.org/koji/buildinfo?buildID=8174 which will contain python-saslwrapper.

bmbouter:
@dkliban LGTM

PR #2041, opened 2015-09-15
created fork of FileDistributor that works with models

mhrivnak:
https://pulp.plan.io/issues/1250

There are two commits in this PR. The first just makes a straight copy of the original distributor, and the second modifies it. Keeping both commits makes it clear what I had to change.

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -18,6 +18,8 @@ _logger = logging.getLogger(__name__)
 
 class FileDistributor(Distributor):
     """
+    DEPRECATED: please use the model_distributor.FileDistributor
+
     Distribute files on the filesystem.
     """
 
--- a/server/pulp/plugins/file/model_distributor.py
+++ b/server/pulp/plugins/file/model_distributor.py
@@ -4,12 +4,13 @@ import errno
 import logging
 import os
 import shutil
-import sys
 import traceback
 
 from pulp.common.plugins.distributor_constants import MANIFEST_FILENAME
-from pulp.common.plugins.progress import ProgressReport
 from pulp.plugins.distributor import Distributor
+from pulp.plugins.file.distributor import FilePublishProgressReport
+from pulp.server.controllers import repository as repo_controller
+from pulp.server.managers.repo import _common as common_utils
 
 
 BUILD_DIRNAME = 'build'
@@ -45,7 +46,7 @@ class FileDistributor(Distributor):
         Please also see the superclass method definition for more documentation on this method.
 
         :param repo: metadata describing the repository
-        :type  repo: pulp.plugins.model.Repository
+        :type  repo: pulp.server.db.model.Repository
 
         :param config: plugin configuration
         :type  config: pulp.plugins.config.PluginCallConfiguration
@@ -72,12 +73,12 @@ class FileDistributor(Distributor):
 
         try:
             progress_report.state = progress_report.STATE_IN_PROGRESS
-            units = publish_conduit.get_units()
+            repo_model = repo.repo_obj
+            units = repo_controller.find_repo_content_units(repo_model, yield_content_unit=True)
 
             # Set up an empty build_dir
-            build_dir = os.path.join(repo.working_dir, BUILD_DIRNAME)
-            # Let's erase the path at build_dir so we can be sure it's a clean directory
-            self._rmtree_if_exists(build_dir)
+            working_dir = common_utils.get_working_directory()
+            build_dir = os.path.join(working_dir, BUILD_DIRNAME)
             os.makedirs(build_dir)
 
             self.initialize_metadata(build_dir)
@@ -95,21 +96,17 @@ class FileDistributor(Distributor):
             # Let's unpublish, and then republish
             self.unpublish_repo(repo, config)
 
-            hosting_locations = self.get_hosting_locations(repo, config)
+            hosting_locations = self.get_hosting_locations(repo_model, config)
             for location in hosting_locations:
                 shutil.copytree(build_dir, location, symlinks=True)
 
-            self.post_repo_publish(repo, config)
-
-            # Clean up our build_dir
-            self._rmtree_if_exists(build_dir)
+            self.post_repo_publish(repo_model, config)
 
             # Report that we are done
             progress_report.state = progress_report.STATE_COMPLETE
             return progress_report.build_final_report()
         except Exception, e:
-            _logger.error(str(e))
-            _logger.error(''.join(traceback.format_exception(*sys.exc_info())))
+            _logger.exception(e)
             # Something failed. Let's put an error message on the report
             progress_report.error_message = str(e)
             progress_report.traceback = traceback.format_exc()
@@ -123,13 +120,15 @@ class FileDistributor(Distributor):
 
         Please also see the superclass method definition for more documentation on this method.
 
-        :param repo: metadata describing the repository
+        :param repo: Repository model object
         :type  repo: pulp.plugins.model.Repository
 
         :param config: plugin configuration
         :type  config: pulp.plugins.config.PluginCallConfiguration
         """
-        hosting_locations = self.get_hosting_locations(repo, config)
+        repo_model = repo.repo_obj
+
+        hosting_locations = self.get_hosting_locations(repo_model, config)
         for location in hosting_locations:
             self._rmtree_if_exists(location)
 
@@ -152,7 +151,7 @@ class FileDistributor(Distributor):
         This should be writing to open file handles from the initialize_metadata call
 
         :param unit: the unit for which metadata needs to be generated
-        :type unit: pulp.plugins.model.AssociatedUnit
+        :type unit: pulp.server.db.model.ContentUnit
         """
         self.metadata_csv_writer.writerow([unit.unit_key['name'], unit.unit_key['checksum'],
                                            unit.unit_key['size']])
@@ -169,7 +168,7 @@ class FileDistributor(Distributor):
         Get the paths within a target directory where this unit should be linked to
 
         :param unit: The unit for which we want to return target paths
-        :type unit: pulp.plugins.model.AssociatedUnit
+        :type unit: pulp.server.db.model.ContentUnit
         :return: a list of paths the unit should be linked to
         :rtype: list of str
         """
@@ -179,8 +178,8 @@ class FileDistributor(Distributor):
         """
         Get the paths on the filesystem where the build directory should be copied
 
-        :param repo: The repository that is going to be hosted
-        :type repo: pulp.plugins.model.Repository
+        :param repo: Repository model object
+        :type  repo: pulp.server.db.model.Repository
         :param config:    plugin configuration
         :type  config:    pulp.plugins.config.PluginConfiguration
         :return : list of paths on the filesystem where the build directory should be copied
@@ -193,8 +192,8 @@ class FileDistributor(Distributor):
         API method that is called after the contents of a published repo have
         been moved into place on the filesystem
 
-        :param repo: The repository that is going to be hosted
-        :type repo: pulp.plugins.model.Repository
+        :param repo: Repository model object
+        :type  repo: pulp.server.db.model.Repository
         :param config: the configuration for the repository
         :type  config: pulp.plugins.config.PluginCallConfiguration
         """
@@ -208,12 +207,11 @@ class FileDistributor(Distributor):
                           This path should already exist.
         :type  build_dir: basestring
         :param unit:     The unit to be symlinked
-        :type  unit:     pulp.plugins.model.AssociatedUnit
+        :type unit:      pulp.server.db.model.ContentUnit
         :param target_paths: The list of paths the unit should be symlinked to.
         :type  target_paths: list of L{str}
         """
         for target_path in target_paths:
-            # symlink_filename = os.path.join(build_dir, unit.unit_key['name'])
             symlink_filename = os.path.join(build_dir, target_path)
             if os.path.exists(symlink_filename) or os.path.islink(symlink_filename):
                 # There's already something there with the desired symlink filename. Let's try and
@@ -228,7 +226,7 @@ class FileDistributor(Distributor):
                     os.remove(symlink_filename)
                 except OSError, e:
                     # This will happen if we attempt to call readlink() on a file that wasn't a
-                    # symlink.  We should remove the file and add the symlink. There error code
+                    # symlink. We should remove the file and add the symlink. The error code
                     # should be EINVAL.  If it isn't, something else is wrong and we should raise.
                     if e.errno != errno.EINVAL:
                         raise e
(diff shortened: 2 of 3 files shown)


server/pulp/plugins/file/distributor.py
```diff
@@ -18,6 +18,8 @@
 
 class FileDistributor(Distributor):
     """
+    DEPRECATED: please use the model_distributor.FileDistributor
```
bmbouter: What do you think about in logged `DeprecationWarning` or `PendingDeprecationWarning` into the `__init__` of `FileDistributor`. That way, users relying on code that will be removed can raise the issue with the plugin developers?

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+import errno
+import logging
+import os
+import shutil
+import traceback
+
+from pulp.common.plugins.distributor_constants import MANIFEST_FILENAME
+from pulp.plugins.distributor import Distributor
+from pulp.plugins.file.distributor import FilePublishProgressReport
+from pulp.server.controllers import repository as repo_controller
```
bmbouter: Consider adding a newline here to group non-standard third-part deps separately from modules provided by this package's entry point.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+    def __init__(self):
+        super(FileDistributor, self).__init__()
+        # Initialize instance variables used for writing out the metadata
+        self.metadata_file = None
+        self.metadata_csv_writer = None
+
+    @classmethod
+    def metadata(cls):
+        """
+        Advertise the capabilities of the mighty FileDistributor.
```
bmbouter: I know you didn't originate this statement, but anyway. `s/mighty //`

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+
+        :return: The description of the impressive FileDistributor's capabilities.
```
bmbouter: same here. `s/impressive //`

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+        :rtype:  dict
+        """
+        raise NotImplementedError()
+
+    def distributor_removed(self, repo, config):
+        """
+        Delete the published files from our filesystem
```
bmbouter: s/our/the/

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+
+        :param repo:            metadata describing the repo
+        :type  repo:            pulp.server.db.model.Repository
+        :param publish_conduit: The conduit for publishing a repo
+        :type  publish_conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+        :param config:          plugin configuration
+        :type  config:          pulp.plugins.config.PluginConfiguration
+        :param config_conduit: Configuration Conduit;
+        :type config_conduit: pulp.plugins.conduits.repo_validate.RepoConfigConduit
+        :return:                report describing the publish operation
```
bmbouter: Consider adding a commented newline between arguments and return values.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+        :type  config: pulp.plugins.config.PluginCallConfiguration
+        """
+        self.unpublish_repo(repo, config)
+
+    def publish_repo(self, repo, publish_conduit, config):
+        """
+        Publish the repository.
+
+        :param repo:            metadata describing the repo
+        :type  repo:            pulp.server.db.model.Repository
```
bmbouter: :+1: to using the new Repository model

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+        :type  repo: pulp.server.db.model.Repository
+
+        :param config: plugin configuration
+        :type  config: pulp.plugins.config.PluginCallConfiguration
+        """
+        hosting_locations = self.get_hosting_locations(repo, config)
+        for location in hosting_locations:
+            self._rmtree_if_exists(location)
+
+    def validate_config(self, repo, config, config_conduit):
```
bmbouter: Also not something you created, but I wouldn't know how to implement this since it has no docstring.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+            self.unpublish_repo(repo_model, config)
+
+            hosting_locations = self.get_hosting_locations(repo_model, config)
+            for location in hosting_locations:
+                shutil.copytree(build_dir, location, symlinks=True)
+
+            self.post_repo_publish(repo_model, config)
+
+            # Clean up our build_dir
+            common_utils.delete_working_directory()
```
bmbouter: We probably don't need states like this anymore. Working dirs are automatically deleted upon task success/failure and also worker start.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+        :param repo:            metadata describing the repo
+        :type  repo:            pulp.server.db.model.Repository
+        :param publish_conduit: The conduit for publishing a repo
+        :type  publish_conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+        :param config:          plugin configuration
+        :type  config:          pulp.plugins.config.PluginConfiguration
+        :param config_conduit: Configuration Conduit;
+        :type config_conduit: pulp.plugins.conduits.repo_validate.RepoConfigConduit
+        :return:                report describing the publish operation
+        :rtype:                 pulp.plugins.model.PublishReport
```
bmbouter: Should a `:raises MissingResource` be added to this method since it can be raised by `get_repo_or_missing_resource`? I'm not sure what is expected since the raise statement isn't actually in this function.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def publish_metadata_for_unit(self, unit):
+        """
+        Publish the metadata for a single unit.
+        This should be writing to open file handles from the initialize_metadata call
+
+        :param unit: the unit for which metadata needs to be generated
+        :type unit: pulp.server.db.model.ContentUnit
+        """
+        self.metadata_csv_writer.writerow([unit.unit_key['name'], unit.unit_key['checksum'],
```
bmbouter: This way will work, but I usually use `unit.name` instead of `unit.unit_key['name']`. Similarly for `checksum` and `size`.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+    def get_paths_for_unit(self, unit):
+        """
+        Get the paths within a target directory where this unit should be linked to
+
+        :param unit: The unit for which we want to return target paths
+        :type unit: pulp.server.db.model.ContentUnit
+        :return: a list of paths the unit should be linked to
+        :rtype: list of str
+        """
+        return [unit.unit_key['name'], ]
```
bmbouter: Consider using `unit.name` instead of `unit.unit_key['name']`.

bmbouter: You don't need the comma in this list definition.

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+        self.temp_dir = tempfile.mkdtemp()
+
+        self.target_dir = os.path.join(self.temp_dir, "target")
+        self.repo = MagicMock(spec=Repository)
+        self.repo.id = "foo"
+        self.repo.working_dir = self.temp_dir
+        self.unit = MagicMock(unit_type_id='RPM', name=SAMPLE_RPM, size=1, checksum='sum1',
+                              unit_key={'name': SAMPLE_RPM, 'checksum': 'sum1', 'size': 1})
+        self.unit.storage_path = os.path.join(__file__, DATA_DIR, SAMPLE_RPM)
+        self.publish_conduit = get_publish_conduit(existing_units=[self.unit, ])
```
bmbouter: Maybe this is a style thing and not a PEP8 guideline I can't remember, but you don't need the comma here.

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.temp_dir)
+
+    def create_distributor_with_mocked_api_calls(self):
+        distributor = FileDistributor()
+        distributor.get_hosting_locations = Mock()
+        distributor.get_hosting_locations.return_value = [self.target_dir, ]
```
bmbouter: Also no comma needed here

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def test_get_hosting_locations_not_implemented(self):
+        distributor = FileDistributor()
+        host_locations = distributor.get_hosting_locations(None, None)
+        self.assertEquals(0, len(host_locations))
+
+    def test_post_repo_publish_not_broken(self):
+        distributor = FileDistributor()
+        # ensure that this doesn't raise an error
+        distributor.post_repo_publish(None, None)
```
bmbouter: Consider having this call wrapped in a try/except that calls self.fail() in except so that the test fails with a failure instead of an error.

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @patch('pulp.server.db.model.Repository.objects')
+    @patch('pulp.plugins.file.model_distributor._logger')
+    def test_repo_publish_handles_errors(self, mock_logger, mock_get):
+        """
+        Make sure that publish() does the right thing with the report when there is an error.
+        """
+        distributor = self.create_distributor_with_mocked_api_calls()
+
+        mock_get.get_repo_or_missing_resource.side_effect = Exception('Rawr!')
```
bmbouter: Rawr! :rabbit2:

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Now let's remove a_directory
+        distributor._rmtree_if_exists(a_directory)
+        self.assertFalse(os.path.exists(a_directory))
+
+    def test__symlink_units(self):
+        """
+        Make sure that the _symlink_units creates all the correct symlinks.
+        """
+
```
bmbouter: Consider removing this extra whitespace

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+    def test__rmtree_if_exists(self):
+        """
+        Let's just make sure this simple thing doesn't barf.
+        """
+        a_directory = os.path.join(self.temp_dir, 'a_directory')
+        test_filename = os.path.join(a_directory, 'test.txt')
+        os.makedirs(a_directory)
+        with open(test_filename, 'w') as test:
+            test.write("Please don't barf.")
```
bmbouter: :hamburger:

server/test/unit/plugins/file/test_model_distributor.py
```diff
@@ -0,0 +1,256 @@
...
+        distributor = self.create_distributor_with_mocked_api_calls()
+
+        # There's some logic in _symlink_units to handle preexisting files and symlinks, so let's
+        # create some fakes to see if it does the right thing
+        build_dir = os.path.join(self.temp_dir, BUILD_DIRNAME)
+        os.makedirs(build_dir)
+        os.symlink('/some/weird/path',
+                   os.path.join(build_dir, self.unit.unit_key['name']))
+
+        distributor._symlink_unit(build_dir, self.unit, [self.unit.unit_key['name'], ])
```
bmbouter: comma not needed in list definition

server/pulp/plugins/file/model_distributor.py
bmbouter: Actually how can this work? The caller would need to also switch its type, but I don't see where that is done in this PR. Also when I run this code with the mongoengine-converted puppet branch I see repo is typed as `pulp.server.db.model.Repository`.

In my stack trace the caller into publish_repo comes from [_do_publish()](https://github.com/bmbouter/pulp/blob/18430255f42e854549fe684a2d19e8ea149e6feb/server/pulp/server/controllers/repository.py#L766) which does have both the `repo_obj` and `transfer_repo`. I think switching it will cause the type to switch for other polymorphic distributors so I'm not sure we can actually change this now. What do you think?

bmbouter:
@mhrivnak I tested the code and it works for me. I left a lot of comments, most of them aren't important. The things I do thing should be fixed/investigated are:

[Adding deprecation warnings](https://github.com/pulp/pulp/pull/2041#discussion_r39647414)
[Making sure the documented types throughout the new code aligns with the actual type being used](https://github.com/pulp/pulp/pull/2041#discussion_r39660013)
[Moving away from unit.unit_key['foo']](https://github.com/pulp/pulp/pull/2041#discussion_r39657969)

If you can clearly resolve all of the issues LGTM, or if you want me to look at it again after revision I can do that. I'm not putting on the LGTM tag because some things do need resolution before merging, but I am also not requiring a re-review if you're confident on the resolution.

mhrivnak:
Thanks for the thorough review. As you observed, most of this distributor and unit test code pre-dates this PR. I think we should avoid the temptation to make improvements outside the scope of the problem at hand, which is getting the pre-existing FileDistributor to work with mongoengine models. I tried to make sure the second of the two commits is clearly focused on that functional change. In that spirit, I hope you don't mind that I'm going to leave the stylistic rough spots as they are.

I will look carefully though at the behavioral questions you raised.

server/pulp/plugins/file/model_distributor.py
```diff
@@ -0,0 +1,249 @@
...
+        :type config_conduit: pulp.plugins.conduits.repo_validate.RepoConfigConduit
+        :return:                report describing the publish operation
+        :rtype:                 pulp.plugins.model.PublishReport
+        """
+        progress_report = FilePublishProgressReport(publish_conduit)
+        _logger.info(_('Beginning publish for repository <%(repo)s>') % {'repo': repo.id})
+
+        try:
+            progress_report.state = progress_report.STATE_IN_PROGRESS
+            repo_model = model.Repository.objects.get_repo_or_missing_resource(repo.id)
```
bmbouter: If the argument `repo` ends up as type `pulp.server.db.model.Repository` then this query can be removed since it returns that type already (I believe).

asmacdo: `repo` must be a repo transfer object here. `model.Repository.id` is an Object id, not the `repo_id`.

server/pulp/plugins/file/distributor.py
mhrivnak: We decided in-person to not do this.

bmbouter:
@mhrivnak The type changes and `repo_obj` look good. LGTM

PR #2042, opened 2015-09-15
Introduce unsafe_autoretry setting

asmacdo:
re #1 

https://pulp.plan.io/issues/1

PR #2043, opened 2015-09-15
Introduce unsafe_autoretry setting

asmacdo:
re #1

https://pulp.plan.io/issues/1

Reopened against 2.7-testing from https://github.com/pulp/pulp/pull/2042

### DIFF
--- a/server/pulp/server/db/connection.py
+++ b/server/pulp/server/db/connection.py
@@ -149,33 +149,64 @@ class PulpCollectionFailure(PulpException):
     """
 
 
-def retry_decorator(full_name=None):
+class UnsafeRetry(object):
     """
-    Collection instance method decorator providing retry support for pymongo
-    AutoReconnect exceptions
-
-    :param full_name: the full name of the database collection
-    :type  full_name: str
+    Class that decorates PyMongo to retry in the event of AutoReconnect exceptions.
     """
 
-    def _decorator(method):
+    _decorated_methods = ('get_lasterror_options', 'set_lasterror_options',
+                          'unset_lasterror_options', 'insert', 'save', 'update', 'remove', 'drop',
+                          'find', 'find_one', 'count', 'create_index', 'ensure_index',
+                          'drop_index', 'drop_indexes', 'reindex', 'index_information', 'options',
+                          'group', 'rename', 'distinct', 'map_reduce', 'inline_map_reduce',
+                          'find_and_modify')
 
-        @wraps(method)
-        def retry(*args, **kwargs):
-            while True:
+    @classmethod
+    def decorate_instance(cls, instance, full_name):
+        """
+        Decorate the PyMongo methods on instance.
+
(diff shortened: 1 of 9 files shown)


server/pulp/server/db/connection.py
```diff
@@ -149,33 +149,67 @@ class PulpCollectionFailure(PulpException):
...
+        if unsafe_autoretry:
+            for m in cls._decorated_methods:
                 try:
-                    return method(*args, **kwargs)
+                    setattr(instance, m, cls.retry_decorator(full_name)(getattr(instance, m)))
+                except AttributeError:
+                    pass
+                except TypeError, e:
+                    pass
+                    # import ipdb; ipdb.set_trace()
```
bmbouter: This should be removed.

server/pulp/server/db/connection.py
```diff
@@ -186,18 +220,10 @@ class PulpCollection(Collection):
...
-                          'group', 'rename', 'distinct', 'map_reduce', 'inline_map_reduce',
-                          'find_and_modify')
-
     def __init__(self, database, name, create=False, **kwargs):
         super(PulpCollection, self).__init__(database, name, create=create, **kwargs)
-
-        for m in self._decorated_methods:
-            setattr(self, m, retry_decorator(self.full_name)(getattr(self, m)))
+        UnsafeRetry.decorate_instance(instance=self, full_name=self.full_name)
+        # UnsafeRetry.decorate_instance(instance=self, full_name="test")
```
bmbouter: This should be removed

server/pulp/server/db/model/__init__.py
```diff
@@ -20,7 +21,23 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize a document and decorate the appropriate methods with the retry_decorator.
+        """
+        super(AutoRetryDocument, self).__init__(*args, **kwargs)
+        UnsafeRetry.decorate_instance(instance=self, full_name=type(self))
+
+    meta = {'abstract': True}
```
bmbouter: Do we need `allow_inheritance: True` ? I'm not really sure

bmbouter:
@asmacdo I put one question about `allow_inheritance` and some commented out code lines to be removed. Handle those and its LGTM. Also pulpbot needs to be happy. This code looks good. :checkered_flag:

server/pulp/server/db/model/__init__.py
asmacdo: @bmbouter, I tried both ways, allowing inheritance or to declare an abstract document. They each work equally well, but I read up a bit, and `allow_inheritance` seems to be designed for Documents that can be instantiated but may also allow others to be subclasses of them (in the same collection). Abstract documents are simple base documents that cannot be instantiated on their own but can still be the parent class of other documents-- this seemed to be more appropriate.

PR #2044, opened 2015-09-16
rm safe=true

asmacdo:
Remove the deprecated safe=True from all calls to PyMongo.

closes #1065


bmbouter:
@asmacdo This also needs to be done throughout the plugins before it's merged since it contains a `closes #1065` statement and not a `re #1065` in the commit message.

PR #2047, opened 2015-09-16
convert consumer register to bindings

asmacdo:
The recommit of double-shame :)

Originally https://github.com/pulp/pulp/pull/2031, but this was negative commited because it was branched from 2.6-dev rather than the merge-base of 2.7-testing and 2.6-dev.


asmacdo:
And to show that the merge-base is now correct:
https://github.com/pulp/pulp/compare/pulp:2.7-testing...asmacdo:base-consumer-register-binding?expand=1

rbarlow:
LGTM.

PR #2050, opened 2015-09-17
Users should download our repo files via TLS.

rbarlow:


ipanova:
Also don't forget to make the change here https://github.com/pulp/pulp/blob/2.6-release/docs/dev-guide/contributing/dev_setup.rst#dependencies 1st step

rbarlow:
@ipanova, I added the change as you suggested. Thanks for the review!

PR #2053, opened 2015-09-18
Tell Vagrant devs to use dnf instead of yum.

rbarlow:
The Vagrant environment requires Fedora, so we should instruct Vagrant
developers to use dnf instead of yum.


asmacdo:
ok test

PR #2055, opened 2015-09-18
removing trivial code that we don't need

mhrivnak:


server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -122,7 +123,7 @@ def get_units(self, repo_id, criteria=None, as_generator=False):
             units_cursors = self._associated_units_cursors_with_skip(units_cursors, criteria.skip)
             units_cursors = self._associated_units_cursors_with_limit(units_cursors, criteria.limit)
 
-        units_generator = self._units_from_chained_cursors(units_cursors)
+        units_generator = itertools.chain(*units_cursors)
```
asmacdo: :clap:

PR #2058, opened 2015-09-21
Revert change from 351fcc56c146a261ea4d8716f5726436bb97ec60

dkliban:
The fix for https://pulp.plan.io/issues/1097 introduced a new issue.
This patch reverses this fix so as to not introduce a regression in
existing behavior.


dkliban:
we decided not to do this

PR #2059, opened 2015-09-21
Update Vagrant provisioning for Crane with Docker v2.

rbarlow:


playpen/vagrant-setup.sh
```diff
@@ -47,8 +47,10 @@ debug: true
 endpoint: pulp-devel:5001
 EOF
 
-    mkdir -p metadata
-    sudo ln -s $HOME/devel/crane/metadata/ /var/lib/pulp/published/docker/app
+    mkdir -p metadata/v1 metadata/v2
+    sudo mkdir -p /var/lib/pulp/published/docker/v1 /var/lib/pulp/published/docker/v2
```
ipanova: is not it supposed to be '/var/lib/pulp/published/docker/v1/app'? same for v2 version.

rbarlow: I don't believe so, because the two lines below this create "app" as a symlink.

rbarlow:
@ipanova Can I get an LGTM?

ipanova:
ok test

PR #2060, opened 2015-09-22
ref #1180 - add redirect view and url signing.

jortel:
https://pulp.plan.io/issues/1180

Create the RedirectView which is used to redirect content GET requests that 404 to the streamer using a signed URL.

After researching signed URLs,  I have concluded that they are a concept for which there is no standard.  Without a standard, it appears that providers such as Amazon have invented there own.  As a result, there are no standard python libs for dealing with signed URLs.  I have included a collection of classes that provide URL signing and validation based on Amazon signed URLs for EC2.  I have included functional tests but have not yet finished the unit tests .. which will follow.  For now, I just wanted to get this work onto the feature branch.

The auth handler looks like this:
```
from pulp.server.config import config
from pulp.server.lazy.url import SignedURL, NotValid, Key

def allow_access(environ, host):
    key_path = config.get('authentication', 'rsa_pub')
    key = Key.load(key_path)
    url = SignedURL(environ['REQUEST_URI'])
    try:
        url.validate(key)
        return True
    except NotValid, le:
        log.info(str(le))
        return False
```

### DIFF
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -337,3 +337,26 @@
 # port: 25
 # from: no-reply@your.domain
 # enabled: false
+
+# = Lazy =
+#
+# Settings for lazy content loading.
+#
+# enabled:
+#   This controls whether redirect is enabled
+#
+# redirect_host:
+#   The host FQDN or IP to which requests are redirected.
+#
+# redirect_port:
+#   The TCP port to which requests are redirected
+#
+# redirect_path:
+#   The base path to which requests are redirected
+#
+
+[lazy]
+# enabled: false
+# redirect_host:
+# redirect_port:
+# redirect_path:
--- a/server/pulp/server/lazy/alias.py
+++ b/server/pulp/server/lazy/alias.py
@@ -1,0 +1,59 @@
+import os
+
+
+class AliasTable(object):
+    """
+    Represents a collection of Apache *Alias* directives.
+
+    :ivar table: A table of alias => path mappings.
+    :type table: dict
+    """
+
+    def __init__(self):
+        self.table = {}
+
+    def load(self):
+        """
+        Load the alias table.
+        Read .conf files in /etc/httpd/conf.d/ and build the table
+        using *Alias* directives.
+        """
+        root = '/etc/httpd/conf.d'
+        for path in [os.path.join(root, f) for f in os.listdir(root)]:
+            if not path.endswith('.conf'):
+                continue
+            with open(path) as fp:
+                while True:
+                    line = fp.readline()
+                    if not line:
+                        break
+                    if not line.startswith('Alias'):
+                        # Not an Alias
+                        continue
+                    parts = line.split(' ')
+                    if len(parts) != 3:
+                        # Malformed
+                        continue
+                    self.table[parts[1].strip()] = parts[2].strip()
+
+    def translate(self, path):
+        """
+        Translate the specified *path* component of a URL by
+        replacing the matched alias with the path mapped to the alias.
+        The original *path* is returned when it cannot be translated.
+
+        :param path: The *path* component of a URL.
+        :type path: str
+        :return: A translated path.
+        :rtype: str
+        """
+        translated = None
+        for alias, real in sorted(self.table.items()):
+            if path.startswith(alias):
+                real = os.path.realpath(real)
+                translated = path.replace(alias, real)
+                break
+        if translated:
+            return os.path.normpath(translated)
+        else:
+            return os.path.realpath(path)
--- a/server/pulp/server/lazy/url.py
+++ b/server/pulp/server/lazy/url.py
@@ -1,0 +1,563 @@
+"""
+Classes that provide URL signing and validation.
+The approach is modeled after Amazon EC2 signing.
+
+Example:
+
+unsigned:
+
+http://redhat.com/content/test.rpm
+
+signed:
+
+http://redhat.com/content/test.rpm?policy=eyJyZXNvdXJjZSI6ICIvY29udGVudC90ZXN0LnJwbSIsICJleHBpcmF0a
+W9uIjogMTQ0MjkzMTc4OH0%3D;signature=PTgdjh0uq1lXDfpvAbYXxF_TO8BKHT4Eo6OYo6M45UL0jIokZ2P5s-tAclzAx-E
+Lv3x7kfjAADA8Cpsk9iHBeeq3pKTX6Ogv_R-BkORxJAAzeM7et9M1HkyGKwBl8BTYAA3POQT0Ips4Zb0LN5uDHFPwclS-W1jvKD
+095g__lhvLmiiVLw5biVM3WFvlUl0-hd4ljzbvB4qrXRGbNtw1rnbCIowM0h2_SsoS_WzetJWsiZNdMVySfUXKJ4m3-du89wWB7
+zDFYRlosVNTom5YiKprVkJEAo3cbTOMoZss_NS-vpzauuk-qXwqj05gI7fJVsVD915gGcY0xPUbpCbVTg%3D%3D
+"""
+
+from base64 import urlsafe_b64encode, urlsafe_b64decode
+from gettext import gettext as _
+from hashlib import sha256
+from time import time
+from urllib import quote, unquote
+from urlparse import ParseResult, urlparse, urlunparse
+
+from M2Crypto import RSA, BIO
+
+from pulp.server.compat import json
+
+
+class NotValid(Exception):
+    """
+    URL or policy is not valid.
+    """
+    pass
+
+
+class DecodingError(NotValid):
+    """
+    The json or base64 decoding is invalid.
+    """
+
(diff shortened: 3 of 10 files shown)


jortel:
The auth handler looks like this:
```
from pulp.server.config import config
from pulp.server.lazy.url import SignedURL, NotValid, Key

def allow_access(environ, host):
    key_path = config.get('authentication', 'rsa_pub')
    key = Key.laod(key_path)
    url = SignedURL(environ['REQUEST_URI'])
    try:
        url.validate(key)
        return True
    except NotValid, le:
        log.info(str(le))
        return False
```

server/etc/pulp/server.conf
```diff
@@ -337,3 +337,26 @@
...
+# Settings for lazy content loading.
+#
+# enabled:
+#   This controls whether redirect is enabled
+#
+# redirect_host:
+#   The host FQDN or IP to which requests are redirected.
+#
+# redirect_port:
+#   The TCP port to which requests are redirected
```
jeremycline: I'm not sure if here is the best place, but we'll definitely want to make it clear somewhere that the hostname and port need to match the ones configured in the Apache reverse proxy. I can definitely see people messing this up, especially when it comes to TLS where the hostname needs to match whatever common name they have in their reverse proxy certificate.

server/pulp/server/lazy/alias.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self):
+        self.table = {}
+
+    def load(self):
+        """
+        Load the alias table.
+        Read .conf files in /etc/httpd/conf.d/ and build the table
+        using *Alias* directives.
+        """
+        root = '/etc/httpd/conf.d'
```
jeremycline: It'd be nice if this was configurable in some way. I know we've got a lot standing in the way of Pulp running on non-RHEL systems, but I'd rather not add to the pile.

server/pulp/server/lazy/alias.py
```diff
@@ -0,0 +1,59 @@
...
+                    if len(parts) != 3:
+                        # Malformed
+                        continue
+                    self.table[parts[1].strip()] = parts[2].strip()
+
+    def translate(self, path):
+        """
+        Translate the specified *path* component of a URL by
+        replacing the matched alias with the path mapped to the alias.
+        The original *path* is returned when it cannot be translated.
```
jeremycline: Is there a particular reason the original path is returned instead of `None`? 

One thing that confuses me here is that it says the original path is returned, but it's `realpath(path)`. If `path` is a symbolic link that doesn't seem true to me.

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+        return policy
+
+    @staticmethod
+    def validate(key, encoded, signature):
+        """
+        Validate and decode a policy.
+
+        :param key: A public RSA key.
+        :type key: RSA.RSA
+        :param encoded: A base42 encoded json policy.
```
jeremycline: s/base42/base64

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+    @property
+    def params(self):
+        """
+        :return: The *params* component of the URL.
+        :rtype: str
+        """
+        return self.content.params
+
+    @property
+    def query(self):
```
jeremycline: Is there a reason to use `@property` instead of just unpacking the tuple returned by urlparse in the `__init__` method? To me this seems more verbose and I can't figure out why it would be better.

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+        :type key: RSA.RSA
+        :param expiration: The signature expiration in seconds.
+        :type expiration: int
+        :return: The signed URL.
+        :rtype: SignedURL
+        """
+        expiration = int(time() + expiration)
+        policy = Policy(self.uri, expiration)
+        policy, signature = policy.sign(key)
+        query = Query.decode(self.content.query)
```
jeremycline: Any particular reason to get the query from inside `self.content` again rather than `self.query`?

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+    @property
+    def query(self):
+        """
+        :return: The *query* component of the URL.
+        :rtype: str
+        """
+        return self.content.query
+
+    @property
+    def uri(self):
```
jeremycline: Nitpicky, but I got to wondering what the difference is, exactly, between a URL and a URI. Going off of http://www.ietf.org/rfc/rfc3986.txt section 1.1.3, all URLs are URIs. Looking at the code you're using it to mean the path and any parameters and query which doesn't seem to match up with my understanding of a URI. That being said, I'm not sure what the best thing to call this is.

server/pulp/server/lazy/alias.py
jortel: Yes.  because the path may not be aliased but could be a symlink.  realpath resolves symlinks.  Also, we always want to redirect and let 404 happen naturally.

server/pulp/server/lazy/url.py
jortel: Making these properties instead of unpacking supports manipulation by subclasses.

server/pulp/server/lazy/url.py
jortel: I originally used *resource* instead of *uri*.  It's my understanding that a URL is a URI + a network location so assumed that meant the: *part* component + params + query.  Would *resource* be better?

server/pulp/server/lazy/alias.py
jeremycline: Ah, makes sense now.

server/pulp/server/lazy/url.py
jeremycline: I like resource.

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+        """
+        Sign the URL.
+        Has the format of: <url>?policy=<policy>;signature=<signature>.
+        The policy is: {uri: <URI>, expiration: <expiration>}
+        The uri is the path?query in the original URL.
+        The expiration is: seconds since epoch.
+        The signature is RSA signature of the SHA256 digest of the
+        json/base64 encoded policy.
+
+        :param key: A public RSA key.
```
jeremycline: This should be the private key, right?

server/pulp/server/lazy/url.py
```diff
@@ -0,0 +1,564 @@
...
+            raise NotSigned()
+
+    def validate(self, key):
+        """
+        Validate the URL *content* using the RSA signature and the
+        public key specified by *key*.  The policy is validated.
+        Then, the URI in the policy is matched against the URI
+        specified in the URL.
+
+        :param key: A private RSA key.
```
jeremycline: This should be the public key, I believe.

server/pulp/server/webservices/views/lazy.py
```diff
@@ -0,0 +1,89 @@
...
+    requests to the lazy streamer.
+
+    :ivar alias: An apache alias table.
+    :type alias: AliasTable
+    :ivar key: The public RSA key used for URL signing.
+    :type key: M2Crypto.RSA.RSA
+    """
+
+    @staticmethod
+    def urljoin(scheme, host, port, base, path, query):
```
jeremycline: Would it be better to use `urlparse.urlunparse` here? It does expect `host:port` as a single argument, but everything else seems to be the same.

server/pulp/server/webservices/views/lazy.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        Join a base URL and path component.
+
+        :param base: A base URL.
```
jeremycline: If you do decide to keep this there are a few parameter docs missing here

server/pulp/server/webservices/views/lazy.py
jortel: You'd think so (and I intended to) but `urlparse.urljoin()` makes really stupid assumptions about the parts that it is joining with respect to leading and trailing `/`.  Much like `os.path.join`.  So, by the time you *prepare* the url parts to be joined correctly by `urljoin()` you don't gain anything my using it.  We can talk about this unintuitive behavior sometime.

server/pulp/server/lazy/url.py
jortel: Yep.  just documented incorrectly.

server/pulp/server/lazy/url.py
jortel: Yep.  just documented incorrectly.

server/pulp/server/webservices/views/lazy.py
jeremycline: Okay, cool.

server/pulp/server/lazy/url.py
rbarlow: Jeremy Cline wrote:
>> +        :param encoded: A base42 encoded json policy.
> 
> s/base42/base64

But base42 is the most meaningful base!

PR #2063, opened 2015-09-23
Do not delete worker if unresponsive.

asmacdo:
closes #1222
The worker may not be on the same machine as the scheduler, and this
cleanup will occur on worker startup anyway.

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -182,9 +182,6 @@ def _delete_worker(name, normal_shutdown=False):
                                           state__in=constants.CALL_INCOMPLETE_STATES):
         cancel(task_status['task_id'])
 
-    # Delete working directory
-    common_utils.delete_worker_working_directory(name)
-
 
 @task
 def _release_resource(task_id):
@@ -593,4 +590,5 @@ def cleanup_old_worker(*args, **kwargs):
     name = kwargs['sender'].hostname
     _delete_worker(name, normal_shutdown=True)
     # Recreate a new working directory for worker that is starting now
+    common_utils.delete_worker_working_directory(name)
     common_utils.create_worker_working_directory(name)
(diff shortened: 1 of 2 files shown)


dkliban:
ok teast

dkliban:
ok test

mhrivnak:
There are two behaviors in question.

Bug 1222 refers to the deletion of a worker's working directory, which may only be present on one specific machine. The scheduler cannot reliably delete that for the reason you cited.

A different pulp action is deletion of the worker and its reservations from pulp's database, which the scheduler can do, and has been doing successfully for quite some time. This deletion is important, because it prevents new work from being assigned to a worker that we do not reasonably expect will become available. Consider a worker that stays down for hours, days, or forever.

Given this proposed change, if a worker goes offline and stays offline for an extended period, pulp may continue to assign work to it, especially for any resources it already had assigned. This would be a very poor user experience. Let me know if you think my summary is incorrect.

I suggest that the ``_delete_worker`` task should have its last line removed, which makes the call to delete the working directory. That call may need to be moved into the worker startup workflow. Does that make sense? It is a much smaller behavioral change that addresses only the problematic behavior cited in the bug report, but no the useful part of the ``_delete_worker`` workflow.

dkliban:
@mhrivnak You make a good point. We should definitely make the change you are suggesting. 

I would like to also point out that the reason this bug was filed was because Partha and I both kept seeing in our logs that a worker is offline. Then it would reappear very quickly. After it would reappear, it's working directory would be wiped out and any sync operation would fail on it because it didn't have a working directory anymore. 

I would like to learn why this is happening ... it might be a Qpid problem or something else. I really have no idea.

asmacdo:
@dkliban, @mhrivnak ready for re-review

asmacdo:
ok test

PR #2067, opened 2015-09-24
ensure datetimes on distributors

asmacdo:
closes https://pulp.plan.io/issues/1254

When saving datetimes to Mongo, tz information is stripped. We have to reconstruct the tzinfo (assuming UTC) so it can be compared with other datetimes.


asmacdo:
ok test

dkliban:
I synced and published repo zoo. 
I removed camel-0.1-1-noarch rpm.
I published ... got a failure.

I applied the patch and was able to successfully publish. 


LGTM

PR #2071, opened 2015-09-24
Lazy content catalog model object.

jortel:
https://pulp.plan.io/issues/1231

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -15,7 +15,7 @@ from pulp.server.content.storage import FileStorage, SharedStorage
 from pulp.plugins.model import Repository as plugin_repo
 from pulp.server.async.emit import send as send_taskstatus_message
 from pulp.server.db.connection import UnsafeRetry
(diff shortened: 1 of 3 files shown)


jeremycline:
I am by no means an expert (or amateur) at mongoengine so you might want a second opinion, but going by the mongoengine documentation and my prior experience with ORMs this looks good to me.

server/pulp/server/db/model/__init__.py
```diff
@@ -645,3 +645,60 @@ class CeleryBeatLock(Document):
 
     # For backward compatibility
     _ns = StringField(default='celery_beat_lock')
+
+
+class ContentCatalogEntry(Document):
+    """
+    A catalog of content that can be downloaded by the specified plugin.  Entries
+    are grouped by revision.  Each revision represents a platform event that
```
jeremycline: Where is this revision stored?

jortel: The revision concept was replace with the concept that importers would add/update the catalog entries instead.  I just need to purge it from the docstrings.

PR #1698, opened 2015-03-05
added -v and -vv options to admin and consumer clients to get info and d...

skarmark:
Debug level output which was going to the logs before

https://pulp.plan.io/issues/145

The behavior now can be summarized as -
------------------------Normal behavior ---------------------------------------
$ pulp-admin rpm repo create --repo-id test
A resource with the ID "test" already exists.

----------------------- with -v flag ---------------------------------------------
$ pulp-admin -v rpm repo create --repo-id test
2015-03-05 01:34:37,817 - ERROR - Exception occurred:
        href:      /pulp/api/v2/repositories/
        method:    POST
        status:    409
        error:     Duplicate resource: test
        traceback: None
        data:      {u'resource_id': u'test', u'error': {u'code': u'PLP0018', u'data': {u'resource_id': u'test'}, u'description': u'Duplicate resource: test', u'sub_errors': []}}
     
A resource with the ID "test" already exists.

-------------------------- with multiple -v flags ----------------------------------

$ pulp-admin -vv rpm repo create --repo-id test
2015-03-05 01:34:42,109 - DEBUG - sending POST request to /pulp/api/v2/repositories/
2015-03-05 01:34:42,325 - INFO - POST request to /pulp/api/v2/repositories/ with parameters {"display_name": null, "description": null, "distributors": [{"distributor_id": "yum_distributor", "auto_publish": true, "distributor_config": {"http": false, "relative_url": "test", "https": true}, "distributor_type_id": "yum_distributor"}, {"distributor_id": "export_distributor", "auto_publish": false, "distributor_config": {"http": false, "https": true}, "distributor_type_id": "export_distributor"}], "notes": {"_repo-type": "rpm-repo"}, "importer_type_id": "yum_importer", "importer_config": {}, "id": "test"}
2015-03-05 01:34:42,325 - INFO - Response status : 409 

2015-03-05 01:34:42,325 - INFO - Response body :
 {
  "exception": null, 
  "traceback": null, 
  "_href": "/pulp/api/v2/repositories/", 
  "resource_id": "test", 
  "error_message": "Duplicate resource: test", 
  "http_request_method": "POST", 
  "http_status": 409, 
  "error": {
    "code": "PLP0018", 
    "data": {
      "resource_id": "test"
    }, 
    "description": "Duplicate resource: test", 
    "sub_errors": []
  }
}

2015-03-05 01:34:42,325 - ERROR - Exception occurred:
        href:      /pulp/api/v2/repositories/
        method:    POST
        status:    409
[...]

PR #2074, opened 2015-09-25
verbose should register a api_log_handler

asmacdo:
https://pulp.plan.io/issues/145

### DIFF
--- a/client_lib/pulp/client/launcher.py
+++ b/client_lib/pulp/client/launcher.py
@@ -192,7 +192,7 @@ def _create_bindings(config, cli_logger, username, password, verbose=None):
         cert_filename = None
 
     api_logger = None
-    if verbose and verbose > 1:
+    if verbose and verbose >= 1:
         api_log_handler = logging.StreamHandler(sys.stderr)
         api_log_handler.setFormatter(logging.Formatter('%(asctime)s - %(levelname)s - %(message)s'))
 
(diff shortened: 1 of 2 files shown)


dkliban:
ok test

client_lib/pulp/client/launcher.py
```diff
@@ -192,7 +192,7 @@ def _create_bindings(config, cli_logger, username, password, verbose=None):
         cert_filename = None
 
     api_logger = None
-    if verbose and verbose > 1:
+    if verbose and verbose >= 1:
```
rbarlow: It was my understanding that we only wanted to log API output with -vv and not with just -v.

rbarlow:
I don't think we want to log the API calls with -v, and the Redmine issue doesn't seem to indicate this either. What non-API info() calls are there during a sync?

ipanova:
@rbarlow when i've been reviewing the PR that introduced this functionality a had same question
https://github.com/pulp/pulp/pull/1698#issuecomment-77535323
@asmacdo maybe this will help to clear the situation.

asmacdo:
This [failed QA](https://pulp.plan.io/issues/145#note-27), but based on your comments, this looks like the correct behavior. Do you agree then that the issue should go back to QA and that this PR should be closed? @ipanova @rbarlow

ipanova:
@asmacdo i am ok to close this pr, if we add to bug description the info mentioned above.
let's see what @rbarlow  will say.

rbarlow:
@asmacdo @ipanova I think we should put it back to ON_QA and close this.

PR #2078, opened 2015-09-25
Updated gofer to 2.6.6

dkliban:
This PR has a conflict because the version of gofer changed between the common ancestor of 2.6-dev and 2.7-testing and now.


beav:
LGTM, the conflict makes sense IMO.

PR #2081, opened 2015-09-28
Adds support for write concern

dkliban:
Pulp now accepts a write concern for the MongoDB connection. All writes
to the database are now acknowledged. For replica sets, 'majority' or 'all'
members can be asked to acknowledge the write.

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -100,6 +100,8 @@ PLP0040 = Error("PLP0040", _("Database 'seeds' config must include at least one 
 PLP0041 = Error("PLP0041", _("Database 'replica_set' config must be specified when more than one "
                              "seed is provided. Refer to /etc/pulp/server.conf for proper use."),
                 [])
+PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
+                             "Refer to /etc/pulp/server.conf for proper use."), [])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
--- a/docs/user-guide/release-notes/2.7.x.rst
+++ b/docs/user-guide/release-notes/2.7.x.rst
@@ -52,6 +52,10 @@ New Features
 * Pulp properly connects to MongoDB replica sets. The `replica_set` setting in `database` section
   of `/etc/pulp/server.conf` must be provided for Pulp to connect to a replica set.
 
+* Pulp allows users to specify the write concern for MongoDB connection. The two options are
+  'majority' and 'all'. When 'all' is specified, the number of seeds listed will be used as the value
+  for 'w'. For installations without replica sets, all writes to the database will be acknowledged.
+
 
 Deprecation
 -----------
--- a/server/pulp/server/db/connection.py
+++ b/server/pulp/server/db/connection.py
@@ -18,14 +18,16 @@ from pulp.server import config
 from pulp.server.compat import wraps
 from pulp.server.exceptions import PulpCodedException, PulpException
 
+import semantic_version
+
 
 _CONNECTION = None
 _DATABASE = None
 _DEFAULT_MAX_POOL_SIZE = 10
 # please keep this in X.Y.Z format, with only integers.
 # see version.cpp in mongo source code for version format info.
-MONGO_MINIMUM_VERSION = "2.4.0"
-
+MONGO_MINIMUM_VERSION = semantic_version.Version("2.4.0")
+MONGO_WRITE_CONCERN_VERSION = semantic_version.Version("2.6.0")
 
 _logger = logging.getLogger(__name__)
 
@@ -48,6 +50,7 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
 
         if seeds is None:
             seeds = config.config.get('database', 'seeds')
+        seeds_list = seeds.split(',')
 
         if max_pool_size is None:
             # we may want to make this configurable, but then again, we may not
@@ -60,6 +63,12 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
 
         if replica_set is not None:
             connection_kwargs['replicaSet'] = replica_set
+
+        write_concern = config.config.get('database', 'write_concern')
+        if write_concern not in ['majority', 'all']:
+            raise PulpCodedException(error_code=error_codes.PLP0043)
+        elif write_concern == 'all':
+            write_concern = len(seeds_list)
 
         # Process SSL settings
         if config.config.getboolean('database', 'ssl'):
@@ -91,12 +100,24 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
                                                                 itertools.repeat(32))
 
         if seeds != '':
-            seeds_list = seeds.split(',')
             if len(seeds_list) > 1 and not replica_set:
                 raise PulpCodedException(error_code=error_codes.PLP0041)
             while True:
                 _CONNECTION = _connect_to_one_of_seeds(connection_kwargs, seeds_list, name)
                 if _CONNECTION:
+                    db_version = semantic_version.Version(_CONNECTION.server_info()['version'])
+                    if db_version < MONGO_MINIMUM_VERSION:
+                        raise RuntimeError(_("Pulp requires Mongo version %s, but DB is reporting"
+                                             "version %s") % (MONGO_MINIMUM_VERSION,
+                                                              db_version))
+                    elif db_version >= MONGO_WRITE_CONCERN_VERSION or replica_set:
+                        # Write concern of 'majority' only works with a replica set or when using
+                        # MongoDB >= 2.6.0
+                        _CONNECTION.write_concern['w'] = write_concern
+                    else:
+                        _CONNECTION.write_concern['w'] = 1
+                    _logger.info(_("Write concern for Mongo connection: %s") %
+                                 _CONNECTION.write_concern)
                     break
                 else:
                     next_delay = min(mongo_retry_timeout_seconds_generator.next(), max_timeout)
@@ -120,16 +141,6 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
         # Query the collection names to ensure that we are authenticated properly
         _logger.debug(_('Querying the database to validate the connection.'))
         _DATABASE.collection_names()
-
-        db_version = _CONNECTION.server_info()['version']
-        _logger.info(_("Mongo database for connection is version %s") % db_version)
-
-        db_version_tuple = tuple(db_version.split("."))
-        db_min_version_tuple = tuple(MONGO_MINIMUM_VERSION.split("."))
-        if db_version_tuple < db_min_version_tuple:
-            raise RuntimeError(_("Pulp requires Mongo version %s, but DB is reporting version %s") %
-                               (MONGO_MINIMUM_VERSION, db_version))
-
     except Exception, e:
         _logger.critical(_('Database initialization failed: %s') % str(e))
         _CONNECTION = None
--- a/server/test/unit/server/db/test_connection.py
+++ b/server/test/unit/server/db/test_connection.py
@@ -2,17 +2,23 @@ import unittest
 
 from mock import call, patch, MagicMock, Mock
 from pymongo.errors import AutoReconnect
+import unittest2
+
+from pulp.common import error_codes
 
 from pulp.server import config
 from pulp.server.db import connection
 from pulp.server.exceptions import PulpCodedException
 
 
+MONGO_MIN_TEST_VERSION = "2.4.0"
+
+
 class MongoEngineConnectionError(Exception):
     pass
 
 
-class TestDatabaseSeeds(unittest.TestCase):
+class TestDatabaseSeeds(unittest2.TestCase):
 
     def tearDown(self):
         # Reload the configuration so that things are cleaned up properly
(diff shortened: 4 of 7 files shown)


docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -52,6 +52,10 @@ New Features
 * Pulp properly connects to MongoDB replica sets. The `replica_set` setting in `database` section
   of `/etc/pulp/server.conf` must be provided for Pulp to connect to a replica set.
 
+* Pulp allows user to specify the write concern for MongoDB connection. The two options are
```
rbarlow: s/user/users/

rbarlow: s/for/for the/

docs/user-guide/release-notes/2.7.x.rst
```diff
@@ -52,6 +52,10 @@ New Features
...
+  'majority' and 'all'. When 'all' is specified, number of seeds listed will be used as the value
```
rbarlow: s/number/the number/

server/pulp/server/db/connection.py
```diff
@@ -91,12 +98,26 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
 
         if seeds != '':
-            seeds_list = seeds.split(',')
             if len(seeds_list) > 1 and not replica_set:
                 raise PulpCodedException(error_code=error_codes.PLP0041)
             while True:
                 _CONNECTION = _connect_to_one_of_seeds(connection_kwargs, seeds_list, name)
                 if _CONNECTION:
+                    db_min_version_tuple = tuple(MONGO_MINIMUM_VERSION.split("."))
+                    wc_min_version_tuple = tuple(MONGO_WRITE_CONCERN_VERSION.split("."))
```
rbarlow: Why not just create these constants as tuples above so you can drop these two lines?

server/pulp/server/db/connection.py
```diff
@@ -91,12 +98,26 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
+                    db_version_tuple = _get_mongo_version()
+                    if db_version_tuple < db_min_version_tuple:
+                        raise RuntimeError(_("Pulp requires Mongo version %s, but DB is reporting"
+                                             "version %s") % (MONGO_MINIMUM_VERSION,
+                                                              '.'.join(db_version_tuple)))
+                    elif db_version_tuple >= wc_min_version_tuple or replica_set:
+                        # Write concern of 'majority' only works with a replica set or when using
+                        # MongoDB >= 2.6.0
+                        _CONNECTION.write_concern['w'] = write_concern
```
rbarlow: It seems that if a user sets write concern while using mongo 2.4, we should raise an error. This code will set the write concern to 1 instead.

server/test/unit/server/db/test_connection.py
```diff
@@ -72,6 +72,28 @@ def test_seeds_from_config(self, mock_connect_seeds, mock_mongoengine):
...
+    @patch('pulp.server.db.connection._connect_to_one_of_seeds')
+    def test_invalid_write_concern(self, mock_connect_seeds, mock_mongoengine):
+        mock_mongoengine.connect.return_value.server_info.return_value = {'version': '2.6.0'}
+        mock_connect_seeds.return_value.server_info.return_value = {'version': '2.6.0'}
+        seeds = "firsthost:1234,secondhost:5678"
+        replica_set = 'fakeReplica'
+        config.config.set('database', 'write_concern', 'blah')
+        config.config.set('database', 'seeds', seeds)
+        config.config.set('database', 'replica_set', replica_set)
+        self.assertRaises(PulpCodedException, connection.initialize)
```
rbarlow: We should assert that the correct PulpCodedException is raised, not just that any of them gets raised.

server/test/unit/server/db/test_connection.py
```diff
@@ -72,6 +72,28 @@ def test_seeds_from_config(self, mock_connect_seeds, mock_mongoengine):
...
 
+    @patch('pulp.server.db.connection.mongoengine')
+    @patch('pulp.server.db.connection._connect_to_one_of_seeds')
+    def test_multiple_seeds_no_replica_set(self, mock_connect_seeds, mock_mongoengine):
+        mock_mongoengine.connect.return_value.server_info.return_value = {'version': '2.6.0'}
+        mock_connect_seeds.return_value.server_info.return_value = {'version': '2.6.0'}
+        seeds = "firsthost:1234,secondhost:5678"
+        config.config.set('database', 'write_concern', 'majority')
+        config.config.set('database', 'seeds', seeds)
+        self.assertRaises(PulpCodedException, connection.initialize)
```
rbarlow: Why is this raising an Exception? It seems that 2.6.0 with majority and a list of seeds should be allowed.

rbarlow:
A few things to fix, and I'll re-review:

0) We need to raise an Exception if the user is using mongo 2.4 and sets write concern without a replica set.

1) We need unit tests that test all the cases, success and error. These tests didn't assert normal cases (mongo 2.4 with replica set, mongo 2.6 with replica set, mongo 2.6 without replica set, for example.)

2) The test assertions need to be stronger. These tests only assert that an error was raised, but not that it was the correct error.

dkliban:
@rbarlow If a user is using Mongo 2.4, he does not have to use a replica set. Our default of 'majority' does not work without using a replica set. However, when MongoDB is running in single node mode, a write concern of 1 behaves the same way in MongoDB 2.4 as 'majority' in MongoDB 2.6.

dkliban:
ok test

rbarlow:
ok test

dkliban:
ok test

common/pulp/common/error_codes.py
```diff
@@ -100,6 +100,9 @@
 PLP0041 = Error("PLP0041", _("Database 'replica_set' config must be specified when more than one "
                              "seed is provided. Refer to /etc/pulp/server.conf for proper use."),
                 [])
+PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
```
mhrivnak: why not 42?

server/pulp/server/db/connection.py
```diff
@@ -24,8 +24,8 @@
 _DEFAULT_MAX_POOL_SIZE = 10
 # please keep this in X.Y.Z format, with only integers.
 # see version.cpp in mongo source code for version format info.
-MONGO_MINIMUM_VERSION = "2.4.0"
-
+MONGO_MINIMUM_VERSION = ("2", "4", "0")
```
mhrivnak: these should be integers. Strings work for single characters, but...

```
>>> "1" < "2"
True
>>> "10" < "2"
True
```

server/pulp/server/db/connection.py
```diff
@@ -163,6 +174,14 @@ def _connect_to_one_of_seeds(connection_kwargs, seeds_list, db_name):
             _logger.info(msg % {'url': seed, 'e': str(e)})
 
 
+def _get_mongo_version():
+    db_version = _CONNECTION.server_info()['version']
+    _logger.info(db_version)
+    _logger.info(_("Mongo database for connection is version %s") % db_version)
+    db_version_tuple = tuple(db_version.split("."))
```
mhrivnak: FWIW, a handy way to turn these into ints is ```tuple(map(int, db_version.split('.')))```

server/pulp/server/db/connection.py
```diff
@@ -163,6 +174,14 @@ def _connect_to_one_of_seeds(connection_kwargs, seeds_list, db_name):
             _logger.info(msg % {'url': seed, 'e': str(e)})
 
 
+def _get_mongo_version():
```
mhrivnak: doc block

server/pulp/server/db/connection.py
```diff
@@ -163,6 +174,14 @@ def _connect_to_one_of_seeds(connection_kwargs, seeds_list, db_name):
...
+    db_version = _CONNECTION.server_info()['version']
+    _logger.info(db_version)
```
mhrivnak: seems redundant given the next line

server/pulp/server/db/connection.py
```diff
@@ -163,6 +174,14 @@ def _connect_to_one_of_seeds(connection_kwargs, seeds_list, db_name):
...
+    _logger.info(_("Mongo database for connection is version %s") % db_version)
```
mhrivnak: I'm being picky, but mongo only manages our database, and that database is not versioned (at least not in this way). You're logging the version of mongo itself, so I'd change the language to something like "Connected to MongoDB version..."

server/test/unit/server/db/test_connection.py
```diff
@@ -72,6 +77,34 @@ def test_seeds_from_config(self, mock_connect_seeds, mock_mongoengine):
...
+        mock_connect_seeds.return_value.server_info.return_value = {'version': '2.6.0'}
+        seeds = "firsthost:1234,secondhost:5678"
+        replica_set = 'fakeReplica'
+        config.config.set('database', 'write_concern', 'blah')
+        config.config.set('database', 'seeds', seeds)
+        config.config.set('database', 'replica_set', replica_set)
+        try:
+            connection.initialize()
+        except PulpCodedException as connection_error:
+            self.assertEqual(connection_error.error_code.message, error_codes.PLP0043.message)
```
mhrivnak: This test will pass if the exception doesn't get raised. I suggest using ``unittest2`` and ``self.assertRaises`` as a context manager. That lets you assert the exception happens, and then make assertions about the exception itself.

Same for the test above.

common/pulp/common/error_codes.py
```diff
@@ -100,6 +100,9 @@
...
+                             "Refer to /etc/pulp/server.conf for proper use."),
+                [])
```
mhrivnak: Looks like you copy-pasted the one above. This one doesn't need this final argument on a separate line.

server/pulp/server/db/connection.py
```diff
@@ -91,12 +98,26 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
+                                             "version %s") % ('.'.join(MONGO_MINIMUM_VERSION),
+                                                              '.'.join(db_version_tuple)))
+                    elif db_version_tuple >= MONGO_WRITE_CONCERN_VERSION or replica_set:
+                        # Write concern of 'majority' only works with a replica set or when using
+                        # MongoDB >= 2.6.0
+                        _CONNECTION.write_concern['w'] = write_concern
+                    else:
+                        _logger.info('VERSION IS 2.4')
+                        _CONNECTION.write_concern['w'] = 1
+                        _logger.info(_CONNECTION.write_concern)
```
mhrivnak: seems redundant given the line below

common/pulp/common/error_codes.py
dkliban: @mhrivnak Because we already have 42 on master

server/test/unit/server/db/test_connection.py
dkliban: @mhrivnak I was using self.assertRaises to check for PulpCodedException. I wasn't able to check that it was a specific one actually. So I changed the test to this format so I could be more specific. Unless you can suggest how I can use self.assertRaises with a particular PulpCodedException,  I will add a failure statement after line 104.

dkliban:
ok test

dkliban:
ok test

server/pulp/server/db/connection.py
```diff
@@ -91,12 +100,25 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
 
         if seeds != '':
-            seeds_list = seeds.split(',')
             if len(seeds_list) > 1 and not replica_set:
                 raise PulpCodedException(error_code=error_codes.PLP0041)
             while True:
                 _CONNECTION = _connect_to_one_of_seeds(connection_kwargs, seeds_list, name)
                 if _CONNECTION:
+                    db_version = semantic_version.Version(_CONNECTION.server_info()['version'])
+                    if cmp(db_version, MONGO_MINIMUM_VERSION) == -1:
```
mhrivnak: I think this (and below) would be more readable with normal comparison operators, which you can use with these objects (unless you found some edge case?). I suggest

```db_version < MONGO_MINIMUM_VERSION```

File: server/pulp/server/db/connection.py [hunk 17]
```diff
@@ -91,12 +100,25 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
...
+                        raise RuntimeError(_("Pulp requires Mongo version %s, but DB is reporting"
```
mhrivnak: I'd be clear here that we require that version *or greater*. We don't want someone to be confused that we might require one specific version.

server/test/unit/server/db/test_connection.py
```diff
@@ -72,6 +77,35 @@ def test_seeds_from_config(self, mock_connect_seeds, mock_mongoengine):
...
+        mock_mongoengine.connect.return_value.server_info.return_value = {'version': '2.6.0'}
+        mock_connect_seeds.return_value.server_info.return_value = {'version': '2.6.0'}
+        seeds = "firsthost:1234,secondhost:5678"
+        replica_set = 'fakeReplica'
+        config.config.set('database', 'write_concern', 'blah')
+        config.config.set('database', 'seeds', seeds)
+        config.config.set('database', 'replica_set', replica_set)
+        try:
+            connection.initialize()
+            self.fail("Should raise PLP0043 PulpCodedException")
```
mhrivnak: This is a reasonable approach if you want to avoid using ```unittest2``` to get ```assertRaises``` as a context manager, although the same should be applied in the test above.

dkliban:
ok test

server/test/unit/server/db/test_connection.py
```diff
@@ -1,18 +1,23 @@
-import unittest
+import unittest, unittest2
```
mhrivnak: last comment, I swear. ;) ```unittest2``` is a third-party module, so it should be in the second group of imports.

dkliban:
ok test

PR #2084, opened 2015-09-28
PEP-8 some more of Nodes.

rbarlow:


nodes/extensions/admin/pulp_node/extensions/admin/commands.py
```diff
@@ -23,9 +23,11 @@
 from pulp.client.commands.repo.cudl import ListRepositoriesCommand
 
 from pulp_node import constants
-from pulp_node.extension import missing_resources, node_activated, repository_enabled, ensure_node_section
+from pulp_node.extension import (missing_resources, node_activated, repository_enabled,
```
ipanova: seems like you forgot to remove the copyright header

rbarlow: Fixed.

PR #2087, opened 2015-09-29
Fixed documentation around nodes activation

mibanescu:


mibanescu:
Argh. I was installing nodes from some older directions, and I didn't properly install the yum/dnf group. I was missing pulp-nodes-consumer-extensions

PR #2089, opened 2015-09-30
Lazy RedirectView no longer returns HTTP 500.

jeremycline:
If the URL that mapped to the Lazy RedirectView was queried
directly rather than being the result of an internal redirect
via Apache's ErrorDocument, the client would get a 500.

A 404 felt like the best response to me, but I would welcome other suggestions.

### DIFF
--- a/server/pulp/server/webservices/views/lazy.py
+++ b/server/pulp/server/webservices/views/lazy.py
@@ -73,7 +73,12 @@ class RedirectView(View):
         :return: A redirect or not-found response.
         :rtype: django.http.HttpResponse
         """
-        path = request.environ['REDIRECT_URL']
+        try:
+            path = request.environ['REDIRECT_URL']
+        except KeyError:
+            # Not redirected by ErrorDocument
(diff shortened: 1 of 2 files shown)


server/pulp/server/webservices/views/lazy.py
```diff
@@ -73,7 +73,12 @@ def get(self, request):
         :return: A redirect or not-found response.
         :rtype: django.http.HttpResponse
         """
-        path = request.environ['REDIRECT_URL']
+        try:
+            path = request.environ['REDIRECT_URL']
+        except KeyError:
+            # Someone arrived here without the aid of an ErrorDocument
```
jortel: Comment seems wordy.  suggest:
```
# Not redirected by ErrorDocument
```

jortel:
Please add a test for this:
```
pulp.server.webservices.views.lazy                                             44      2    95%   78-80
```

jeremycline:
Oops.

PR #2090, opened 2015-09-30
Add content download task.

jortel:
https://pulp.plan.io/issues/1181

Still needs unit tests but wanted to get some pre-review.

### DIFF
--- a/server/pulp/server/controllers/content.py
+++ b/server/pulp/server/controllers/content.py
@@ -1,17 +1,31 @@
 # -*- coding: utf-8 -*-
 from gettext import gettext as _
+from logging import getLogger
+import os
 import threading
 
 import celery
+from nectar.request import DownloadRequest
+from nectar.listener import AggregatingEventListener
 
 from pulp.common import error_codes
 from pulp.common.plugins import reporting_constants
+from pulp.common.tags import resource_tag, action_tag, RESOURCE_CONTENT_UNIT_TYPE
 from pulp.plugins.conduits.mixins import (ContentSourcesConduitException, StatusMixin,
                                           PublishReportMixin)
+from pulp.plugins.loader import api as plugin_api
+from pulp.plugins.loader.exceptions import PluginNotFound
 from pulp.plugins.util.publish_step import Step
 from pulp.server.async.tasks import Task
+from pulp.server.controllers import repository
 from pulp.server.content.sources.container import ContentContainer
 from pulp.server.exceptions import PulpCodedTaskException
+from pulp.server.managers.repo._common import get_working_directory
+
+
+
+
+_logger = getLogger(__name__)
 
 
 class ContentSourcesConduit(StatusMixin, PublishReportMixin):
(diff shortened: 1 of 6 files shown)


server/pulp/server/controllers/content.py
```diff
@@ -96,3 +110,87 @@ def refresh_content_source(content_source_id=None):
...
     step = ContentSourcesRefreshStep(conduit, content_source_id=content_source_id)
     step.process_lifecycle()
+
+
+@celery.task(base=Task)
+def _download(plugin_id, unit_locator, url, options):
+    """
+    Download the specified content unit.
+
+    :param plugin_id: An importer ID.
```
jeremycline: I assume this is the importer object id, but since we also have the `importer_type_id` it would be nice if this was very explicit

server/pulp/server/controllers/content.py
```diff
@@ -96,3 +110,87 @@ def refresh_content_source(content_source_id=None):
...
+    downloader.event_listener = listener
+
+    # Download and store the unit
+    downloader.download_one(request)
+    if listener.succeeded_reports:
+        _logger.info(_('Download {url}, succeeded').format(url=url))
+        unit.set_content(destination)
+        unit.save()
+    else:
+        _logger.info(_('Download {url}, failed').format(url=url))
```
jeremycline: It might be nice to include some information about the failure from the listener report like the HTTP status code.

server/pulp/server/controllers/content.py
```diff
@@ -96,3 +110,87 @@ def refresh_content_source(content_source_id=None):
...
+
+
+def queue_download(plugin_id, unit_locator, url, options):
+    """
+    Queue task to download the specified content unit.
+
+    :param plugin_id: An importer ID.
```
jeremycline: Same thing as L120.

server/pulp/server/controllers/repository.py
```diff
@@ -271,6 +272,33 @@ def queue_delete(repo_id):
     return async_result
 
 
+def get_importer_by_id(object_id):
+    """
+    Get a plugin and call configuration using the document ID
+    of the repository-importer association document.
+
+    :param object_id: The document ID .
```
jeremycline: There seems to be a stray space after 'ID'

server/pulp/server/controllers/repository.py
```diff
@@ -271,6 +272,33 @@ def queue_delete(repo_id):
...
+    :type object_id: str
+    :return: A tuple of:
+        (pulp.plugins.importer.Importer, pulp.plugins.config.PluginCallConfiguration)
+    :rtype: tuple
+    :raise pulp.plugins.loader.exceptions.PluginNotFound: not found.
+    :raise
```
jeremycline: Incomplete doc string.

server/pulp/server/controllers/repository.py
```diff
@@ -271,6 +272,33 @@ def queue_delete(repo_id):
...
+    """
+    collection = RepoImporter.get_collection()
+    try:
+        object_id = ObjectId(object_id)
+    except InvalidId:
+        raise plugin_exceptions.PluginNotFound()
```
jeremycline: I'm on the fence about this. On the one hand it might be better to let the InvalidId exception be raised because that lets the user know they aren't providing the correct input, but on the other hand the PluginNotFound exception isn't *wrong* and will still make the user go check their input.

server/pulp/server/db/model/__init__.py
```diff
@@ -647,6 +648,36 @@ class CeleryBeatLock(Document):
...
+class Digest(EmbeddedDocument):
+    """
+    Embedded digest (checksum) document.
+
+    :ivar algorithm: The digest algorithm.
+    :type algorighm: str
+    :ivar value: The digest value.
+    :type value: str
+    """
+    ALG_REGEX = r'({a})'.format(a='|'.join(hashlib.algorithms))
```
jeremycline: Sadly, hashlib.algorithms is new in 2.7. I'm not sure if there's a good replacement that is 2.6 compatible.

jeremycline:
The only thing this needs is a replacement for the `hashlib.algorithms` and then it'll look good to me.

PR #2091, opened 2015-10-01
Removing shutil.move and copytree where /var/cache/pulp is involved.

ipanova:
closes #1269 

I did not change move--> copy when
1) move was performed from /var/lib/pulp/ to /var/lib/pulp/
2) move was performed from /var/cache/pulp to /var/cache/pulp
3) in tests where it was not testing the code itself
4) under rel-eng directory

I did not change copytree in case where only symlinks were copied and param symlinks  was set to True https://docs.python.org/2/library/shutil.html#shutil.copytree for more info read 2nd paragraph


ipanova:
https://github.com/pulp/pulp_docker/pull/97
https://github.com/pulp/pulp_rpm/pull/718

PR #2092, opened 2015-10-02
Apache 2.4 now starts without a conf file in /etc/pulp/vhost80/

jeremycline:
In Apache 2.4 when an 'Include' command fails to match any file
Apache will fail to start. This is avoided by using the
'IncludeOptional' command. In 2.2 the 'Include' command acts in a
similar fashion to the 'IncludeOptional' command in 2.4, so that
configuration remains unchanged.

Closes #293

### DIFF
--- a/server/etc/httpd/conf.d/pulp_apache_24.conf
+++ b/server/etc/httpd/conf.d/pulp_apache_24.conf
@@ -51,7 +51,7 @@ WSGIImportScript /srv/pulp/webservices.wsgi process-group=pulp application-group
 </Files>
 
 <VirtualHost *:80>
-    Include /etc/pulp/vhosts80/*.conf
+    IncludeOptional /etc/pulp/vhosts80/*.conf
 </VirtualHost>
 
 


jeremycline:
One thing to note here is that Apache will still fail to start if /etc/pulp/vhost80 doesn't exist, but going by the spec file (https://github.com/pulp/pulp/blob/master/pulp.spec#L172) that shouldn't be a problem

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

bmbouter:
@jeremycline what about apache 2.2?

https://github.com/jeremycline/pulp/blob/issue-293/server/etc/httpd/conf.d/pulp_apache_22.conf#L50

jeremycline:
@bmbouter In 2.2 the 'Include' command acts in a similar fashion to the 'IncludeOptional' command in 2.4, so that configuration remains unchanged.

bmbouter:
:ear: :+1: :footprints: :last_quarter_moon_with_face: LGTM

PR #2094, opened 2015-10-02
Updates dev-guide with dnf commands

dkliban:


docs/dev-guide/contributing/dev_setup.rst
```diff
@@ -10,7 +10,7 @@ For the Impatient
 These instructions will create a developer install of Pulp on a dedicated
 development instance.
 
-* Start a Fedora 20 x86_64 instance that will be dedicated for development with
+* Start a RHEL 7 or Fedora 21/22 x86_64 instance that will be dedicated for development with
```
bmbouter: Should this be agnostic about the Fedora version so we don't need to constantly update this. It could refer to them as a `current Fedora` instead maybe?

docs/dev-guide/contributing/dev_setup.rst
```diff
@@ -42,8 +42,8 @@ pull requests into the Pulp repositories as described in :doc:`merging`.
 Dependencies
 ^^^^^^^^^^^^
 
-The easiest way to download the other dependencies is to install Pulp through yum, which will pull in
-the latest dependencies according to the spec file.
+The easiest way to download the other dependencies is to install Pulp through yum or dnf, which
+will pull in the latest dependencies according to the spec file.
```
bmbouter: s/will pull in/pulls in/

docs/dev-guide/contributing/dev_setup.rst
```diff
@@ -53,9 +53,14 @@ the latest dependencies according to the spec file.
        $ sudo wget https://repos.fedorapeople.org/repos/pulp/pulp/fedora-pulp.repo
 
 #. Edit the repo and enable the most recent testing repository.
-#. Install the main Pulp groups to get all of the dependencies.
+
+#. When using dnf, you can install the dependencies with this command.
```
bmbouter: s/you can //

bmbouter:
Should this PR target 2.6-testing instead?
Also will it apply with just this one commit against 2.7-dev+?

bmbouter:
@dkliban Actually this needs to be published soon because both 2.6.3 and 2.6.4 released versions when run on Fedora 22 would need these dnf based docs. LGTM, handle the comments how you want.

PR #2097, opened 2015-10-04
Delete server public key as well

pcreech:
Delete imported server public key

Fixes [#1260](https://pulp.plan.io/issues/1260)

### DIFF
--- a/client_consumer/pulp/client/consumer/cli.py
+++ b/client_consumer/pulp/client/consumer/cli.py
@@ -316,8 +316,10 @@ class UnregisterCommand(PulpCliCommand):
         id_cert_dir = self.context.config['filesystem']['id_cert_dir']
         id_cert_name = self.context.config['filesystem']['id_cert_filename']
         cert_filename = os.path.join(id_cert_dir, id_cert_name)
-        if os.path.exists(cert_filename):
-            os.remove(cert_filename)
+        server_rsa = self.context.config['server']['rsa_pub']
+        for filename in [cert_filename, server_rsa]:
+            if os.path.exists(filename):
+                os.remove(filename)
 
 
 class HistoryCommand(PulpCliCommand):


pcreech:
This seemed the simplest way of ensuring the cleanup.  I'm not entirely sure if it's the most appropriate approach.  I'm open to any feedback.

client_consumer/pulp/client/consumer/cli.py
```diff
@@ -316,8 +316,10 @@ def _delete_cert(self):
         id_cert_dir = self.context.config['filesystem']['id_cert_dir']
         id_cert_name = self.context.config['filesystem']['id_cert_filename']
         cert_filename = os.path.join(id_cert_dir, id_cert_name)
-        if os.path.exists(cert_filename):
-            os.remove(cert_filename)
+        server_rsa = self.context.config['server']['rsa_pub']
+        for filename in {cert_filename, server_rsa}:
```
mhrivnak: this set notation is new in python 2.7, but we need to run on 2.6.

mhrivnak: actually, it's even worse: since this is in the consumer CLI, it needs to run in python 2.4!

pcreech:
Updated to use [] instead of {}

mhrivnak:
ok test

mhrivnak:
It looks like the tests failures are not your fault. We'll get them straightened out on master and then re-run the tests here.

jortel:
This looks good.  

@mhrivnak, I had a comment on the issue that suggested this code be shared with the agent.  However, I really don't think that the current behavior of invoking the Consumer.unregistered() RMI on the agent when a consumer is unregistered using API/pulp-admin is valuable.  Instead, I suggest that registration status and associated artifacts (key, certificate) should only be managed by pulp-consumer and that we discontinue invoking Consumer.unregistered().  I users really want to remotely unregister a consumer, I think we should support a new API `pulp/consumers/<id>/actions/unregister/` that basically does a `pulp-consumer unregister`.

FWIW, In satellite, the registration status and certificate on the content host are only managed by subscription manager.

jortel:
Added story: https://pulp.plan.io/issues/1307

jortel:
ok test

jortel:
It's a little concerning that this changes does not cause a unit test to fail.  I'll review the tests.

PR #2098, opened 2015-10-05
python-requests 2.6+ is now available on all platforms.

jeremycline:
Originally, we carried python-requests because EL6 and EL7
had versions less than 2.2, which is required by Nectar.
This is no longer the case[0], so we can stop building requests.

I verified that the spec files for EL6 and EL7 are unbundling urllib3, chardet, and the certificate bundle from Mozilla. It also contains the patch to set the CA bundle to the system-provided one.

[0] https://access.redhat.com/errata/RHEA-2015:1418

closes #816


mhrivnak:
This imposes a limit on which versions of el6.x and el7.x pulp will support, and I think we should capture that somehow. Perhaps add a platform release note stating which versions are supported?

PR #2101, opened 2015-10-06
Redirect redoux

jortel:
Lazy redirecting redoux.
* Added separate WSGI application for serving content.  Mainly defined in: `pulp.server.content.web`.  This application is the endpoint for `/pulp/content/*` URLs.
* Moved RedirectView to `pulp.server.content.web` and added X-SENDFILE functionality and authorization.
* Eliminated need for Alias parsing.  See: `<Location/>` below.
* Updated pulp-dev.py.

Plugins participating in lazy loading need to update their .conf files in `/etc/httpd/conf.d/` as follows:

For example in RPM, This:
```
# -- HTTPS Repositories ---------
Alias /pulp/repos /var/www/pub/yum/https/repos

<Directory /var/www/pub/yum/https>
    WSGIAccessScript /srv/pulp/repo_auth.wsgi
    SSLRequireSSL
    SSLVerifyClient optional_no_ca
    SSLVerifyDepth 2
    SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
    Options FollowSymLinks Indexes
</Directory>
```
Changed to:
```
<Location /pulp/repos/>
    RewriteEngine On
    RewriteRule (.+/pulp/repos/)(.+) /pulp/content/var/www/pub/yum/https/repos/$2
    SSLRequireSSL
    SSLVerifyClient optional_no_ca
    SSLVerifyDepth 2
    SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
</Location>
```


server/etc/httpd/conf.d/pulp_content.conf
```diff
@@ -0,0 +1,25 @@
+
+SSLInsecureRenegotiation on
```
jeremycline: In my opinion we should leave this commented out or remove it. Apparently this was added because very old yum clients didn't support secure renegotiation[0] and although the version mentioned in the bug (3.2.25) is newer than RHEL5's version (3.2.22), I believe the Apache configuration we provide should be secure by default. If a user wants to support RHEL5, they can uncomment/add it.

I do also wonder if the RHEL5 version of yum (or rather openssl because I suspect yum uses openssl) has been patched to support secure renegotiation.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=647828

server/pulp/server/content/web/views.py
```diff
@@ -1,16 +1,18 @@
...
 
 
-class RedirectView(View):
+class ContentView(View):
     """
-    The content redirect view redirects unsatisfied content
-    requests to the lazy streamer.
+    The content delivery view provides content.
 
     :ivar alias: An apache alias table.
```
jeremycline: Stale docblock

server/pulp/server/content/web/views.py
```diff
@@ -51,52 +53,91 @@ def urljoin(scheme, host, port, base, path, query):
...
+        if not enabled:
+            # Not Redirected
+            return HttpResponseNotFound(path)
 
-        query = request.environ.get('REDIRECT_QUERY_STRING', '')
         scheme = request.environ['REQUEST_SCHEME']
         host = request.environ['SERVER_NAME']
         port = request.environ['SERVER_PORT']
-        enabled = parse_bool(pulp_conf.get('lazy', 'enabled'))
+        query = request.environ.get('REDIRECT_QUERY_STRING', '')
```
jeremycline: Reading http://httpd.apache.org/docs/2.4/custom-error.html#variables makes me believe (though I have not actually tested it) that this variable will never be provided since we aren't using the ErrorDocument command.

server/pulp/server/content/web/views.py
```diff
@@ -51,52 +53,91 @@ def urljoin(scheme, host, port, base, path, query):
...
+
+        # Authorization
+        if not allow_access(request.environ, host):
+            # Not Authorized
+            return HttpResponseForbidden()
+
+        # Delivery
+        if os.path.exists(path):
+            return self.x_send(path)
         else:
```
jeremycline: Completely personal preference (and very quibbly), but I feel like it would read a bit more smoothly if the check to see if lazy is enabled happened here rather than in the `redirect` method, so something like 
```
if path exists:
    send file
elif lazy not enabled:
    404
else:
    redirect
```
and then remove the enabled check from redirect.

jeremycline:
Once we decide what do to with the Apache configuration it'll LGTM.

server/pulp/server/content/web/views.py
jortel: The REDIRECT_ vars seem to be populated by mod_rewirte.  Almost positive I saw them in the debugger but will investigate to be sure.

server/pulp/server/content/web/views.py
jortel: Yeah, I like that better.

server/pulp/server/lazy/__init__.py
```diff
@@ -0,0 +1,2 @@
+from pulp.server.lazy.alias import AliasTable
```
jeremycline: Why is this here?

jortel: The AliasTable specifically?

PR #2104, opened 2015-10-08
Add an Apache configuration file for the lazy reverse proxy.

jeremycline:


pulp-dev.py
```diff
@@ -104,6 +104,7 @@
         # Lazy Streamer components
         ('server/srv/pulp/streamer.tac', '/srv/pulp/streamer.tac'),
         ('server/srv/pulp/streamer_auth.wsgi', '/srv/pulp/streamer_auth.wsgi'),
+        ('server/etc/httpd/conf.d/pulp_lazy.conf', '/etc/httpd/conf.d/pulp_lazy.conf'),
```
jortel: I think `pulp_streamer.conf` would be more appropriate.  The term `lazy` is a broader term for the general concept which includes more than the streamer.

jortel:
:+1:  for making this a separate configuration.

pulp-dev.py
jeremycline: I like that better, too. I'm ashamed I'm getting so lazy with my names.

PR #2105, opened 2015-10-08
flatten the catalog model.

jortel:
Changes:
- Flatten embedded documents.  Mainly because they won't be supported by postgres in the future.
- Renamed: `plugin_id` => `importer_id`.  More accurate.
- Renamed `relative_path` => `path`.  The value is an absolute path.

At @jeremycline request, the referencing code has not be fixed to match these changes.  He will make the necessary changes in subsequent PR.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -4,8 +4,8 @@ import os
 import uuid
 from collections import namedtuple
 
-from mongoengine import (DateTimeField, DictField, Document, DynamicField, EmbeddedDocument,
-                         EmbeddedDocumentField, IntField, ListField, StringField)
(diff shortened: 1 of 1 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -647,65 +647,42 @@ class CeleryBeatLock(Document):
...
 class LazyCatalogEntry(Document):
     """
     A catalog of content that can be downloaded by the specified plugin.
 
-    :ivar relative_path: The unit storage path that is relative to /var/lib/pulp/content.
-    :type relative_path: str
-    :ivar plugin_id: The ID of the plugin that contributed the catalog entry.
+    :ivar path: The content unit storage path.
+    :type path: str
+    :ivar importer_id: The ID of the plugin that contributed the catalog entry.
```
jeremycline: :+1:

PR #2107, opened 2015-10-08
Fixes unit tests for database connection initialize() method

dkliban:


server/test/unit/server/db/test_connection.py
```diff
@@ -84,28 +82,27 @@ def test_seeds_from_config(self, mock_connect_seeds, mock_mongoengine):
...
+    @mock_config.patch({'database': {'seeds': 'firsthost:1234,secondhost:5678'}})
+    @patch('pulp.server.db.connection._CONNECTION', None)
+    @patch('pulp.server.db.connection._DATABASE', None)
     @patch('pulp.server.db.connection.mongoengine')
     @patch('pulp.server.db.connection._connect_to_one_of_seeds')
     def test_multiple_seeds_no_replica_set(self, mock_connect_seeds, mock_mongoengine):
         mock_mongoengine.connect.return_value.server_info.return_value = {'version': '2.6.0'}
         mock_connect_seeds.return_value.server_info.return_value = {'version': '2.6.0'}
-        seeds = "firsthost:1234,secondhost:5678"
-        config.config.set('database', 'write_concern', 'majority')
```
asmacdo: I assume that this one doesn't need to be in there?

dkliban: It does not. I am surprised it was not causing a test failure.

PR #2108, opened 2015-10-09
Update the download_one task to use the streamer.

jeremycline:
This also updates the streamer to check for a special HTTP header to
determine whether or not it should dispatch a task and changes the
streamer to use the new LazyCatalogEntry model.

I'm hesitant to write tests for the task just yet, since we may end up changing it quite a bit depending on whether we download entire content units or not.

### DIFF
--- a/server/pulp/server/constants.py
+++ b/server/pulp/server/constants.py
@@ -2,3 +2,5 @@ LOCAL_STORAGE = "/var/lib/pulp/"
 
 PULP_USER_METADATA_FIELDNAME = 'pulp_user_metadata'
 PULP_DJANGO_SETTINGS_MODULE = 'pulp.server.webservices.settings'
+
+PULP_STREAM_REQUEST_HEADER = 'Pulp-Stream-Request'
--- a/server/pulp/server/controllers/content.py
+++ b/server/pulp/server/controllers/content.py
@@ -5,6 +5,7 @@ import os
 import threading
 
 import celery
+from mongoengine import DoesNotExist
 from nectar.request import DownloadRequest
 from nectar.listener import AggregatingEventListener
 
@@ -17,13 +18,30 @@ from pulp.plugins.loader import api as plugin_api
 from pulp.plugins.loader.exceptions import PluginNotFound
 from pulp.plugins.util.publish_step import Step
 from pulp.server.async.tasks import Task
+from pulp.server.config import config as pulp_conf
+from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
 from pulp.server.controllers import repository
 from pulp.server.content.sources.container import ContentContainer
+from pulp.server.content.web.views import ContentView
 from pulp.server.exceptions import PulpCodedTaskException
+from pulp.server.lazy import URL, Key
 from pulp.server.managers.repo._common import get_working_directory
 
 
 _logger = getLogger(__name__)
+
+URL_SIGNING_KEY = Key.load(pulp_conf.get('authentication', 'rsa_pub'))
+
+
+class ModelNotFound(Exception):
+    """
+    Raised when a content unit Model was not found.
+    """
+    def __init__(self, model_type):
+        self.model_type = model_type
+
+    def __str__(self):
+        return _('Model type "{type}" not found.').format(type=self.model_type)
 
 
 class ContentSourcesConduit(StatusMixin, PublishReportMixin):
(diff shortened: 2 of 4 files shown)


server/pulp/server/controllers/content.py
```diff
@@ -111,83 +126,110 @@ def refresh_content_source(content_source_id=None):
...
+    scheme = 'https'
+    host = pulp_conf.get('lazy', 'redirect_host')
+    port = pulp_conf.get('lazy', 'redirect_port')
+    path_prefix = pulp_conf.get('lazy', 'redirect_path')
+    return ContentView.urljoin(scheme, host, port, path_prefix,
+                               catalog_entry.path, '')
 
 
-def queue_download(plugin_id, unit_locator, url, options):
+def queue_download(catalog_entry):
```
jortel: should this be `queue_download_one()`?

server/pulp/server/controllers/content.py
```diff
@@ -111,83 +126,110 @@ def refresh_content_source(content_source_id=None):
...
+    :param catalog_entry: The catalog entry to get the URL for.
+    :type  catalog_entry: pulp.server.db.model.LazyCatalogEntry
+
+    :return: The streamer URL which corresponds to the content unit.
+    """
+    scheme = 'https'
+    host = pulp_conf.get('lazy', 'redirect_host')
+    port = pulp_conf.get('lazy', 'redirect_port')
+    path_prefix = pulp_conf.get('lazy', 'redirect_path')
+    return ContentView.urljoin(scheme, host, port, path_prefix,
```
jortel: Shouldn't this be be a signed URL?

server/pulp/server/constants.py
```diff
@@ -2,3 +2,5 @@
 
 PULP_USER_METADATA_FIELDNAME = 'pulp_user_metadata'
 PULP_DJANGO_SETTINGS_MODULE = 'pulp.server.webservices.settings'
+
+PULP_STREAM_REQUEST_HEADER = 'Pulp-Stream-Request'
```
jortel: I wonder if this should begin with 'X-'?

server/pulp/server/controllers/content.py
```diff
@@ -111,83 +126,110 @@ def refresh_content_source(content_source_id=None):
...
-    :type url: str
-    :param options: Passed to the Importer.get_downloader().
-    :type options: dict
+    :param catalog_entry: The catalog entry to download content for.
+    :type  catalog_entry: pulp.server.db.model.LazyCatalogEntry
     """
-    _logger.info(_('Downloading: {url}').format(url=url))
+    _logger.info(_('Downloading {url} via the Pulp streamer').format(url=catalog_entry.url))
 
     # Find the plugin
```
jortel: Comment seems unrelated.

server/pulp/server/lazy/streamer.py
```diff
@@ -102,51 +103,64 @@ def render_GET(self, request):
...
         :param request: The content request.
         :type  request: twisted.web.server.Request
         """
-        relative_path = urlparse(request.uri).path.lstrip('/')
+        catalog_path = urlparse(request.uri).path
         with Responder(request) as responder:
             try:
                 catalog_entry = model.LazyCatalogEntry.objects(
-                    relative_path=relative_path).order_by('-plugin_id').first()
+                    path=catalog_path).order_by('-plugin_id').first()
```
jortel: s/-plugin_id/importer_id/

server/pulp/server/lazy/streamer.py
```diff
@@ -102,51 +103,64 @@ def render_GET(self, request):
...
-                downloader.event_listener = StreamerListener(request)
-                downloader.download_one(download_request, events=True)
-
-                # Dispatch the Pulp task to download the unit.
-                content_controller.queue_download(catalog_entry.plugin_id,
-                                                  catalog_entry.unit_locator,
-                                                  catalog_entry.url,
-                                                  catalog_entry.data)
+                # Only dispatch the task if the request didn't originate from Pulp.
+                if not request.getHeader(PULP_STREAM_REQUEST_HEADER):
```
jortel: :+1:

server/pulp/server/controllers/content.py
jeremycline: Oops,  it definitely should be.

server/pulp/server/constants.py
jeremycline: I did a bit of research about this because I knew at one point non-standard headers were supposed to start with 'X-'. According to http://tools.ietf.org/html/rfc6648#section-3 new headers should not use the 'X-'.

PR #2109, opened 2015-10-09
Add pulp-smash setup to the Vagrant environment.

jeremycline:


playpen/vagrant-setup.sh
```diff
@@ -55,6 +55,28 @@ EOF
...
+    setvirtualenvproject
+    # Install dependencies
+    pip install -r requirements.txt -r requirements-dev.txt
+    python setup.py develop
+    mkdir -p $HOME/.config/pulp_smash/
+    cat << EOF > $HOME/.config/pulp_smash/settings.json
+{
+    "default": {
+        "base_url": "https://dev.example.com",
+        "auth": ["admin", "admin"]
```
dkliban: I would  add 

>"verify": false

to the config

PR #2111, opened 2015-10-12
Update the ContentUnit model with a `downloaded` field.

jeremycline:
Furthermore, the `download_catalog_entry` method now checks
to see if all files in the content unit are present, and if they are, it 
will set the `downloaded` field on the ContentUnit to `True`.

### DIFF
--- a/server/pulp/server/controllers/content.py
+++ b/server/pulp/server/controllers/content.py
@@ -23,6 +23,7 @@ from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
 from pulp.server.controllers import repository
 from pulp.server.content.sources.container import ContentContainer
 from pulp.server.content.web.views import ContentView
+from pulp.server.db.model import LazyCatalogEntry
 from pulp.server.exceptions import PulpCodedTaskException
 from pulp.server.lazy import URL, Key
 from pulp.server.managers.repo._common import get_working_directory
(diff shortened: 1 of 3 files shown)


server/pulp/server/controllers/content.py
```diff
@@ -176,14 +177,30 @@ def _download_catalog_entry(catalog_entry):
     if downloader.event_listener.succeeded_reports:
         _logger.info(_('Download {url} via {streamer_url} succeeded').format(
             url=catalog_entry.url, streamer_url=request.url))
-        content_unit.set_content(request.destination)
+        content_unit.set_content(request.destination, downloaded=False)
```
jortel: FALSE?

server/pulp/server/db/model/__init__.py
```diff
@@ -563,13 +566,15 @@ def pre_save_signal(cls, sender, document, **kwargs):
         with FileStorage() as storage:
             storage.put(document, document._source_location)
 
-    def set_content(self, source_location):
+    def set_content(self, source_location, downloaded=True):
```
jortel: This method is used to copy downloaded content into the proper location in the platform.  How can this ever be FALSE?

server/pulp/server/controllers/content.py
jeremycline: Yes, it's quite confusing. Suppose we're dealing with a multi-file content unit: we download one file and copy it to its final resting place, but we don't know if that's the last file for that content unit, so we have to say that it's false. Later, we check to see if all the files are present, and if they are, we set it to True.

This is clearly not what you were thinking about though, so maybe we should sync up.

jeremycline:
I'm going to go ahead and close this since no matter what we do, this will need significant revision.

PR #2112, opened 2015-10-13
Adds group_id to TaskStatus model

dkliban:
TaskStatus has a group_id field. The group_id is used to group a set
of tasks together. The TaskStatus view selects only TaskStatus documents
that don't have a group_id set. The task_id field should only be set
when performing parallelized work.

https://pulp.plan.io/issues/1204
closes #1204

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -300,6 +300,8 @@ class ReservedTaskMixin(object):
         :param tags:          A list of tags (strings) to place onto the task, used for searching
                               for tasks by tag
         :type  tags:          list
+        :param group_id:      The id to identify which group of tasks a task belongs to
+        :type  group_id:      uuid.UUID
         :return:              An AsyncResult instance as returned by Celery's apply_async
         :rtype:               celery.result.AsyncResult
         """
@@ -309,10 +311,12 @@ class ReservedTaskMixin(object):
         inner_task_id = str(uuid.uuid4())
         task_name = self.name
         tag_list = kwargs.get('tags', [])
+        group_id = kwargs.get('group_id', None)
 
         # Create a new task status with the task id and tags.
         task_status = TaskStatus(task_id=inner_task_id, task_type=task_name,
-                                 state=constants.CALL_WAITING_STATE, tags=tag_list)
+                                 state=constants.CALL_WAITING_STATE, tags=tag_list,
+                                 group_id=group_id)
         # To avoid the race condition where __call__ method below is called before
         # this change is propagated to all db nodes, using an 'upsert' here and setting
         # the task state to 'waiting' only on an insert.
@@ -344,12 +348,15 @@ class Task(CeleryTask, ReservedTaskMixin):
         :param tags:        A list of tags (strings) to place onto the task, used for searching for
                             tasks by tag
         :type  tags:        list
+        :param group_id:    The id that identifies which group of tasks a task belongs to
+        :type group_id:     uuid.UUID
         :return:            An AsyncResult instance as returned by Celery's apply_async
         :rtype:             celery.result.AsyncResult
         """
         routing_key = kwargs.get('routing_key',
                                  defaults.NAMESPACES['CELERY']['DEFAULT_ROUTING_KEY'].default)
         tag_list = kwargs.pop('tags', [])
+        group_id = kwargs.pop('group_id', None)
 
         async_result = super(Task, self).apply_async(*args, **kwargs)
         async_result.tags = tag_list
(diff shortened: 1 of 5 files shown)


dkliban:
ok test

server/pulp/server/async/tasks.py
```diff
@@ -344,12 +348,15 @@ def apply_async(self, *args, **kwargs):
         :param tags:        A list of tags (strings) to place onto the task, used for searching for
                             tasks by tag
         :type  tags:        list
+        :param group_id:    The id that identifies which group of tasks a task belongs to
+        :type group_id:     basestring
```
bmbouter: Consider making this a uuid instead of a basestring. If it's just a basestring we'll have to write a generator, when it would be better just use UUIDs since they are great and we don't have to maintain them. It would be good to add a doc about usage here so that callers know this needs to be a UUID.

server/pulp/server/async/tasks.py
```diff
@@ -344,12 +348,15 @@ def apply_async(self, *args, **kwargs):
...
         :return:            An AsyncResult instance as returned by Celery's apply_async
         :rtype:             celery.result.AsyncResult
         """
         routing_key = kwargs.get('routing_key',
                                  defaults.NAMESPACES['CELERY']['DEFAULT_ROUTING_KEY'].default)
         tag_list = kwargs.pop('tags', [])
+        group_id = kwargs.pop('group_id', None)
```
bmbouter: Consider also adding an `isinstance` check that will raise an exception if this argument is not None and is not an instance of `uuid.UUID`.

server/pulp/server/async/tasks.py
```diff
@@ -300,6 +300,8 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         :param tags:          A list of tags (strings) to place onto the task, used for searching
                               for tasks by tag
         :type  tags:          list
+        :param group_id:      The id to identify which group of tasks a task belongs to
+        :type  group_id:      basestring
```
bmbouter: same here as below, regarding UUIDs

server/pulp/server/async/tasks.py
```diff
@@ -309,10 +311,12 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         inner_task_id = str(uuid.uuid4())
         task_name = self.name
         tag_list = kwargs.get('tags', [])
+        group_id = kwargs.get('group_id', None)
```
bmbouter: consider adding the `isinstance` call (similar to the comment below).

bmbouter:
@dkliban back to you. I'm waiting on the LGTM until we figure out the UUIDs.

dkliban:
ok test

server/pulp/server/async/tasks.py
```diff
@@ -300,6 +300,8 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         :param tags:          A list of tags (strings) to place onto the task, used for searching
                               for tasks by tag
         :type  tags:          list
+        :param group_id:      The id to identify which group of tasks a task belongs to
+        :type  group_id:      UUID
```
bmbouter: this should be `uuid.UUID`

server/pulp/server/db/model/__init__.py
```diff
@@ -315,6 +317,7 @@ class TaskStatus(AutoRetryDocument, ReaperMixin):
     start_time = ISO8601StringField()
     finish_time = ISO8601StringField()
     result = DynamicField()
+    group_id = StringField(default=None)
```
bmbouter: Should this be a UUIDField instead?

http://docs.mongoengine.org/en/latest/apireference.html#mongoengine.fields.UUIDField

server/pulp/server/async/tasks.py
```diff
@@ -344,12 +352,19 @@ def apply_async(self, *args, **kwargs):
         :param tags:        A list of tags (strings) to place onto the task, used for searching for
                             tasks by tag
         :type  tags:        list
+        :param group_id:    The id that identifies which group of tasks a task belongs to
+        :type group_id:     UUID
```
bmbouter: This should also be `uuid.UUID`

server/pulp/server/db/model/__init__.py
bmbouter: If this is switched to a UUID the mongoengine validation can take care of it and we don't need the `isinstance` or the `str(group_id)` anymore. That would be the best. We need to ensure two things:

1) that a UUIDField can be serialized by the celery JSON serializer (for an upcoming change)
2) that postgres supports UUID fields natively for when we switch to that

dkliban:
I checked that we can use UUIDField with with PostgreSQL later.

dkliban:
ok test

dkliban:
ok test

server/pulp/server/db/model/__init__.py
```diff
@@ -296,6 +296,8 @@ class TaskStatus(AutoRetryDocument, ReaperMixin):
     :type start_time:  basestring
     :ivar finish_time: ISO8601 representation of the time the task completed
     :type finish_time: basestring
+    :ivar group_id:    The id used to identify which  group of tasks a task belongs to
+    :type group_id:    basestring
```
bmbouter: This should be UUID not basestring

bmbouter:
@dkliban LGTM after fixing the docstring

dkliban:
ok test

dkliban:
ok test

PR #2114, opened 2015-10-14
Convert repo importers collection to mongoengine.

asmacdo:
closes: #782


server/pulp/plugins/conduits/mixins.py
```diff
@@ -266,37 +266,32 @@ def get_scratchpad(self):
...
+        :type  value: dict
 
-        @raise ImporterConduitException: wraps any exception that may occur
-               in the Pulp server
+        :raises ImporterConduitException: wraps any exception that may occur in the Pulp server
         """
         try:
-            importer_manager = manager_factory.repo_importer_manager()
-            importer_manager.set_importer_scratchpad(self.repo_id, value)
+            importer = model.Importer.objects(repo_id=self.repo_id).first()
```
mhrivnak: can you use ``.get()`` instead?

http://docs.mongoengine.org/apireference.html#mongoengine.queryset.QuerySet.get

server/pulp/server/controllers/importer.py
```diff
@@ -0,0 +1,281 @@
...
+    :return: key-value pairs describing the importer in use
+    :rtype:  dict
+
+    :raises pulp.server.exceptions.MissingResource: if repo/importer combination cannot be found
+    """
+    model.Repository.objects.get_repo_or_missing_resource(repo_id)
+    try:
+        importer = model.Importer.objects.get_or_404(repo_id=repo_id)
+        if importer.importer_type_id != importer_type_id:
+            raise exceptions.MissingResource(importer_id=importer_type_id)
```
mhrivnak: This will get caught on the next line, thrown away, and replaced by a new but identical exception. One solution is to move these two lines below the ``except`` block. Then you at least have two distinct code paths that end up raising the same exception.

server/pulp/plugins/conduits/mixins.py
mhrivnak: just as another option that's a bit more readable, this could be
```
 if not isinstance(distributor_list, (list, tuple, type(None))):
```

or

```
 import types
 if not isinstance(distributor_list, (list, tuple, types.NoneType)):
```

server/pulp/server/controllers/repository.py
```diff
@@ -202,7 +203,7 @@ def create_repo(repo_id, display_name=None, description=None, notes=None, import
     """
 
     # Prevalidation.
-    if not isinstance(distributor_list, (list, tuple)):
+    if distributor_list is not None and not isinstance(distributor_list, (list, tuple)):
```
mhrivnak: just as another option that's a bit more readable, this could be
```
 if not isinstance(distributor_list, (list, tuple, type(None))):
```

or

```
 import types
 if not isinstance(distributor_list, (list, tuple, types.NoneType)):
```

server/pulp/server/controllers/repository.py
```diff
@@ -229,24 +229,25 @@ def create_repo(repo_id, display_name=None, description=None, notes=None, import
...
-            auto_publish = distributor.get('auto_publish', False)
-            distributor_id = distributor.get('distributor_id')
-            distributor_manager.add_distributor(repo_id, type_id, plugin_config, auto_publish,
-                                                distributor_id)
-        except Exception:
-            _logger.exception('Exception adding distributor to repo [%s]; the repo will be '
-                              'deleted' % repo_id)
-            repo.delete()
-            raise
+    if distributor_list:
```
mhrivnak: looks fine, and there's not a need to change it, but another option that avoids an extra "if" block and also extra indentation would be this:

```
for distributor in distributor_list or []:
```

server/pulp/server/controllers/repository.py
```diff
@@ -229,24 +229,25 @@ def create_repo(repo_id, display_name=None, description=None, notes=None, import
...
+        distributor_manager = manager_factory.repo_distributor_manager()
+        for distributor in distributor_list:
+            if not isinstance(distributor, dict):
```
mhrivnak: It would be better to do this check above, before creating the repo. You're already doing some validation on this value at line 206, so consider putting this there.

server/pulp/server/db/migrations/0021_remove_extra_importer_fields.py
```diff
@@ -0,0 +1,20 @@
...
+    Perform the migration as described in this module's docblock.
+
+    :param args:   unused
+    :type  args:   list
+    :param kwargs: unused
+    :type  kwargs: dict
+    """
+    db = connection.get_database()
+    collection = db['repo_importers']
+    collection.update({}, {"$unset": {"id": True}})
```
mhrivnak: does this also need ``multi=True`` ?

server/pulp/server/controllers/importer.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+
+    if config is not None:
+        return dict([(k, v) for k, v in config.items() if v is not None])
+    else:
+        return None
+
+
+@celery.task(base=Task, name='pulp.server.managers.repo.importer.set_importer')
+def set_importer(repo_id, importer_type_id, repo_plugin_config):
```
mhrivnak: It looks like this function only gets called from one place, and that place has the repo object. It could pass in that object, instead of making this function look it up below.

server/pulp/server/controllers/repository.py
```diff
@@ -293,15 +294,13 @@ def delete(repo_id):
     # the server logs for more information.
     error_tuples = []  # tuple of failed step and exception arguments
 
-    importer_manager = manager_factory.repo_importer_manager()
     distributor_manager = manager_factory.repo_distributor_manager()
 
     # Inform the importer
-    importer_coll = RepoImporter.get_collection()
-    repo_importer = importer_coll.find_one({'repo_id': repo_id})
+    repo_importer = model.Importer.objects(repo_id=repo_id).first()
```
mhrivnak: same question as elsewhere: can you use ``.get()`` instead, which ensures only one is found?

server/pulp/server/controllers/repository.py
asmacdo: :heart:  python

server/pulp/server/db/migrations/0021_remove_extra_importer_fields.py
asmacdo: good catch!

server/pulp/server/controllers/importer.py
asmacdo: Thanks! I really hated this block, now it can be a little less ugly.

server/pulp/server/controllers/repository.py
asmacdo: For the other places where we know that the importer does exist or if it doesn't the exception will be caught, `.get()` works well. But in this case, we only need to do something if the importer exists, so it doesn't seem worth catching the `DoesNotExist` and having a try/except inside of another try/except.

asmacdo:
@mhrivnak, comments taken, mind having another look?

PR #2115, opened 2015-10-14
ContentUnit now models pulp_user_metadata as pulp_user_metadata

bmbouter:
The first-pass implementation of ContentUnit for mongoengine
used an attribute name of 'user_metadata' to model the database
field named 'pulp_user_metadata'. This is a nice chance, but it
forces multiple plugins to handle name translation for both
Criteria objects and serializers.

By modeling the attribute as the actual database name everything
becomes simpler. We can rename pulp_user_metadata to
user_metadata in a future X release.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -398,8 +398,8 @@ class ContentUnit(AutoRetryDocument):
     :type id: mongoengine.StringField
     :ivar last_updated: last time this unit was updated (since epoch, zulu time)
     :type last_updated: mongoengine.IntField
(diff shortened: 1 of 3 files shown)


bmbouter:
ok test

ipanova:
@bmbouter i think in the past Randy and Michael had some conversation about this https://pulp.plan.io/issues/86#note-15

bmbouter:
Getting mhrivnak or rbarlow's input would be good. As I see it, this PR makes makes the internal and external representation of `pulp_user_metadata`. That should be consistent with this comment: https://pulp.plan.io/issues/86#note-17

PR #2116, opened 2015-10-14
Content sources container API safe to be used within Twisted.

jortel:
Added non-threaded API for use within reactor frameworks such as Twisted.
Usage:
```
container = ContentContainer()
container.threaded = False
...
```
Simplified code with better Listener/Event model.
Removed cancel logic.

### DIFF
--- a/server/pulp/server/content/sources/container.py
+++ b/server/pulp/server/content/sources/container.py
@@ -4,9 +4,10 @@ from threading import Thread, RLock
 from Queue import Queue, Empty, Full
 
 from nectar.listener import DownloadEventListener
-from nectar.report import DownloadReport as NectarDownloadReport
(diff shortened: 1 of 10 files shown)


server/pulp/server/content/sources/container.py
```diff
@@ -189,28 +154,149 @@ def download_failed(self, report):
...
+            event(self.listener)
+        return report
+
+    @staticmethod
+    def _download(url, destination, source):
+        """
+        Download the URL using the source.
+
+        :param source: A content source used for the download.
+        :type source: ContentSource
```
jeremycline: Docstring ordering doesn't match parameter ordering

jeremycline:
For such a large PR, I feel like I should have more complaints, but it looks good to me!

PR #2117, opened 2015-10-15
Renames last_updated to _last_updated on ContentUnit

bmbouter:
The first-pass implementation of ContentUnit for mongoengine
used an attribute name of 'last_updated' to model the database
field named '_last_updated'. This is a nice chance, but it
forces multiple plugins to handle name translation for both
Criteria objects and serializers.

By modeling the attribute as the actual database name everything
becomes simpler. We can rename _last_updated to last_updated in
a future X release.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -396,8 +396,8 @@ class ContentUnit(AutoRetryDocument):
 
     :ivar id: content unit id
     :type id: mongoengine.StringField
-    :ivar last_updated: last time this unit was updated (since epoch, zulu time)
-    :type last_updated: mongoengine.IntField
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -551,8 +551,7 @@ def __init__(self, *args, **kwargs):
     def pre_save_signal(cls, sender, document, **kwargs):
         """
         The signal that is triggered before a unit is saved, this is used to
-        support the legacy behavior of generating the unit id and setting
-        the last_updated timestamp
+        move the unit file or directory into place.
```
bmbouter: This docstring was wrong, but the change is unrelated to `_last_updated`.

PR #2118, opened 2015-10-15
Adds pulp.common.util.copytree method

dkliban:
Implements copytree method without use of copy2() and copystat() to prevent
copying of metadata. This ensures files have proper selinux security context
based on the new location.

https://pulp.plan.io/issues/1253
re: 1253
https://pulp.plan.io/issues/1312
re: 1312

### DIFF
--- a/server/pulp/server/util.py
+++ b/server/pulp/server/util.py
@@ -14,6 +14,9 @@
 # XXX this is not a dumping grounds for any random code. It is a place to put
 # paradigm-changing code that allows you to get unique or more efficient behaviors
 
+import os
+from shutil import copy
+
 from gettext import gettext as _
 
 from pulp.server.exceptions import PulpExecutionException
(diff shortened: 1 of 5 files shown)


dkliban:
ok test

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
         for k,v in obj.items():
             if k in filter:
                 self[k] = v
+
+def copytree(src, dst, symlinks=False, ignore=None):
+    """
+    Copies src tree to dst
+
+    shutil.copytree method uses copy2() and copystat() to perform the recursive copy of a
+    directory. Both of these methods copy the attributes associated with the files. We are When copying
```
jeremycline: s/we are//

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
...
+    files from /var/cache/pulp to /var/lib/pulp, we don't want to copy the SELinux security context
+    labels. See: https://pulp.plan.io/issues/1253
```
jeremycline: I feel the doc-block explains the reason for the method's existence well enough to not need a reference to the issue. That URL may one day not be around and leave the reader wondering what else they might be missing. Personal preference, though.

I think there's also usually a space between the doc block and the param list.

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
...
+    :param src: Source directory rooted at src
+    :type  src: basestring
+    :param dst: Destination directory, a new directory and any parent directories are created if
+                any are missing
+    :type  dst: basestring
+    :param symlinks: If true, symlinks are copied as symlinks and metadata of original links is not
+                     copied. If false, the contents and metadata of symlinks is coppied to the new
```
jeremycline: s/coppied/copied/

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
...
+                     tree.
+    :type  symlinks: boolean
+    :param ignore: If provided, it receives as it's arguments the directory being visited by
```
jeremycline: s/it's/its/

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
...
+                   must return a sequence of directory and file names relative to the current
+                   directory (i.e. a subset of the items in its second argument); these names will
+                   then be ignored in the copy process.
+    :type  ignore: Callable
+    """
+
+    names = os.listdir(src)
+    if ignore is not None:
+        ignored_names = ignore(src, names)
+        print("list of ignored: %s" % ignored_names)
```
jeremycline: Is this supposed to be here?

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
...
+                copy(srcname, dstname)
+            # XXX What about devices, sockets etc.?
+        except (IOError, os.error) as why:
+            errors.append((srcname, dstname, str(why)))
+        # catch the Error from the recursive copytree so that we can
+        # continue with other files
+        except Error as err:
+            errors.extend(err.args[0])
+    if errors:
+        raise Error(errors)
```
jeremycline: `shutil.copytree` returns `dst` rather than `None` so I think this should, too.

server/pulp/server/util.py
```diff
@@ -182,3 +185,66 @@ def __init__(self, obj, filter=()):
         for k,v in obj.items():
             if k in filter:
                 self[k] = v
+
+def copytree(src, dst, symlinks=False, ignore=None):
```
jeremycline: `shutil.copytree` accepts the `ignore_dangling_symlinks` keyword argument. The docs say:

```
If the file pointed by the symlink doesn't
exist, an exception will be added in the list of errors raised in
an Error exception at the end of the copy process.

You can set the optional ignore_dangling_symlinks flag to true if you
want to silence this exception. Notice that this has no effect on
platforms that don't support os.symlink.
```
I suppose we can add it later if we need it, though.

server/test/unit/server/test_util.py
```diff
@@ -0,0 +1,135 @@
...
+
+        The mock 'src' directory has following structure:
+            - src
+               - dir1
+                  - file1
+               - dir2
+                  - file2
+               - file3
+        """
+        mock_list_dir.side_effect = [['dir1', 'dir2', 'file3'], ['file2']]
```
jeremycline: The doc-block talks about 'file1', but it's not present here.

jeremycline:
There are quite a few print statements that need removal and a few minor comments.

server/test/unit/server/test_util.py
dkliban: The directory and it's content are supposed to be ignored.

dkliban:
ok test

PR #2128, opened 2015-10-21
Add files to content unit.

jortel:
https://pulp.plan.io/issues/1302

The *files* list on a unit is optional and only used by *lazy* content feature.  The importer should add files when creating the unit.  The lazy download tasks should update files when downloaded.
something like this:
```
try:
    file = UnitFile.objects.get(path=path)
    file.downloaded = True
    file.save()
except: DoesNotExist:
    # ignored
    pass
```

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -5,7 +5,7 @@ import uuid
 from collections import namedtuple
 
 from mongoengine import (DateTimeField, DictField, Document, DynamicField, IntField, ListField,
-                         StringField)
+                         StringField, BooleanField)
 from mongoengine import signals
 
 from pulp.common import constants, dateutils, error_codes
@@ -382,6 +382,34 @@ class TaskStatus(AutoRetryDocument, ReaperMixin):
 
 
 signals.post_save.connect(TaskStatus.post_save, sender=TaskStatus)
+
+
+class UnitFile(AutoRetryDocument):
+    """
+    A file associated with a content unit.
+
+    :ivar path: The absolute path to the file.
+    :type path: str
+    :ivar unit_id: A unit ID.
+    :type unit_id: str
+    :ivar unit_type_id: The unit type.
+    :type unit_type_id: str
+    :ivar downloaded: Indicates the file has been downloaded.
+    :type downloaded: bool
+    """
+
+    meta = {
+        'collection': 'unit_files',
+        'indexes': [
+            {'fields': ['path'], 'unique': True},
+            ('unit_id', 'unit_type_id')
+        ]
+    }
+
+    path = StringField(required=True)
+    unit_id = StringField(required=True)
+    unit_type_id = StringField(required=True)
+    downloaded = BooleanField(required=True, default=False)
 
 
 class ContentUnit(AutoRetryDocument):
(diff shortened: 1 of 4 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -471,6 +495,19 @@ def get_repositories(self):
...
 
     @property
+    def downloaded(self):
+        """
+        Whether all of the files associated with the unit have been downloaded.
+
+        :return: True if all downloaded.
+        :rtype: bool
+        """
+        for file in self.files:
```
jeremycline: `file` is a builtin type, so I'd prefer a different variable name.

server/pulp/server/db/model/__init__.py
```diff
@@ -471,6 +495,19 @@ def get_repositories(self):
         return Repository.objects(repo_id__in=id_list)
 
     @property
+    def downloaded(self):
```
jeremycline: I don't think this method has test coverage.

server/pulp/server/db/model/__init__.py
jortel: Grrr!  I hate the way python floods the global namespace with builtins.  It takes up so many useful names.  Python should require explicit import like:
```
from builtins import file, list
```

server/pulp/server/db/model/__init__.py
jortel: Tests added.

mhrivnak:
This looks simple and correct. My concern though is the number of queries that will have to happen to run ``find_units_not_downloaded``. It looks like a unique query will have to happen for each unit, and that query will always return all the files for that unit.

If we stored the unit ID and type on the file collection, we could do batched queries as we've done elsewhere (there's a ``paginate`` function in some util library). Then you could do a query that's basically "find all UnitFile entries with one of these unit IDs and where downloaded is False". That further gives us the opportunity to slap a ``.count()`` method on the query object, so none of the documents ever get actually transferred out of mongo. I think it's likely that the most common question being answered will be "are all the units in this repo downloaded?" or "how many of the units in the repo are downloaded?". Especially from a REST API perspective, it's less likely that a client will be interested in knowing which specific units haven't been downloaded.

Thoughts? many-to-one relationships in mongoengine are a bit different than what I'm used to from django, so I'm not sure if there's a graceful/idiomatic way to do this. If this was the django ORM, I'd want a ForeignKey field on the UnitFile model, referencing the unit.

Although that might fall apart when it comes to units with shared content?

jortel:
Agreed.  If doing this with an with a relation DB and ORM, I would have definitely model it the way you suggested.  And really, should have done it that way using Mongoengine.  Looking at what Mongoengine did (under the covers) here it's really bad.  I made a poor assumption that it would do something sensible but sadly it just added a list of ObjectId to the ContentUnit.  Glad you raised the question.  Reworking.

jortel:
Bench marked on my laptop and with a 5k unit repository where each unit has 10 files:
- find_repo_content_units()      632 (ms)
- find_units_not_downloaded() 2.285 (seconds)
- has_all_units_downloaded()   632 (ms) - 2.264 (seconds)

jortel:
Pretty sure I can rewrite has_all_units_downloaded() and find_units_not_downloaded() to be much faster.  find_units_not_downloaded() should probably be: find_files_not_downloaded() instead.

jeremycline:
This looks good to me, though I am not particularly experienced at optimizing mongo queries.

jortel:
No longer needed.  Different plan.

PR #2130, opened 2015-10-22
Makes some README file changes to Pulp dependencies

bmbouter:
* Consolidates several README files into a new master README
* Updates rbarlow as the maintainer for mongoengine and pymongo
* Updates bmbouter as the maintainer amqp, billiard, and celery


deps/README
```diff
@@ -0,0 +1,109 @@
...
+kombu>=3.0.24
+
+
+Permanent Patches:
+
+1212200.patch - This patch will need to continue to be with us forever as
+                it works around a permanent downstream issue whereby
+                ordereddict is installed as part of the Python system
+                library, and not in site-packages. This is not
+                contributed to upstream Kombu.
```
rbarlow: Could we just cary the OrderedDict as a dependency for Python 2.6 systems? Looks like someone has added it to pypi:

https://pypi.python.org/pypi/ordereddict

deps/README
```diff
@@ -0,0 +1,109 @@
...
+The spec file was originally adapted from the spec file inside of
+python-mongoengine-0.8.4-3.el7.src.rpm [0].
+
+[0]: http://koji.fedoraproject.org/koji/buildinfo?buildID=612617
+
+
+python-pymongo
+--------------
+
+Team owner: bmbouter
```
rbarlow: s/bmbouter/rbarlow/

bmbouter: indeed! Updated

asmacdo:
ok test

deps/README
bmbouter: I'd like to do something to improve this, but I don't think it' that simple. It's not that the bits are missing it's that [setuptools doesn't know it's there](https://bugzilla.redhat.com/show_bug.cgi?id=1212200#c7)! Since it's already there we can't add bits to solve the problem. Given that, what do you think?

PR #2131, opened 2015-10-22
refactor the search class to simplify use of GET and POST searches

asmacdo:
The base class for search was originally implemented for POST searches only and expanded to also apply to GET searches. This resulted in some mess and some bugs which are corrected here.

closes: #312


asmacdo:
ok test

PR #2133, opened 2015-10-22
wrap the use of criteria to catch invalid criteria objects

asmacdo:
https://pulp.plan.io/issues/981
closes: #981

### DIFF
--- a/server/test/unit/server/webservices/views/test_search.py
+++ b/server/test/unit/server/webservices/views/test_search.py
@@ -5,6 +5,7 @@ import unittest
 
 import mock
 from django import http
+from pymongo.errors import OperationFailure
 
 from base import assert_auth_READ
 from pulp.server import exceptions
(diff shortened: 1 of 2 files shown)


server/test/unit/server/webservices/views/test_search.py
```diff
@@ -266,6 +267,17 @@ class FakeSearchView(search.SearchView):
...
                          ['big money', 'bigger money'])
 
+    def test__generate_response_with_invalid_criteria(self):
+        """
+        Test that a pymongo exception is handled correctly.
+        """
+        class FakeSearchView(search.SearchView):
+            model = mock.MagicMock()
+
+        query = {'filters': {'money': {'$gt': 1000000}}}
```
mhrivnak: If only github had a Dr. Evil emoji...

seandst: image

It lets you drag and drop images, which I think is good enough...

PR #2134, opened 2015-10-22
log a warning when a content source config file cannot be read

mhrivnak:
fixes #1278 

https://pulp.plan.io/issues/1278

### DIFF
--- a/server/pulp/server/content/sources/model.py
+++ b/server/pulp/server/content/sources/model.py
@@ -9,6 +9,7 @@
 # have received a copy of GPLv2 along with this software; if not, see
 # http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
 
+from gettext import gettext as _
 import sys
 import os
 import re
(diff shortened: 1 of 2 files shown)


server/pulp/server/content/sources/model.py
```diff
@@ -139,7 +139,10 @@ def load_all(conf_d=None):
             if not os.path.isfile(path):
                 continue
             cfg = ConfigParser()
-            cfg.read(path)
+            list_of_read_files = cfg.read(path)
+            if not list_of_read_files:
+                log.warn("Could not read content source config file at path %s", path)
```
jortel: Should be using gettext.  Suggest *error* instead of warning because by default I don't think *warning* shows up in syslog.

rbarlow: On Fri, Oct 23, 2015 at 07:29:54AM -0700, Jeff Ortel wrote:
> Suggest *error* instead of warning because by default I don't think *warning* shows up in syslog.

I believe that warning does show up in syslog, since it is a higher
level than info and info does show up. debug is the only one that should
be missing from syslog by default.

jortel: I think I've observed *warn* not being included in the log while debugging satellite cases.

PR #2135, opened 2015-10-22
Support incremental copy of downloaded directories.

jortel:
This is needed to support downloading and copying into place - files related to content units associated with directories.  Example: distributions, docker images.  Using shutil.copytree won't work because it requires the destination directory to be empty.  This will likely have much better performance as well.


jortel:
Not need.  Different plan.

PR #2136, opened 2015-10-23
move all remaining dep information to single file

asmacdo:


deps/README
```diff
@@ -92,6 +120,16 @@ python-mongoengine-0.8.4-3.el7.src.rpm [0].
 [0]: http://koji.fedoraproject.org/koji/buildinfo?buildID=612617
 
 
+python-nectar
+-------------
+
+Team owner: ipanova
```
bmbouter: @ipanova you ok with this?
@asmacdo are you still co-maintainer?

I'm OK w/ whatever you work out in this area, and you can merge it as-is and adjust it later if you want.

PR #2137, opened 2015-10-23
Update the streamer rpm package name.

jeremycline:
Since the streamer is a Python package, its RPM package name should have the
'python-' prefix.


jortel:
I missed that :/  python- is more appropriate.  And, required by fedora packaging.

PR #2139, opened 2015-10-26
This commit fixes the permissions on the Docker publish folder.

rbarlow:
Previously, the /var/lib/pulp/published/docker/v{1,2}/app folders were owned by root instead of apache, which
caused some problems during publishing. This commit fixes them to be owned by apache for Vagrant
installations.


rbarlow:
ok test

PR #2140, opened 2015-10-26
Adds a new MongoEngine model for deferred downloads.

jeremycline:
Deferred downloads is used by the passive lazy workflow to queue up
downloads processed by the streamer. A celerybeat task will use this
queue to download the units later.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -700,3 +700,30 @@ class LazyCatalogEntry(Document):
     checksum = StringField()
     checksum_algorithm = StringField(regex=ALG_REGEX)
     data = DictField()
+
+
+class DeferredDownload(Document):
+    """
+    A collection of units that have been handled by the streamer in the
+    passive lazy workflow that Pulp should download.
+
+    :ivar unit_id:      The associated content unit ID.
+    :type unit_id:      str
+    :ivar unit_type_id: The associated content unit type.
+    :type unit_type_id: str
+    """
+    meta = {
+        'collection': 'deferred_download',
+        'indexes': [
+            {
+                'fields': ['unit_id', 'unit_type_id'],
+                'unique': True
+            }
+        ]
+    }
+
+    unit_id = StringField(required=True)
+    unit_type_id = StringField(required=True)
+
+    # For backward compatibility
+    _ns = StringField(default='deferred_download')
--- a/server/test/unit/server/db/test_model.py
+++ b/server/test/unit/server/db/test_model.py
@@ -590,3 +590,29 @@ class TestCeleryBeatLock(unittest.TestCase):
         Assert that the collection name is correct.
         """
         self.assertEquals(model.CeleryBeatLock._meta['collection'], 'celery_beat_lock')
+
+
+class TestDeferredDownload(unittest.TestCase):
+    """
+    Test the DeferredDownload class.
+    """
+
+    def test_model_superclass(self):
+        sample_model = model.DeferredDownload()
+        self.assertTrue(isinstance(sample_model, Document))
+
+    def test_attributes(self):
+        self.assertTrue(isinstance(model.DeferredDownload.unit_id, StringField))
+        self.assertTrue(model.DeferredDownload.unit_id.required)
+
+        self.assertTrue(isinstance(model.DeferredDownload.unit_type_id, StringField))
+        self.assertTrue(model.DeferredDownload.unit_type_id.required)
+
+        self.assertTrue(isinstance(model.DeferredDownload._ns, StringField))
+        self.assertEqual('deferred_download', model.DeferredDownload._ns.default)
+
+    def test_meta_collection(self):
+        """
+        Assert that the collection name is correct.
+        """
+        self.assertEquals(model.DeferredDownload._meta['collection'], 'deferred_download')


server/pulp/server/db/model/__init__.py
```diff
@@ -700,3 +700,21 @@ class LazyCatalogEntry(Document):
...
+    """
+    A collection of units that have been handled by the streamer in the
+    passive lazy workflow that Pulp should download.
+
+    :ivar unit_id:      The associated content unit ID.
+    :type unit_id:      str
+    :ivar unit_type_id: The associated content unit type.
+    :type unit_type_id: str
+    """
+
```
jortel: This needs:
```
meta = {
    'collection': 'deferred_downloads',
    'indexes': {
        'fields': ['unit_id', 'unit_type_id'],
        'unique': True
    }
}
```

server/pulp/server/db/model/__init__.py
```diff
@@ -700,3 +700,21 @@ class LazyCatalogEntry(Document):
...
+    # For backward compatibility
+    _ns = StringField(default='deferred_download')
+
+    unit_id = StringField(required=True)
+    unit_type_id = StringField(required=True, unique_with='unit_id')
```
jortel: Don't need the *unique_with* constraint in addition to the unique index.

server/pulp/server/db/model/__init__.py
```diff
@@ -700,3 +700,21 @@ class LazyCatalogEntry(Document):
...
+    passive lazy workflow that Pulp should download.
+
+    :ivar unit_id:      The associated content unit ID.
+    :type unit_id:      str
+    :ivar unit_type_id: The associated content unit type.
+    :type unit_type_id: str
+    """
+
+    # For backward compatibility
+    _ns = StringField(default='deferred_download')
```
jortel: May not need the *_ns*.  Check with @bmbouter.  I think he's in the process of getting rid of the need for this in all concrete classes.

server/test/unit/server/db/test_model.py
```diff
@@ -590,3 +590,30 @@ def test_meta_collection(self):
...
+        self.assertEqual(['unit_id'], model.DeferredDownload.unit_type_id.unique_with)
+
+        self.assertTrue(isinstance(model.DeferredDownload._ns, StringField))
+        self.assertEqual('deferred_download', model.DeferredDownload._ns.default)
+
+    def test_meta_collection(self):
+        """
+        Assert that the collection name is correct.
+        """
+        self.assertEquals(model.DeferredDownload._meta['collection'], 'deferred_download')
```
jortel: Where is _meta defined?

server/pulp/server/db/model/__init__.py
bmbouter: All concrete classes still need the `_ns` defined. It's confusing because we've talked about getting rid of it and gone back and for. For now we're going to keep it because otherwise we would have to also write a migration that affects all collections and it would be a change to the API. We have a task to remove all of these with Pulp 3.0.

server/test/unit/server/db/test_model.py
jeremycline: It's created as part of the ``Document`` class, but in retrospect I don't really like this test since it's poking about in a private variable. I wrote it because I didn't explicitly define the collection name (and mongoengine does it for you based on the class name), but I regret that choice. Explicit is better than implicit, as they say.

jortel: Suggest testing *meta* here instead of *_meta*.

jeremycline: I did some poking around because all the other tests were also using _meta, and the superclass actually deletes ``meta`` and only has ``_meta``.

server/pulp/server/db/model/__init__.py
```diff
@@ -700,3 +700,30 @@ class LazyCatalogEntry(Document):
     checksum = StringField()
     checksum_algorithm = StringField(regex=ALG_REGEX)
     data = DictField()
+
+
+class DeferredDownload(Document):
```
bmbouter: All documents must inherit from AutoRetryDocument and never from `Document` directly

bmbouter: Also every new model introduced that lives in platform must have ensure_indexes called in [`ensure_database_indexes()`](https://github.com/bmbouter/pulp/blob/e022f247f841800b0be18f7fcd0bdd5bb2f4ab3a/server/pulp/server/db/manage.py#L120)

server/test/unit/server/db/test_model.py
jortel: Really?  That seems like odd behavior for a superclass.

bmbouter: It would be good to make less assertions on `_meta`. One way this cold be done using public methods would be to make assertions on the response output from `list_indexes`. Consider this snippet from your model definition.

```
DeferredDownload.list_indexes()
[[('unit_id', 1), ('unit_type_id', 1)], [(u'_id', 1)]]
```

bmbouter: @jortel I verified that `meta` is removed and `_meta` is added. Here's my snippet.

```
from mongoengine import Document, StringField

class DeferredDownload(Document):
    meta = {
        'collection': 'deferred_download',
        'indexes': [
            {
                'fields': ['unit_id', 'unit_type_id'],
                'unique': True
            }
        ]
    }

    unit_id = StringField(required=True)
    unit_type_id = StringField(required=True)

    # For backward compatibility
    _ns = StringField(default='deferred_download')


print DeferredDownload.meta
```

When I run that snippet it outputs:

```
  File "/tmp/foo.py", line 21, in <module>
    print DeferredDownload.meta
AttributeError: type object 'DeferredDownload' has no attribute 'meta'
```

server/pulp/server/db/model/__init__.py
jeremycline: Hey, thanks for the review. I'll make a new PR tomorrow morning!

PR #2145, opened 2015-10-29
Udates ContentUnit and adjusts platform code to match

bmbouter:
* renames source_location to _source_location
    * renames unit_type_id to _content_type_id
    * adds validation to AutoRetryDocument for the _ns field definition
    * adds on-startup validation of subclasses of ContentUnit
    * adds tests for new code and fixes other tests

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -1,4 +1,5 @@
 import copy
+from gettext import gettext as _
 import logging
 import os
 import random
@@ -7,7 +8,7 @@ from collections import namedtuple
 from hmac import HMAC
 
 from mongoengine import (DateTimeField, DictField, Document, DynamicField, IntField,
-                         ListField, StringField, UUIDField)
+                         ListField, StringField, UUIDField, ValidationError)
 from mongoengine import signals
 
 from pulp.common import constants, dateutils, error_codes
@@ -38,6 +39,8 @@ class AutoRetryDocument(Document):
     """
     Base class for mongoengine documents, includes auto retry functionality,
     if unsafe_autoretry is set to true in the server config.
+
+    All classes inheriting from this class must define a _ns field.
     """
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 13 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -49,6 +52,25 @@ def __init__(self, *args, **kwargs):
 
     meta = {'abstract': True}
 
+    def clean(self):
```
asmacdo: Is this function called automatically by mongoengine at document creation time? I am not sure that I like the idea raising a `ValidationError` if it is the document that is broken rather than an instance.

server/pulp/server/db/model/__init__.py
```diff
@@ -224,8 +244,7 @@ class Worker(AutoRetryDocument):
     """
     Represents a worker.
 
-    This inherits from mongoengine.Document and defines the schema for the documents
-    in the worker collection.
+    Defines the schema for the documents in the worker collection.
```
asmacdo: Maybe put "workers" into quotes to indicate that this is the name of the collection, not just "the collection that contains workers".

jortel:
LGTM but removed label until others have finished review.

server/pulp/server/db/model/__init__.py
bmbouter: `clean()` is called by mongoengine as [part of the save process](http://docs.mongoengine.org/en/latest/guide/document-instances.html?highlight=clean#pre-save-data-validation-and-cleaning).

The point about this raising an exception when an instance is saved when the issue is actually on the document definition is well stated and correct. I still decided to do it this way for 3 reasons:

1. We lack a mechanism to validate that "all pulp mongoengine based models" validate in a specific way. Even putting it here is a poor attempt to address that problem because we can't be guaranteed that all pulp mongoengine models will even inherit from `AutoRetryDocument`.

2. As developers and testers, we will see the errors before actual users do so it will have the intended effect.

3. We get to delete this once we handle [issue 625](https://pulp.plan.io/issues/625).

This is a quick-and-dirty way for us to get some confidence that subclasses defined the right fields. Currently there is only 1 global field that should be defined, `_ns`.

asmacdo: :+1:

PR #2147, opened 2015-11-03
Add downloaded field to file content unit.

jortel:
https://pulp.plan.io/issues/1302

Mainly added the *downloaded* field to content unit.  Added *downloaded* to ContentUnit.  Replaced *set_content()* with *import_content()* and change behavior to *import* the content when it's called instead of setting _source_path and copying when the unit is saved.  Support for importing directories has been removed.  The *path* passed to import_content() must be an existing file.  Importers (rpm, puppet, docker) will be updated to call import_content() for each file downloaded instead of with the entire directory after it has all been downloaded.  This should not be an issue as both the puppet and rpm plugins are in the process of conversion to mongoengine on master.

I decided to leave *_storage_path* on ContentUnit.  After looking at orphan clean up, I became concerned that there were other parts of the system that expected *_storage_path* to exist on all content units and did not want to break things.  Or, take this on as part of the Lazy effort.

I also added controller functions to query for downloaded status of repositories.

### DIFF
--- a/server/pulp/server/content/storage.py
+++ b/server/pulp/server/content/storage.py
@@ -28,7 +28,7 @@ class ContentStorage(object):
     Base class for content storage.
     """
 
-    def put(self, unit, path):
+    def put(self, unit, path, location=None):
         """
         Put the content (bits) associated with the specified content unit into storage.
         The file (or directory) at the specified *path* is transferred into storage.
@@ -37,6 +37,9 @@ class ContentStorage(object):
         :type unit: pulp.sever.db.model.ContentUnit
         :param path: The absolute path to the file (or directory) to be stored.
         :type path: str
+        :param location: The (optional) location within the path
+            where the content is to be stored.
+        :type location: str
         """
         raise NotImplementedError()
 
@@ -74,30 +77,47 @@ class ContentStorage(object):
 
 class FileStorage(ContentStorage):
     """
-    Direct storage for files and directories.
+    Direct file storage.
     """
 
-    def put(self, unit, path):
+    @staticmethod
+    def get_path(unit):
         """
-        Put the content defined by the content unit into storage.
-        The file (or directory) at the specified *path* is transferred into storage.
+        Get the appropriate storage path for the specified unit.
 
-        :param unit: The content unit to be stored.
-        :type unit: pulp.sever.db.model.ContentUnit
-        :param path: The absolute path to the file (or directory) to be stored.
-        :type path: str
+        :param unit: A content unit.
+        :type unit: pulp.server.db.model.FileContentUnit
+        :return: An absolute path to where associated content files will be stored.
+        :rtype: str
         """
         storage_dir = os.path.join(
             config.get('server', 'storage_dir'),
             'content',
             'units')
-        destination = os.path.join(storage_dir, unit._content_type_id, unit.id[0:4], unit.id)
+        return os.path.join(
+            storage_dir,
+            unit.type_id,
+            unit.id[0:4],
+            unit.id)
+
+    def put(self, unit, path, location=None):
+        """
+        Put the content defined by the content unit into storage.
+        The file at the specified *path* is transferred into storage.
+
+        :param unit: The content unit to be stored.
+        :type unit: pulp.sever.db.model.ContentUnit
+        :param path: The absolute path to the file (or directory) to be stored.
+        :type path: str
+        :param location: The (optional) location within the path
+            where the content is to be stored.
+        :type location: str
+        """
+        destination = unit.storage_path
+        if location:
+            destination = os.path.join(destination, location.lstrip('/'))
         mkdir(os.path.dirname(destination))
-        if os.path.isdir(path):
-            shutil.copytree(path, destination)
-        else:
-            shutil.copy(path, destination)
-        unit._storage_path = destination
+        shutil.copy(path, destination)
 
     def get(self, unit):
         """
@@ -135,7 +155,7 @@ class SharedStorage(ContentStorage):
         self.storage_id = sha256(storage_id).hexdigest()
         self.provider = provider
 
-    def put(self, unit, path=None):
+    def put(self, unit, path=None, location=None):
         """
         Put the content (bits) associated with the specified content unit into storage.
         The file (or directory) at the specified *path* is transferred into storage.
@@ -144,6 +164,9 @@ class SharedStorage(ContentStorage):
         :type unit: pulp.sever.db.model.ContentUnit
         :param path: The absolute path to the file (or directory) to be stored.
         :type path: str
+        :param location: The (optional) location within the path
+            where the content is to be stored.
+        :type location: str
         """
         self.link(unit)
 
@@ -212,6 +235,8 @@ class SharedStorage(ContentStorage):
 
         :param unit: The content unit to be linked.
         :type unit: pulp.sever.db.model.ContentUnit
+        :return: The absolute path to the link.
+        :rtype: str
         """
         target = self.content_dir
         link = os.path.join(self.links_dir, unit.id)
@@ -222,4 +247,4 @@ class SharedStorage(ContentStorage):
                 pass  # identical
             else:
                 raise
-        unit._storage_path = link
+        return link
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -15,7 +15,7 @@ from pulp.plugins.config import PluginCallConfiguration
 from pulp.plugins.loader import api as plugin_api
 from pulp.plugins.loader import exceptions as plugin_exceptions
 from pulp.plugins.model import SyncReport
-from pulp.plugins.util import misc
+from pulp.plugins.util.misc import paginate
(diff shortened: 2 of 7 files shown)


server/pulp/server/content/storage.py
```diff
@@ -74,30 +74,45 @@ def __exit__(self, *unused):
...
+    def put(self, unit, path, location=None):
         """
         Put the content defined by the content unit into storage.
-        The file (or directory) at the specified *path* is transferred into storage.
+        The file at the specified *path* is transferred into storage.
 
         :param unit: The content unit to be stored.
         :type unit: pulp.sever.db.model.ContentUnit
         :param path: The absolute path to the file (or directory) to be stored.
         :type path: str
```
jeremycline: ``location`` is missing docs.

server/pulp/server/content/storage.py
```diff
@@ -74,30 +74,45 @@ def __exit__(self, *unused):
...
+        if location:
+            location.lstrip('/')
+            destination = os.path.join(destination, location)
         mkdir(os.path.dirname(destination))
-        if os.path.isdir(path):
-            shutil.copytree(path, destination)
-        else:
-            shutil.copy(path, destination)
-        unit._storage_path = destination
+        shutil.copy2(path, destination)
```
jeremycline: I believe we want to use ``shutil.copy`` here, as ``copy2`` uses ``copystat``[0] which will result in an incorrect SELinux context[1].

[0] https://docs.python.org/2/library/shutil.html#shutil.copy2
[1] https://pulp.plan.io/issues/1312

server/pulp/server/content/storage.py
```diff
@@ -74,30 +74,45 @@ def __exit__(self, *unused):
...
+        :return: An absolute path to where associated content files will be stored.
+        :rtype: str
+        """
+        storage_dir = os.path.join(
+            config.get('server', 'storage_dir'),
+            'content',
+            'units')
+        return os.path.join(
+            storage_dir,
+            unit.type_id,
```
jeremycline: Should this be ``unit._content_type_id``?

jeremycline: Never mind, I noticed there's a property.

server/pulp/server/controllers/repository.py
```diff
@@ -105,6 +105,46 @@ def find_repo_content_units(
...
+    Find content units that have not been fully downloaded.
+
+    :param repository: A repository.
+    :type repository: model.Repository
+    :return: The requested units.
+    :rtype: generator
+    """
+    types = {}
+    for association in model.RepositoryContentUnit.objects(repo_id=repository.repo_id):
+        unit_ids = types.setdefault(association.unit_type_id, [])
```
jeremycline: I did not know about ``setdefault``. Interesting!

server/test/unit/server/content/test_storage.py
```diff
@@ -62,45 +62,53 @@ def test_exit(self):
...
         config.get = lambda s, p: {'server': {'storage_dir': storage_dir}}[s][p]
+        unit = Mock(id='0123456789', type_id='ABC')
+
+        # test
+        path = FileStorage.get_path(unit)
+
+        # validation
+        self.assertEqual(
+            path,
+            os.path.join(os.path.join(storage_dir, 'content', 'units'),
```
jeremycline: Out of curiosity, what's the reason for the nested call to ``join``? 
Is ``os.path.join(storage_dir, 'content', 'units', unit.type_id, unit.id[0:4], unit.id)`` not equivalent?

server/test/unit/server/content/test_storage.py
```diff
@@ -62,45 +62,53 @@ def test_exit(self):
...
-        storage.put(unit, path_in)
+        storage.put(unit, path_in, location)
 
         # validation
-        destination = os.path.join(
-            os.path.join(storage_dir, 'content', 'units', unit._content_type_id),
-            unit.id[0:4], unit.id)
-        shutil.copy.assert_called_once_with(path_in, destination)
-        self.assertEqual(unit._storage_path, destination)
+        location.lstrip('/')
```
jeremycline: I believe this is a no-op, since ``lstrip`` returns a copy of the string and does not modify the string pointed to by ``location``:
```
In [1]: x = 'hi'

In [2]: x.lstrip('h')
Out[2]: 'i'

In [3]: x
Out[3]: 'hi'
```

server/pulp/server/controllers/repository.py
```diff
@@ -105,6 +105,46 @@ def find_repo_content_units(
...
+    """
+    Find content units that have not been fully downloaded.
+
+    :param repository: A repository.
+    :type repository: model.Repository
+    :return: The requested units.
+    :rtype: generator
+    """
+    types = {}
+    for association in model.RepositoryContentUnit.objects(repo_id=repository.repo_id):
```
mhrivnak: I think it's reasonable to use ``.only('unit_id')`` here to avoid loading extra data about every association.

server/pulp/server/controllers/repository.py
```diff
@@ -105,6 +105,46 @@ def find_repo_content_units(
...
+        unit_ids = types.setdefault(association.unit_type_id, [])
+        unit_ids.append(association.unit_id)
+    for type_id, unit_ids in types.items():
+        _model = plugin_api.get_unit_model_by_id(type_id)
+        for unit in _model.objects(id__in=unit_ids, downloaded=False):
```
mhrivnak: check out ``pulp.plugins.util.misc.paginate``, which would avoid creating a gigantic query here that might exceed size limits.

server/pulp/server/controllers/repository.py
```diff
@@ -105,6 +105,46 @@ def find_repo_content_units(
...
+
+
+def has_all_units_downloaded(repository):
+    """
+    Get whether a repository contains units that have all been downloaded.
+
+    :param repository: A repository.
+    :type repository: model.Repository
+    :return: True if completely downloaded
+    :rtype: generator
```
mhrivnak: bool

server/pulp/server/controllers/repository.py
```diff
@@ -105,6 +105,46 @@ def find_repo_content_units(
...
+    """
```
mhrivnak: same suggestions as the above function, to use ``only`` and ``paginate``.

server/test/unit/server/content/test_storage.py
mhrivnak: Good catch! This is true for all string manipulation in python.

mhrivnak:
I'm just adding some feedback, but will let @jeremycline be the primary reviewer and LGTMer.

jeremycline:
Thanks for the help, @mhrivnak!

server/test/unit/server/controllers/test_repository.py
```diff
@@ -145,6 +148,131 @@ def test_skip(self, mock_get_model, mock_demo_objects, mock_rcu_objects):
...
+        self.assertEqual(
+            sorted([u.id for u in found]),
+            sorted([u.id for u in units if not u.downloaded]))
+
+
+class TestHasAllUnitsDownloaded(unittest.TestCase):
+
+    @patch(MODULE + '.plugin_api.get_unit_model_by_id')
+    @patch(MODULE + '.model.RepositoryContentUnit')
+    def test_all_downlaoded(self, association, get_model):
```
jeremycline: s/downlaoded/downloaded/

server/test/unit/server/controllers/test_repository.py
```diff
@@ -145,6 +148,131 @@ def test_skip(self, mock_get_model, mock_demo_objects, mock_rcu_objects):
...
+            id__in=[a.unit_id for a in associations[0:3]],
+            downloaded=False)
+        cat.objects.assert_called_once_with(
+            id__in=[a.unit_id for a in associations[3:6]],
+            downloaded=False)
+        self.assertTrue(result)
+
+    @patch(MODULE + '.plugin_api.get_unit_model_by_id')
+    @patch(MODULE + '.model.RepositoryContentUnit')
+    def test_not_all_downlaoded(self, association, get_model):
```
jeremycline: s/downlaoded/downloaded/

jeremycline:
Just a few things need tweaking and then it'll LGTM.

server/test/unit/server/content/test_storage.py
jortel: Oh, no reason.  just a copy-paste mistake.

server/pulp/server/controllers/repository.py
jortel: Good idea.

server/pulp/server/content/storage.py
jortel: @bmbouter can you comment on this?  IIRC, you advised me to use copy2() unless I misunderstood.

bmbouter: @jortel use `shutil.copy` instead of `copy2`. I didn't remember which we wanted, except that we needed to be sure we used the right one. I double checked the docs just now, and what @jeremycline indicated is correct.

PR #2148, opened 2015-11-04
Provide means for including the file name in storage path

jortel:
There are publishers and other components that expect the unit.storage_path to include the file name for single file units.  On master, `set_content()` updates the storage path when single files are copied to include the file name.  To support `lazy`, we need to provide a way for the plugin to be more involved in setting the storage path to include the file name.  This is hacky and I don't really care for it much but it get's the job done until we can fix unit/file modeling.  Unfortunately, we'll need to live with this unitl 3.0.

Importers will need to do the following for single file units:
```
unit = MyUnit()
unit.set_storage_path('dog.rpm')
unit.save()
if not lazy:
    unit.import_content(path_to_downloaded_file)
```
Publishers can work like they do currently.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -457,7 +457,7 @@ class ContentUnit(AutoRetryDocument):
     :type _storage_path: mongoengine.StringField
     """
 
-    id = StringField(primary_key=True)
+    id = StringField(primary_key=True, default=lambda: str(uuid.uuid4()))
     pulp_user_metadata = DictField()
     _last_updated = IntField(required=True)
     _storage_path = StringField()
@@ -554,8 +554,6 @@ class ContentUnit(AutoRetryDocument):
         :param document: Document that sent the signal
         :type document: ContentUnit
         """
-        if not document.id:
-            document.id = str(uuid.uuid4())
         document._last_updated = dateutils.now_utc_timestamp()
 
     def get_repositories(self):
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -667,6 +665,20 @@ def pre_save_signal(cls, sender, document, **kwargs):
...
+        """
+        Update the storage path to include the specified file name.
+        This should only be called when the content unit is associated with
+        a single file.
+
+        :param name: A file name.
+        :type name: str
+        """
+        if self.storage_path:
+            return
```
rbarlow: Should this raise an Exception instead of returning? The method description makes it sound to me like doing this a second time would be an error.

server/pulp/server/db/model/__init__.py
```diff
@@ -667,6 +665,20 @@ def pre_save_signal(cls, sender, document, **kwargs):
...
+        path = FileStorage.get_path(self)
+        self._storage_path = os.path.join(path, name)
```
rbarlow: Is self.storage_path a property that returns self._storage_path? If so, can we make that property also be the setter so this doesn't need to access _storage_path directly? It might be a little cleaner and less confusing.

server/pulp/server/db/model/__init__.py
```diff
@@ -667,6 +665,20 @@ def pre_save_signal(cls, sender, document, **kwargs):
         if not document._storage_path:
             document._storage_path = FileStorage.get_path(document)
 
+    def set_file(self, name):
```
rbarlow: I feel like the names "set_file" and "import_content" are a little confusing together. They both sound like they might do the same thing (based only on the names). Do you think we could come up with more descriptive names so it's more obvious which one developers should use in which cases? I don't have great suggestions so this is a dumb complaint. Here's some terrible brainstorming:

set_complete_unit_file()
import_complete_content() # Maybe we could allow this to be used for any number of files but they must all be present?
set_file_on_content_unit_that_only_has_one_file_and_can_never_have_any_more()

OK I'm done. It's no big deal if you stay with set_file(), but if you can think of a way to better distinguish these names I think that would be helpful to developers.

server/test/unit/server/db/test_model.py
```diff
@@ -336,6 +330,23 @@ def test_pre_save_signal_already_has_storage_path(self, get_path):
...
+        unit.set_file(name)
+        get_path.assert_called_once_with(unit)
+        self.assertEqual(unit._storage_path, os.path.join(path, name))
+
+    def test_set_file_ignored(self):
+        path = '/tmp/9876'
+        unit = TestFileContentUnit.TestUnit()
+        unit._storage_path = path
+        unit.set_file('1234')
+        self.assertEqual(unit._storage_path, path)
```
rbarlow: IMO, it would be good to assert that this situation raises an Exception, perhaps a RuntimeError?

server/pulp/server/db/model/__init__.py
jortel: Agreed.
Further, it should be raise *only* if the exception if the unit has been saved.  It should be fine to call multiple times before it is saved.

server/test/unit/server/db/test_model.py
jortel: Yes :)

PR #2150, opened 2015-11-05
Add lazy option to pulp admin.

jortel:
https://pulp.plan.io/issues/1201

I'm thinking we should revisit the branding on this feature.  The term *lazy* is fine for an informal reference to the feature (or internal code) but thinking we should use something better for CLI and API as it pertains to repositories. 

Perhaps something like:

CLI:
```
--download-policy=(immediate | background | on_demand)
```

API (importer configuration):
```
download_policy=(immediate | background | on_demand)
```

Thoughts?

### DIFF
--- a/client_lib/pulp/client/commands/repo/importer_config.py
+++ b/client_lib/pulp/client/commands/repo/importer_config.py
@@ -2,6 +2,7 @@ from gettext import gettext as _
 
 from pulp.client import arg_utils, parsers
 from pulp.client.extensions.extensions import PulpCliOption, PulpCliOptionGroup
+from pulp.client.validators import download_policy_validator
 from pulp.common.plugins import importer_constants as constants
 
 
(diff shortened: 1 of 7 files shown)


seandst:
I think 'demand' or something like that is definitely better than 'lazy' for the external interfaces (API, pulp-admin, etc). I have two *trivial* thoughts:

1. I think I'd prefer 'ondemand'; I think it reads a little better: "immediate download policy, background download policy, ondemand download policy"
1. Confusion might still result ("Lazy? Are you talking about demand downloading?", or vice versa) by having more than one name for something, so if we do put a "public" name on it, we should be careful to limit ourselves to having no more than two names for one thing.

Like I said...trivial. :)

jeremycline:
I much prefer ``--download-policy=(immediate | background | on_demand)`` to ``--lazy=(off|active|passive)`` or similar. I think they describe the feature in a much more user-friendly way.

Basically @seandst++

client_lib/pulp/client/commands/repo/importer_config.py
```diff
@@ -96,6 +97,12 @@ def __init__(self):
             parse_func=parsers.pulp_parse_optional_nonnegative_int
         )
 
+        d = _('content downloading policy ({m})'.format(
+            m=' | '.join((constants.DOWNLOAD_IMMEDIATE,
+                          constants.DOWNLOAD_BACKGROUND,
+                          constants.DOWNLOAD_ON_DEMAND))))
+        self.opt_download_policy = PulpCliOption('--download-policy', d, required=False)
```
jeremycline: Since there is a known set of valid values, it'd be nice to have a validator on the option.

jeremycline:
Other than the missing validator I have no complaints.

PR #2156, opened 2015-11-06
Removes master.rst from release notes.

dkliban:


bmbouter:
@dkliban you also need to remove the 1 line from index.rst

https://raw.githubusercontent.com/pulp/pulp/master/docs/user-guide/release-notes/index.rst

bmbouter:
@dkliban LGTM, you'll also merge this PR branch directly into 2.7-dev -> master

when you merge from 2.7-dev to master if you are *sure* this is the only commit that will go with it you can use a git merge strategy that will cause the change to not actually modify master if that is what you want.

PR #2161, opened 2015-11-12
Translate _id field specially to allow for ObjectIds

asmacdo:
Using a criteria search by string representations of an ObjectId was broken in master for collections that have been converted to mongoengine. There is not a redmine issue for this because though this code exists in 2.7, no collections are affected, thus there has never been a release with this issue.

Thanks to @Ichimonji10 for smashing up my search code :)

### DIFF
--- a/server/pulp/server/webservices/views/search.py
+++ b/server/pulp/server/webservices/views/search.py
@@ -140,9 +140,6 @@ class SearchView(generic.View):
         :raises exceptions.InvalidValue: if pymongo is unable to use the criteria object
         """
         query = criteria.Criteria.from_client_input(query)
-        if query.fields:
-            if 'id' not in query.fields:
-                query.fields.append('id')
 
         # Our MongoEngine SearchViews will have cls.model set to the MongoEngine model, while the
         # "old" style objects will have the cls.manager attribute set to the model's manager. While
@@ -154,6 +151,12 @@ class SearchView(generic.View):
         if hasattr(cls, 'model'):
             search_method = cls.model.objects.find_by_criteria
         else:
+            # id field should always be in query.fields. We are forced to do this twice, here and
+            # in the queryset for the mongoengine work. This is necessary because in mongoengine
+            # id is an alias to _id, and queries must be translated to reflect that. If we append
+            # id before this split, our serializers will translate this id also.
+            if query.fields and 'id' not in query.fields:
+                query.fields.append('id')
             search_method = cls.manager.find_by_criteria
 
         # We do not validate all aspects of the criteria object, so if pymongo has a problem we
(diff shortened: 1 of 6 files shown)


asmacdo:
This will admittedly not work for all possible queries on `_id`. We cannot support all arbitrary criteria searches because we have to translate fields to accommodate internal db changes. This fixes the functionality of some possible use cases and will at least raise an error if we do are unable to translate the search term.

server/pulp/server/webservices/views/search.py
```diff
@@ -154,6 +151,12 @@ def _generate_response(cls, query, options, *args, **kwargs):
         if hasattr(cls, 'model'):
             search_method = cls.model.objects.find_by_criteria
         else:
+            # id field should always be in query.fields. We are forced to do this twice, here and
+            # in the queryset for the mongoengine work. This is necessary because in mongoengine
+            # id is an alias to _id, and queries must be translated to reflect that. If we append
+            # id before this split, our serializers will translate this id also.
+            if query.fields and 'id' not in query.fields:
```
mhrivnak: Just for entertainment value, another very-slightly-less-wordy way to do this is:

```
if 'id' not in query.fields or []:
```

Definitely no need to change, but I thought the alternative is worth at least knowing about.

server/pulp/server/webservices/views/serializers/__init__.py
```diff
@@ -327,3 +364,8 @@ def get_href(self, instance):
         :rtype:  str
         """
         return reverse('user_resource', kwargs={'login': instance.login})
+
+    def _translate_filters(self, model, filters):
```
mhrivnak: doc block

server/pulp/server/webservices/views/search.py
asmacdo: `in` comes before `or` in the Python the order of precedence, so that won't work.

``` python
if 'a' in None or ['a']:
    print "or first"

>>> TypeError: argument of type 'NoneType' is not iterable


```

asmacdo:
ok test

server/pulp/server/webservices/views/search.py
seandst: `'x' in (is_none or [])` is as easy to read or easier, for me, than `if is_none and x in is_none:`, with the added benefit of being more elegant code.

asmacdo: @seandst, since this a `not in` scenario and 'id' is not in `[]`, it will attempt to run an append on `NoneType`. I can get the correct behavior hard coding `['id]` like I did above, but that is really just shorthand for:

``` python
if query.fields:
    if 'id' not in query.fields:
        query.fields.append('id')
else:
    if 'id' not in ['id']:
        query.fields.append('id')
```
Obviously, the last line will never run, but it still seems like a weird way to do it to me.

PR #2164, opened 2015-11-12
Update deps readme for createrepo

jeremycline:


deps/README
```diff
@@ -11,8 +11,16 @@ createrepo
 
 Team owner: jcline
 
-Note: createrepo-0.9.9 is available in EL7.
+Note: createrepo-0.9.9 is available in EL6, EL7. createrepo-0.10.3 is available in F21+
```
rbarlow: s/,/ and/

rbarlow:
Thanks @jeremycline!

deps/README
jeremycline: That list dreamed of being so much longer. Alas!

PR #2168, opened 2015-11-16
add scripts to switch branches

asmacdo:
~~These need to be used with caution. Note that the 2.6 script also drops the db. I can remove the db drop and migrations part of it if people think that is a safer option.~~

I have removed the database stuff. Users will still need to migrate (and drop if rolling backwards), but this should not be a part of the script.

### DIFF
--- a/playpen/all_26.sh
+++ b/playpen/all_26.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/bash -e
+
+#  When using this script, you will need to fix your database accordingly. If moving backward, you
+#  will need to drop the database, and you will always need to run the migrations.
+
+# You need the .bashrc that the Vagrant or dev-setup.sh environment creates. You can get it from
+# playpen/ansible/roles/dev/files/bashrc if you aren't using one of those envs.
+. ~/.bashrc
+
+# Switch back to pulp-2.6
+pstop
+
+pushd ~/devel
+for r in {pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_python,pulp_rpm,pulp}; do
+    if [ -d $r ]; then
+        pushd $r
+        sudo ./pulp-dev.py --uninstall
+        popd
+    fi
+done
+
+for repo in pulp pulp_puppet pulp_rpm
+do
+    pushd /home/vagrant/devel/$repo
+    git checkout 2.6-dev
+    find ~/devel -name "*.py[c0]" -delete
+    sudo ./pulp-dev.py --install
+    popd
+done
+
+popd # ~/devel
+
+sudo dnf install -y python-webpy
+
+workon pulp
+pip install paste
--- a/playpen/all_27.sh
+++ b/playpen/all_27.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+#  When using this script, you will need to fix your database accordingly. If moving backward, you
+#  will need to drop the database, and you will always need to run the migrations.
+
+
+# pulp_docker, pulp_python, pulp_ostree excluded because they are versioned differently
+for repo in pulp pulp_puppet pulp_rpm
+do
+    pushd /home/vagrant/devel/$repo
+    git checkout 2.7-dev
+    find ~/devel -name "*.py[c0]" -delete
+    sudo ./pulp-dev.py --install
+    popd
+done
+
+for repo in pulp_docker pulp_python
+do
+    pushd /home/vagrant/devel/$repo
+    sudo ./pulp-dev.py --uninstall
+    popd
+done
--- a/playpen/all_master.sh
+++ b/playpen/all_master.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+#  When using this script, you will need to fix your database accordingly. If moving backward, you
+#  will need to drop the database, and you will always need to run the migrations.
+
+
+for repo in pulp_docker pulp_puppet pulp_python pulp_rpm pulp
+do
+    pushd /home/vagrant/devel/$repo
+    sudo ./pulp-dev.py --uninstall
+    git checkout master
+    find ~/devel -name "*.py[c0]" -delete
+    popd
+done
+
+for repo in pulp pulp_docker pulp_puppet pulp_python pulp_rpm
+do
+    pushd /home/vagrant/devel/$repo
+    sudo ./pulp-dev.py --install
+    popd
+done


asmacdo:
ok test

playpen/all_27.sh
```diff
@@ -0,0 +1,17 @@
+for repo in pulp pulp_puppet pulp_rpm
```
Ichimonji10: How about adding a shebang to the top of this script?

What if I also have `pulp_python`, `pulp_docker`, etc cloned? Even if nothing needs to change, it'd probably be good to add a comment and explain that.

playpen/all_27.sh
```diff
@@ -0,0 +1,17 @@
...
+done
+
+for repo in pulp_docker pulp_python
+do
+    pushd /home/vagrant/devel/$repo
+    sudo ./pulp-dev.py -U
+    popd
+done
+
+
```
Ichimonji10: Nitpick: drop whitespace.

playpen/all_27.sh
```diff
@@ -0,0 +1,17 @@
+for repo in pulp pulp_puppet pulp_rpm
+do
+    pushd /home/vagrant/devel/$repo
+    git checkout 2.7-dev
+    find ~/devel -name "*.pyc" -delete
+    sudo ./pulp-dev.py -I
```
Ichimonji10: How about using switches like `--install` instead of `-I`? These scripts will be read by many people in many different contexts, and I imagine some extra verbosity will help avoid confusion. Terseness is great when you're typing out things at a shell by hand, and that isn't happening here.

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# You need the .bashrc that the Vagrant or dev-setup.sh environment creates. You can get it from
+# playpen/ansible/roles/dev/files/bashrc if you aren't using one of those envs.
+. ~/.bashrc
+
+# Switch back to pulp-2.6
+pstop
+
+pushd ~/devel
+for r in {pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_python,pulp_rpm,pulp}; do
```
Ichimonji10: Brace expansion is a feature only available to certain shells, like bash and csh. How about either dropping it:

```sh
for r in pulp_deb pulp_docker pulp_openstack pulp_ostree ... ; do
```

Or taking full advantage of it:

```sh
for r in pulp pulp_{deb,docker,openstack,ostree,puppet,python,rpm}; do
```

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+    if [ -d $r ]; then
+        pushd $r
+        sudo ./pulp-dev.py -U
+        popd
+    fi
+done
+
+pushd pulp
+git checkout 2.6-testing
```
Ichimonji10: [vagrant@dev pulp]$ git checkout 2.6-testing
    error: pathspec '2.6-testing' did not match any file(s) known to git.

Do you mean `2.6-dev`?

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+find ~/devel -name "*.pyc" -delete
+sudo ./pulp-dev.py -I
+popd # pulp
+
+# This one is necessary to work around a Pulp httpd config bug, where some plugin MUST install an
+# httpd config.
+pushd pulp_rpm
+git checkout 2.6-release
```
Ichimonji10: Elsewhere, `-dev` branches are checked out. Is there any particular reason `-release` is checked out here?

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+find ~/devel -name "*.pyc" -delete
+sudo ./pulp-dev.py -I
+popd # pulp_rpm
+
+popd # ~/devel
+
+sudo dnf install -y python-webpy
+
+workon pulp
+pip install paste
```
Ichimonji10: Is there any particular explanation for installing python-webpy and paste?

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+    if [ -d $r ]; then
+        pushd $r
+        sudo ./pulp-dev.py -U
+        popd
+    fi
+done
+
+pushd pulp
+git checkout 2.6-testing
+find ~/devel -name "*.pyc" -delete
```
rbarlow: You might consider dropping pyos as well.

playpen/all_26.sh
```diff
@@ -0,0 +1,39 @@
...
+sudo ./pulp-dev.py -I
+popd # pulp
+
+# This one is necessary to work around a Pulp httpd config bug, where some plugin MUST install an
+# httpd config.
+pushd pulp_rpm
+git checkout 2.6-release
+find ~/devel -name "*.pyc" -delete
```
rbarlow: Might drop pyos here too.

playpen/all_26.sh
rbarlow: webpy is required in pulp <= 2.6.z. I'm not sure about paste though.

playpen/all_27.sh
rbarlow: It's a fine suggestion, but that would be separate of this script since pulp-dev has been around a long time.

Ichimonji10: `pulp-dev` already supports `--install`.

playpen/all_26.sh
asmacdo: Didn't really look into it, but here is where `paste` is used (only on 2.6)

https://github.com/pulp/pulp/blob/2.6-dev/nodes/test/unit/base.py#L23

playpen/all_26.sh
asmacdo: I am only using bash for this so I am going to leave it as is. Feel free to add support for other shells in a separate PR.

Ichimonji10: If bash is expected, how about the second form? More concise.

playpen/all_26.sh
```diff
@@ -0,0 +1,41 @@
...
+popd # pulp
+
+# This one is necessary to work around a Pulp httpd config bug, where some plugin MUST install an
+# httpd config.
+pushd pulp_rpm
+git checkout 2.6-dev
+find ~/devel -name "*.py[c0]" -delete
+sudo ./pulp-dev.py --install
+popd # pulp_rpm
+
```
Ichimonji10: Should this script also go checkout the 2.6-dev branches in each of the other repositories, such as `pulp_puppet`? When I run this script as is, it leaves at least that repository alone and on the master branch, which of couse causes breakage when I try to use the rolled-back system.

PR #2169, opened 2015-11-16
The lazy catalog is now purged when an importer is deleted.

jeremycline:
This commit adds a ``pre_delete`` signal to the Importer model which
handles the removal of catalog entries.

closes #1327

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -233,6 +233,23 @@ class Importer(AutoRetryDocument):
             'indexes': [{'fields': ['-repo_id', '-importer_type_id'], 'unique': True}],
             'queryset_class': CriteriaQuerySet}
 
+    @classmethod
+    def pre_delete(cls, sender, document, **kwargs):
+        """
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -233,6 +233,23 @@ class Importer(AutoRetryDocument):
...
+        Purge the lazy catalog of all entries for the importer being deleted.
+
+        :param sender:   class of sender (unused)
+        :type  sender:   object
+        :param document: mongoengine document being deleted.
+        :type  document: pulp.server.db.model.Importer
+        """
+        query_set = LazyCatalogEntry.objects(importer_id=str(document.id))
+        _logger.debug(_('Deleting {n} lazy catalog entries for the {repo} repository.').format(
+            n=query_set.count(), repo=document.repo_id))
```
mhrivnak: Calling ``count()`` will do a database query every time this function runs, even if the log level is not at debug. Even though it's fairly inexpensive, I don't think we should add db calls that run all the time but only add value when in debug mode. I suggest just leaving out the specific number.

PR #2170, opened 2015-11-16
Removing migrations from prior to 2.4.0.

mhrivnak:
Changing the rules to allow gaps in migration numbers, and to fast-forward
brand new migration trackers instead of applying migrations to empty collections.

re #1305 
https://pulp.plan.io/issues/1305

This does not completely satisfy the requirements for the task, but is the first step, and includes the majority of the work. Once this is merged, I'll go into the plugins and remove migrations.


mhrivnak:
I thought about an important use case that isn't covered, so I'm going to close this for now.

PR #2171, opened 2015-11-17
Trim down `pstatus` function in .bashrc

Ichimonji10:
The various systemd commands accept multiple arguments. Why use a for loop if
a simpler expression suffices?


seandst:
Manually tested this with the latest vagrantfile from master, and it (somewhat predictably) works perfectly.

PR #2175, opened 2015-11-17
Add a preset command to the dev environment.

rbarlow:
This command stops Pulp, erases the Pulp database, /var/lib/pulp, reruns the migrations, and then starts Pulp.

### DIFF
--- a/playpen/ansible/roles/dev/tasks/main.yml
+++ b/playpen/ansible/roles/dev/tasks/main.yml
@@ -55,3 +55,5 @@
 
 - name: Install custom ~/.bashrc
   copy: src=bashrc dest=/home/{{ ansible_env.SUDO_USER }}/.bashrc
+
+- copy: src=drop_database.js dest=/home/{{ ansible_env.SUDO_USER }}/drop_database.js
--- a/playpen/ansible/roles/dev/files/bashrc
+++ b/playpen/ansible/roles/dev/files/bashrc
@@ -33,6 +33,25 @@ pstatus() {
     _paction status
 }
 
+preset() {
+    pstop;
+    mongo pulp_database ~/drop_database.js;
+    sudo rm -rf /var/lib/pulp/*;
+    sudo -u apache pulp-manage-db;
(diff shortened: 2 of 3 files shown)


playpen/ansible/roles/dev/tasks/main.yml
```diff
@@ -55,3 +55,5 @@
 
 - name: Install custom ~/.bashrc
   copy: src=bashrc dest=/home/{{ ansible_env.SUDO_USER }}/.bashrc
+
```
seandst: This empty line is weird. :)

jeremycline:
@rbarlow It'd be cool if this also made some testing repos (like how the zoo one is made initially).

rbarlow:
@jeremycline That's a great suggestion. I also realized that it needs to rebuild the Crane symlinks if Crane is present. So this PR needs a little more love.

seandst:
If you don't mind, I'd love to see this get merged, and get new fanciness in other PRs. :)

playpen/ansible/roles/dev/tasks/main.yml
rbarlow: The empty line is there to separate the second copy command from the first. The "name" only applies to the first. I've been contemplating removing all uses of "name" in this and my personal playbooks because I don't think it's very valuable.

rbarlow:
@jeremycline Per @seandst's request I went ahead and merged this PR without the zoo repo. Would you make a PR to add that to the function? Make sure to test that the rpm plugins are installed first so that it only runs on systems it can work with.

playpen/ansible/roles/dev/tasks/main.yml
seandst: Probably it's just me, but the yaml already does that, and this just seems redundant.

meh :)

seandst: (submitted before I was done, sorry)

...but it lines up with the rest of the formatting, so it's for the best.

PR #2177, opened 2015-11-18
retry package install to allow provisioning with poor bandwidth

asmacdo:
If the connection dropped, the whole install would hang indefinitely. We can now provision from coffee shops!!

### DIFF
--- a/playpen/ansible/roles/core/tasks/main.yml
+++ b/playpen/ansible/roles/core/tasks/main.yml
@@ -4,6 +4,7 @@
 
 - name: Install packages
   dnf: name={{ item }} state=present
+  retries: "{{ install_retries }}"
   with_items:
       - bash-completion
       - vim-enhanced
--- a/playpen/ansible/roles/db/tasks/main.yml
+++ b/playpen/ansible/roles/db/tasks/main.yml
@@ -1,6 +1,7 @@
 ---
 - name: Install packages
   dnf: name={{ item }} state=present
+  retries: "{{ install_retries }}"
   with_items:
       - mongodb
       - mongodb-server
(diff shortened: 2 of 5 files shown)


rbarlow:
There are several places that packages get installed, so doing this in this one place doesn't cover all the possible problems. Furthermore, you can always just re-run vagrant provision and it will try again. Ansible is idempotent, so it will only do what is needed each time (i.e., it goes faster the second time). Thus, I don't think we should merge this as is and I don't think it's needed to add this line to all the places that packages get installed due to the vagrant provision workaround.

seandst:
/me agrees

I don't know if the shell stuff is idempotent, but the use-case here is the initial provision, which doesn't need to be strictly idempotent; it just need to (eventually) work.

`vagrant provision` can be run multiple times to get the desired result if package installs fail, and since ansible definitely is idempotent, you might want to `vagrant provision --provision-with ansible`if all you're really trying to do is re-run the playbook.

asmacdo:
I have removed the hard coded line to a variable to solve the problem more generally, but it does require adding a line to each role that installs pack

I would prefer not to rely on impotency because when the connection is dropped the install doesn't fail, it hangs. Provisioning takes long enough (particularly when the bandwidth is low) that it is very undesirable to have to monitor it.

seandst:
I'm confused by that. If the install hangs, what's the difference between ansible retrying and the human? If it's truly hung, it seems like both would fail.

asmacdo:
@seandst You make a good point. I don't know what is happening under the hood here (a quick search didn't reveal much, I don't want to spend more time on this right now). I can say that when I disconnect the internet, it hangs in this branch with retries and in master, so I guess adding `retries` appeared to work due to random chance. Closing for now, I will reopen if I figure it out later.

seandst:
I'll keep an eye out as well, and let you know if I have any ideas.

PR #2179, opened 2015-11-18
Removing migrations from prior to 2.4.0.

mhrivnak:
Changing the rules to allow gaps in migration numbers, and to fast-forward
brand new migration trackers instead of applying migrations to empty collections.

re #1305 
https://pulp.plan.io/issues/1305

This does not completely satisfy the requirements for the task, but is the first step, and includes the majority of the work. Once this is merged, I'll go into the plugins and remove migrations.


server/test/unit/server/db/test_manage.py
```diff
@@ -199,7 +227,7 @@ def test_current_version_too_high(self, mocked_file_config, initialize, getLogge
         # Make sure we start out with a clean slate
         self.assertEquals(MigrationTracker.objects().count(), 0)
         # Make sure that our mock works. There are four valid packages.
```
asmacdo: 5 now. maybe this comment should just be removed.

server/test/unit/server/db/test_manage.py
```diff
@@ -238,7 +266,7 @@ def test_migrate(self, mock_file_config, initialize, mocked_apply_migration, get
         # Make sure we start out with a clean slate
         self.assertEquals(MigrationTracker.objects().count(), 0)
         # Make sure that our mock works. There are three valid packages.
```
asmacdo: same

server/test/unit/server/db/test_manage.py
```diff
@@ -741,16 +754,11 @@ def test_migration_version_cant_be_zero(self):
...
         """
-        error_message = 'Migration version 2 is missing in ' +\
-            'unit.server.db.migration_packages.version_gap.'
-        try:
-            models.MigrationPackage(migration_packages.version_gap)
-            self.fail('The MigrationPackage.MissingVersion exception should have been raised, '
-                      'but was not raised.')
-        except models.MigrationPackage.MissingVersion, e:
-            self.assertEquals(str(e), error_message)
+        models.MigrationPackage(migration_packages.version_gap)
```
asmacdo: Would it be preferable to mock these migration_packages?

docs/dev-guide/newtypesupport/plugin/migrations.rst
```diff
@@ -10,7 +10,9 @@ Registration
 ============
 
 In order to write migrations for your Pulp plugin, you will need to register your plugin's
-migrations package with the Pulp server.
+migrations package with the Pulp server. It is advantageous to register even a brand-new plugin
```
asmacdo: It might be nice to put a similar note somewhere that a new plugin writer is guaranteed to see it. The intended audience of this particular note might not have a reason to look here until they need to write a migration for a plugin that is not brand-new.

server/test/unit/server/db/test_manage.py
mhrivnak: Probably, but since that would be a fundamental change to the way this entire test module works, I think we're better off at this point leaving it the way it is.

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

PR #2181, opened 2015-11-18
The lazy catalog is purged during an orphan purge.

jeremycline:
This change appears in both the mongoegine version and the
non-mongoengine version of the "delete orphan by type" method because
although lazy is only supported by mongoengine types, the
``delete_orphan_by_id`` method works with mongoengine types and calls
the non-mongoegine version of "delete orphan by type".

It also updates the mongoengine version of "delete by type" to match the
signature of the non-mongoengine version.

closes #1339

### DIFF
--- a/server/pulp/server/managers/content/orphan.py
+++ b/server/pulp/server/managers/content/orphan.py
@@ -1,4 +1,5 @@
 from gettext import gettext as _
+import itertools
 import logging
 import os
 import re
@@ -220,6 +221,10 @@ class OrphanManager(object):
             if content_unit_ids is not None and content_unit['_id'] not in content_unit_ids:
                 continue
 
+            model.LazyCatalogEntry.objects(
+                unit_id=content_unit['_id'],
+                unit_type_id=content_type_id
+            ).delete()
             content_units_collection.remove(content_unit['_id'], safe=False)
 
             storage_path = content_unit.get('_storage_path', None)
@@ -227,7 +232,7 @@ class OrphanManager(object):
                 OrphanManager.delete_orphaned_file(storage_path)
 
     @staticmethod
-    def delete_orphan_content_units_by_type(type_id):
+    def delete_orphan_content_units_by_type(type_id, content_unit_ids=None):
         """
         Delete the orphaned content units for the given content type.
         This method only applies to new style content units that are loaded via entry points
@@ -236,10 +241,19 @@ class OrphanManager(object):
 
         :param type_id: id of the content type
         :type type_id: basestring
+        :param content_unit_ids: list of content unit ids to delete; None means delete them all
+        :type content_unit_ids: iterable or None
         """
         # get the model matching the type
         content_model = plugin_api.get_unit_model_by_id(type_id)
-        content_units = content_model.objects().only('id', '_storage_path')
+        if content_unit_ids:
+            query_sets = []
+            for page in plugin_misc.paginate(content_unit_ids):
+                qs = content_model.objects(id__in=page).no_cache().only('id', '_storage_path')
+                query_sets.append(qs)
+            content_units = itertools.chain(*query_sets)
+        else:
+            content_units = content_model.objects.no_cache().only('id', '_storage_path')
 
         # Paginate the content units
         for units_group in plugin_misc.paginate(content_units):
@@ -252,12 +266,16 @@ class OrphanManager(object):
 
             # Clear the units that are currently associated from unit_dict
             non_orphan = model.RepositoryContentUnit.objects(unit_id__in=id_list)\
-                .distinct('unit_id')
+                .no_cache().distinct('unit_id')
             for non_orphan_id in non_orphan:
                 unit_dict.pop(non_orphan_id)
 
-            # Remove the unit and any references on disk
+            # Remove the unit, lazy catalog entries, and any content in storage.
             for unit_to_delete in unit_dict.itervalues():
+                model.LazyCatalogEntry.objects(
+                    unit_id=str(unit_to_delete.id),
+                    unit_type_id=str(type_id)
+                ).delete()
                 unit_to_delete.delete()
                 if unit_to_delete._storage_path:
                     OrphanManager.delete_orphaned_file(unit_to_delete._storage_path)
--- a/server/test/unit/server/managers/content/test_orphan.py
+++ b/server/test/unit/server/managers/content/test_orphan.py
@@ -7,17 +7,18 @@ import string
 import tempfile
 import traceback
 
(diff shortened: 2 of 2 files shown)


server/pulp/server/managers/content/orphan.py
```diff
@@ -256,13 +259,42 @@ def delete_orphan_content_units_by_type(type_id):
             for non_orphan_id in non_orphan:
                 unit_dict.pop(non_orphan_id)
 
-            # Remove the unit and any references on disk
+            # If there's a list of unit ids to delete, filter the orphan list
+            if content_unit_ids is not None:
+                for orphaned_unit_id in unit_dict.keys():
+                    if orphaned_unit_id not in content_unit_ids:
```
mhrivnak: Unless you only expect to have a handful at a time, this will perform better if ``content_unit_ids`` is a set.

server/pulp/server/managers/content/orphan.py
```diff
@@ -256,13 +259,42 @@ def delete_orphan_content_units_by_type(type_id):
             for non_orphan_id in non_orphan:
                 unit_dict.pop(non_orphan_id)
 
-            # Remove the unit and any references on disk
+            # If there's a list of unit ids to delete, filter the orphan list
```
mhrivnak: Why not do the filtering on line 245 above? It looks like that could avoid loading a lot of data.

server/pulp/server/managers/content/orphan.py
```diff
@@ -256,13 +259,42 @@ def delete_orphan_content_units_by_type(type_id):
...
+        :type  unit_id:      str
+        :param unit_type_id: The content type of the content unit.
+        :type  unit_type_id: str
+        """
+        catalog_entries = model.LazyCatalogEntry.objects(
+            unit_id=str(unit_id),
+            unit_type_id=str(unit_type_id)
+        )
+        msg = _('Deleting {n} catalog entries for orphaned content unit {type}:{id}')
+        _logger.error(msg.format(
```
mhrivnak: This is an error?

server/pulp/server/managers/content/orphan.py
```diff
@@ -256,13 +259,42 @@ def delete_orphan_content_units_by_type(type_id):
...
+            n=len(catalog_entries),
```
mhrivnak: Calling ``len()`` causes the ``QuerySet`` to be evaluated, so results are loaded from the DB into python. This is not worth the expense just for a log message. I wouldn't even call ``.count()`` on it. I suggest leaving out the number, and then you only have to do one database action (the delete) instead of two.

server/pulp/server/managers/content/orphan.py
```diff
@@ -236,10 +237,16 @@ def delete_orphan_content_units_by_type(type_id):
...
         :param type_id: id of the content type
         :type type_id: basestring
+        :param content_unit_ids: list of content unit ids to delete; None means delete them all
+        :type content_unit_ids: iterable or None
         """
         # get the model matching the type
         content_model = plugin_api.get_unit_model_by_id(type_id)
-        content_units = content_model.objects().only('id', '_storage_path')
+        if content_unit_ids:
+            query_set = content_model.objects(id__in=content_unit_ids)
```
mhrivnak: Will ``content_unit_ids`` be large enough possibly that we need to paginate this query? We don't want to hit the max query size.

jeremycline: Although it is *very* unlikely that anyone would do that, it is technically possible.

mhrivnak: Well then, you know what to do. ;)

I suggest a combination of the ``paginate`` function and ``itertools.chain()``, so you can use it the same as you are now, but the requests are broken up behind the scenes.

jeremycline: I have made it so, I think. Better double check my work as I'm quite new to MongoEngine and itertools (which is very neat).

server/pulp/server/managers/content/orphan.py
```diff
@@ -236,10 +238,18 @@ def delete_orphan_content_units_by_type(type_id):
...
+        :param content_unit_ids: list of content unit ids to delete; None means delete them all
+        :type content_unit_ids: iterable or None
         """
         # get the model matching the type
         content_model = plugin_api.get_unit_model_by_id(type_id)
-        content_units = content_model.objects().only('id', '_storage_path')
+        if content_unit_ids:
+            query_sets = []
+            for page in plugin_misc.paginate(content_unit_ids):
+                query_sets.append(content_model.objects(id__in=page).only('id', '_storage_path'))
```
mhrivnak: I learned recently about mongoengine's caching behavior, and it's non-intuitive. Would you expect that each QuerySet instance keeps a copy of every result it yields? Me either. You'll need to call ``.no_cache()`` on each QuerySet instance to prevent that.

server/pulp/server/managers/content/orphan.py
```diff
@@ -236,10 +238,18 @@ def delete_orphan_content_units_by_type(type_id):
...
+            content_units = itertools.chain(*query_sets)
+        else:
+            content_units = content_model.objects().only('id', '_storage_path')
```
mhrivnak: same here

server/pulp/server/managers/content/orphan.py
```diff
@@ -256,13 +266,32 @@ def delete_orphan_content_units_by_type(type_id):
             for non_orphan_id in non_orphan:
                 unit_dict.pop(non_orphan_id)
 
-            # Remove the unit and any references on disk
+            # Remove the unit, lazy catalog entries, and any content in storage.
             for unit_to_delete in unit_dict.itervalues():
+                OrphanManager.delete_catalog_entry(unit_to_delete.id, type_id)
```
mhrivnak: What if you replace this line with:

```
model.LazyCatalogEntry.objects(unit_id=unit_id, unit_type_id=unit_type_id).delete()
```

and get rid of the function below? I don't think the extra function makes things easier.

server/pulp/server/managers/content/orphan.py
```diff
@@ -220,6 +221,7 @@ def delete_orphans_by_type(content_type_id, content_unit_ids=None):
             if content_unit_ids is not None and content_unit['_id'] not in content_unit_ids:
                 continue
 
+            OrphanManager.delete_catalog_entry(content_unit['_id'], content_type_id)
```
mhrivnak: same suggestion here as below, to just do the delete using mongoengine's API instead of using a custom but equivalent API.

PR #2183, opened 2015-11-20
Ensure all values in ScheduledCall are serialized

asmacdo:
https://pulp.plan.io/issues/1381

`ScheduledCall` model uses `as_dict` to serialize in preparation for a pymongo insert which cannot handle Mongoengine objects.

The `SystemUser` is such a basic object that the simplest path was just to have it inherit from dict (via the old base model) instead of writing a serializer for it. Treating it as a dict has always worked before, this is the most obvious way to ensure that the behavior doesn't change.

### DIFF
--- a/server/pulp/server/db/model/dispatch.py
+++ b/server/pulp/server/db/model/dispatch.py
@@ -201,7 +201,15 @@ class ScheduledCall(Model):
         :return:    dictionary of public keys and values
         :rtype:     dict
         """
-        return {
+
+        # If principal is a User object, serialize it. If it does not have a serializer, it is
+        # a SystemUser which is already a dict.
+        try:
+            serial_principal = self.principal.serializer(self.principal).data
+        except AttributeError:
+            serial_principal = self.principal
+
+        dict_repr = {
             '_id': str(self._id),
             'args': self.args,
             'consecutive_failures': self.consecutive_failures,
@@ -213,13 +221,15 @@ class ScheduledCall(Model):
             'last_run_at': self.last_run_at,
             'last_updated': self.last_updated,
             'next_run': self.calculate_next_run(),
-            'principal': self.principal,
+            'principal': serial_principal,
             'remaining_runs': self.remaining_runs,
             'resource': self.resource,
             'schedule': self.schedule,
             'task': self.task,
             'total_run_count': self.total_run_count,
         }
+
+        return dict_repr
 
     def for_display(self):
         """
--- a/server/test/unit/server/db/model/test_dispatch.py
+++ b/server/test/unit/server/db/model/test_dispatch.py
@@ -35,12 +35,7 @@ SCHEDULE = {
     u'kwargs': {u'overrides': {}},
     u'last_run_at': u'2013-12-17T00:35:53Z',
     u'last_updated': 1387218569.811224,
-    u'principal': u"(dp0\nV_id\np1\nccopy_reg\n_reconstructor\np2\n(cbson.objectid\nObjectId\np3\n"
-                  u"c__builtin__\nobject\np4\nNtp5\nRp6\nS'R \\xab\\x06\\xe1\\x9a\\x00\\x10\\xe1i"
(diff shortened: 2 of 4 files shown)


rbarlow:
Task calls should not try to serialize any objects going into their parameters. Instead, the primary keys of the obects that are needed by the task should be passed, and the task should query the database for the native Mongoengine object.

asmacdo:
ok test

rbarlow:
Looks like you've got a test failure to fix before review on RHEL 7.

dkliban:
ok test

dkliban:
ok test

asmacdo:
@rbarlow, that was a jenkies problem. This one is ready for review!

rbarlow:
@asmacdo is the purpose of this to serialize the arguments to tasks?

asmacdo:
@rbarlow the `ScheduledCall` object has other objects that are a part of it. The `as_dict` function is essentially a serializer that is used prior to saving this object to the db. The internal objects must be serialized before they can be a part of an object inserted into the db.
https://github.com/asmacdo/pulp/blob/fix-task-user-serialization/server/pulp/server/db/model/dispatch.py#L252

I would prefer that these foreign objects were not saved as part of the `ScheduledCall`, but that is a bigger change.

server/pulp/server/db/model/dispatch.py
```diff
@@ -221,6 +222,14 @@ def as_dict(self):
             'total_run_count': self.total_run_count,
         }
 
+        for key, value in dict_repr.iteritems():
+            try:
+                dict_repr[key] = value.serializer(value).data
+            except AttributeError:
+                continue
```
seandst: This is effectively `except AttributeError: pass`. I think I understand why it's find to just `continue` here, but I'd appreciate a comment about why missing keys are probably just fine.

server/pulp/server/db/model/dispatch.py
```diff
@@ -221,6 +222,14 @@ def as_dict(self):
             'total_run_count': self.total_run_count,
         }
 
+        for key, value in dict_repr.iteritems():
```
seandst: You're not really saving anything with `iteritems` here; might as well stick with `items` since `iteritems` is deprecated in py3.

server/pulp/server/db/model/dispatch.py
asmacdo: Yeah, that could use a comment. I'll add something like this:

The idea is that any value that has a `.serializer` attribute has been converted to Mongoengine. Because of this, it does not inherit from `dict` like the old models and will need to use that serializer. If value is not a mongoengine object it will not have `.serializer` and will raise an attribute error. In this case, don't try to serialize it and check the next item.

server/test/unit/server/db/model/test_dispatch.py
```diff
@@ -614,6 +614,28 @@ def test_returns_dict(self):
...
+    def test_mongoengine_objects_serialized(self):
+        """
+        Ensure that mongoengine objects in the scheduled call were serialized.
+        """
+        schedule = bson.SON(SCHEDULE)
+        call = ScheduledCall.from_db(schedule)
+        call['principal'] = model.User('test_user')
+
+        result = call.as_dict()
+        self.assertTrue(isinstance(result['principal'], dict))
```
seandst: This appears to only be checking that `result['principal']` is a dict, and possibly relying on `as_dict()` to throw exceptions if serialization fails. I don't see that anything is doing what I'd expect, based on the test name and docstring, which is making sure SCHEDULE is serialized properly, with all the expected keys, values, etc.

seandst: Assuming `model.User` is the only mongoengine object, it's probably enough to just make sure `result['principal']` was serialized properly.

asmacdo:
@seandst, when I wrote this, I attempted to be overly clever to handle future cases. Looking back at this, none of the other parts of this object will be Mongoengine objects, so I have switched from general to specific. This now only attempts to serialize the User object.

PR #2185, opened 2015-11-20
Removed unused code, added docs, consolidated workflows, fixed inconsistencies

mhrivnak:


server/pulp/plugins/util/publish_step.py
```diff
@@ -577,9 +602,11 @@ def get_distributor_type(self):
 
     def publish(self):
         """
+        DEPRECATED
```
bmbouter: Do we want to just remove this method?

mhrivnak: I'd love to remove it, but I'm sure there are plugins that are calling it. I think it'll be easier to deprecate a bunch of methods, then remove them later (like in 3.0).

server/pulp/plugins/util/publish_step.py
```diff
@@ -248,7 +280,9 @@ def on_error(self):
 
     def _process_block(self, item=None):
         """
-        This block is called for the main processing loop
+        This is an implementation detail that should not be overridden unless you are sure of what
```
bmbouter: Consider s/ `an implementation detail`/`part of the step processing internals`.

server/pulp/plugins/util/publish_step.py
```diff
@@ -248,7 +280,9 @@ def on_error(self):
...
+        you are doing. If you want somewhere to generally perform work in your step, this is not
```
bmbouter: Should we refer people to `process_main()` here specifically?

server/pulp/plugins/util/publish_step.py
seandst: For this and other deprecated methods, is there a reason not to warnings.warn a DeprecationWarning?

bmbouter:
Jenkins is failing on F21 because of a nectar problem. The unit tests pass everywhere else so I think it's good regarding Jenkins

bmbouter:
What I would like to see is this get merged, and then migrate the plugins to use all non-deprecated objects and methods, and then later fully remove all deprecated objects and methods. LGTM

server/pulp/plugins/util/publish_step.py
mhrivnak: Only because it's historically been difficult to prevent those warnings from showing up in front of users and customers, who can't do anything about them, but definitely get concerned.

seandst: I figured it was something like that, and that makes perfect sense. Thanks!

```python
raise EverythingIsTotallyFineException
```

mhrivnak:
I'll proceed with what we discussed. I'll merge this now, try to fix up the plugins to not use deprecated stuff, and then see about removing the deprecated parts.

server/pulp/plugins/util/publish_step.py
rbarlow: Michael Hrivnak wrote:
> I'd love to remove it, but I'm sure there are plugins that are calling
> it. I think it'll be easier to deprecate a bunch of methods, then remove
> them later (like in 3.0).

We already have significantly broken backwards compatibility that I
don't think it's worth trying to preserve what little compatibility is
left. I think removing it is better since all the plugins have to be
touched as is anyway.

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

PR #2186, opened 2015-11-23
Add total/downloaded units to repository view.

jeremycline:
This adds the following fields to the repository object returned from
the API ``v2/repositories/<repo-id>/?details=true``:
``total_content_units`` and ``total_units_downloaded``.

closes #1354

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from gettext import gettext as _
+from itertools import chain
 import logging
 import sys
 
@@ -82,6 +83,47 @@ def get_unit_model_querysets(repo_id, model_class, repo_content_unit_q=None):
         yield model_class.objects(id__in=chunk)
 
 
+def get_repo_unit_models(repo_id):
+    """
+    Retrieve all the MongoEngine models for units in a given repository. If a unit
+    type is in the repository and does not have a MongoEngine model, that unit type
+    is excluded from the returned list.
+
+    :param repo_id: ID of the repo whose unit models should be retrieved.
+    :type  repo_id: str
+
+    :return: A list of sub-classes of ContentUnit that define a unit model.
+    :rtype:  list of pulp.server.db.model.ContentUnit
+    """
+    unit_types = model.RepositoryContentUnit.objects(
+        repo_id=repo_id).distinct('unit_type_id')
+    unit_models = [plugin_api.get_unit_model_by_id(type_id) for type_id in unit_types]
+    # Filter any non-MongoEngine content types.
+    return filter(None, unit_models)
+
+
+def get_mongoengine_unit_querysets(repo_id, repo_content_unit_q=None):
+    """
+    Retrieve an iterable of QuerySets for all the units in a repository that have
+    MongoEngine models. If a unit type is in the repository and does not have a
+    MongoEngine model, that unit type is excluded from the iterable.
+
+    :param repo_id:             The ID of the repo whose units should be queried
+    :type  repo_id:             str
+    :param repo_content_unit_q: Any additional filters that should be applied to the
+                                RepositoryContentUnit search
+    :type  repo_content_unit_q: mongoengine.Q
+
+    :return: A generator of query sets.
+    :rtype:  generator of mongoengine.queryset.QuerySet
+    """
+    unit_models = get_repo_unit_models(repo_id)
+    for unit_model in unit_models:
+        query_sets = get_unit_model_querysets(repo_id, unit_model, repo_content_unit_q)
+        for query_set in query_sets:
+            yield query_set
+
+
 def find_repo_content_units(
         repository, repo_content_unit_q=None,
         units_q=None, unit_fields=None, limit=None, skip=None,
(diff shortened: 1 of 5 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -105,14 +105,20 @@ def find_repo_content_units(
             yield_count += 1
 
 
-def find_units_not_downloaded(repository):
+def _get_paged_repo_unit_ids_with_model(repository):
```
mhrivnak: I have something similar going in a branch right now that might be more reusable. Let's compare notes tomorrow.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -182,6 +181,10 @@ def get(self, request, repo_id):
         if request.GET.get('distributors', 'false').lower() == 'true' or details:
             _merge_related_objects(
                 'distributors', manager_factory.repo_distributor_manager(), (repo,))
+        if details:
+            repo['total_repository_units'] = sum(repo['content_unit_counts'].itervalues())
+            total_missing = repo_controller.missing_unit_count(repo_obj)
+            repo['total_downloaded_units'] = repo['total_repository_units'] - total_missing
```
mhrivnak: Maybe this should have a different name, since some may be uploaded instead of downloaded. Perhaps ``locally_stored_units`` or something similar?

mhrivnak:
REST API docs need to be updated with the new attributes.

server/pulp/server/controllers/repository.py
```diff
@@ -186,20 +194,39 @@ def has_all_units_downloaded(repository):
...
-            q_set = _model.objects(id__in=page, downloaded=False)
-            if q_set.count():
-                return False
+    for qs in _get_mongoengine_repo_querysets(str(repository.repo_id)):
+        if qs(downloaded=False).count():
+            return False
     return True
 
 
+def _get_mongoengine_repo_querysets(repo_id, repo_content_unit_q=None):
```
mhrivnak: This seems useful as a public function, so I would ditch the leading underscore.

mhrivnak: This function does two things. It figures out what modeled types are in a repo, and then it uses those models to get query sets from another function. Both sound broadly useful, and not necessarily tied to each other, so I suggest breaking this up into two functions. Something like:

```
def get_model_classes(repo_id):
    unit_types = model.RepositoryContentUnit.objects(repo_id=repo_id).distinct('unit_type_id')
    unit_models = [plugin_api.get_unit_model_by_id(type_id) for type_id in unit_types]
    # Filter any non-MongoEngine content types.
    return [m for m in unit_models if m is not None]


def get_unit_model_query_sets_across_types(repo_id, repo_content_unit_q):
    """Maybe you can find a less-wordy name that matches the other names"""
    for unit_model in get_model_classes(repo_id):
        query_sets = get_unit_model_querysets(repo_id, unit_model, repo_content_unit_q)
        for query_set in query_sets:
            yield query_set
```

server/pulp/server/controllers/repository.py
```diff
@@ -165,16 +166,23 @@ def find_units_not_downloaded(repository):
...
-        _model = plugin_api.get_unit_model_by_id(type_id)
-        for page in paginate(unit_ids):
-            for unit in _model.objects(id__in=page, downloaded=False):
-                yield unit
+    query_sets = _get_mongoengine_repo_querysets(str(repository.repo_id))
+    query_sets = [q(downloaded=False) for q in query_sets]
+    return chain(*query_sets)
+
+
+def missing_unit_count(repository):
```
mhrivnak: This is up for debate, but my feeling is that if there is zero chance that this function is going to need the full repository object, then just let it take the ``repo_id`` as its argument. It would be a shame to make someone retrieve the full object from the DB just for this function, which would ignore everything but the id.

server/pulp/server/controllers/repository.py
```diff
@@ -186,20 +194,39 @@ def has_all_units_downloaded(repository):
...
-    for association in q_set.only('unit_id', 'unit_type_id'):
-        unit_ids = types.setdefault(association.unit_type_id, [])
-        unit_ids.append(association.unit_id)
-    for type_id, unit_ids in types.items():
-        _model = plugin_api.get_unit_model_by_id(type_id)
-        for page in paginate(unit_ids):
-            q_set = _model.objects(id__in=page, downloaded=False)
-            if q_set.count():
-                return False
+    for qs in _get_mongoengine_repo_querysets(str(repository.repo_id)):
```
mhrivnak: Just for fun (not necessarily better), another way to do this is:

```
return any(q.count() for q in _get_mongoengine_repo_querysets(str(repository.repo_id)))
```

server/pulp/server/controllers/repository.py
seandst: >     # Filter any non-MongoEngine content types.
>     return [m for m in unit_models if m is not None]

If you want to filter, I think you should `filter`. `filter(None, unit_models)` will return a list of truthies. That's probably(?) good enough for model instances, but if explicit comparison to None is needed `filter(lambda model: model is not None, unit_models)`.

server/pulp/server/controllers/repository.py
seandst: I think the logic is flipped, so it would need to be `return not any(...)`, which I think highlights how using `any` here is not necessarily better, at least in terms of readability.

I [think](https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/queryset/base.py#L170) that MongoEngine QuerySets will do what you want when evaluated for truthiness here, so assuming that qs is an instance of a MongoEngine QuerySet (which I am), a somewhat more readable option might be `return not any(_get_mongoengine_repo_querysets(str(repository.repo_id))`

server/pulp/server/controllers/repository.py
jeremycline: My reason for having it a private function is that it skips any non-mongoengine models, which at the time felt like odd behaviour to have in a public function. Having slept on it, I think it's reasonable given the function name, and it makes sense to break things out into the two functions you suggested :+1:.

Also ``filter`` sounds like a good way to filter.

server/pulp/server/controllers/repository.py
```diff
@@ -82,6 +83,47 @@ def get_unit_model_querysets(repo_id, model_class, repo_content_unit_q=None):
...
+    """
+    Retrieve all the MongoEngine models for units in a given repository. If a unit
+    type is in the repository and does not have a MongoEngine model, that unit type
+    is excluded from the returned list.
+
+    :param repo_id: ID of the repo whose unit models should be retrieved.
+    :type  repo_id: str
+
+    :return: A list of sub-classes of ContentUnit that define a unit model.
+    :rtype:  pulp.server.db.model.ContentUnit
```
mhrivnak: the base thing returned is a ``list``

server/pulp/server/controllers/repository.py
```diff
@@ -82,6 +83,47 @@ def get_unit_model_querysets(repo_id, model_class, repo_content_unit_q=None):
...
+    """
+    unit_types = model.RepositoryContentUnit.objects(
+        repo_id=repo_id).distinct('unit_type_id')
+    unit_models = [plugin_api.get_unit_model_by_id(type_id) for type_id in unit_types]
+    # Filter any non-MongoEngine content types.
+    return filter(None, unit_models)
+
+
+def get_mongoengine_repo_querysets(repo_id, repo_content_unit_q=None):
```
mhrivnak: This is picky, but note that the name here is effectively ``repo_controller.get_mongoengine_repo_querysets``, and nowhere does the word "unit" appear. Of course units are the only thing we want out of these querysets.

This function is in the repo controller, so you don't necessarily need "repo" in the function name. Perhaps just replace it with  the word "unit", to make ``repo_controller.get_mongoengine_unit_querysets``

server/test/unit/server/controllers/test_repository.py
```diff
@@ -223,138 +223,79 @@ def test_skip(self, mock_get_model, mock_demo_objects, mock_rcu_objects):
...
+
+class HasAllUnitsDownloadedTests(unittest.TestCase):
+
+    @patch(MODULE + '.get_mongoengine_repo_querysets')
+    def test_true(self, mock_repo_querysets):
+        repo = Mock(repo_id='mock_repo')
+        query_set = Mock()
+        query_set.return_value.count.return_value = 0
+        mock_repo_querysets.return_value = [query_set]
+        self.assertTrue(repo_controller.has_all_units_downloaded(repo))
```
mhrivnak: The call being tested takes a repo_id now, not a repo object. The same problem appears in other nearby tests.

server/pulp/server/controllers/repository.py
jeremycline: A fair nit-pick

PR #2187, opened 2015-11-23
Handle RuntimeError exceptions raised by finish

jeremycline:


streamer/pulp/streamer/server.py
```diff
@@ -225,7 +226,23 @@ def close(self):
         """
         Forward the call to close the 'file' to the request.finish method.
         """
-        reactor.callFromThread(self.request.finish)
+        reactor.callFromThread(self.finish_wrapper)
+
+    def finish_wrapper(self):
```
mhrivnak: You may be interested to know about this, even though we obviously can't use it: https://docs.python.org/3.4/library/contextlib.html#contextlib.suppress

streamer/pulp/streamer/server.py
```diff
@@ -225,7 +226,23 @@ def close(self):
...
+
+        If finish is called after the client disconnects, a RuntimeError is
+        raised and Twisted logs the stack trace. Clients disconnecting before
+        Twisted gets around to calling ``finish`` is not uncommon. Ultimately
+        there should be a handler that stops streaming if the client disconnects,
+        but that is tricky with the current implementation's use of threads.
+        """
+        try:
+            self.request.finish()
+        except RuntimeError, e:
```
mhrivnak: use ``as`` instead of the comma for python3 compatibility.

https://www.python.org/dev/peps/pep-3110/

PR #2188, opened 2015-11-23
adding UTCDateTimeField, which always returns datetimes with timezone UTC

mhrivnak:
re #1254
re #1267

https://www.redhat.com/archives/pulp-list/2015-November/msg00044.html


mhrivnak:
Looks like the F23 tests are failing for unrelated reasons.

seandst:
LGTM, because...

- Explicit is better than implicit. (datetime objects returned by mongo are always tz-aware UTC)
- Simple is better than complex. (we aren't messing with timezones, just one timezone)
- In the face of ambiguity, refuse the temptation to guess. (naive datetime objects are ambiguous, tz-aware datetime objects are not)
- There should be one-- and preferably only one --obvious way to do it. (use UTCDateTime)

:)

PR #2192, opened 2015-11-24
adds a plugin step that works with mognoengine ContentUnit models

mhrivnak:


bmbouter:
ok test

bmbouter:
I consider jenkins happy enough since the f23 builder is failing for an unknown reason.

server/pulp/server/controllers/repository.py
```diff
@@ -30,6 +30,55 @@
...
+    :param repo_content_unit_q: any additional filters that should be applied to the
+                                RepositoryContentUnit search
+    :type  repo_content_unit_q: mongoengine.Q
+    :return:    generator of unit IDs
+    :rtype:     generator
+    """
+    qs = model.RepositoryContentUnit.objects(q_obj=repo_content_unit_q,
+                                             repo_id=repo_id,
+                                             unit_type_id=unit_type)
+    return (assoc.unit_id for assoc in qs.only('unit_id'))
```
bmbouter: To make the return argument type more obvious, consider using the `yield` statement. This would become:

```
for assoc in qs.only('unit_id')
    yield assoc.unit_id
```

server/pulp/server/controllers/repository.py
```diff
@@ -30,6 +30,55 @@
...
+    Return a generator of unit IDs within the given repo that match the type and query
+
+    :param repo_id:     ID of the repo whose units should be queried
+    :type  repo_id:     str
+    :param unit_type:   ID of the unit type which should be retrieved
+    :type  unit_type:   str
+    :param repo_content_unit_q: any additional filters that should be applied to the
+                                RepositoryContentUnit search
+    :type  repo_content_unit_q: mongoengine.Q
+    :return:    generator of unit IDs
```
bmbouter: Consider adding a space between arguments and return values. It makes a nice visual separation between inputs and output for readability.

server/pulp/plugins/util/publish_step.py
```diff
@@ -562,6 +563,77 @@ def process_lifecycle(self):
         return self._build_final_report()
 
 
+class UnitModelPluginStep(PluginStep):
+    """
+    Useful for performing actions on units that are defined as mongoengine models.
```
bmbouter: I propose the internal caching behavior of this should be documented.

server/pulp/plugins/util/publish_step.py
```diff
@@ -562,6 +563,77 @@ def process_lifecycle(self):
...
+        :param conduit: The conduit for the repo
+        :type  conduit: a conduit from pulp.plugins.conduits
+        :param config: The configuration
+        :type  config: PluginCallConfiguration
+        :param working_dir: The temp directory this step should use for processing
+        :type  working_dir: str
+        :param plugin_type: The type of the plugin
+        :type  plugin_type: str
+        """
+        super(UnitModelPluginStep, self).__init__(step_type, repo, conduit, config, working_dir,
```
bmbouter: Not a change/concer of this PR, but what can be done to remove conduit from the call signature of `PluginStep`. The all-mongoengine objects shouldn't use the conduit. What do you think?

server/pulp/plugins/util/publish_step.py
```diff
@@ -562,6 +563,77 @@ def process_lifecycle(self):
...
+                 passed in to this step.
+        :rtype:  list
+        """
+        if self._unit_querysets is None:
+            self._unit_querysets = []
+            for model_class in self.model_classes:
+                queries = repo_controller.get_unit_model_querysets(self.get_repo().id,
+                                                                   model_class,
+                                                                   self._repo_content_unit_q)
+                self._unit_querysets.extend(queries)
```
bmbouter: Consider having `_unit_querysets` store the querysets themselves instead of the contents of the iterable querysets.

If I'm reading this right, `_unit_querysets` will contain not the querysets themselves (as if you used `append`) but instead will contain the items the iterable provides, which in this case would be subclasses of `ContentUnit`. If we're sticking with that consider renaming `_units_querysets` to `_units` since it's not containing querysets.

I think an even better suggestion would be to remove the caching altogether from this object and let mongoengine's default queryset caching provide it for you. If you store the querysets internally and work out of those, you'll get just that I believe.

mhrivnak: I agree that ``self._unit_querysets`` should be a collection of ``QuerySet`` objects, and I think that it what the code does.

``queries`` will itself be a collection of ``QuerySet`` objects, which is what ``get_unit_model_querysets()`` returns. I do indeed want to take that collection and add its contents to the existing list that is ``self._unit_querysets`` by using the ``extend`` method. Does that make sense?

I need to read about how mongoengine's own caching works, but for the moment I am skeptical. :) If each ``QuerySet`` instance has its own cache, then I'll definitely want to keep those instances as the code is currently doing. If there is a global cache, I assume it has eviction rules, and we may not want to depend on it to ensure that successive calls to ``get_unit_model_queryset`` are cheap.

Keep in mind that making repeated calls to ``get_unit_model_querysets`` would not only make new instances of these ``QuerySet`` objects, but would first query the ``RepoContentUnit`` collection again in order to get the list of unit_ids that are used to make the ContentUnit queries.

mhrivnak: It looks like ``QuerySet`` objects each have their own cache. https://github.com/MongoEngine/mongoengine/blob/v0.10.3/mongoengine/queryset/queryset.py#L23

So I think we should keep references to the same ``QuerySet`` objects for re-use.

server/pulp/plugins/util/publish_step.py
mhrivnak: Good call.

server/pulp/plugins/util/publish_step.py
mhrivnak: I think the conduit is only used for reporting at this point by the steps, so we need a different way to do that.

mhrivnak:
Thanks for the review! I documented the caching behavior, switched from a generator expression to a ``yield`` statement, added a ``.no_cache()`` call in the ``get_associated_unit_ids`` controller function, and added a blank line to that doc block as suggested.

server/pulp/plugins/util/publish_step.py
bmbouter: I see now that `extend` does evaluate the iterable, but the iterable itself is a generator of `Queryset` objects. Thus your caching of them works with the mongoengine queryset caching and not against it. I consider this matter resolved.

server/pulp/plugins/util/publish_step.py
bmbouter: We need to think about how to get this code off of the conduits completely, but that isn't a topic for this PR.

bmbouter:
LGTM, great job.

Out of curiosity, what motivated the `no_cache()` addition? It's fine to leave in, but I'm curious. Is it that you don't want the queryset to cache results after it yields them? I can't see a practical benefit to caching given that it won't be referenced again, so that would improve memory performance right? Even if it wasn't added, it won't be cached for long I think. Once the generator of `get_associated_unit_ids` raises a `StopIteration`, the queryset that would/would-not cache would be deallocated anyway?

mhrivnak:
@bmbouter that's exactly right. The results wouldn't be cached for long, but my reading of mongoengine's cache behavior is that it would cache the entire collection of results until the ``QuerySet`` object gets garbage collected, which would cause an unnecessary spike in memory consumption. I strongly suspect we'll need to add a ``.no_cache()`` call to many other places.

bmbouter:
@mhrivnak I agree that `no_cache()` should be used in more places. Especially where `Queryset` objects are long-lived (ie: generators!). I'll keep this in mind for future code production and review.

mhrivnak:
This conversation made me realize that we need to call ``.no_cache`` also on the QuerySet objects that contain the ``ContentUnit`` results. Otherwise, all ``ContentUnit`` results would be held in memory until the publish finishes, which would be **bad**. I updated the PR accordingly.

mhrivnak:
@bmbouter reaffirmed his LGTM via IRC, accounting for the latest change.

PR #2193, opened 2015-11-25
Document preset and psmash commands in the dev motd.

rbarlow:


rbarlow:
We can ignore the Jenkies failure - it's due to F23 not being available yet.

seandst:
I'm vaguely concerned about the length of the motd, and that it might not fit on someone screen if we keep packing stuff in there, but not so concerned (at this point) that I'll do more than mention it.

lgtm

PR #2195, opened 2015-11-25
Removes qpid 0.26 stack from external_deps.json

bmbouter:
https://pulp.plan.io/issues/1340
re #1340


seandst:
This looks a little alarming at first glance, but all the answers are in #1340 on redmine. :)

PR #2203, opened 2015-12-01
Add a ppopulate command that creates some basic repos.

jeremycline:


rbarlow:
Can you document this in the development motd file? It's here:

https://github.com/pulp/pulp/blob/master/playpen/vagrant-motd.txt

Which reminds me, I really should move it to the ansible folder…

playpen/ansible/roles/dev/files/bashrc
```diff
@@ -76,5 +77,25 @@ psmash() {
     deactivate;
 }
 
+ppopulate() {
+    pulp-admin login -u admin -p admin
+    if [ -d $HOME/devel/pulp_rpm ]; then
+        pulp-admin rpm repo create --repo-id=zoo \
+            --feed=https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/
+    fi
```
rbarlow: Can you remove the creation of zoo from the vagrant-setup.sh file and have it call ppopulate there instead? In general we don't want to add new stuff to that file but in this case I think it makes sense.

playpen/ansible/roles/dev/files/bashrc
```diff
@@ -76,5 +77,25 @@ psmash() {
...
+        pulp-admin puppet repo create --repo-id=forge \
+            --feed=http://forge.puppetlabs.com \
+            --queries=torssh
+    fi
+
+    if [ -d $HOME/devel/pulp_python ]; then
+        pulp-admin python repo create --repo-id=pypi \
+            --feed=https://pypi.python.org/ \
+            --package-names=pip
+    fi
```
rbarlow: Very nice!

rbarlow:
I keep saying ppopulate in my head. nice.

PR #2204, opened 2015-12-01
Remove any pyc files before running pulp-smash.

jeremycline:
This also restarts Pulp before running the tests.

### DIFF
--- a/playpen/ansible/roles/dev/files/bashrc
+++ b/playpen/ansible/roles/dev/files/bashrc
@@ -26,6 +26,7 @@ pstop() {
 }
 
 prestart() {
+    find ~/devel/pulp* -name *.py[co] -delete;
     _paction restart
 }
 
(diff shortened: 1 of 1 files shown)


playpen/ansible/roles/dev/files/bashrc
```diff
@@ -72,6 +72,8 @@ _paction() {
 
 psmash() {
     workon pulp-smash;
+    prestart;
+    find ~/devel/pulp-smash/ -name *.pyc -delete;
```
seandst: `-name *.py[co]` to also account for optimized compilation artifacts.

All these semicolons need to go, but that should be its own PR.

rbarlow: It might be good to swap the order of these two lines, and have it rm py[co]'s from ~/devel instead of just smash. Or perhaps even better, why not modify prestart itself to rm the py[co]'s from ~/devel/pulp*? That might be nice. It could be a separate PR if you prefer, or if you don't want to do it let me know and I'll make the PR.

rbarlow:
I'm not sure what is up with the RHEL 6 builder, but I am confident that it is not due to this change since the bashrc has nothing to do with unit tests. LGTM.

PR #2207, opened 2015-12-01
Move the motd file to be managed by Ansible.

rbarlow:


dkliban:
ok test

seandst:
Not sure what github is complaining about, this merges without conflicts for me, but meh

This definitely does conflict with my changes in #2215, but I'll deal with that once this gets merged.

PR #2212, opened 2015-12-02
Add the importer ObjectId to the RepoSyncConduit.

jeremycline:
Conduits are going away someday, but in the mean time the importer
ObjectId is needed during sync so the lazy catalog can be built and this
seems the best place to stash it.

### DIFF
--- a/server/pulp/plugins/conduits/repo_sync.py
+++ b/server/pulp/plugins/conduits/repo_sync.py
@@ -53,9 +53,13 @@ class RepoSyncConduit(RepoScratchPadMixin, ImporterScratchPadMixin, AddUnitMixin
     allowed to do whatever threading makes sense to optimize its sync process.
     Calls into this instance do not have to be coordinated for thread safety,
     the instance will take care of it itself.
+
+    :ivar importer_object_id: The ObjectId of the importer for the given repo_id
+                              and importer_id.
+    :type importer_object_id: bson.objectid.ObjectId
     """
 
-    def __init__(self, repo_id, importer_id):
+    def __init__(self, repo_id, importer_id, importer_object_id):
         RepoScratchPadMixin.__init__(self, repo_id, ImporterConduitException)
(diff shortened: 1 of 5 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -699,6 +699,7 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
     call_config = PluginCallConfiguration(imp_config, repo_importer.config, sync_config_override)
     transfer_repo.working_dir = common_utils.get_working_directory()
     conduit = RepoSyncConduit(repo_id, repo_importer.importer_type_id)
+    conduit.importer_object_id = repo_importer.id
```
jeremycline: This is unpleasant and hacky, but I'm not sure if we want to change the RepoSyncConduit __init__ signature. If that's okay, I'd prefer to pass it in and document it there.

jortel: Yes, It would be better to pass to __init__().

server/pulp/plugins/conduits/repo_sync.py
```diff
@@ -53,9 +53,13 @@ class RepoSyncConduit(RepoScratchPadMixin, ImporterScratchPadMixin, AddUnitMixin
...
     Calls into this instance do not have to be coordinated for thread safety,
     the instance will take care of it itself.
+
+    :ivar importer_object_id: The ObjectId of the importer for the given repo_id
+                              and importer_id.
+    :type importer_object_id: bson.objectid.ObjectId
     """
 
-    def __init__(self, repo_id, importer_id):
+    def __init__(self, repo_id, importer_id, importer_object_id=None):
```
jortel: I'm concerned about defaulting this to be `None`.  Since importers will depend on this to create catalog entries, I think it should be required.

jeremycline:
Related: https://github.com/pulp/pulp_rpm/pull/739

PR #2217, opened 2015-12-03
Update the streamer to use alternate content sources.

jeremycline:


server/pulp/server/content/sources/container.py
```diff
@@ -285,7 +287,7 @@ def _download(url, destination, source):
         """
         downloader = source.get_downloader()
         request = DownloadRequest(url, destination)
-        report = downloader.download_one(request)
+        report = downloader.download_one(request, events=True)
```
jeremycline: I looked through the relevant Nectar code and I think having this always ``True`` is okay, but I'd appreciate input from @ipanova since she added it.

PR #2221, opened 2015-12-04
Change all QuerySet objects to be QuerySetNoCache objects.

jeremycline:
In order to not cache query results by default, all models must
use QuerySet objects that are subclasses of QuerySetNoCache. This
modifies all our models that don't have custom QuerySet objects to
use the QuerySetNoCache by default (by inheriting from
AutoRetryDocument) and introduces a superclass for all custom QuerySet
classes to use which derives from QuerySetNoCache.

closes #1395

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -630,7 +630,7 @@ class UnitModelPluginStep(PluginStep):
                 queries = repo_controller.get_unit_model_querysets(self.get_repo().id,
                                                                    model_class,
                                                                    self._repo_content_unit_q)
-                self._unit_querysets.extend(q.no_cache() for q in queries)
+                self._unit_querysets.extend(queries)
         return self._unit_querysets
 
     def get_total(self):
--- a/server/pulp/server/db/querysets.py
+++ b/server/pulp/server/db/querysets.py
@@ -1,10 +1,39 @@
-from mongoengine.queryset import DoesNotExist, QuerySet
+from gettext import gettext as _
+
+from mongoengine.queryset import DoesNotExist, QuerySetNoCache
 from pymongo import ASCENDING
 
 from pulp.server import exceptions as pulp_exceptions
 
 
-class CriteriaQuerySet(QuerySet):
+class QuerySetPreventCache(QuerySetNoCache):
+    """
+    All custom QuerySet classes should inherit from this class rather than QuerySet
+    or QuerySetNoCache. There are two reasons for this. Firstly, all QuerySets should
+    not cache their query results by default, so QuerySetNoCache must be inherited.
+    Secondly, QuerySetNoCache's implementation of the `cache` instance method
+    explicitly returns an instance of QuerySet which effectively removes any
+    customized behavior for that QuerySet.
+    """
+
+    def cache(self):
+        """
+        Currently, there is no reason to have a caching version of any of our
+        custom QuerySet classes. If that ever changes, this method should be
+        implemented to return a caching version of the custom QuerySet. This
+        implementation exists in order to avoid bugs resulting from the
+        unexpected default behavior of `cache`.
+
+        As of mongoengine 0.10, the default implementation of ``cache`` cannot
+        be used as it explicitly creates a QuerySet object, thus removing any
+        custom behavior of the custom QuerySet.
+        """
+        msg = _('In order to support caching QuerySets, you must implement the'
+                ' cache method on ' + self.__class__.__name__)
+        raise NotImplementedError(msg)
+
+
+class CriteriaQuerySet(QuerySetPreventCache):
     """
     This class defines a custom QuerySet to support searching by Criteria object
     which is a Pulp custom query object.
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -47,7 +47,7 @@ def get_associated_unit_ids(repo_id, unit_type, repo_content_unit_q=None):
     qs = model.RepositoryContentUnit.objects(q_obj=repo_content_unit_q,
                                              repo_id=repo_id,
                                              unit_type_id=unit_type)
-    for assoc in qs.no_cache().only('unit_id'):
+    for assoc in qs.only('unit_id'):
         yield assoc.unit_id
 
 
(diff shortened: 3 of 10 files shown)


server/pulp/server/db/querysets.py
```diff
@@ -1,10 +1,44 @@
...
+        implementation exists in order to avoid bugs resulting from the
+        unexpected default behavior of `cache`.
+
+        As of mongoengine 0.10, the default implementation of ``cache`` cannot
+        be used as it explicitly creates a QuerySet object, thus removing any
+        custom behavior of the custom QuerySet.
+        """
+        raise NotImplementedError()
+
+    def __len__(self):
```
jeremycline: I was a little hesitant to include this since it deviates from the behavior documented in MongoEngine, but it seemed like a nice thing to have. If whoever reviews agrees with the inclusion of ``__len__`` here, perhaps the AutoRetryDocument should use this class. The downside is that we cannot call ``cache`` on non-custom QuerySets, though that may not be much of a downside.

seandst: I think this is probably the best solution...

The docstring  does a good job explaining why, but perhaps a shorter version might get passed to NotImplementError.

I also wonder if something like 'QuerySetPreventCache' might be better than 'CustomQuerySet', though.

Regarding __len__ calling count, I did dig into this a (very) little bit, and find this commit interesting:
https://github.com/MongoEngine/mongoengine/commit/9251ce312bc0545d3b86224e35a913029a86695e

I haven't yet found more information about what "unusual issues" that caused, but it might be worth at least an eyebrow raise *particularly* since the documentation update says outright that you should probably use `.count()` instead.

jeremycline: I've skimmed the two issues mentioned in the commit message you linked. If I'm reading them correctly, https://github.com/MongoEngine/mongoengine/issues/247 is about there being an extra call to the database when doing things like ``list(qs)`` since list apparently calls ``__len__`` (which called ``count``). Personally I feel that this is okay, since if you're regularly calling ``list`` on a generator...

It seems https://github.com/MongoEngine/mongoengine/issues/311 was about the removal of ``__len__``, so it came back, but made use of the caching behavior (so no extra database call).

jeremycline: One other thing to note: QuerySetNoCache (and its ancestors) do not provide a ``__len__`` method at all.

jeremycline: @seandst with respect to the name, would ``CustomQuerySetNoCache`` be acceptable? I'd prefer ``CustomNoCacheQuerySet``, but since the MongoEngine version is ``QuerySetNoCache`` it matches.

seandst: I'm honestly fine with any of those; the main thing is that the name be a little more descriptive than `CustomQuerySet`, but not so descriptive that we have to rename it to add a method without feeling bad about having to refactor it later.

That's where `QuerySetPreventCache` came from...not only does it not cache, it goes so far as preventing you from converting it to a caching QuerySet by exploding when you call `.cache()`. But like I said, all the names are probably fine, and far less important than the docstrings, which are great. Your call.

jeremycline: I find that argument compelling.

server/pulp/plugins/util/publish_step.py
```diff
@@ -630,7 +630,7 @@ def unit_querysets(self):
                 queries = repo_controller.get_unit_model_querysets(self.get_repo().id,
                                                                    model_class,
                                                                    self._repo_content_unit_q)
-                self._unit_querysets.extend(q.no_cache() for q in queries)
+                self._unit_querysets.extend(q for q in queries)
```
mhrivnak: I think this can now just be ``.extend(queries)``

server/pulp/server/db/querysets.py
mhrivnak: I'm in favor of sticking with one way to do it. ``.count()`` is the generally accepted and supported way to get the count from the db, so let's not confuse the matter. I don't think ``len()`` is sufficiently more convenient, readable, or easier to use to justify a divergence of behavior from the rest of the library.

"There should be one-- and preferably only one --obvious way to do it."

rbarlow: On Tue, Dec 08, 2015 at 01:00:58PM -0800, Michael Hrivnak wrote:
> I'm in favor of sticking with one way to do it. ``.count()`` is the generally accepted and supported way to get the count from the db, so let's not confuse the matter. I don't think ``len()`` is sufficiently more convenient, readable, or easier to use to justify a divergence of behavior from the rest of the library.
> 
> "There should be one-- and preferably only one --obvious way to do it."

I don't believe that Python "mantra" is about this problem. The
mongoengine implementation of __len__ is problematic, and even has some
controversy surrounding it upstream. This implementation was not about
making it so that Pulp devs *will* use it, it was more about making sure
that we don't balloon our RAM in the event that someone does use it.

Upstream is arguably the place that shouldn't have provided more than
one way to do it, so for us as a downstream user that ship has sailed.

I'm a +1 for overriding the upstream __len__(). I don't believe there is
a reasonable chance that we would ever want their behavior for this
function and this implementation makes sure that we keep our RAM usage
down in the event that someone accidentally does this.

An alternative could be to override it and have it raise a
NotImplemented("Please use the count() method instead."). I think that
would make both of us happy (you because then we're enforcing that
there's only one way, and me because we won't have any future
"accidents" in using len()). What do you think about that?

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

jeremycline: One thing to note is that although the caching version of QuerySet provides an implementation of ``__len__``, the non-caching variety does not. Personally I feel like ``__len__`` is the "obvious way" to get the length of an iterable, but given that upstream went with both ``count`` and occasionally ``__len__`` (with different behaviors for each)...

rbarlow: Jeremy Cline wrote:
> One thing to note is that although the caching version of QuerySet
> provides an implementation of |__len__|, the non-caching variety does
> not. Personally I feel like |__len__| is the "obvious way" to get the
> length of an iterable, but given that upstream went with both |count|
> and occasionally |__len__| (with different behaviors for each)...

Ah that does sway my opinion. I'd prefer not to implement __len__ here
if the superclass does not implement it.

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

PR #2222, opened 2015-12-04
Use Celery's local_received time for worker availability.

rbarlow:
Pulp had been using the timestamp attribute of Celery's Events to determine when the workers were most
recently available. Due to a bug in Celery, this cause an off-by-one-hour bug for users who live in a locale
that does daylight savings time that is not currently observing daylight savings time (such as New York in
the winter):

https://github.com/celery/celery/issues/1802#issuecomment-161916587

This fixes the issue by using the local_received time which is computed in the receiver by Celery with
time.time() without any timezone shifting attempts.

https://pulp.plan.io/issues/1380

fixes #1380

### DIFF
--- a/server/pulp/server/async/worker_watcher.py
+++ b/server/pulp/server/async/worker_watcher.py
@@ -28,37 +28,35 @@ def _parse_and_log_event(event):
     """
     Parse and return the event information we are interested in. Also log it.
 
-    A new dict is returned containing the keys 'timestamp', 'type', and 'worker_name'. The
-    only data transformation here is on the timestamp. The timestamp arrives as seconds since
-    the epoch, and is converted to UTC time, and returned as a naive datetime.datetime object.
+    A new dict is returned containing the keys 'timestamp', 'local_received', 'type', and
+    'worker_name'. The data transformations here are on the timestamp and local_received. They
+    both arrive as seconds since the epoch, and are converted to a naive datetime.datetime object in
+    UTC. The timestamp is set by the sender, and the local_received time is set by the receiver.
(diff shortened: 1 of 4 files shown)


rbarlow:
The F23 Jenkies failure should be ignored as it is not related to this pull request. All other builds passed.

server/pulp/server/async/worker_watcher.py
```diff
@@ -28,37 +28,35 @@ def _parse_and_log_event(event):
...
-    Logging is done through a call to _log_event()
+    A new dict is returned containing the keys 'timestamp', 'local_received', 'type', and
+    'worker_name'. The data transformations here are on the timestamp and local_received. They
+    both arrive as seconds since the epoch, and are converted to a naive datetime.datetime object in
+    UTC. The timestamp is set by the sender, and the local_received time is set by the receiver.
+    Beware of a bug in the value of the timestamp as of the time of this commit as it suffers from
+    issues in localities that use daylight savings time during the non-daylight savings time part of
+    the year. See https://github.com/celery/celery/issues/1802#issuecomment-161916587 for discussion
+    around this issue. Until that issue is resolved, consider using the local_received time instead
+    of the timestamp.
```
bmbouter: I expected more language about how horrible DST and timezones are.

bmbouter:
I like the new log statement. LGTM.

jeremycline:
Based on my limited understanding of how everything works, this looks good to me.

server/pulp/server/async/worker_watcher.py
rbarlow: Haha, well I am sorry to disappoint!

PR #2224, opened 2015-12-04
Removes URLs from migration 0009 error messages

dkliban:


kfiresmith:
Well, I rather like the error messages having a helpful URL, and with the docs linking whatever the latest version is at the time to "https://pulp.readthedocs.org/en/latest/", it seems safe to just update the source code of the error message to point to the "/latest/" version of the docs and that should work forever.

dkliban:
I also updated the comps.xml [0] so that you should just be able to group install @pulp-server-qpid and get qpid-tools installed. I would prefer not having any links because even if the page is there in the future, the content might not be there anymore.

[0] https://github.com/pulp/pulp/pull/2220

server/pulp/server/db/migrations/0009_qpid_queues.py
```diff
@@ -41,17 +35,13 @@ def migrate(*args, **kwargs):
 
     if not QPID_MESSAGING_AVAILABLE:
         msg = _('Migration 0009 did not run because the python package qpid.messaging is not '
-                'installed. Please install qpid.messaging and rerun the migrations. See %s'
-                'for more information.')
-        msg = msg % QPID_MESSAGING_URL
+                'installed. Please group install @pulp-server-qpid.')
```
bmbouter: Maybe offer them two choices and with some rewording.

```
installed. Pulp's Qpid client dependencies can be installed with the "pulp-server-qpid" package group. See the installation docs for more information. Alternatively, you may reconfigure Pulp to use RabbitMQ.
```

server/pulp/server/db/migrations/0009_qpid_queues.py
```diff
@@ -41,17 +35,13 @@ def migrate(*args, **kwargs):
...
         logger.error(msg)
         raise Exception(msg)
 
     if not QPIDTOOLLIBS_AVAILABLE:
         msg = _('Migration 0009 did not run because the python package qpidtoollibs is not '
-                'installed. Please install qpidtoollibs and rerun the migrations. See %s for more '
-                'information.')
-        msg = msg % QPIDTOOLLIBS_URL
+                'installed. Please group install @pulp-server-qpid.')
```
bmbouter: Same change as above.

kfiresmith:
Fair enough points @dkliban.

bmbouter:
@kfiresmith It's very likely Pulp will migrate out of read the docs at some point. Thanks for the detailed bug report!

server/pulp/server/db/migrations/0009_qpid_queues.py
```diff
@@ -41,17 +35,17 @@ def migrate(*args, **kwargs):
 
     if not QPID_MESSAGING_AVAILABLE:
         msg = _('Migration 0009 did not run because the python package qpid.messaging is not '
-                'installed. Please install qpid.messaging and rerun the migrations. See %s'
-                'for more information.')
-        msg = msg % QPID_MESSAGING_URL
+                'installed. Pulp\'s Qpid client dependencies can be installed with the '
+                '\"pulp-server-qpid\" package group. See the installation docs for more '
```
bmbouter: Since the single quote is being used, the escaping slashes in front of the double quotes aren't needed. They also don't hurt except being slightly less readable when string aren't being interpreted. Not required to change.

PR #2231, opened 2015-12-08
ref #1373 - changes needed to support plugins.

jortel:
General fixes and changes needed to better support plugins.
- The repository controller: `find_repo_content_units()` is called on repositories containing multiple types.  The query is not always valid for all types and raises `InvalidQueryError`.  Not sure if this is a good way to deal with this but seemed like the only option.
- For backwards compatibility we need the `storage_path` on single-file units to include the file name.  To facilitate this, the FileContentUnit needs to provide a way for plugins to specify what the file name is.  The semantics were challenging here.  A `set_filename()` method would suggest that it sets a `filename` attribute on the unit (which is does not).  So, don't love it but went with a *setter* instead.  Long term, when we can change file unit modeling, we can fix this.
- The storage_path is calculated when the unit is created.  This is really needed to support plugin work flows.  It's safe because all of the required attributes are already required to be class attributes (or defaults) for each derived model class.
- Added `list_files()` on file content units.  Provides the authoritative list of files for a unit (instead of the catalog).  The default is tailored for single-file-units.
- The last thing is to add a method for inserting a new revision of a catalog record.  The revision is necessary to provide read consistency for *download tasks*.  The tasks will select catalog entries ordered by revision.  This way, the newest entry is used and downloads can safely happen concurrently with catalog updates.  This PR does not adding the `order_by('revision')` to the appropriate places.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -668,38 +668,47 @@ class FileContentUnit(ContentUnit):
         ]
     }
 
-    @classmethod
-    def pre_save_signal(cls, sender, document, **kwargs):
-        """
-        The signal that is triggered before a unit is saved.
+    def __init__(self, *args, **kwargs):
+        super(FileContentUnit, self).__init__(*args, **kwargs)
+        self._storage_path = FileStorage.get_path(self)
 
-        :param sender: sender class
-        :type sender: object
-        :param document: Document that sent the signal
-        :type document: FileContentUnit
+    @property
+    def storage_path(self):
         """
-        if not document._storage_path:
-            document.set_storage_path()
-        super(FileContentUnit, cls).pre_save_signal(sender, document, **kwargs)
+        The content storage path.
 
-    def set_storage_path(self, filename=None):
+        :return: The absolute path to stored content.
+        :rtype: str
         """
-        Set the storage path including an optional file name.
-        The path is calculated as: content/units/<type_id>/unit.id[0:4]/unit.id/<filename>.
-        A *filename* should only be specified when the content unit is associated
-        with a single file.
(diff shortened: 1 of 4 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -847,8 +859,31 @@ class LazyCatalogEntry(AutoRetryDocument):
...
     checksum_algorithm = StringField(regex=ALG_REGEX)
+    revision = IntField(default=0)
     data = DictField()
 
+    def save_revision(self):
+        """
+        Add the entry using the next revision number.
+        Previous revisions are deleted.
+        """
+        revisions = {0}
```
jeremycline: I believe this needs to be ``set([0])`` since set literals are Python 2.7+[0].

[0] https://docs.python.org/3/whatsnew/2.7.html#python-3-1-features

server/pulp/server/db/model/__init__.py
```diff
@@ -847,8 +859,31 @@ class LazyCatalogEntry(AutoRetryDocument):
...
+        query = dict(
+            unit_id=self.unit_id,
+            unit_type_id=self.unit_type_id,
+            importer_id=self.importer_id)
+        # Find revisions
+        qs = LazyCatalogEntry.objects.filter(**query)
+        for revision in qs.distinct('revision'):
+            revisions.add(revision)
+        # Add new revision
+        last_revision = sorted(revisions, reverse=True)[0]
```
jeremycline: Not that it matters, but you could do ``sorted(revisions)[-1]``. I'm not sure which I find more readable, though.

jeremycline:
After the set literal notation goes away, LGTM

PR #2232, opened 2015-12-08
Fix a unit test that was broken during branch merges.

rbarlow:


server/test/unit/server/async/test_scheduler.py
```diff
@@ -123,7 +124,7 @@ def test_calls_handle_heartbeat(self, mock_celerybeatlock, mock_worker_watcher,
 
         expected_event = {'timestamp': 1449261335.275528, 'local_received': 1449261335.275528,
                           'type': 'scheduler-event', 'hostname': 'scheduler@some_host'}
-        mock_heartbeat.assert_called_once_with(expected_event)
+        mock_worker_watcher.handle_worker_heartbeat.assert_called_once_with(expected_event)
```
seandst: Was this just a merge gone weird? I would have expected flake8 to pick this up...

PR #2233, opened 2015-12-09
Convert the distributor manager to mongoengine friendly controller.

asmacdo:
Replaces the old RepoDistributor model and all its uses with new model,
pulp.server.db.model.Distributor. Also replaces all instances of using
pymongo to access the the distributors collection with mongoengine.

https://pulp.plan.io/issues/780


asmacdo:
ok test

nodes/test/nodes_tests/test_plugins.py
```diff
@@ -169,7 +168,7 @@ def setUp(self):
         self.alias = (self.parentfs, self.parentfs)
         Consumer.get_collection().remove()
         Bind.get_collection().remove()
-        RepoDistributor.get_collection().remove()
+        model.Distributor.drop_collection()
```
mhrivnak: Does dropping the collection also remove indicies? If so, then sticking with a remove call would be better.

server/pulp/plugins/conduits/mixins.py
```diff
@@ -305,40 +305,33 @@ def __init__(self, repo_id, distributor_id):
...
+        :raises DistributorConduitException: wraps any exception that may occur in the Pulp server
         """
         try:
-            distributor_manager = manager_factory.repo_distributor_manager()
-            value = distributor_manager.get_distributor_scratchpad(self.repo_id,
-                                                                   self.distributor_id)
-            return value
+            qs = model.Distributor.objects.get_or_404
+            distributor = qs(repo_id=self.repo_id, distributor_id=self.distributor_id)
+            return distributor.scratchpad or None
```
mhrivnak: Why the ``or None`` ? Do you specifically want to replace a return value of ``{}`` with ``None``, or something like that?

server/pulp/plugins/conduits/mixins.py
```diff
@@ -305,40 +305,33 @@ def __init__(self, repo_id, distributor_id):
...
-        @raises DistributorConduitException: wraps any exception that may occur
-                in the Pulp server
+        :param value: will overwrite the existing scratchpad
+        :type  value: dict
+        :raises DistributorConduitException: wraps any exception that may occur in the Pulp server
         """
         try:
-            distributor_manager = manager_factory.repo_distributor_manager()
-            distributor_manager.set_distributor_scratchpad(self.repo_id, self.distributor_id, value)
+            qs = model.Distributor.objects.get_or_404
```
mhrivnak: It's worth noting that the ``except`` statement below will catch the ``MissingResource`` exception and turn it into whatever a ``DistributorConduitException`` is. :( You've preserved previous behavior, which is totally fine, but I thought I'd at least point it out that the result of calling ``get_or_404`` is being diminished.

server/pulp/plugins/conduits/repo_config.py
```diff
@@ -5,7 +5,8 @@
 
 import os
 
-from pulp.server.db.model.repository import RepoDistributor
+from pulp.server.db import model
```
mhrivnak: ``pulp`` goes in the third group of imports, since this file is part of it.

server/pulp/plugins/conduits/repo_publish.py
```diff
@@ -64,11 +63,9 @@ def last_publish(self):
         :raises DistributorConduitException: if any errors occur
         """
         try:
-            collection = RepoDistributor.get_collection()
-            distributor = collection.find_one({'repo_id': self.repo_id, 'id': self.distributor_id})
-            if distributor is None:
-                raise pulp_exceptions.MissingResource(self.repo_id)
-            return dateutils.ensure_tz(distributor['last_publish'])
+            dist = model.Distributor.objects.get_or_404(repo_id=self.repo_id,
+                                                        distributor_id=self.distributor_id)
```
mhrivnak: I recommend adding ``.only('last_publish')``, if ``.get()`` supports that. Let me know if you know for sure whether it does or doesn't support that.

server/pulp/server/controllers/distributor.py
```diff
@@ -1,109 +1,294 @@
...
-    Get the itinerary for deleting a repository distributor.
-      1. Delete the distributor on the sever.
-      2. Unbind any bound consumers.
-    :param repo_id: A repository ID.
-    :type repo_id: str
-    :param distributor_id: A distributor id
-    :type distributor_id: str
-    :return: Any errors that may have occurred and the list of tasks spawned for each consumer
-    :rtype TaskResult
+    Removes a distributor from a repository and unbinds any bount consumers.
```
mhrivnak: s/bount/bound

server/pulp/server/controllers/distributor.py
```diff
@@ -1,109 +1,294 @@
...
+    async_result = update.apply_async_with_reservation(
+        tags.RESOURCE_REPOSITORY_TYPE, distributor.repo_id,
+        [distributor, config, delta],
+        tags=task_tags)
+    return async_result
+
+
 @celery.task(base=Task, name='pulp.server.tasks.repository.distributor_update')
-def update(repo_id, distributor_id, config, delta):
+def update(distributor, config, delta):
```
mhrivnak: For tasks, we are much better off passing IDs than objects. Otherwise the object has to be serialized and stuffed into an AMQP message, which is usually doable, but not the best approach. @shubham90 will thank you.

server/pulp/plugins/conduits/repo_publish.py
asmacdo: `.only` will work if used before the `.get_or_404`.

server/pulp/server/controllers/distributor.py
```diff
@@ -1,109 +1,294 @@
...
+        if v is None:
+            distributor.config.pop(k)
+        else:
+            distributor.config[k] = v
 
-    distributor = manager.update_distributor_config(repo_id, distributor_id, config, auto_publish)
+    auto_publish = delta.get('auto_publish') if delta else None
+    if not isinstance(auto_publish, (bool, type(None))):
+        raise exceptions.InvalidValue(['auto_publish'])
+    distributor.auto_publish = auto_publish or distributor.auto_publish
```
mhrivnak: If ``distributor.auto_publish == True`` and ``delta['auto_publish'] == False``, I think this will neglect to set the value to ``False`` on the object.

server/pulp/server/db/model/__init__.py
```diff
@@ -863,6 +862,36 @@ def _pbkdf_sha256(self, password, salt, iterations):
...
+class Distributor(AutoRetryDocument):
+    """
+    Defines schema for a Distributor in the 'repo_distributors' collection.
+    """
+    repo_id = StringField(required=True)
+    distributor_id = StringField(required=True, regex=r'^[\-_A-Za-z0-9]+$')
+    distributor_type_id = StringField(required=True)
+    config = DictField()
+    auto_publish = BooleanField(default=False)
+    last_publish = DateTimeField()
```
mhrivnak: You should use the brand-new ``pulp.server.db.fields.UTCDateTimeField``

server/pulp/server/db/model/repository.py
```diff
@@ -5,72 +5,6 @@
 import pulp.common.dateutils as dateutils
 
 
-class RepoDistributor(Model):
```
mhrivnak: Another one bites the dust!

server/pulp/server/managers/schedule/repo.py
```diff
@@ -178,22 +175,20 @@ def create(cls, repo_id, distributor_id, publish_options, schedule,
...
-        resource = RepoDistributor.build_resource_tag(repo_id, distributor_id)
         schedule = ScheduledCall(schedule, task, args=args, kwargs=kwargs,
-                                 resource=resource, failure_threshold=failure_threshold,
+                                 resource=dist.resource_tag, failure_threshold=failure_threshold,
                                  enabled=enabled)
         schedule.save()
 
         try:
-            cls.validate_distributor(repo_id, distributor_id)
+            model.Distributor.objects.get_or_404(repo_id=repo_id, distributor_id=distributor_id)
```
mhrivnak: For a future effort, it seems like it might be helpful to have a method ``objects.exists_or_404`` that does the cheapest operation possible to determine if the thing exists, without actually retrieving it.

server/pulp/server/managers/schedule/repo.py
```diff
@@ -250,22 +241,15 @@ def delete_by_distributor_id(repo_id, distributor_id):
...
-
-        :raise: pulp.server.exceptions.MissingResource
         """
-        distributor_manager = managers_factory.repo_distributor_manager()
-        distributor_manager.get_distributor(repo_id, distributor_id)
+        try:
+            dist = model.Distributor.objects.get_or_404(repo_id=repo_id,
+                                                        distributor_id=distributor_id)
+        except exceptions.MissingResource:
+            pass
```
mhrivnak: If the repo or distributor doesn't exist, I think that's a problem the caller should know about. I'd let this bubble up.

nodes/test/nodes_tests/test_plugins.py
rbarlow: I believe dropping the collection does remove the indices as well.

server/pulp/plugins/conduits/mixins.py
asmacdo: I specifically wanted to replace the default `{}` with `None` to preserve the previous behavior. 

https://github.com/pulp/pulp/blob/master/server/pulp/server/managers/repo/distributor.py#L381

server/pulp/server/managers/schedule/repo.py
asmacdo: I think that the intent was that this should be idempotent.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -683,26 +660,15 @@ def delete(self, request, repo_id, distributor_id):
...
-            tags.resource_tag(tags.RESOURCE_REPOSITORY_TYPE, repo_id),
-            tags.resource_tag(tags.RESOURCE_REPOSITORY_DISTRIBUTOR_TYPE, distributor_id),
-            tags.action_tag('remove_distributor')
-        ]
-        async_result = dist_controller.delete.apply_async_with_reservation(
-            tags.RESOURCE_REPOSITORY_TYPE, repo_id, [repo_id, distributor_id],
-            tags=task_tags)
-        raise pulp_exceptions.OperationPostponed(async_result)
+        dist = model.Distributor.objects.get_or_404(repo_id=repo_id, distributor_id=distributor_id)
+        async_result = dist_controller.queue_delete(dist)
```
mhrivnak: +1 centralizing this queuing work really pays off.

server/pulp/plugins/conduits/mixins.py
mhrivnak: Gotcha. The new behavior is slightly different than the old. In the previous case, if the scratchpad was an empty dict, ``get_scratchpad()`` would return an empty dict. In your case, it would return None instead.

The difference is that calling ``.get('scratchpad', None)`` only returns the default None if the key ``'scratchpad'`` doesn't exist on the document. In your code, the value for ``.scratchpad`` is evaluated as a ``bool`` and replaced with ``None`` if that evaluates to ``False``.

Does that make sense?

server/pulp/server/managers/schedule/repo.py
mhrivnak: I agree that deleting the schedules should be idempotent. Call this as many times as you want on a distributor, and we'll make sure that distributor doesn't have any schedules. But this function is not responsible for deleting the distributor itself, so it should complain if the distributor doesn't exist. If the caller happens to also be trying to delete the distributor, it can ignore that error if it wants.

server/pulp/plugins/conduits/mixins.py
asmacdo: yup, nice catch.

asmacdo:
@mhrivnak thanks for the super fast review!

asmacdo:
ok test

PR #2234, opened 2015-12-09
Two improvements related to checksum_type validation

bmbouter:
* The checksum_type now validates while being sanitized
* If its invalid a coded exception is raised
* A coded exception for checksum not being correct is introduced

### DIFF
--- a/server/pulp/plugins/util/verification.py
+++ b/server/pulp/plugins/util/verification.py
@@ -4,6 +4,10 @@ Functions for verifying files.
 
 import hashlib
 
+from pulp.common import error_codes
+
+from pulp.server.exceptions import PulpCodedException
+
 
 # Number of bytes to read into RAM at a time when validating the checksum
 VALIDATION_CHUNK_SIZE = 32 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


server/pulp/plugins/util/verification.py
```diff
@@ -40,26 +44,32 @@ class VerificationException(ValueError):
...
+
     :param checksum_type: The checksum type we are sanitizing
     :type  checksum_type: basestring
-    :return:              A sanitized checksum type, converting "sha" to "sha1", otherwise returning
-                          the given checksum_type in lowercase.
-    :rtype:               basestring
-    """
-    # In our unit tests, there are examples of checksum_type being None. Since None doesn't have
-    # string-like operations we cannot continue, so we should return.
-    if checksum_type is None:
```
jeremycline: :+1:

server/test/unit/plugins/util/test_verification.py
```diff
@@ -49,6 +36,10 @@ def test_SHA256_to_sha256(self):
 
         self.assertEqual(checksum_type, 'sha256')
 
+    def test_invalid_type_raises_coded_exception(self):
+        self.assertRaises(PulpCodedException,
+                          verification.sanitize_checksum_type, 'not_a_real_checksum')
```
jeremycline: You're going to be sorry when my new checksum algorithm, 'not_a_real_checksum', becomes widely accepted and breaks this unit test.

bmbouter: That competes with my new checksum library `didnt_check`. It's inspiration comes from the very excellent [Didn't Test Framework](https://github.com/daviddavis/didnt) and has similar benefits.

PR #2235, opened 2015-12-09
Parallelizes applicability regeneration for a repo

dkliban:
This patch provides a new Celery task for performing applicability regenration
for a batch of applicability profiles. The ApplicabilityRegenerationManager
dispatches a series of tasks with the same group id. Each task is dispatched with
a list of up to 100 RepoProfileApplicabilities to reevaluate.

The API endpoint for generating content applicability for updated repositories
changes as part of this patch. Instead of returning 202 with a call report, the
server returns 202 with a group call report.

This patch does not make any changes to the algorithm used to calculate content
applicability.

https://pulp.plan.io/issues/20
closes #20


dkliban:
ok test

rbarlow:
Dennis Kliban wrote:
> Instead of returning 202 with a call report, the
> server returns 200 with a group call report.

I am a little concerned about semantic versioning here. Is there a way
to maintain API compatibility by having the call report list the group
report as one of the spawned tasks?

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

dkliban:
@rbarlow I actually changed this documentation after the original PR was submitted. Though in the code, it still does return 200. I would like to continue sending out a 202 because it is more appropriate.

docs/dev-guide/integration/rest-api/consumer/applicability.rst
```diff
@@ -87,6 +85,13 @@ and postponed until the current task is completed.
   }
  }
 
+:group_call_report:`_` ::
+
+ {
+     "_href": "/pulp/api/v2/task_groups/16412fcb-06fa-4caa-818b-b103e2a9bf44/state_summary/",
```
mhrivnak: FWIW I'm not crazy about this, but I'm not sure we can do better right now. In theory the ``_href`` should reference the root of the resource, but I recognize that we probably haven't implemented that endpoint yet (or have we?). In an ideal world, our REST API would have a section for links, one of which could be ``_href``, among others. Like this:

```
{
  "_links": {
    "href": "/pulp/api/v2/task_groups/16412fcb-06fa-4caa-818b-b103e2a9bf44/",
    "state_summary": "/pulp/api/v2/task_groups/16412fcb-06fa-4caa-818b-b103e2a9bf44/state_summary/",
  }
}

In which case we could actually just leave out the ``href`` for now. But we don't have that structure, so we probably can't.

On balance, I think it's better to do the most correct option, have the ``_href`` reference the root of the resource, and make the API client add ``state_summary`` if they want it.

File: server/pulp/server/managers/consumer/applicability.py [hunk 2]
```diff
@@ -1,9 +1,11 @@
+
```
mhrivnak: I assume this was an accident and can be removed.

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -118,29 +120,52 @@ def regenerate_applicability_for_repos(repo_criteria):
         repo_criteria.fields = ['id']
         repo_ids = [r.repo_id for r in model.Repository.objects.find_by_criteria(repo_criteria)]
 
+        task_group_id = uuid4()
+
         for repo_id in repo_ids:
             # Find all existing applicabilities for given repo_id. Setting batch size of 5 ensures
```
mhrivnak: This comment can be removed. Since this function no longer does the work synchronously, it should never cause that timeout.

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -118,29 +120,52 @@ def regenerate_applicability_for_repos(repo_criteria):
...
             # the MongoDB cursor does not time out. See https://pulp.plan.io/issues/998#note-6 for
             # more details.
             existing_applicabilities = RepoProfileApplicability.get_collection().find(
-                {'repo_id': repo_id}).batch_size(5)
+                {'repo_id': repo_id})
+            counter = 0
```
mhrivnak: This will be much simpler using https://github.com/pulp/pulp/blob/master/server/pulp/plugins/util/misc.py#L14

15 lines of code basically becomes:

```
for batch in paginate(existing_applicabilities, 100):
  regenerate_applicability_for_repos_task.apply_async(repo_id, batch, group_id=task_group_id)
```

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -118,29 +120,52 @@ def regenerate_applicability_for_repos(repo_criteria):
...
+
+            if len(existing_applicabilities_batch) > 0:
+                # Dispatch the remaining batch
+                task_kwargs = {"group_id": task_group_id}
+                regenerate_applicability_for_repos_task.apply_async((repo_id,
+                                                                    existing_applicabilities_batch),**task_kwargs)
+        return task_group_id
+
+    @staticmethod
+    def regenerate_applicability_for_repos_task(repo_id, existing_applicabilities, group_id=None):
```
mhrivnak: need a doc block. For tasks, it's preferable to pass IDs instead of objects, so we don't have to serialize objects and stuff them in an AMQP message. A list of unique IDs would be ideal for the ``existing_applicabilities`` parameter.

mhrivnak: How do you feel about renaming this function to ``batch_regenerate_applicability`` ? I think that's more accurate for what it does.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -938,11 +937,11 @@ def post(self, request):
             invalid_criteria.add_child_exception(e)
             raise invalid_criteria
 
-        regeneration_tag = tags.action_tag('content_applicability_regeneration')
-        async_result = regenerate_applicability_for_repos.apply_async_with_reservation(
-            tags.RESOURCE_REPOSITORY_PROFILE_APPLICABILITY_TYPE, tags.RESOURCE_ANY_ID,
-            (repo_criteria.as_dict(),), tags=[regeneration_tag])
-        raise pulp_exceptions.OperationPostponed(async_result)
+        async_result = ApplicabilityRegenerationManager.regenerate_applicability_for_repos(
```
mhrivnak: This got me thinking about the name of the function being called. I suggest ``queue_regenerate_applicability_for_repos``, since all it really does it make some tasks. That also makes it more obvious that it's going to return some kind of async job reference.

docs/dev-guide/integration/rest-api/consumer/applicability.rst
dkliban: @mhrivnak  I would love to add a link to the root of the resource, but it would look ugly. We currently use the reverse() method provided by django formulate the URL for the task_group_summary view. Since we don't have a view that represents the task_group resource, we can't do a reverse look up for it. I could just manually remove the state_summary from the URL we have now. Thoughts?

mhrivnak: How about making a 1-line view function for the root of the resource that raises 404, but then you at least have something to reference with ``reverse`` ?

dkliban: Works for me. I'll add it to this PR.

server/pulp/server/webservices/views/task_groups.py
```diff
@@ -11,6 +11,28 @@
...
 
+class TaskGroupView(View):
+    """
+    View for a task group summary.
+    """
+
+    @auth_required(authorization.READ)
+    def get(self, request, group_id):
+        """
+        Return a response containing a summary of task states for task group.
```
dkliban: I will update this docblock shortly

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -118,29 +121,48 @@ def regenerate_applicability_for_repos(repo_criteria):
...
         for repo_id in repo_ids:
-            # Find all existing applicabilities for given repo_id. Setting batch size of 5 ensures
-            # the MongoDB cursor does not time out. See https://pulp.plan.io/issues/998#note-6 for
-            # more details.
-            existing_applicabilities = RepoProfileApplicability.get_collection().find(
-                {'repo_id': repo_id}).batch_size(5)
-            for existing_applicability in existing_applicabilities:
+            existing_applicability_ids = RepoProfileApplicability.get_collection().find(
+                {'repo_id': repo_id}, {'_id': 1})
+            for batch in paginate(existing_applicability_ids, 100):
```
mhrivnak: I recall that someone has measured this calculation at about 12 seconds per profile. I don't know how representative that is of production use, but it's probably in the neighborhood. 100 of those would take 20 minutes, which is a long time for a single task to run. I suggest going with more granularity, so workers have more opportunity to alternate with other kinds of work. For example, a batch size of 20 would take 4 minutes at that speed. I'd rather err on the side of being too granular, than to have users notice a long-running task and get annoyed by it.

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -118,29 +121,48 @@ def regenerate_applicability_for_repos(repo_criteria):
...
+    def batch_regenerate_applicability(repo_id, existing_applicability_ids, group_id=None):
+        """
+        Regenerate and save applicability data for a batch of existing applicabilities
+
+        :param repo_id: Repository id for which applicability is being calculated
+        :type repo_id: str
+        :param existing_applicability_ids: Tuple of Object Ids for applicability profiles
+        :type existing_applicability_ids: tuple of dicts in form of {"_id": ObjectID('mongo-id')}
+        :param group_id: task group id that this task should belong to
+        :type group_id: UUID
```
mhrivnak: the type is still ``str`` even though the data is a UUID

docs/dev-guide/integration/rest-api/consumer/applicability.rst
mhrivnak: Great. Does the _href doc above need to be updated then to remove the ``state_summary`` part?

mhrivnak:
Looks very good! The last thing it definitely needs is a release note.

server/pulp/server/managers/consumer/applicability.py
dkliban: I was thinking about this last night also. I was going to suggest 5.

server/pulp/server/managers/consumer/applicability.py
dkliban: We are using a UUID field type in the TaskStatus model. So we are expecting this to be of type UUID 4.

dkliban:
ok test

docs/user-guide/release-notes/master.rst
```diff
@@ -15,6 +15,8 @@ New Features
   store rather than the certificate authority trust store bundled with
   ``python-requests``.
 
+* Content applicability for an updated repository is calculated in parallel.
```
mhrivnak: The API change needs to be called out as well. It's minor and probably doesn't affect many people, but we do need to fess up. :)

mhrivnak:
Just need the release note improvement, and we're good to go. Thanks! Good work!

dkliban:
ok test

dkliban:
ok test

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -50,3 +52,24 @@ Example Call Report::
...
+.. _group_call_report:
+
+Group Call Report
+-----------------
+
+A 202 ACCEPTED response returns a **Group Call Report** JSON object as the response body that has
+the following fields:
+
+* **_href** - Path to the root of task group resource. However, this API endpoint currently
+  returns 404 in all cases. You can append `/state-summary/` to the URL and perform a GET request
```
mhrivnak: Drop the leading / to stay within the normal "urljoin" concepts, in which a leading / would cause the entire path of the URL to be overwritten.

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -16,9 +16,11 @@ A success response indicates no conflicts were detected and the REST call
 executed. This is what is typically expected from a REST API call.
 
 A postponed response indicates that some portion of the command has been
-queued to execute asynchronously.  In this case a :ref:`call_report` will be returned
-with the results of the synchronously executed portion of the command, if there are any,
-and a list of the tasks that have been spawned to complete the work in the future.
+queued to execute asynchronously.  In this case a :ref:`call_report` or a :ref:`group_call_report`
+will be returned with the results of the synchronously executed portion of the command. If there
+are any tasks that were dispatched to execute serially, a :ref:`call_report` will contain a list
```
mhrivnak: I don't think spawned tasks have to execute serially.

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -16,9 +16,11 @@ A success response indicates no conflicts were detected and the REST call
...
+of spawned tasks. If a lot of tasks were dispatched to execute in parallel, a
+:ref:`group_call_report` is returned.
```
mhrivnak: When a task queues more tasks that are "next steps", like a publish after a sync, or unbind consumers after a repo delete, those are spawned tasks. The spawned tasks are not necessarily related to each other, now will a client necessarily want to track them together.

When some number of related tasks are queued, and it is useful to track the state of all of them together, that is when a task group is used.

I would not try to make a serial vs. parallel explanation.

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -15,10 +15,16 @@ Any REST API will return one of three responses:
 A success response indicates no conflicts were detected and the REST call
 executed. This is what is typically expected from a REST API call.
 
-A postponed response indicates that some portion of the command has been
-queued to execute asynchronously.  In this case a :ref:`call_report` will be returned
-with the results of the synchronously executed portion of the command, if there are any,
-and a list of the tasks that have been spawned to complete the work in the future.
+A postponed response indicates that some portion of the command has been queued to execute
+asynchronously.  In this case a :ref:`call_report` or a :ref:`group_call_report` will be returned
```
bmbouter: s/case/case,/

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -15,10 +15,16 @@ Any REST API will return one of three responses:
...
+with the results of the synchronously executed portion of the command.
+
+When a single task is queued or a task with follow up tasks, such as a publish after a sync or unbind
```
bmbouter: s/with/dispatches/

bmbouter: consider removing `or unbind consumers after repo delete`. Two examples aren't that much better than 1 and they make it more confusing.

docs/dev-guide/conventions/sync-v-async.rst
```diff
@@ -15,10 +15,16 @@ Any REST API will return one of three responses:
...
+consumers after repo delete, a :ref:`call_report` is returned. The :ref:`call_report` will contain
+a list of spawned tasks.
+
+When some number of related tasks are queued, and it is useful to track the state of all of them
```
bmbouter: s/some number of/many/
s/queued,/queued/

PR #2236, opened 2015-12-09
Add telnet to the development environment.

rbarlow:
I recently added rpdb to the development environment, but it is not very useful without telnet ☺


jeremycline:
*Technically* it is useful without telnet since you can also use ncat, but I'm not one to point out technicalities, you know?

rbarlow:
Jeremy Cline wrote:
> /Technically/ it is useful without telnet since you can also use ncat,
> but I'm not one to point out technicalities, you know?

Haha I suppose you are correct. I could also open a socket to it myself.

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

PR #2238, opened 2015-12-10
make rhsm an external dep

asmacdo:


seandst:
While this LGTM, I'm not sure that I know all of the gotchas that might come with adding external deps. I do see that the dep version matches what we're building, and as far as I can tell it's good. :)

asmacdo:
ok test

PR #2239, opened 2015-12-10
ref #1337 - Discontinue linking files for each unit during node publish.

jortel:
https://pulp.plan.io/issues/1337

See details ^^.


jortel:
ok test

PR #2240, opened 2015-12-11
Add lazy to the ansible playbook

jeremycline:
This doesn't enable or start the pulp streamer since that gets set up in the vagrant-setup.sh file which happens after the ansible run. When that changes, we can enable the streamer and start it in the lazy role.

### DIFF
--- a/playpen/ansible/dev-playbook.yml
+++ b/playpen/ansible/dev-playbook.yml
@@ -7,3 +7,4 @@
     - db
     - qpidd
     - dev
+    - lazy
--- a/playpen/ansible/roles/dev/files/bashrc
+++ b/playpen/ansible/roles/dev/files/bashrc
@@ -67,7 +67,7 @@ ptests() {
 }
 
 _paction() {
-    sudo systemctl $1 goferd httpd pulp_workers pulp_celerybeat pulp_resource_manager
+    sudo systemctl $1 goferd httpd pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer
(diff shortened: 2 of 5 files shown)


playpen/ansible/dev-playbook.yml
```diff
@@ -6,4 +6,5 @@
     - core
     - db
     - qpidd
-    - dev
\ No newline at end of file
+    - dev
+    - lazy
```
rbarlow: I didn't realize we had a separate dev and vagrant playbook these days. I had actually been planning to rename vagrant-playbook to dev-playbook so now that's confusing.

rbarlow:
Should we make this an optional role? I'm fine either way, but it's worth considering. LGTM.

rbarlow:
We could easily make it optional later, so no need to be part of this PR.

playpen/ansible/dev-playbook.yml
jeremycline: I just noticed this a few minutes ago, too. I'll add it in both until one or the other passes into the ether.

jeremycline:
I'm fine with either one, but it's off by default in /etc/pulp/server.conf so people aren't going to be lazy unless they want to be.

playpen/ansible/dev-playbook.yml
jeremycline: Oops. I guess I noticed, forgot, and noticed again. Clearly I already added it to both here. :coffee:

PR #2243, opened 2015-12-11
Add python-dnf to the dev-setup.sh script.

rbarlow:
python-dnf is a prerequisite for Ansible's dnf module to work, and is not included in Fedora 23 by default.

### DIFF
--- a/playpen/dev-setup.sh
+++ b/playpen/dev-setup.sh
@@ -105,6 +105,9 @@ for r in $REPOS; do
   fi
 done
 
+# Install some of Ansible's dependencies
+$HOME/devel/pulp/playpen/bootstrap-ansible.sh
+
 pushd pulp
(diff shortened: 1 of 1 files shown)


jeremycline:
Could this script make use of the playpen/bootstrap-ansible.sh script?

seandst:
> Could this script make use of the playpen/bootstrap-ansible.sh script?

It probably can, though I'm curious if there might be a reason that it doesn't.

rbarlow:
@jeremycline I think that is a good suggestion. I'll see if I can just do that instead.

rbarlow:
@jeremycline @seandst Yeah I was able to get that working, so thanks for the suggestion. Code sharing++!

rbarlow:
I am confident that this Jenkies failure is not due to this code, since this PR only changes playpen/dev-setup.sh. Therefore I am going to ignore Jenkies and merge.

PR #2245, opened 2015-12-11
Update the HTTP callback event publish to use requests.

jeremycline:
This addresses two problems: firstly, the listener would
not validate the signature on the server's x509 certificate. Secondly,
it was willing to perform SSLv2 and SSLv3. python-requests validates
certificates by default. On EL6+ and Fedora, it uses the system trust
store to perform this validation.

The protocols supported by requests is determined by urllib3. As of
Fedora 23, SSLv2 and SSLv3 is not supported.

closes #581

### DIFF
--- a/server/pulp/server/event/http.py
+++ b/server/pulp/server/event/http.py
@@ -8,13 +8,12 @@ url
 
 Eventually this should be enhanced to support authentication credentials as well.
 """
-
-import base64
-import httplib
+from gettext import gettext as _
 import logging
 import threading
 
-from pulp.server.compat import json, json_util
+from requests import post
+from requests.auth import HTTPBasicAuth
 
 
 TYPE_ID = 'http'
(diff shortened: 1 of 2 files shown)


server/pulp/server/event/http.py
```diff
@@ -25,56 +24,36 @@
 def handle_event(notifier_config, event):
     # fire the actual http push function off in a separate thread to keep
     # pulp from blocking or deadlocking due to the tasking subsystem
-
     data = event.data()
-
     _logger.info(data)
-
-    body = json.dumps(data, default=json_util.default)
```
rbarlow: What is the reason to remove this line? I'm not particularly familiar with Pulp's json_util module.

server/pulp/server/event/http.py
```diff
@@ -25,56 +24,36 @@
...
+        auth = HTTPBasicAuth(notifier_config['username'], notifier_config['password'])
     else:
-        connection = httplib.HTTPConnection(server)
-    return connection
+        auth = None
+
+    response = post(url, data=data, auth=auth)
+    if response.status_code != 200:
+        _logger.error(_('Received HTTP {code} from HTTP notifier to {url}.').format(
+            code=response.status_code, url=url))
```
rbarlow: Much cleaner!

server/test/unit/server/event/test_http.py
```diff
@@ -1,129 +1,54 @@
...
-        # Test HTTP
-        conn = http._create_connection('http', 'foo')
-        self.assertTrue(isinstance(conn, httplib.HTTPConnection))
+    @mock.patch(MODULE_PATH + 'HTTPBasicAuth')
+    @mock.patch(MODULE_PATH + 'post')
+    def test_send_post_auth(self, mock_post, mock_basic_auth):
+        notifier_config = {
+            'url': 'https://localhost/api/',
+            'username': 'jcline',
+            'password': 'hunter2'
```
rbarlow: nice password

server/test/unit/server/event/test_http.py
```diff
@@ -1,129 +1,54 @@
...
+        }
+        data = {'head': 'feet'}
+
+        http._send_post(notifier_config, data)
+        mock_post.assert_called_once_with(
+            'https://localhost/api/',
+            data=data,
+            auth=mock_basic_auth.return_value
+        )
+        mock_basic_auth.assert_called_once_with('jcline', 'hunter2')
```
rbarlow: It would be good to test the two error-level log statements too, especially the one where requests returns a non-200 response.

rbarlow:
LGTM, so much cleaner than before! Only blocker is that it would be ideal to at least test the case where requests returns a non-200. No need for re-review once you've got that, so merge once that's in place!

server/pulp/server/event/http.py
jeremycline: python-requests handles it all - we just have to hand it a `dict` rather than a string of JSON.

server/test/unit/server/event/test_http.py
jeremycline: Ah, didn't even notice I had missed the two if statements. That's what happens at the end of the day (literally).

server/pulp/server/event/http.py
rbarlow: On Wed, Dec 16, 2015 at 07:39:14PM -0800, Jeremy Cline wrote:
> python-requests handles it all - we just have to hand it a `dict` rather than a string of JSON.

So much better.

-- 
Randy Barlow
irc:   bowlofeggs
phone: +1-919-890-8150

PR #2249, opened 2015-12-14
the associate function failed to limit by unit type ID

mhrivnak:
This bug would not have been noticed for plugins with only one unit type.

This fixes the bug and also pulls the criteria-to-units logic out to a separate
function, so it can more easily be reasoned about and tested.

### DIFF
--- a/server/pulp/server/managers/repo/unit_association.py
+++ b/server/pulp/server/managers/repo/unit_association.py
@@ -139,7 +139,30 @@ class RepoUnitAssociationManager(object):
         return unique_count
 
     @staticmethod
-    def associate_from_repo(source_repo_id, dest_repo_id, criteria=None,
+    def _units_from_criteria(source_repo, criteria):
+        """
+        Given a criteria, return an iterator of units
+
+        :param source_repo: repository to look for units in
+        :type  source_repo: pulp.server.db.model.Repository
+        :param criteria:    criteria object to use for the search parameters
+        :type  criteria:    pulp.server.db.model.criteria.UnitAssociationCriteria
+
+        :return:    generator of pulp.server.db.model.ContentUnit instances
+        :rtype:     generator
+        """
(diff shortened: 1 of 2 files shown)


server/pulp/server/managers/repo/unit_association.py
```diff
@@ -139,7 +139,30 @@ def associate_all_by_ids(self, repo_id, unit_type_id, unit_id_list):
...
+        :param source_repo: repository to look for units in
+        :type  source_repo: pulp.server.db.model.Repository
+        :param criteria:    criteria object to use for the search parameters
+        :type  criteria:    pulp.server.db.model.criteria.UnitAssociationCriteria
+
+        :return:    generator of pulp.server.db.model.ContentUnit instances
+        :rtype:     generator
+        """
+        association_q = mongoengine.Q(__raw__=criteria.association_spec)
+        if criteria.type_ids:
```
mhrivnak: These two lines are the change that fixed the bug.

server/pulp/server/managers/repo/unit_association.py
```diff
@@ -139,7 +139,30 @@ def associate_all_by_ids(self, repo_id, unit_type_id, unit_id_list):
         return unique_count
 
     @staticmethod
-    def associate_from_repo(source_repo_id, dest_repo_id, criteria=None,
+    def _units_from_criteria(source_repo, criteria):
```
jeremycline: In my opinion, this would be a nice thing to have on a custom QuerySet. I can point you at all the places you need to make changes, if you aren't already familiar with them.

mhrivnak: That sounds like a fine idea for another day. Making this support a more general case would require adding support for limit, skip and sort at a minimum, which takes the scope well beyond my 2-line bugfix. Thanks for the suggestion!

PR #2251, opened 2015-12-14
Ansible now installs the RPM dependencies for each project.

jeremycline:
There's certainly a lot of room for improvement, but this is a start. One thing I've noticed is that quite often we need to make a list of projects (as you will see, I made another such list here), and it'd be nice if there was only one such list. I'm not sure where it should live though, so I'm open to suggestions.


playpen/ansible/library/pulp_facts.py
```diff
@@ -25,12 +26,41 @@
...
+    'pulp_python',
+    'pulp_rpm'
+]
+rpm_dependency_list = []
+# This is run using sudo, but the code is checked out in the normal user directory.
+user_homedir = pwd.getpwuid(int(os.environ['SUDO_UID'])).pw_dir
+for project in projects:
+    project_path = os.path.join(user_homedir, 'devel', project)
+    if os.path.isdir(project_path):
+        rpmspec_command = r"rpmspec -q --queryformat '[%{REQUIRENEVRS}\n]' " + project_path + \
```
asmacdo: Could you throw a comment on this?

jeremycline: No, but I can gently place a comment on it. There's no need for throwing.

PR #2252, opened 2015-12-14
Add pdebug to the Vagrant bashrc.

jeremycline:
This currently just uses telnet to connect to localhost port 4444. When
rpdb is updated, it could be used to send a SIGTRAP to a specified
process and attach to it.


rbarlow:
I think we can ignore jenkies since this is all in playpen.

PR #2257, opened 2015-12-15
Update Vagrant to use the nightly repository.

jeremycline:
Each night Jenkies re-builds Pulp and the dependencies we carry. Our
vagrant environment should pull from these (rather than a beta repo)
when on master.

### DIFF
--- a/playpen/ansible/roles/dev/tasks/main.yml
+++ b/playpen/ansible/roles/dev/tasks/main.yml
@@ -16,19 +16,9 @@
       url: https://repos.fedorapeople.org/repos/pulp/pulp/fedora-pulp.repo
       dest: /etc/yum.repos.d/fedora-pulp.repo
 
-- name: Enable Pulp 2.7 beta repository
-  command: dnf config-manager --set-enabled pulp-2.7-beta
-  when: pulp_27_beta_repo_enabled == false
-
-# This and its related fact should go away when we do have an f23 build, since it'll be a noop at that point
-- name: Rejigger the repo file if fedora 23 builds aren't available
-  replace:
-      dest: /etc/yum.repos.d/fedora-pulp.repo
-      regexp: 'fedora-\$releasever'
-      replace: fedora-22
-  when: not pulp_27_beta_f23_repo_available and
-        ansible_distribution == 'Fedora' and
-        ansible_distribution_version == '23'
+- name: Enable Pulp Nightly repository
+  command: dnf config-manager --set-enabled pulp-nightlies
+  when: pulp_nightly_repo_enabled == false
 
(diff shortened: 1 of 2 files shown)


playpen/ansible/roles/dev/files/pulp-nightlies.repo
```diff
@@ -0,0 +1,7 @@
+[pulp-nightlies]
+name=Pulp Nightlies $releasever - $basearch
+baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/master/fedora-$releasever/$basearch/
+enabled=1
+gpgcheck=0
+metadata_expire=6h
+skip_if_unavailable=0
```
rbarlow: What if we added this to the repo files we host on fedorapeople instead? That way it would be more advertised for community people who might want to help test.

playpen/ansible/roles/dev/tasks/main.yml
```diff
@@ -11,24 +11,7 @@
...
   notify: restart sshd
 
-- name: Install Pulp dnf repository
-  get_url:
-      url: https://repos.fedorapeople.org/repos/pulp/pulp/fedora-pulp.repo
-      dest: /etc/yum.repos.d/fedora-pulp.repo
-
-- name: Enable Pulp 2.7 beta repository
-  command: dnf config-manager --set-enabled pulp-2.7-beta
-  when: pulp_27_beta_repo_enabled == false
```
rbarlow: If you decide to put the repo on fedorapeople, then you can just change this to enable the nightly repo, but keep it instead of removing it.

rbarlow:
I like the idea here, but I think it might be slightly better to just add this repo definition to our repo files on fedorapeople.org. What do you think? It'll make this diff a little smaller as a side effect too.

playpen/ansible/roles/dev/tasks/main.yml
```diff
@@ -16,19 +16,9 @@
       url: https://repos.fedorapeople.org/repos/pulp/pulp/fedora-pulp.repo
       dest: /etc/yum.repos.d/fedora-pulp.repo
 
-- name: Enable Pulp 2.7 beta repository
-  command: dnf config-manager --set-enabled pulp-2.7-beta
-  when: pulp_27_beta_repo_enabled == false
-
-# This and its related fact should go away when we do have an f23 build, since it'll be a noop at that point
-- name: Rejigger the repo file if fedora 23 builds aren't available
```
rbarlow: Let us mourn that the word "rejigger" no longer will appear in our code.

jeremycline: It'll always be in our hearts (and git history)

PR #2260, opened 2015-12-15
Run the psmash linter.

rbarlow:


Ichimonji10:
Is the idea here that a developer can write tests in parallel with test code? Cool! If that's the case, consider adding in `make test && ` before the `make lint` part. `make test` causes Pulp Smash to run its own internal unit tests. It's extremely fast, and doesn't trigger any interactions with your Pulp system.

playpen/ansible/roles/dev/files/bashrc
```diff
@@ -75,6 +75,7 @@ _paction() {
 psmash() {
     workon pulp-smash;
     prestart;
+    make lint || return 1;
```
seandst: returning 1 here prevents  the later call to deactivate

I recommend something like

```bash
if make lint
then
    python -m unittest2 discover ~/devel/pulp-smash/pulp_smash;
fi
deactivate
```

seandst:
> make test && before the make lint

I think `make lint test` would probably be a more readable option, and I think I agree that if we're linting the pulp smash tests, we might as well run them, too.

Ichimonji10:
> I think `make lint test` would probably be a more readable option

I somehow did not know that you could specify multiple make targets with a single invocation of `make`. Thanks.

seandst:
Not only that, but make doesn't continue after a failed target (by default, at least), so the behavior is identical to `make lint && make test`

playpen/ansible/roles/dev/files/bashrc
rbarlow: Oh I like that, much nicer. My bash foo is so terrible.

rbarlow:
I have updated this PR to use make lint test with an if statement.

Ichimonji10:
:+1:

rbarlow:
ok test

rbarlow:
I am confident that the failing test here is not due to this pull request.

PR #2263, opened 2015-12-16
Fix failing unit tests on the lazy branch.

jeremycline:
There were a few unit tests failing on the lazy branch, some from the
merge of master into lazy-content. This ensures flake8 is happy and all
the tests pass.


mhrivnak:
Not realizing this PR existed, Jeff and I ended up fixing all of these same problems in a merge commit while merging master into lazy. I just emailed both of you about it, so please hold this PR until we decide what to do.

mhrivnak:
I think this should be closed, and then just visually confirm that similar fixes are in-place in master. Sound good?

PR #2264, opened 2015-12-16
Vagrant's `preset` now makes use of pulp-dev.py

jeremycline:
Several files were going missing after preset, including the server's
public RSA key in /var/lib/pulp/static. This ensures everything is
re-linked and re-installed when resetting Pulp.

### DIFF
--- a/playpen/ansible/roles/dev/files/bashrc
+++ b/playpen/ansible/roles/dev/files/bashrc
@@ -35,9 +35,32 @@ pstatus() {
 
 preset() {
     pstop;
+
+    pushd ~/devel/
+    # Remove and unlink all development files
+    for r in {pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_python,pulp_rpm,pulp}; do
+        if [ -d $r ]; then
+            pushd $r
+            sudo ./pulp-dev.py -U
+            popd
+        fi
+    done
+
     mongo pulp_database ~/drop_database.js;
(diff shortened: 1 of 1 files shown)


playpen/ansible/roles/dev/files/bashrc
```diff
@@ -35,9 +35,32 @@ pstatus() {
 
 preset() {
     pstop;
+
+    pushd ~/devel/
+    # Remove and unlink all development files
+    for r in {pulp_deb,pulp_docker,pulp_openstack,pulp_ostree,pulp_puppet,pulp_python,pulp_rpm,pulp}; do
```
rbarlow: Suggestion: since this PR is adding two more instances of this repeated list, I think now is a good time to make this list a variable. Otherwise when we add new plugins, we have to hunt down this list in many places. I know this list is also repeated in other places (Ansible and dev-setup.sh for example) and that it would be nice to solve this problem once and for all globally. However, I think for now that just making it a variable in only this file and only solving the problem for this file would be an improvement. We can think about the bigger problem later.

I also suggest leaving "pulp" out of the variable, since you need to uninstall pulp last and install it first and it would be nice to not have two variables to represent those two orderings.

rbarlow:
My suggestion is optional, so choose your own adventure. Nice work!

jeremycline:
I will admit I battled with bash for a while in an effort to get your suggestion working, but it would not behave (and I don't know much bash...)

It's certainly a good suggestion, but I'd like to get this in without spending several hours digesting bash syntax.

PR #2265, opened 2015-12-16
remove importer objects from tasks

asmacdo:
Distributors/importers conversion mixed with recent serialization work caused some problems. This removes passing importer objects through the tasking system. Additionally, recent changes to the distributor controller mean that it should be passing a model.Repository object to the plugin instead of a repo transfer unit.


asmacdo:
ok test

asmacdo:
@jortel, these tests are failing because of this commit:
https://github.com/pulp/pulp/commit/f21e76cbada239dae613c5c405271127f2e26272

I'm guessing that lazy needs an extra arg?

PR #2266, opened 2015-12-17
Fix config patcher; fix nodes tests.

jortel:
The `mock_config.patch` decorator only worked when code that imported the `config` module and used like `config.config.get()`.  I needed to be fixed to work when imported like:
```
```

Various fixes in test_plugins related to changes in master.


jortel:
@mhrivnak The failed tests seem unrelated.  Want me to merge it anyway?

PR #2268, opened 2015-12-17
The auth decorator no longer causes a 404s for consumers.

jeremycline:
It also no longer causes a 404 for missing users, which is an incorrect
behaviour since this decorator is used on all views.

closes #1399

### DIFF
--- a/server/pulp/server/webservices/views/decorators.py
+++ b/server/pulp/server/webservices/views/decorators.py
@@ -138,7 +138,6 @@ def _verify_auth(self, operation, super_user_only, method, *args, **kwargs):
     :param super_user_only: Only authorize a user if they are a super user.
     """
     # Check Authentication
-
     # Run through each registered and enabled auth function
     is_consumer = False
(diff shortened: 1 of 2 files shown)


dkliban:
ok test

server/pulp/server/webservices/views/decorators.py
```diff
@@ -167,12 +166,15 @@ def _verify_auth(self, operation, super_user_only, method, *args, **kwargs):
 
     principal_manager = factory.principal_manager()
 
-    user = model.User.objects.get_or_404(login=login)
-    if super_user_only and not user.is_superuser():
-        raise PulpCodedAuthenticationException(error_code=error_codes.PLP0026, user=login,
-                                               operation=OPERATION_NAMES[operation])
+    # Consumers are not part of the User collection
+    if not is_consumer:
+        user = model.User.objects.get(login=login)
```
asmacdo: Ah, right. A consumer wouldn't have a login.

PR #2286, opened 2015-12-21
Update the lazy-content branch based on the PR review.

jeremycline:
This commit addresses some of the issues brought up during the
lazy-content branch review. Others have been fixed in a separate PR,
while some have been written as tasks for the future.

### DIFF
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -11,7 +11,6 @@ from twisted.web import resource
 from twisted.web.server import NOT_DONE_YET
 
 from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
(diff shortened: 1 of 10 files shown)


streamer/pulp/streamer/server.py
```diff
@@ -84,14 +85,18 @@ class Streamer(resource.Resource):
     """
     Define the web resource that streams content from the upstream repository
     to the client.
-
-    :ivar config: The configuration for this streamer instance.
-    :type config: ConfigParser.SafeConfigParser
```
jortel: Why was removed?

jeremycline: It was moved down to the ``__init__`` block since mhrivnak wanted that doc-blocked.

PR #2287, opened 2015-12-21
Fix unit test failures on EL5/6 for lazy-content.

jeremycline:
Several unit tests are failing on lazy-content due to features missing
from Python 2.4 or 2.6.

### DIFF
--- a/client_lib/pulp/client/commands/repo/importer_config.py
+++ b/client_lib/pulp/client/commands/repo/importer_config.py
@@ -98,10 +98,10 @@ class OptionsBundle(object):
             parse_func=parsers.pulp_parse_optional_nonnegative_int
         )
 
-        d = _('content downloading policy ({m})'.format(
-            m=' | '.join((constants.DOWNLOAD_IMMEDIATE,
-                          constants.DOWNLOAD_BACKGROUND,
-                          constants.DOWNLOAD_ON_DEMAND))))
+        methods = ' | '.join((constants.DOWNLOAD_IMMEDIATE,
+                              constants.DOWNLOAD_BACKGROUND,
+                              constants.DOWNLOAD_ON_DEMAND))
+        d = _('content downloading policy (%(method)s)' % {'method': methods})
         self.opt_download_policy = PulpCliOption(
             '--download-policy',
             d,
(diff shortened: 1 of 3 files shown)


jeremycline:
okay test

jeremycline:
ok test

streamer/test/unit/streamer/test_config.py
```diff
@@ -1,4 +1,4 @@
-from unittest import TestCase
+from unittest2 import TestCase
```
mhrivnak: ``from pulp.common.compat import unittest`` will do the "right thing"

jeremycline: Ah, I looked around in there, but didn't noticed it because I looked for the import at the top. Thanks!

client_lib/pulp/client/commands/repo/importer_config.py
```diff
@@ -98,10 +98,10 @@ def __init__(self):
...
         )
 
-        d = _('content downloading policy ({m})'.format(
-            m=' | '.join((constants.DOWNLOAD_IMMEDIATE,
-                          constants.DOWNLOAD_BACKGROUND,
-                          constants.DOWNLOAD_ON_DEMAND))))
+        methods = ' | '.join((constants.DOWNLOAD_IMMEDIATE,
+                              constants.DOWNLOAD_BACKGROUND,
+                              constants.DOWNLOAD_ON_DEMAND))
+        d = _('content downloading policy (%(method)s)' % {'method': methods})
```
jortel: pretty sure I've seen str.format() used in lots of other parts of sever code.  why an issue here?

streamer/test/unit/streamer/test_server.py
```diff
@@ -13,7 +13,7 @@
 MODULE_PREFIX = 'pulp.streamer.server.'
 
 
-class TestStreamerListener(TestCase):
+class TestStreamerListener(unittest.TestCase):
```
jortel: Just curious, what's driving this change?

jeremycline: Oh, there were several asserts I was using (`assertIn`, `assertIs`) which I thought I could use in rhel6 with our compat module. However, that was not the case. I updated them to `assertTrue` in my second test PR.

PR #2290, opened 2015-12-22
Changing mod_xsendfile required version

pcreech:
Modifying mod_xsendfile required version 

closes #[1465](https://pulp.plan.io/issues/1465)


jeremycline:
The Jenkies failure is a known problem and is not related to this change.

PR #2292, opened 2015-12-22
fix regex to remove content and empty directories on orphan cleanup

asmacdo:
https://pulp.plan.io/issues/1238

This also adds a script into playpen that can be run to manually clean up any remaining directories that were left behind by orphan cleanup in pulp < 2.8.0


asmacdo:
ok test

seandst:
ok test

mhrivnak:
I suggest rebasing on current master to see if the tests will pass.

PR #2293, opened 2015-12-23
Create tasks to download lazy repository content.

jeremycline:
This introduces two new tasks, one to download deferred content units,
and one to download a repository using the lazy catalog.

closes #1199

### DIFF
--- a/server/pulp/server/controllers/content.py
+++ b/server/pulp/server/controllers/content.py
@@ -7,41 +7,41 @@ from urlparse import urlunsplit
 
 import celery
 from mongoengine import DoesNotExist
+from nectar.config import DownloaderConfig
 from nectar.request import DownloadRequest
-from nectar.listener import AggregatingEventListener
+from nectar.downloaders.threaded import HTTPThreadedDownloader, SkipLocation
+from nectar.listener import DownloadEventListener
 
 from pulp.common import error_codes
+from pulp.common import tags as pulp_tags
 from pulp.common.plugins import reporting_constants
-from pulp.common.tags import resource_tag, action_tag, RESOURCE_CONTENT_UNIT_TYPE
-from pulp.plugins.conduits.mixins import (ContentSourcesConduitException, StatusMixin,
+from pulp.common.config import parse_bool, Unparsable
+from pulp.plugins.conduits.mixins import (ContentSourcesConduitException,
+                                          LazyStatusConduitException, StatusMixin,
                                           PublishReportMixin)
 from pulp.plugins.loader import api as plugin_api
-from pulp.plugins.loader.exceptions import PluginNotFound
 from pulp.plugins.util.publish_step import Step
+from pulp.plugins.util.verification import (InvalidChecksumType, VerificationException,
+                                            verify_checksum)
 from pulp.server.async.tasks import Task
 from pulp.server.config import config as pulp_conf
 from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
-from pulp.server.controllers import repository
 from pulp.server.content.sources.container import ContentContainer
+from pulp.server.content.sources.constants import MAX_CONCURRENT, HEADERS, SSL_VALIDATION
+from pulp.server.content.storage import FileStorage, mkdir
+from pulp.server.controllers import repository as repo_controller
+from pulp.server.db.model import LazyCatalogEntry, DeferredDownload
 from pulp.server.exceptions import PulpCodedTaskException
 from pulp.server.lazy import URL, Key
 from pulp.server.managers.repo._common import get_working_directory
 
-
 _logger = getLogger(__name__)
 
-URL_SIGNING_KEY = Key.load(pulp_conf.get('authentication', 'rsa_pub'))
-
-
-class ModelNotFound(Exception):
-    """
-    Raised when a content unit Model was not found.
-    """
-    def __init__(self, model_type):
-        self.model_type = model_type
-
-    def __str__(self):
-        return _('Model type "{type}" not found.').format(type=self.model_type)
+PATH_DOWNLOADED = 'downloaded'
+CATALOG_ENTRY = 'catalog_entry'
+UNIT_ID = 'unit_id'
+TYPE_ID = 'type_id'
+UNIT_FILES = 'unit_files'
 
 
 class ContentSourcesConduit(StatusMixin, PublishReportMixin):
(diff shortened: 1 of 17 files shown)


server/pulp/server/controllers/content.py
```diff
@@ -128,113 +137,392 @@ def refresh_content_source(content_source_id=None):
...
-    except PluginNotFound:
-        msg = _('Download of {url} failed - plugin not found')
-        _logger.info(msg.format(url=catalog_entry.url))
-    except DoesNotExist:
-        msg = _('Download of {url} failed - unit not found')
-        _logger.info(msg.format(url=catalog_entry.url))
-    except ModelNotFound, e:
-        msg = _('Download {url}, failed - {reason}')
-        _logger.info(msg.format(url=catalog_entry.url, reason=str(e)))
+    tags = [pulp_tags.action_tag(pulp_tags.ACTION_DEFERRED_DOWNLOADS_TYPE)]
```
jeremycline: I feel like the tags on both these new tasks are not the best so I welcome other suggestions. Maybe this one should carry the 'schedule' resource tag?

jeremycline:
The tests have only been partially written for this PR

server/pulp/server/controllers/content.py
```diff
@@ -128,113 +137,392 @@ def refresh_content_source(content_source_id=None):
...
+            unit_files[temporary_destination] = {
+                CATALOG_ENTRY: catalog_entry,
+                PATH_DOWNLOADED: None,
+            }
+            request = DownloadRequest(signed_url, temporary_destination)
+            request.data = {
+                CONTENT_UNIT: content_unit,
+                UNIT_FILES: unit_files,
+            }
+            requests.append(request)
```
mhrivnak: This is a great opportunity for making a generator and yielding requests one at a time. Each iteration of this loop is relatively expensive, and you can imagine that it might take substantial time if there are tens of thousands of content units to work through. That could also take a lot of RAM, especially for RPMs.

server/pulp/server/controllers/content.py
```diff
@@ -128,113 +137,392 @@ def refresh_content_source(content_source_id=None):
...
+            MAX_CONCURRENT: int(pulp_conf.get('lazy', 'download_concurrency')),
+            HEADERS: {PULP_STREAM_REQUEST_HEADER: 'true'},
+            SSL_VALIDATION: True
+        }
+        self.downloader = HTTPThreadedDownloader(
+            DownloaderConfig(**self.download_config),
+            self
+        )
+
+    def _process_block(self, item=None):
```
mhrivnak: The new in-line docs in the step module help clarify things, and probably didn't exist when you wrote this.

This method shouldn't be overridden, and instead you should override the ``process_main()`` method to do this work.

server/pulp/server/controllers/content.py
```diff
@@ -128,113 +137,392 @@ def refresh_content_source(content_source_id=None):
...
+        """
+        The total number of download requests so progress reporting occurs at
+        the file level.
+
+        Inherited from Step.
+
+        :return: The number of download requests this step will process.
+        :rtype:  int
+        """
+        return len(self.download_requests)
```
mhrivnak: I see that this will be difficult if ``self.download_requests`` is a generator. But the memory implications alone of this not being a generator are I think worth the cost of not reporting a denominator in the progress. I wonder if anything would complain if this returned 0 or -1? Or ``None``?

Otherwise, I suggest storing the number of files on the ``DeferredDownload`` model, allowing you to quickly add them up with a relatively inexpensive DB call.

server/pulp/server/controllers/content.py
```diff
@@ -128,113 +137,392 @@ def refresh_content_source(content_source_id=None):
...
+        self.report_progress()
+
+        # Mark the entire unit as downloaded, if necessary.
+        download_flags = [entry[PATH_DOWNLOADED] for entry in
+                          report.data[UNIT_FILES].values()]
+        if all(download_flags):
+            _logger.debug(_('Marking content unit {type}:{id} as downloaded.').format(
+                type=content_unit.type_id, id=content_unit.id))
+            content_unit.downloaded = True
+            content_unit.save()
```
mhrivnak: It's better to just set that attribute. This is probably a cheaper operation, and it will help avoid race conditions. Something like:

```
content_unit.__class__.objects.filter(id=content_unit.id).update_one(set__downloaded=True)
```

server/pulp/server/controllers/content.py
mhrivnak: Actually, do you need the entire unit on the request? It looks like the task just uses the unit's ``id`` and ``type_id``.

jeremycline: Initially I had this, but ``list_files`` on platform makes use of ``_storage_path`` and in the rpm plugin that relies on ``files`` being available. I wouldn't object to loading only a subset, but it would mean changing things when a plugin gets added and needs more to be loaded to determine its files.

server/pulp/server/controllers/content.py
jeremycline: After an in-person talk with mhrivnak, he's okay with this.

server/pulp/server/controllers/content.py
jeremycline: After an in-person talk we decided it would be a good first step to have the CONTENT_UNIT key to hold the unit_id and type_id only, and to free the content_unit object (del or garbage collection).

server/pulp/server/controllers/content.py
jeremycline: This method uses several methods defined on the content unit. I see your point about race conditions, but it looks like under the hood, MongoEngine is doing the operation you suggested[0], and we're only modifying ``downloaded`` and ``last_modified`` on multi-file units. I can't think if a situation where this will be problematic (although that doesn't mean it doesn't exist, obviously).

One thing that could be done is to reload the content unit from the database in this method from the ``unit_id`` and ``type_id`` saved in the ``DownloadRequest.data`` dict (which was suggested above to avoid holding the entire list of content units in memory). 

[0] http://docs.mongoengine.org/apireference.html#mongoengine.Document.save

server/pulp/server/controllers/content.py
mhrivnak: Thoughts? I'm ok with waiting to test this feature with tens of thousands of requests to see how much memory it uses before going to the trouble of optimizing it, but we need to either have that on the radar, or do the generator conversion.

jeremycline: I've removed the ContentUnit reference in each download request. My preference is to do some performance profiling on it to see if it's going to be troublesome. If we make it a generator we can get away from using the steps entirely (not unappealing), but we'll either need to be okay with not reporting progress, or coming up with a good progress reporting format/method for tasks with an unknown number of items.

jeremycline: I've filed an issue to track this: https://pulp.plan.io/issues/1468

PR #2294, opened 2016-01-04
prevent step from incrementing progress beyond total

mhrivnak:


seandst:
ok test

seandst:
This change is disappointing, in that it prevents pulp from giving 110%...

This looks good, but (assuming you actually saw this happening), is there a bug to be filed about something inappropriately incrementing the counter?

seandst:
ok test

mhrivnak:
To my knowledge, this hasn't caused a bug seen in a pulp release, but only on master. And the inappropriate incrementation was caused by this very class overloading the role of method ``process_main``, which made it challenging to auto-increment when appropriate to do so, but not when it's not.

More concretely, in a case where a subclass just wanted to do a bunch of work in the ``process_main`` method, and handle incrementing on its own, when the work was all done, this method would add one extra "success". The logic was originally intended for a case where the ``process_main`` method does something simple, like download a single file, and then it would theoretically be helpful to auto-increment that one success.

PR #2295, opened 2016-01-05
Adds screen to core ansible role

bmbouter:


jeremycline:
Something something tmux. I'll accept screen, though.

bmbouter:
@jeremycline thanks!

PR #2296, opened 2016-01-05
Ansible dev role now pip installs httpie

bmbouter:


rbarlow:
Let's use the RPM instead of pip installing it:

```
$ dnf info httpie
Last metadata expiration check performed 0:00:10 ago on Tue Jan  5 11:33:32 2016.
Available Packages
Name        : httpie
Arch        : noarch
Epoch       : 0
Version     : 0.9.2
Release     : 2.fc23
Size        : 128 k
Repo        : fedora
Summary     : A Curl-like tool for humans
URL         : http://httpie.org/
License     : BSD
Description : HTTPie is a CLI HTTP utility built out of frustration with existing tools. The
            : goal is to make CLI interaction with HTTP-based services as human-friendly as
            : possible.
            : 
            : HTTPie does so by providing an http command that allows for issuing arbitrary
            : HTTP requests using a simple and natural syntax and displaying colorized
            : responses.
```

seandst:
yessssssss

rbarlow:
posthumous LGTM

PR #2297, opened 2016-01-05
Fix 'Unknown OpenSSL error' unit test failure on F23.

jeremycline:
Unit tests for the streamer are failing because one or more unit tests
in `server` are making use of OpenSSL and failing to clean the error
stack. This is probably a bug in M2Crypto. This commit simply ensures
the streamer tests run before the server tests.

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -37,8 +37,8 @@ TESTS_ALL_PLATFORMS = [
 TESTS_NON_RHEL5 = [
     'client_admin/test/unit',
     'nodes/test/nodes_tests',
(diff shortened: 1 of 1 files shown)


bmbouter:
This does avoid the exception, but it doesn't solve the root cause. I'm OK to merge it because the problem isn't in the lazy work recently added so its not a new problem.

We have some related issues about test independence, but it still doesn't resolve having dirty state when a test runs after another given test which I'm not sure how to identify without a test runner that shuffles the test order in an organized way.

https://pulp.plan.io/issues/132

PR #2298, opened 2016-01-05
Update Repository Units from platform

pcreech:
Updating repository units from platform, due to them
not being reliably updated by plugins.  The code to do
this lives in platform, so we should be able to utilize it
from there

https://pulp.plan.io/issues/1467
re #[1467](https://pulp.plan.io/issues/1467)

### DIFF
--- a/server/pulp/server/managers/content/upload.py
+++ b/server/pulp/server/managers/content/upload.py
@@ -15,6 +15,7 @@ from pulp.server.async.tasks import Task
 from pulp.server.db import model
 from pulp.server.exceptions import (PulpDataException, MissingResource, PulpExecutionException,
                                     PulpException)
+from pulp.server.controllers import repository as repo_controller
 
 
 logger = logging.getLogger(__name__)
@@ -204,8 +205,11 @@ class ContentUploadManager(object):
 
         # Invoke the importer
         try:
-            return importer_instance.upload_unit(transfer_repo, unit_type_id, unit_key,
-                                                 unit_metadata, file_path, conduit, call_config)
+            result = importer_instance.upload_unit(transfer_repo, unit_type_id, unit_key,
+                                                   unit_metadata, file_path, conduit, call_config)
+            repo_controller.rebuild_content_unit_counts(repo_obj)
+            return result
+
         except PulpException:
             msg = _('Error from the importer while importing uploaded unit to repository [%(r)s]')
             msg = msg % {'r': repo_id}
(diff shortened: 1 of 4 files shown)


bmbouter:
A few things on the commit message. I'm not sure if Redmine cares about `Closes #1467` versus `closes #1467`. Also the issue should stay at modified until the plugin ones are also merged, so consider making the association to the redmine issue using `re #1467` instead. This is per https://pulp.plan.io/issues/103

Also consider putting a link to the issue itself in the commit message. Mine usually have a couplet of lines like:

```
https://pulp.plan.io/issues/1467
re #1467
```

bmbouter:
What about the removal of units? The original issue describes how removal needs to maintain the count also. I'm not sure where it goes, but I expect something like:

`repo_controller.update_unit_count(repo_obj, unit_type_id, -1)`

@pcreech Is that already in the existing code or should it be added with this?

pcreech:
Fixed formatting of the commit message.

import and remove pathways already were updating the unit counts

[remove](https://github.com/pulp/pulp/blob/master/server/pulp/server/managers/repo/unit_association.py#L358)

[import](https://github.com/pulp/pulp/blob/master/server/pulp/server/managers/repo/unit_association.py#L259)

server/pulp/server/managers/content/upload.py
```diff
@@ -204,8 +205,11 @@ def import_uploaded_unit(repo_id, unit_type_id, unit_key, unit_metadata, upload_
 
         # Invoke the importer
         try:
-            return importer_instance.upload_unit(transfer_repo, unit_type_id, unit_key,
-                                                 unit_metadata, file_path, conduit, call_config)
+            result = importer_instance.upload_unit(transfer_repo, unit_type_id, unit_key,
+                                                   unit_metadata, file_path, conduit, call_config)
+            repo_controller.update_unit_count(repo_obj, unit_type_id, 1)
```
mhrivnak: Let's chat about the case where the unit already existed in the repo. I'm not sure if all of the importers would raise an error in that case. If any of them silently let that slide in the name of idempotency, we could end up incorrectly incrementing.

pcreech:
Per conversation in IRC, incrementing the unit count by 1 turned out to not be reliable.  Switched to doing a full rebuild for now, until we can find out a way to report back from plugin what was updated

mhrivnak:
In addition to mocking the controller in the tests to make sure they don't try to do it "for real", we should get a test for each place that asserts the rebuild is being called.

PR #2299, opened 2016-01-06
Add a new REST API endpoint to download lazy repositories.

jeremycline:
This new API is located at ``v2/repositories/<repo-id>/actions/download/``.
This commit also adds support for this API in pulp-admin.

closes #1355

### DIFF
--- a/client_lib/pulp/client/commands/repo/sync_publish.py
+++ b/client_lib/pulp/client/commands/repo/sync_publish.py
@@ -9,6 +9,8 @@ from pulp.client.commands import options, polling
 from pulp.client.extensions.extensions import PulpCliOptionGroup
 from pulp.common import tags
 
+from pulp.client.extensions.extensions import PulpCliFlag
+
 
 # Command Descriptions
 DESC_SYNC_RUN = _('triggers an immediate sync of a repository')
@@ -16,6 +18,14 @@ DESC_SYNC_STATUS = _('displays the status of a repository\'s sync tasks')
 
 DESC_PUBLISH_RUN = _('triggers an immediate publish of a repository')
 DESC_PUBLISH_STATUS = _('displays the status of a repository\'s publish tasks')
+
+DESC_DOWNLOAD = _('queues a full download of all missing units in a repository '
+                  'that is using a "background" or "on_demand" download policy.')
+DESC_VERIFY_ALL_FLAG = _('check all units in the repository for corrupted or '
+                         'missing files and re-download files as necessary '
+                         'rather than just downloading files that are known '
+                         'to be missing.')
+FLAG_VERIFY_ALL = PulpCliFlag('--verify-all', DESC_VERIFY_ALL_FLAG, aliases=['-a'])
 
 
 class StatusRenderer(object):
@@ -285,7 +295,7 @@ class PublishStatusCommand(SyncPublishCommand):
 def _get_repo_tasks(context, repo_id, action):
     """
     Retrieve a list of incomplete Task objects for the given repo_id and action. action must be one
-    of 'sync' or 'publish'.
+    of 'sync', 'download', or 'publish'.
 
     :param context: The CLI context from Okaara
     :type  context: pulp.client.extensions.core.ClientContext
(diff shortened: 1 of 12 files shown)


jeremycline:
ok test

client_lib/pulp/client/commands/repo/sync_publish.py
```diff
@@ -17,6 +19,14 @@
 DESC_PUBLISH_RUN = _('triggers an immediate publish of a repository')
 DESC_PUBLISH_STATUS = _('displays the status of a repository\'s publish tasks')
 
+DESC_DOWNLOAD = _('triggers an immediate download of a repository that is using '
```
mhrivnak: to emphasize that it goes in the task queue with everything else, I'd adjust this to "queues a full download of all units in a repository..."

client_lib/pulp/client/commands/repo/sync_publish.py
```diff
@@ -301,9 +311,72 @@ def _get_repo_tasks(context, repo_id, action):
...
+        :param kwargs: The user inputs
+        :type  kwargs: dict
+        """
+        repo_id = kwargs[options.OPTION_REPO_ID.keyword]
+        background = kwargs[polling.FLAG_BACKGROUND.keyword]
+        verify_all_units = kwargs[FLAG_VERIFY_ALL.keyword]
+
+        self.prompt.render_title(_('Downloading Repository [%(r)s]') % {'r': repo_id})
+
+        # See if an existing publish is running for the repo. If it is, resume
```
mhrivnak: s/publish/download task/

jeremycline:
I've filed https://pulp.plan.io/issues/1481 which will fix a problem with the lazy tasks not aborting downloads if the file is already present.

server/pulp/server/webservices/views/repositories.py
```diff
@@ -1069,6 +1070,40 @@ def post(self, request, repo_id):
...
+    """
+    View for downloading a lazy (background or on-demand download method) repository.
+    """
+
+    @auth_required(authorization.EXECUTE)
+    @json_body_allow_empty
+    def post(self, request, repo_id):
+        """
+        Dispatch a task to publish a repository. The JSON body may contain a key,
+        `force`, that forces the task to attempt to download all content units again
```
mhrivnak: should this be talking about ``verify_all_units`` instead of ``force`` ?

server/test/unit/server/webservices/views/test_repositories.py
```diff
@@ -1719,6 +1720,99 @@ def test_post_publish_repo_missing_distributor(self, mock_repo_qs, mock_factory)
...
+                new=assert_auth_EXECUTE())
+    @mock.patch('pulp.server.webservices.views.repositories.content_controller')
+    @mock.patch('pulp.server.webservices.views.repositories.model.Repository.objects')
+    def test_post_download_repo(self, mock_repo_qs, mock_content_controller):
+        """Test that a repo download task is dispatched."""
+        # Setup
+        download_repo = RepoDownload()
+
+        # Tests
+        try:
```
mhrivnak: You can use ``assertRaises`` as a context manager, which would improve readability. You probably need to use ``unittest2`` for that, in which case you can just import ``unittest`` from our ``compat`` module. It goes something like:

```
with self.assertRaises(exception.OperationPostponed) as exception:
  do_thing()

# now make assertions about the exception
self.assertEqual(exception.http_status_code, 202)
```

server/test/unit/server/controllers/test_content.py
```diff
@@ -157,13 +158,25 @@ class TestDownloadRepo(TestCase):
...
-    def test_download_deferred(self, mock_missing_units, mock_create_requests, mock_step):
-        """Assert the download step is initialized and called."""
+    def test_download_repo_no_verify(self, mock_missing_units, mock_create_requests, mock_step):
+        """Assert the download step is initialized and called with missing units."""
         content_controller.download_repo('fake-id')
         mock_missing_units.assert_called_once_with('fake-id')
         mock_create_requests.assert_called_once_with(mock_missing_units.return_value)
         mock_step.return_value.process_lifecycle.assert_called_once_with()
 
+    @patch(MODULE_PATH + 'itertools.chain')
```
mhrivnak: As you've written it, this is definitely the most concise way to do the test, which is valuable. Mocking ``chain`` raised an eyebrow though. I think tests are more valuable when they can make assertions that desired behaviors happened, not that specific implementations were executed. In a case like this, I prefer to assign a more representative return value to ``mock_units_qs``, and then assert that the expected transformation of that was passed into ``mock_create_requests``.

More concretely, something along these lines:

```
qs1 = mock.Mock()
qs2 = mock.Mock()
mock_units_qs.return_value = [[qs1], [qs2]]
...
self.assertEqual(list(mock_create_requests.call_args[0][0]), [qs1, qs2])
```

It's a small difference especially in this case, but the closer we can stick to letting representative data flow through our unit tests, the more valuable those tests will be.

server/test/unit/server/webservices/views/test_repositories.py
jeremycline: I think my reasoning was that all the other tests were using the old style so it'd be better to match existing tests. However, I'd much prefer the context manager method and since you suggested it, I'll happily change it.

jeremycline:
ok test

jeremycline:
ok test

jeremycline:
ok test

jeremycline:
Jenkies seems to be having trouble installed a few dependencies, but I think over the last three runs each platform has passed at least once. I'm going to go ahead and merge this.

PR #2305, opened 2016-01-10
Add a shebang to the selinux relabel.sh script.

rbarlow:


server/selinux/server/relabel.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/bash
```
seandst: On fedora, maybe...

On debian:
```
$ which bash
/bin/bash
```

Any reason not to `/usr/bin/env bash` here?

rbarlow: Nooooope. Fixed.

PR #2306, opened 2016-01-10
Remove unused pulp.server.managers.pluginwrapper.

rbarlow:


seandst:
supergrep confirms no plugins that I know of use this. old-style classes for the win!

PR #2318, opened 2016-01-11
Dispatch a download task after a sync if policy is 'background'.

jeremycline:
This PR is large because the task and its tests need to be moved to avoid import trouble, and steps had to be removed from the task.

closes #1445


server/pulp/server/controllers/repository.py
```diff
@@ -757,6 +782,9 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
         raise pulp_exceptions.PulpExecutionException(_('Importer indicated a failed response'))
 
     spawned_tasks = _queue_auto_publish_tasks(repo_obj.repo_id, scheduled_call_id=scheduled_call_id)
+    download_policy = call_config.get(importer_constants.DOWNLOAD_POLICY)
```
jeremycline: This bit here is what closes issue 1445

server/pulp/server/controllers/repository.py
```diff
@@ -1007,3 +1035,423 @@ def _now_timestamp():
...
+    port = pulp_conf.get('lazy', 'redirect_port')
+    path_prefix = pulp_conf.get('lazy', 'redirect_path')
+    netloc = (host + ':' + port) if port else host
+    path = os.path.join(path_prefix, catalog_entry.path.lstrip('/'))
+    unsigned_url = urlunsplit((retrieval_scheme, netloc, path, None, None))
+    # Sign the URL for a year to avoid the URL expiring before the task completes
+    return str(URL(unsigned_url).sign(signing_key, expiration=31536000))
+
+
+class LazyUnitDownloadStep(DownloadEventListener):
```
bmbouter: I expected that a class named 'xxxxxStep' would inherit from a step. The original one inherited from `Step` and `DownloadEventListener` why did that change?

bmbouter: After some in-person discussion, `LazyUnitDownloadStep` should be renamed to `LazyUnitDownloader`. I think that would be good to do in this PR.

jeremycline: I filed https://pulp.plan.io/issues/1508 in response to this problem. The whole class needs to be refactored, but since we want a beta in a few minutes...

bmbouter:
Can an issue be filed so that `LazyUnitDownloader` be broken apart into two functional pieces. The `LazyUnitDownloader` which will just handle the downloads, and the stop parts to a Step of some sort. This was described as a next step action, so that should be written up as a task or refactor.

bmbouter:
@jeremycline I left some feedback; handle as you see fit.

PR #2321, opened 2016-01-11
ref #1487 - update node conduit to handle both legacy and mongoengine units.

jortel:
https://pulp.plan.io/issues/1487

This conduit was written long before the platform provided generator based methods for fetching content units.  Also, the conduit manipulates the data a little.  And, the `UnitIterator` implements `__len__` which is used to reporting progress.  Long term, I would advocate that nodes use the `SingleRepoUnitsMixin` and find a way to deal with the lack of `__len__()` support.  But, given the future of nodes, I'm just fixing what's here.

I also added some missing docstrings and renamed some variables to make the code clearer.

### DIFF
--- a/nodes/common/pulp_node/conduit.py
+++ b/nodes/common/pulp_node/conduit.py
@@ -1,11 +1,14 @@
-from pulp.plugins.types import database as types_db
+from pulp.plugins.types.database import type_units_collection
+from pulp.plugins.util.misc import paginate
+from pulp.server.controllers.units import get_unit_key_fields_for_type
 from pulp.server.db.model.repository import RepoContentUnit
 from pulp.server.config import config as pulp_conf
 
 
 class NodesConduit(object):
 
-    def get_units(self, repo_id):
+    @staticmethod
+    def get_units(repo_id):
         """
         Get all units associated with a repository.
         :param repo_id: The repository ID used to query the units.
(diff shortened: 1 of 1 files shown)


jortel:
ok test

nodes/common/pulp_node/conduit.py
```diff
@@ -13,77 +16,99 @@ def get_units(self, repo_id):
...
-            unit_list = types.setdefault(type_id, [])
-            unit_list.append(unit['unit_id'])
-        return UnitsIterator(units, types)
-
-
-class Typedef(object):
-
-    def __init__(self):
-        self.cached = {}
+        for association in collection.find({'repo_id': repo_id}):
```
bmbouter: Can this loop be redone using mongoengine based queries? We shouldn't continue to use direct pymongo based code anymore since the collection being accessed is converted to mongoengine. I bet this whole thing would become perhaps one or two lines.

bmbouter:
I have a similar request for the rest of the code. Can it be converted to use mongoengine based querying. I can point out some examples of how to convert each section if that is helpful to make resolving this straightforward.

bmbouter:
FYI, a side-effect of having the `ref #` in the title of the commit message, two commits will be associated with the issue. This is because the merge commit contains the title of the commit being merged. This creates confusion on the issue about which commit is the one downstream should pick. This commit could be ammended to fix that or issue cleanup can be done after the merge.

nodes/common/pulp_node/conduit.py
jortel: It could be but I question the return on investment given the future of nodes.

jortel:
I appreciate your suggestion here but given the future of nodes, I do not think that effort beyond the minimum necessary to keep it running is a good return on investment.  Do you?  In any case, I think converting nodes to use mongoengine models falls outside the scope of this issue/PR.  If we decide it is worth the effort, I think we should open another issue to do this work.

bmbouter:
After thinking if over some, I agree that converting this to mongoengine should not be done because we're removing nodes soon anyway. It's not worth the effort.

I read through the code twice and it looks right, but since it's a bunch of dictionary operations, I don't actually know if it *is* right. If it works for you, then go ahead and merge it.

PR #2325, opened 2016-01-12
Add flag to filter for units inheriting from FileContentUnit.

jeremycline:
This is because not all units have a downloaded flag anymore, and we
don't want them for lazy-related queries.

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -127,7 +127,7 @@ def get_repo_unit_models(repo_id):
     return filter(None, unit_models)
 
 
-def get_mongoengine_unit_querysets(repo_id, repo_content_unit_q=None):
+def get_mongoengine_unit_querysets(repo_id, repo_content_unit_q=None, file_units=False):
     """
     Retrieve an iterable of QuerySets for all the units in a repository that have
     MongoEngine models. If a unit type is in the repository and does not have a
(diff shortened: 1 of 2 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -138,11 +138,17 @@ def get_mongoengine_unit_querysets(repo_id, repo_content_unit_q=None):
...
+    :param file_units:          Retrieve QuerySets exclusively for units inheriting
+                                from pulp.server.db.model.FileContentUnit.
+    :type  file_units:          bool
 
     :return: A generator of query sets.
     :rtype:  generator of mongoengine.queryset.QuerySet
     """
     unit_models = get_repo_unit_models(repo_id)
+    if file_units:
+        unit_models = filter(lambda m: 'FileContentUnit' in m._superclasses, unit_models)
```
bmbouter: Let's do this without accessing `_superclasses`. This will work right?

filter(lambda m: isinstance(m, FileContentUnit), unit_models)

bmbouter:
Once the filter statement is rewritten, LGTM

server/pulp/server/controllers/repository.py
jeremycline: Nope. I tried that initially:
```
(Pdb) p rpm_model
<class 'pulp_rpm.plugins.db.models.RPM'>
(Pdb) p rpm_model.__class__
<class 'mongoengine.base.metaclasses.TopLevelDocumentMetaclass'>
```

bmbouter:
Please add a test covering this `if` statement. There are already existing tests so one should be added for this one.

bmbouter:
Once the filter statement is updated and a test is added it can be merged.

server/pulp/server/controllers/repository.py
mhrivnak: I'm seeing this:

```
>>> from pulp_rpm.plugins.db.models import RPM
>>> RPM
<class 'pulp_rpm.plugins.db.models.RPM'>
>>> from pulp.server.db.model import ContentUnit
>>> issubclass(RPM, ContentUnit)
True
>>> isinstance(RPM(), ContentUnit)
True
```

jeremycline:
```
In [1]: import mongoengine

In [2]: from pulp.server.db.connection import initialize as mongo_initialize

In [3]: from pulp.server.db import model

In [4]: from pulp.plugins.loader import api as plugin_api

In [5]: mongo_initialize()

In [6]: mongoengine.connect('pulp_database')
Out[6]: MongoClient('localhost', 27017)

In [7]: plugin_api.initialize()

In [8]: rpm_plugin = plugin_api.get_unit_model_by_id('rpm')

In [9]: print(isinstance(rpm_plugin, model.FileContentUnit))
False

In [10]: print(isinstance(rpm_plugin, model.ContentUnit))
False
```

PR #1496, opened 2015-01-09
Fixing typo in docs

AndreaGiardini:

PR #2327, opened 2016-01-13
Move contents of /srv/pulp to /usr/share/pulp/srv.

rbarlow:
It is an error for an RPM package to install content into /srv as
this path is reserved for the local administrator.

https://pulp.plan.io/issues/1496

fixes #1496


server/selinux/server/relabel.sh
```diff
@@ -21,6 +20,7 @@ fi
 # If upgrading from before 2.7.0
 if [[ $1 < '2.7.0' ]]
 then
+    /sbin/restorecon -i -R /usr/share/pulp/srv
```
bmbouter: This should get moved to its own if block which is guarded at '2.8.0'. Otherwise it won't run when users upgrade from 2.7.0 -> 2.8.0. This is required for correctness

bmbouter:
I think this needs a release note since the files moving and users may have come to expect them to the old place. I'll suggest that is also a required change.

bmbouter:
@rbarlow LGTM with the release note added and the selinux block added for '2.8.0'.

seandst:
Out of curiosity, why `/usr/share/pulp/srv/`, and not `/usr/share/pulp/` or `/usr/share/pulp/wsgi/`?

server/selinux/server/relabel.sh
rbarlow: Ah, yes. d'oh!

rbarlow:
@seandst it's arbirary to me. I'm not sure if we might want to put other things in /usr/share/pulp in the future, so I think it's a good idea to at least put it in a subdir. Do you prefer wsgi?

seandst:
Yeah, I think I prefer wsgi.

bmbouter:
+1 to wsgi

rbarlow:
I have made the requested changes.

I will need to make a few changes in some plugins too, so I will hold off on merging this until they are also LGTM'd.

rbarlow:
This and these other pull requests must all be merged at the same time:

https://github.com/pulp/pulp_ostree/pull/58
https://github.com/pulp/pulp_puppet/pull/210
https://github.com/pulp/pulp_rpm/pull/769

rbarlow:
ok test

rbarlow:
ok test

PR #2328, opened 2016-01-13
Ensures setting an importer serializes into the task result

bmbouter:
The importer conversion introduced a regression where when you
update an importer it saved a reference to the new importer
itself in the TaskStatus instead of a serializer verison of it.

When that TaskStatus was fetched, it could not be serialized
by the webservices layer and caused a 500 exception.

https://pulp.plan.io/issues/1483
closes #1483


asmacdo:
Needs the return type to be updated in the docblock. I am guessing also that some tests will need to mock the serializer and modify some assertions.

parthaa:
+1 . Seemed to work with the runcible tests...

bmbouter:
@asmacdo I pushed an update which fixes the doc block, adds test assertions relevant to the new code, and updates the tests so they all pass. Thanks for the review. I'm merging now.

PR #2329, opened 2016-01-14
1356 - Pulp returns HTTP 500 when a bad request is made during repo c…

goosemania:
…reation.

closes #1356
https://pulp.plan.io/issues/1356

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -156,3 +156,5 @@ PLP1013 = Error("PLP1013", _("Checksum does not match calculated value."), [])
 PLP1014 = Error("PLP1014", _("The configuration value for the key '%(key)s' in "
                              "section '%(section)s' is not valid for the following "
                              "reason: %(reason)s"), ["key", "section", "reason"])
+PLP1015 = Error("PLP1015", _("The JSON data must be of type '%(data_type)s'."),
+                ['data_type'])
--- a/server/pulp/server/webservices/views/util.py
+++ b/server/pulp/server/webservices/views/util.py
@@ -93,25 +93,43 @@ def _ensure_input_encoding(input):
         raise InputEncodingError(input), None, sys.exc_info()[2]
 
 
-def json_body_required(func, allow_empty=False):
+def parse_json_body(allow_empty=False, json_type=None):
+    """
+    Ensure request body carries valid JSON of specified data type
 
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        request = args[1]
-        if allow_empty and not request.body:
-            request.body_as_json = {}
+    :param allow_empty: if True, allows request body to be empty, defaults to False.
+    :type  allow_empty: bool
+    :param json_type: required type of request body. If None, any type is allowed,
+                      defaults to None.
+    :type  json_type: type
+
+    :raises PulpCodedValidationException: if request body contains invalid JSON or
+                                          of incorrect data type.
+
+    :return: decorator with proper parameters applied.
+    :rtype: function
+    """
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            request = args[1]
+            if allow_empty and not request.body:
+                request.body_as_json = {}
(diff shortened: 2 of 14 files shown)


asmacdo:
Very nice, this adds some much needed robustness to the API. Rather than creating a new decorator, how would you feel about making the required type an optional argument for the `@json_body_required` decorator?

common/pulp/common/error_codes.py
```diff
@@ -156,3 +156,5 @@
 PLP1014 = Error("PLP1014", _("The configuration value for the key '%(key)s' in "
                              "section '%(section)s' is not valid for the following "
                              "reason: %(reason)s"), ["key", "section", "reason"])
+PLP1015 = Error("PLP1015", _("The required type of JSON data is '%(data_type)s'."),
```
bmbouter: Consider rewording `The required type of JSON data is` => `The JSON data must be of type `

bmbouter:
+1 to this much needed robustness! Great first PR. :palm_tree: :cake: :candy:

goosemania:
@asmacdo
Probably it is better to add a new parameter, I was a little misled by the second decorator which is a partial of the basic one (did not want to create lots of them :)). 

    json_body_allow_empty = functools.partial(json_body_required, allow_empty=True)

So is it better to have only one decorator with two optional params,  or because the allow_empty parameter is boolean it is better to leave it as is? What do you think?

    @json_body_required(allow_empty=True, json_type=dict)
or

    @json_body_allow_empty(json_type=dict)
    and
    @json_body_required(json_type=dict)

goosemania:
ok test

goosemania:
@bmbouter 
Thanks! :sparkles:

asmacdo:
If I remember correctly, I was concerned that `@json_body_required(allow_empty=True)` might be confusing. "Is the body required or not?" 

If you prefer a single decorator that's fine with me. The potential confusion really is caused by a poor choice of decorator name to begin with. This looks pretty clear to me, what do you think?

    @parse_json_body(allow_empty=True, json_type=dict)

server/pulp/server/webservices/views/util.py
```diff
@@ -114,6 +114,21 @@ def wrapper(*args, **kwargs):
 json_body_allow_empty = functools.partial(json_body_required, allow_empty=True)
 
 
+def json_type_required(json_type):
```
mhrivnak: need a doc block. See the one below for an example. It needs to have a brief description of what this function is supposed to do or why it exists, then describe the argument and declare its type.

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -167,6 +168,45 @@ def mock_function(self, request):
         self.assertEqual(response.error_code.code, 'PLP1009')
 
 
+class TestJsonTypeRequired(unittest.TestCase):
```
mhrivnak: Great tests!

goosemania:
ok test

server/test/unit/server/webservices/views/test_util.py
```diff
@@ -166,6 +187,63 @@ def mock_function(self, request):
...
+        Test that incorrect type of JSON results in a 400.
+        """
+        mock_request = mock.MagicMock()
+        mock_request.body = json.dumps(["unexpected datatype"])
+
+        @parse_json_body(json_type=dict)
+        def mock_function(self, request):
+            return request
+
+        with self.assertRaises(PulpCodedValidationException) as cm:
```
mhrivnak: This failed in the jenkins run because on python 2.6, it doesn't have the context manager capability. We handle that sort of problem by using ``unittest2``, which is a backport of the fancy new test tools to python 2.6. You can do this to get the "right" one for your platform:

``from pulp.common.compat import unittest``

goosemania:
ok test

goosemania:
ok test

server/test/unit/server/webservices/views/test_util.py
goosemania: Thank you for the hint. Fixed.

PR #2332, opened 2016-01-18
Fix unit test to use ISO8601 date string

pcreech:
Unit tests had invalid ISO8601 date string, causing
tests to fail under mongoengine-0.10

fixes #[1495](https://pulp.plan.io/issues/1495)
https://pulp.plan.io/issues/1495


seandst:
indicated jenkins failure is due to an issue in jenkins having trouble spawning nodes, tests passed on all distros with this change

lgtm

PR #2335, opened 2016-01-19
1541 - Retrieve bindings by consumer and repository call fails

goosemania:
closes #1541
https://pulp.plan.io/issues/1541


ipanova:
ok test

PR #2336, opened 2016-01-19
Remove the requestor's IP address from the signature.

jeremycline:
I encountered problems with this because one request was being made via
IPv6 and a second request was being made via IPv4.

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -91,7 +91,6 @@ class ContentView(View):
         host = request.environ['SERVER_NAME']
         port = request.environ['SERVER_PORT']
         query = request.environ['QUERY_STRING']
-        remote_ip = request.environ['REMOTE_ADDR']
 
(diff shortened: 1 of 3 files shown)


jeremycline:
ok test

dkliban:
ok test

dkliban:
This looks good to me. However, I am concerned that we are removing the IP address. The consequence is that the redirect URL can be used by anyone that sniffs HTTP traffic.

jeremycline:
So I've given this some thought. It's unlikely that someone would run into the issue I ran into in the real world, but it could happen. If, for example, you make your request to http://192.168.1.100/repo/some.rpm and then get redirected to http://hostname/content/path/with/signature/ where hostname resolves to an IPv6 address, you will receive a 403.

I'm on the fence about whether or not we should merge this.

jeremycline:
After some discussion it was decided that it would be best to just add some notes to the documentation on how to avoid this problem.

PR #2337, opened 2016-01-19
Load `_storage_path` when reloading the content unit.

jeremycline:
`_storage_path` is required when saving the unit. The reason this
already worked with single file units is that `set_storage_path` is
called on them and regenerates the _storage_path. This is required for
multi-file units.

closes #1545


dkliban:
ok test

dkliban:
ok test

PR #2338, opened 2016-01-19
Update Pulp to work with the pymongo 3.0.0 API.

rbarlow:
https://pulp.plan.io/issues/1528

fixes #1528 

This pull request does not include the effort needed to provide pymongo >= 3.0.0 in the Pulp RPM repository. That will be done in a separate effort. Until that is done, this pull request will not pass in Jenkies.

https://github.com/pulp/pulp_rpm/pull/775 must be merged before this is merged.

### DIFF
--- a/docs/user-guide/release-notes/master.rst
+++ b/docs/user-guide/release-notes/master.rst
@@ -23,11 +23,12 @@ New Features
 Deprecation
 -----------
 
-Supported Platforms Changes
+Dependency/Platform Changes
 ---------------------------
 
 * If run on CentOS or Red Hat Enterprise Linux, the Pulp server now requires either
   version 7.1+ or 6.7+.
+* pymongo >= 3.0.0 is now required.
 
 Client Changes
 --------------
--- a/pulp.spec
+++ b/pulp.spec
@@ -350,8 +350,8 @@ Requires: python-%{name}-repoauth = %{pulp_version}
 Requires: python-blinker
 Requires: python-celery >= 3.1.0
 Requires: python-celery < 3.2.0
-Requires: python-pymongo >= 2.7.1
-Requires: python-mongoengine >= 0.9.0
+Requires: python-pymongo >= 3.0.0
+Requires: python-mongoengine >= 0.10.0
 Requires: python-setuptools
 Requires: python-oauth2 >= 1.5.211
 Requires: python-httplib2
--- a/server/setup.py
+++ b/server/setup.py
@@ -27,6 +27,6 @@ setup(
     },
     install_requires=[
         'blinker', 'celery >=3.1.0, <3.2.0', DJANGO_REQUIRES, 'httplib2', 'iniparse',
-        'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.7.10', 'oauth2>=1.5.211', 'pymongo>=2.5.2',
+        'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.10.0', 'oauth2>=1.5.211', 'pymongo>=3.0.0',
         'semantic_version>=2.2.0', 'setuptools']
 )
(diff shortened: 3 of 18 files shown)


rbarlow:
It was identified in today's scrum meeting that we need to test the migrations to make sure that they also work with pymongo 3. I did not do that while making this change.

ipanova:
ok test

ipanova:
Handtested with pymongo 3.2 + mongoengine 0.10. Migrations and connections to mongodb performed  with no issues

docs/user-guide/release-notes/master.rst
```diff
@@ -23,11 +23,12 @@ New Features
 Deprecation
 -----------
 
-Supported Platforms Changes
+Dependency/Platform Changes
 ---------------------------
 
 * If run on CentOS or Red Hat Enterprise Linux, the Pulp server now requires either
   version 7.1+ or 6.7+.
+* pymongo >= 3.0.0 is now required.
```
ipanova: can you update this with 3.2 dependency requirement, also in spec

pulp.spec
```diff
@@ -350,7 +350,7 @@ Requires: python-%{name}-repoauth = %{pulp_version}
 Requires: python-blinker
 Requires: python-celery >= 3.1.0
 Requires: python-celery < 3.2.0
-Requires: python-pymongo >= 2.7.1
+Requires: python-pymongo >= 3.0.0
 Requires: python-mongoengine >= 0.9.0
```
ipanova: i guess we will need to update mongoengine requirement >= 0.10.5

server/setup.py
```diff
@@ -27,6 +27,6 @@
     },
     install_requires=[
         'blinker', 'celery >=3.1.0, <3.2.0', DJANGO_REQUIRES, 'httplib2', 'iniparse',
-        'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.7.10', 'oauth2>=1.5.211', 'pymongo>=2.5.2',
+        'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.7.10', 'oauth2>=1.5.211', 'pymongo>=3.0.0',
```
ipanova: update please to pymongo>=3.2

ipanova:
@rbarlow I found several occurrences that have been left during update of the pulp code. Please check if they also have been intended to be updated or you just missed them.
```
$ git grep 'find(fields'
server/pulp/server/db/migrations/0007_scheduled_task_conversion.py:    for importer in importer_collection.find(fields=['scheduled_syncs', 'id', 'repo_id']):
server/pulp/server/db/migrations/0007_scheduled_task_conversion.py:    for distributor in distributor_collection.find(fields=['scheduled_publishes', 'id', 'repo_id']):
server/pulp/server/db/migrations/lib/managers.py:            repo_ids = [repo['id'] for repo in repo_collection.find(fields=['id'])]
[ipanova@ina pulp]$ git grep 'safe=True'
server/pulp/server/db/migrations/lib/managers.py:                                   safe=True)
server/test/unit/server/db/migrations/lib/test_managers.py:            safe=True
```

docs/user-guide/release-notes/master.rst
rbarlow: Technically we don't require 3.2, just 3.0. We are shipping 3.2, but I don't think we need to mention that in our upstream release notes, which I think should target the Python audience more than the repo users.

pulp.spec
rbarlow: Probably to 0.10.0, but yeah good catch!

server/setup.py
rbarlow: I'm going to leave it as it does work with 3.0.0's API.

rbarlow:
Hello @ipanova! First off, I did not know about git grep, so thanks for teaching me about a new tool!

Thanks for catching those remaining changes, I have now fixed them too. I updated the mongoengine dependencies to 0.10.0, and I left pymongo at 3.0.0 as I discussed. Requirements should be used to pin down API incompatibilities, but should otherwise be lax so that we can give an easier task to packagers from other distributions.

PR #2346, opened 2016-01-22
Disables all other celery serializers except JSON

bmbouter:
This removes a scary, multiline warning from celery workers when
they start up

https://pulp.plan.io/issues/23
re #23


bmbouter:
ok test

bmbouter:
ok test

PR #2354, opened 2016-01-23
Serialize content units

asmacdo:


asmacdo:
Should not be used, based on http://pulp.readthedocs.org/en/latest/dev-guide/integration/rest-api/repo/content.html which are apparently incorrect.

PR #2355, opened 2016-01-24
when _id is not an ObjectId, leave it alone

asmacdo:


asmacdo:
https://github.com/pulp/pulp/pull/2358 fixes this issue in a better way.

PR #2358, opened 2016-01-25
Do not attempt to translate an _id if it is not in remapped_fields

asmacdo:


asmacdo:
Waiting on some tests to pass from Partha before merge.

PR #2359, opened 2016-01-25
Share http sessions in content sources.

jortel:
https://pulp.plan.io/issues/1579

Mainly to support reusing http sessions when used with the streamer.


jortel:
ok test

jortel:
ok test

PR #2360, opened 2016-01-25
Do not error in preset if user is not logged in.

rbarlow:


rbarlow:
ok test

rbarlow:
I am confident that this Jenkins error is not related to this change since it is contained in the development .bashrc.

PR #2361, opened 2016-01-25
Adds validation to the UTCDateTimeField for mongoengine

dkliban:
Due to a bug in datetime utils, we can't rely on mongonegine to
convert a string representation of a timestamp with timezone represented
as Z on end. As a result we now require the date to be supplied as
a datetime object with tzinfo set.

https://pulp.plan.io/issues/1565
closes #1565

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -994,7 +994,7 @@ def _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_confi
               % {'r': repo_obj.repo_id}))
         raise
 
-    publish_end_timestamp = _now_timestamp()
+    publish_end_timestamp = _now_timestamp(string=False)
 
     # Reload the distributor in case the scratchpad is set by the plugin
     dist = model.Distributor.objects.get_or_404(repo_id=repo_obj.repo_id, distributor_id=dist_id)
@@ -1046,7 +1046,7 @@ def publish_history(start_date, end_date, repo_id, distributor_id):
     return RepoPublishResult.get_collection().find(search_params)
 
 
-def _now_timestamp():
+def _now_timestamp(string=True):
     """
     Return a current timestamp in iso8601 format.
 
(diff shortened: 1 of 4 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -1046,7 +1046,7 @@ def publish_history(start_date, end_date, repo_id, distributor_id):
     return RepoPublishResult.get_collection().find(search_params)
 
 
-def _now_timestamp():
+def _now_timestamp(string=True):
```
bmbouter: `string` needs `:param:` and `:type:` documented in the docstring

server/pulp/server/controllers/repository.py
```diff
@@ -1054,7 +1054,10 @@ def _now_timestamp():
     :rtype:  str
     """
     now = dateutils.now_utc_datetime_with_tzinfo()
-    now_in_iso_format = dateutils.format_iso8601_datetime(now)
+    if string:
+        now_in_iso_format = dateutils.format_iso8601_datetime(now)
+    else:
+        return now
     return now_in_iso_format
```
bmbouter: Consider making L#1057 - L#1061:

```
   if string:
        return dateutils.format_iso8601_datetime(now)
    else:
        return now
```

server/pulp/server/db/fields.py
```diff
@@ -48,3 +50,16 @@ def to_python(self, value):
...
         ret = super(UTCDateTimeField, self).to_python(value)
         return dateutils.ensure_tz(ret)
+
+    def _validate(self, value, **kwargs):
+        """
+        Ensures that the value is a datetime.datetime object that has tzinfo set.
+
+        :param value: datetime.datetime object
+        :param kwargs: any extra kwargs that should be passed to the super method
+        :raises: mongoengine.errors.ValidationError
```
bmbouter: Consider adding a newline separating the inputs and outputs of the function's docstring.

server/pulp/server/db/fields.py
```diff
@@ -48,3 +50,16 @@ def to_python(self, value):
         """
         ret = super(UTCDateTimeField, self).to_python(value)
         return dateutils.ensure_tz(ret)
+
+    def _validate(self, value, **kwargs):
```
bmbouter: We should use `clean()` instead of `_validate()`. The docs say `clean()` is good for cleaning and pre-validation and it's part of the public API.

http://docs.mongoengine.org/en/latest/guide/document-instances.html#pre-save-data-validation-and-cleaning

bmbouter:
Some changes are needed, but I don't need to re-review. LGTM

PR #2363, opened 2016-01-25
pulp_streamer no longer forwards hop-by-hop HTTP headers.

jeremycline:
Forwarding incorrect headers was causing Squid to fail to pool TCP
connections. This stops the streamer from forwarding hop-by-hop HTTP
headers (defined in RFC 2069).

This is part of issue #1524

### DIFF
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -17,6 +17,24 @@ from pulp.plugins.loader.exceptions import PluginNotFound
 
 logger = logging.getLogger(__name__)
 
+# These HTTP/1.1 headers are defined as being hop-by-hop and should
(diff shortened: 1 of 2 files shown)


streamer/pulp/streamer/server.py
```diff
@@ -17,6 +17,19 @@
...
+# HTTP/1.1 are end-to-end headers and should be passed back to the
+# client. See RFC 2068, section 13.5.1 for more information.
+HOP_BY_HOP_HEADERS = [
+    'connection',
+    'keep-alive',
+    'public',
+    'proxy-authenticate',
+    'transfer-encoding',
+    'upgrade',
+]
```
seandst: nit: These are all lowercase for a reason. That reason should probably be in the comment above this list definition.

PR #2365, opened 2016-01-26
Remove database initialization from content.wsgi.

jeremycline:
Firstly, the content application does not use the database, so it
shouldn't be initializing a connection to the database. Secondly, this
implementation breaks with pymongo-3.x because it was creating a
connection in a parent process (pre-WSGI forking, I believe) which is
not safe (see https://jira.mongodb.org/browse/PYTHON-961).

Without this, it is not possible to download content from Pulp.

### DIFF
--- a/server/pulp/server/content/web/wsgi.py
+++ b/server/pulp/server/content/web/wsgi.py
@@ -2,10 +2,7 @@ import os
 
 from django.core.wsgi import get_wsgi_application
 
-from pulp.server.db import connection
-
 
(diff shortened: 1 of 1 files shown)


seandst:
Tested in a vagrant up'd on the parent of this PR, doesn't appear to break everything...or anything, for that matter. Still looks terrifying, but also LGTM.

seandst:
That el6 failure might be related. :(

jeremycline:
ok test

PR #2366, opened 2016-01-26
Sends back a group call report for summary of task group with 0 tasks

dkliban:
Originally we wanted to return  a 404 when querying for a task group that has no tasks
associated with it. However, we always return a task group id when kicking off the
applicability regeneration task. User should be able to see that 0 tasks were dispatched
as part of that task group.

https://pulp.plan.io/issues/1582
closes #1582


bmbouter:
@dkliban this is OK to me. I do wonder though about a 404 being appropriate since it's in the URL. I'm OK w/ merging this, I'm just wondering.

PR #2368, opened 2016-01-27
Rewrite script for rolling back to Pulp 2.7

Ichimonji10:
Rewrite `playpen/all_27.sh`:

* Make the script behave more safely by adding `set -euo pipefail`.
* Make the script iterate through all directories named `pulp*` in the `~/devel` directory and either install the correct version of that plugin (if we know what that version is), or uninstall the plugin otherwise.
* Make the script disable the pulp-nightlies repo, enable the pulp-2.7-beta repo, and downgrade the python-{mongoengine,pymongo} packages as appropriate.
* Make the script delete the `/etc/httpd/conf.d/pulp_{content,streamer}` symlinks. These links are broken, and if left in place, they prevent httpd from starting.
* Make the script advise the user on how to drop and recreate the mongo database, and how to restart the Pulp services.

### DIFF
--- a/playpen/all_27.sh
+++ b/playpen/all_27.sh
@@ -1,22 +1,64 @@
 #!/bin/bash
+# Downgrade a Pulp system from master to 2.7.
+#
+# WARNING: This script enabled and disables system-wide yum/dnf repositories and
+# performs other system-wide actions. Execute with care.
+set -euo pipefail
 
-#  When using this script, you will need to fix your database accordingly. If moving backward, you
-#  will need to drop the database, and you will always need to run the migrations.
+# This associative array is incomplete. Patches welcome.
+declare -rA repos_branches=(
+    [pulp]=2.7-dev
+    [pulp_ostree]=1.0-dev
+    [pulp_puppet]=2.7-dev
+    [pulp_python]=1.0-dev
+    [pulp_rpm]=2.7-dev
+)
 
+# Reinstall plugins we know how to handle, and uninstall all others.
+pushd ~/devel
+for repo in pulp*; do
+    # Defend against e.g. a file named "pulp_log.txt".
+    if [ ! -d "$repo" ]; then
+        continue
+    fi
 
-# pulp_docker, pulp_python, pulp_ostree excluded because they are versioned differently
-for repo in pulp pulp_puppet pulp_rpm
-do
-    pushd /home/vagrant/devel/$repo
-    git checkout 2.7-dev
-    find ~/devel -name "*.py[c0]" -delete
-    sudo ./pulp-dev.py --install
+    pushd "$repo"
+    if [ -n "${repos_branches[$repo]:-}" ]; then
+        git checkout "${repos_branches[$repo]}"
+        find . -name '*.py[c0]' -delete
+        sudo ./pulp-dev.py --install
+    else
+        sudo ./pulp-dev.py --uninstall
+    fi
     popd
 done
+popd
 
-for repo in pulp_docker pulp_python
-do
-    pushd /home/vagrant/devel/$repo
-    sudo ./pulp-dev.py --uninstall
-    popd
-done
+set -x
+
+# These aren't used by Pulp 2.7.
+sudo rm /etc/httpd/conf.d/pulp_content.conf
+sudo rm /etc/httpd/conf.d/pulp_streamer.conf
+
+# Pulp 2.7 is incompatible with python-mongoengine >= 3 and python-pymongo >=
+# 0.10.
+sudo dnf  -y remove python-{mongoengine,pymongo}
+sudo dnf config-manager --disablerepo pulp-nightlies
+sudo dnf config-manager --enablerepo pulp-2.7-beta
+sudo dnf -y install python-mongoengine-0.8.8 python-pymongo-2.5.2
+
+fmt <<EOF
+This script does not touch the Pulp database or change the state of Pulp
+services. If downgrading from Pulp master, you may wish to execute the
+following:
+EOF
+cat <<EOF
+
+    mongo pulp_database --eval 'db.dropDatabase()'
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I've used this script yesterday and today, and it's worked perfectly for me. My use case is as follows:

1. Clone the `pulp`, `pulp_python`, `pulp_rpm`, etc repositories into a single directory.
2. Edit `pulp/Vagrantfile` and `pulp/playpen/ansible/vagrant-playbook.yml` to make the VM unique. (I often have four VMs running at the same time.)
3. Spin up a new VM with `cd pulp && vagrant up`.
4. Log in to the VM with `vagrant ssh`, and execute this script to roll back the system.

The error checking I've added to this script is especially useful. I've often had to throw away VMs that took an hour or two to provision simply because I forgot to `cd ~/devel/pulp && git checkout -- .` or something else similarly trivial.

playpen/all_27.sh
```diff
@@ -1,22 +1,52 @@
...
+    fi
 done
+popd
 
-for repo in pulp_docker pulp_python
-do
-    pushd /home/vagrant/devel/$repo
-    sudo ./pulp-dev.py --uninstall
-    popd
+# Completely uninstall these plugins. (Why?)
```
Ichimonji10: I'm genuinely confused about this. Can someone give some insight here?

asmacdo: pulp_python and pulp_docker are versioned independently from pulp and the other plugins so they cannot be installed with the loop above which uses `2.7-dev`. When I wrote this script, I wasn't using either of these plugins so I just left them uninstalled. Please feel free to add something to check out the appropriate version for each and install it.

playpen/all_27.sh
```diff
@@ -1,22 +1,52 @@
...
+
+# Pulp 2.7 is incompatible with python-mongoengine >= 3 and python-pymongo >=
+# 0.10.
+sudo dnf  -y remove python-{mongoengine,pymongo}
+sudo dnf config-manager --disablerepo pulp-nightlies
+sudo dnf config-manager --enablerepo pulp-2.7-beta
+sudo dnf -y install python-mongoengine-0.8.8 python-pymongo-2.5.2
+
+# Drop and recreate database.
+mongo pulp_database --eval 'db.dropDatabase()'
```
asmacdo: I removed this from the original script because I would rather an extra step than anyone accidentally losing data.

asmacdo: Also the appropriate db action depends on which way you are going. If you are moving from 2.6 to 2.7 for instance, you might want to keep the db so you can migrate old data.

playpen/all_27.sh
```diff
@@ -1,22 +1,52 @@
 #!/bin/bash
+# Downgrade a Pulp system from master to 2.7.
+#
+# WARNING: This script drops a mongo database, uninstalls software, and
+# performs other destructive actions.
+source ~/.bashrc  # get pstop, pstart, etc
+set -euo pipefail
```
asmacdo: +1

asmacdo:
@Ichimonji10 good work, this is turning into a robust tool rather than the hack script I made to save time :)

The only blocker for me is the db stuff.

asmacdo:
I needed to go back to 2.7, so I tried this out. Worked great!

playpen/all_27.sh
Ichimonji10: If moving from 2.6 to 2.7, would you still need to execute `sudo -u apache pulp-manage-db`?

asmacdo: Yes, but I would still prefer to leave the database stuff seperate.
Paricutlarly when testing migrations it would be useful to have a seperate
script to check out and install from the script that interacts with the
database. I would support a second script or an alias that calls this
script and interacts with the db.

Ichimonji10:
Updated.

Ichimonji10:
Just edited the PR and twiddled with the message at the end of the script for improved formatting. I added in a `cat <<EOF` section in order to preserve formatting for the sample code block.

Ichimonji10:
I've significantly updated this pull request. @asmacdo, you'll want to re-review.

playpen/all_27.sh
```diff
@@ -1,22 +1,64 @@
...
-for repo in pulp_docker pulp_python
-do
-    pushd /home/vagrant/devel/$repo
-    sudo ./pulp-dev.py --uninstall
-    popd
-done
+set -x
+
+# These aren't used by Pulp 2.7.
+sudo rm /etc/httpd/conf.d/pulp_content.conf
```
asmacdo: Does it hurt anything to leave these? The devs who might use this script might "pop over" to 2.7 to see if a bug was present there, and then go right back to master where they would be surprised that their config files have been restored to defaults.

Ichimonji10: Leaving these prevents httpd from starting.

Ichimonji10: See the commit message.

asmacdo:
There is a new problem rolling back though related to mongo dependencies, but that isn't caused by this work. So I am going to go ahead and merge.

Ichimonji10:
@asmaco Thank you!

Ichimonji10:
@asmaco Is the problem handled by this?

```sh
# Pulp 2.7 is incompatible with python-mongoengine >= 3 and python-pymongo >=
# 0.10.
sudo dnf  -y remove python-{mongoengine,pymongo}
sudo dnf config-manager --disablerepo pulp-nightlies
sudo dnf config-manager --enablerepo pulp-2.7-beta
sudo dnf -y install python-mongoengine-0.8.8 python-pymongo-2.5.2
```

Or is the problem newer than that?

asmacdo:
@Ichimonji10 it should be handled by that but it looks like this code doesn't disable/enable the correct repos

PR #2369, opened 2016-01-27
Implement alternate content sources in the streamer.

jeremycline:
Alternate content sources is now supported for repositories using
download policies other than `immediate`.

closes #1451

### DIFF
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -5,13 +5,15 @@ import logging
 
 from mongoengine import DoesNotExist, NotUniqueError
 from nectar import listener as nectar_listener
-from nectar import request as nectar_request
 import requests
 from twisted.internet import reactor
 from twisted.web import resource
 from twisted.web.server import NOT_DONE_YET
 
+from pulp.plugins.loader import api as plugins_api
 from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
+from pulp.server.content.sources import container as content_container
+from pulp.server.content.sources import model as content_models
 from pulp.server.db import model
 from pulp.server.controllers import repository as repo_controller
 from pulp.plugins.loader.exceptions import PluginNotFound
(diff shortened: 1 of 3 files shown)


streamer/pulp/streamer/server.py
```diff
@@ -53,12 +55,16 @@ def __init__(self, request, streamer_config):
         :type  request:         twisted.web.server.Request
         :param streamer_config: The configuration for this streamer instance.
         :type  streamer_config: ConfigParser.SafeConfigParser
+        :param catalog_entry:   Catalog entry for the file requested.
+        :type  catalog_entry:   pulp.server.db.model.LazyCatalogEntry
+        :param pulp_request:    True if this request originated from Pulp.
```
rbarlow: Can you elaborate on when or why a request would originate from outside of Pulp?

jeremycline: A request originates from outside of Pulp when a client requests a piece of content. If Pulp itself is requesting content (to save to permanent storage) it sets an HTTP header so the streamer doesn't create a database entry so that Pulp later goes and downloads the content, which causes a database entry...

streamer/pulp/streamer/server.py
```diff
@@ -204,15 +208,29 @@ def _download(self, catalog_entry, request, responder):
...
+                                                     **catalog_entry.data)
+        pulp_request = request.getHeader(PULP_STREAM_REQUEST_HEADER)
+        listener = StreamerListener(request, self.config, catalog_entry, pulp_request)
+        primary_downloader.session = self.session
+        primary_downloader.event_listener = listener
+
+        # Build the alternate content source download request
+        unit_model = plugins_api.get_unit_model_by_id(catalog_entry.unit_type_id)
+        qs = unit_model.objects.filter(id=catalog_entry.unit_id).only(*unit_model.unit_key_fields)
+        unit = qs.get()
```
rbarlow: .get() can raise two exceptions - are you confident that we don't need to handle those exceptions because something above us in the stack will do whatever is needed?

http://docs.mongoengine.org/en/latest/apireference.html#mongoengine.queryset.QuerySet.get

streamer/pulp/streamer/server.py
rbarlow: Can you add that explanation to the docblock?

PR #2372, opened 2016-01-28
Don't forward requests for files that don't exist to the streamer.

jeremycline:
Requests for directories or files that don't exist in the repository
should not be forwared to the streamer. This PR updates the view for the
content WSGI app to check for the link in the published repository
before forwarding a request.

closes #1551

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -131,11 +131,15 @@ class ContentView(View):
             # Not Authorized
             return HttpResponseForbidden()
 
+        # Immediately 404 if the symbolic link doesn't even exist
+        if not os.path.lexists(request.path_info):
+            return HttpResponseNotFound(request.path_info)
+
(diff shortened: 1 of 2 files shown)


server/pulp/server/content/web/views.py
```diff
@@ -131,13 +131,14 @@ def get(self, request):
...
-        # Redirect
-        if lazy_enabled:
-            return self.redirect(request, self.key)
+        if not os.path.isdir(path):
+            # Already downloaded and is a file
+            if os.path.exists(path):
+                return self.x_send(path)
+
+            # Redirect if lazy is on and the link exists in the repo
+            if lazy_enabled and os.path.lexists(request.path_info):
```
jeremycline: `request.path_info` is the path the client requested (`/var/www/pub/yum/https/repos/.../some_symlink.rpm`) whereas `path` is the file that symlink points to.

PR #2373, opened 2016-01-29
Launch the streamer with its own script in /usr/bin/

jeremycline:
This is to pave the way for the SELinux work (issue #1459).

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -661,6 +661,7 @@ The streamer component of the Pulp Lazy Sync feature.
 
 %files -n python-pulp-streamer
 %defattr(-,root,root,-)
+%{_bindir}/pulp_streamer
 %{python_sitelib}/%{name}/streamer/
 %{python_sitelib}/pulp_streamer*.egg-info
 %config(noreplace) %{_sysconfdir}/%{name}/streamer.conf
--- a/streamer/setup.py
+++ b/streamer/setup.py
@@ -13,5 +13,10 @@ setup(
         'nectar >= 1.4.0',
         'setuptools',
         'twisted',
-    ]
+    ],
+    entry_points={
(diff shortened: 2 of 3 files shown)


pulp.spec
```diff
@@ -409,6 +409,7 @@ Pulp provides replication, access, and accounting for software repositories.
 %{_bindir}/pulp-manage-db
 %{_bindir}/pulp-qpid-ssl-cfg
 %{_bindir}/pulp-gen-ca-certificate
+%{_bindir}/pulp_streamer
```
bmbouter: Does the spec file actually need this entry? What are the files contents? I figured the `entry_point` would populate the file when `setup.py` is run.

jeremycline: The file is generated by setuptools at build time. This declares that this package (which I just noticed is the server and not the streamer, oops) owns that file. The contents of the file (generated by `python-setuptools-18.0.1-2.fc23.noarch`) are:
```
#!/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'pulp-streamer==2.8.0b5','console_scripts','pulp_streamer'
__requires__ = 'pulp-streamer==2.8.0b5'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('pulp-streamer==2.8.0b5', 'console_scripts', 'pulp_streamer')()
    )
```
I'm not crazy about the use of `#!/bin/python` instead of `#!/usr/bin/env python2`. I believe this problem should be addressed upstream, though. I will see if this has been fixed in the current rawhide release. If not, I'll file a bug.

bmbouter: Oh right the file needs to be declared that the package owns it. Thanks for the answer!

PR #2376, opened 2016-02-01
Story 1581: Cancel tasks in a task group

shubham90:
https://pulp.plan.io/issues/1581

closes #1581

### DIFF
--- a/docs/dev-guide/integration/rest-api/tasks.rst
+++ b/docs/dev-guide/integration/rest-api/tasks.rst
@@ -184,6 +184,29 @@ Task Group Management
 
 .. _task_group_summary:
 
+Cancelling Tasks in a Task Group
+------------------------------
+
+All asynchronous tasks in a particular task group may be canceled by the user before they complete. A
+task must be in the *waiting* or *running* state in order to be canceled.
+
+.. Note::
+
+   It is possible for a task to complete or experience an error before the cancellation request is
+   processed, so it is not guaranteed that a task's final state will be 'canceled' as a result of
+   this call. In these instances this method call will still return a response code of 200.
+
+| :method:`delete`
+| :path:`/v2/task_groups/<group_id>/`
+| :permission:`delete`
+
+| :response_list:`_`
+
+* :response_code:`200, if the task group cancellation request was successfully received`
+* :response_code:`404, if the task group is not found`
+
+| :return:`null`
+
 Task Group Summary
 ------------------
 
--- a/docs/user-guide/release-notes/master.rst
+++ b/docs/user-guide/release-notes/master.rst
@@ -86,6 +86,8 @@ Rest API Changes
 * The API for regenerating content applicability for updated repositories no longer returns a
   :ref:`call_report`. Instead a :ref:`group_call_report` is returned.
 
+* Task Groups with tasks having incomplete states can now be canceled.
+
 Binding API Changes
 -------------------
 
--- a/server/pulp/server/webservices/views/task_groups.py
+++ b/server/pulp/server/webservices/views/task_groups.py
@@ -4,11 +4,13 @@ This module contains views related to Pulp's task groups.
 from django.views.generic import View
 
 from pulp.common.constants import CALL_STATES
+from pulp.server.async import tasks
 from pulp.server.auth import authorization
 from pulp.server.db.model import TaskStatus
 from pulp.server.exceptions import MissingResource
 from pulp.server.webservices.views.decorators import auth_required
-from pulp.server.webservices.views.util import generate_json_response_with_pulp_encoder
+from pulp.server.webservices.views.util import generate_json_response, \
+    generate_json_response_with_pulp_encoder
 
 
 class TaskGroupView(View):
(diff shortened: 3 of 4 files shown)


shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

shubham90:
ok test

docs/dev-guide/integration/rest-api/tasks.rst
```diff
@@ -184,6 +184,28 @@ Task Group Management
...
+------------------------------
+
+All asynchronous tasks in a particular task group may be canceled by the user before they complete. A
+task must be in the *waiting* or *running* states in order to be canceled.
+
+.. Note::
+
+   It is possible for a task to complete or experience an error before the cancellation request is
+   processed, so it is not guaranteed that a task's final state will be 'canceled' as a result of
+   this call. In these instances this method call will still return a response code of 200.
```
dkliban: Shouldn't it be 202?

shubham90: @dkliban  I have tried to keep it in consistent with the tasks API. Tasks return 200, I will see more about it.

docs/user-guide/release-notes/master.rst
```diff
@@ -86,6 +86,8 @@ Rest API Changes
 * The API for regenerating content applicability for updated repositories no longer returns a
   :ref:`call_report`. Instead a :ref:`group_call_report` is returned.
 
+* Group tasks with incomplete states can now be canceled.
```
mhrivnak: They are "task groups", not "group tasks".

server/pulp/server/webservices/views/task_groups.py
```diff
@@ -32,6 +33,24 @@ def get(self, request, group_id):
...
+
+        :param request: WSGI request object
+        :type  request: django.core.handlers.wsgi.WSGIRequest
+        :param group_id: The ID of the task group you wish to cancel
+        :type  group_id: basestring
+
+        :return: Response containing None
+        :rtype:  django.http.HttpResponse
+        """
+        raw_tasks = TaskStatus.objects(group_id=group_id)
```
mhrivnak: Since you only need one field, use the "only" method: http://docs.mongoengine.org/apireference.html#mongoengine.queryset.QuerySet.only

shubham90:
@mhrivnak Thanks for the review. I have updated the PR as per your suggestion.

docs/dev-guide/integration/rest-api/tasks.rst
```diff
@@ -184,6 +184,29 @@ Task Group Management
...
+
+| :method:`delete`
+| :path:`/v2/task_groups/<group_id>/`
+| :permission:`delete`
+
+| :response_list:`_`
+
+* :response_code:`200, if the group task cancellation request was successfully received`
```
mhrivnak: s/group task/task group/

shubham90: I always do this.:sweat: My bad

server/test/unit/server/webservices/views/test_task_groups.py
```diff
@@ -27,6 +28,55 @@ def filter(self, state=None):
...
+    def test_delete_task_resource_nonexistant(self, mock_task, mock_objects):
+        """
+        Test delete task_group with no tasks
+        """
+        mock_request = mock.MagicMock()
+        mock_objects.only.return_value.get_or_404.side_effect = MissingResource
+        task_resource = TaskGroupView()
+        try:
+            response = task_resource.delete(mock_request, 'mock_task')
+        except MissingResource, response:
```
mhrivnak: You re-used the variable ``response`` to hold an exception instead of a response object. It's better to use a different variable name.

server/test/unit/server/webservices/views/test_task_groups.py
```diff
@@ -27,6 +28,55 @@ def filter(self, state=None):
...
+    @mock.patch('pulp.server.webservices.views.task_groups.TaskStatus.objects')
+    @mock.patch('pulp.server.webservices.views.task_groups.tasks')
+    def test_delete_task_resource_nonexistant(self, mock_task, mock_objects):
+        """
+        Test delete task_group with no tasks
+        """
+        mock_request = mock.MagicMock()
+        mock_objects.only.return_value.get_or_404.side_effect = MissingResource
+        task_resource = TaskGroupView()
+        try:
```
mhrivnak: It would be better to just use ``self.assertRaises`` here. Don't worry about testing the response code on the exception, because that's a property of the exception itself. That should be tested elsewhere.

server/test/unit/server/webservices/views/test_task_groups.py
```diff
@@ -27,6 +28,55 @@ def filter(self, state=None):
...
+        mock_request = mock.MagicMock()
+        task1 = mock.Mock(task_id='test_foo_path')
+        task2 = mock.Mock(task_id='test_foo_path1')
+        mock_objects.only.return_value.get_or_404.return_value = [task1, task2]
+
+        task_resource = TaskGroupView()
+
+        response = task_resource.delete(mock_request, 'mock_task')
+
+        assert 2 == mock_task.cancel.call_count
```
mhrivnak: use ``self.assertEqual`` instead to keep this in the realm of the test framework.

shubham90:
ok test

shubham90:
@mhrivnak I have made the changes you suggested. Learnt some new things along the way :) Thanks

mhrivnak:
Well done!

Next time you're in the office, let's talk briefly about the pattern here of using DELETE to cancel an event, and a different option we should have probably used in our API. You definitely did the right thing staying consistent with the existing API, but in v3 of our API we'll probably do it differently.

shubham90:
@mhrivnak Sure thing!

shubham90:
Thanks for the review.

shubham90:
ok test

shubham90:
ok test

PR #2378, opened 2016-02-01
Fix orphan purge to not fail when content not downloaded.

jortel:
https://pulp.plan.io/issues/1586

Skip file deletion when unit storage path does not exist.


ipanova:
ok test

jortel:
ok test

PR #2385, opened 2016-02-02
Unit storage path based on sha256 of unit key instead of using UUID.

jortel:
https://pulp.plan.io/issues/1600

Changed `ContentUnit.unit_key_str()` because it did not produce unique results based on unit key.  The problem is that it created a string representation of - a list of sorted *values* that are not associated with their keys.

For example:
```
key_a = {
    'a':1,
    'b':2,
    'c':3,
    'd':3,
    'e':4
}

key_b = {
    'a':3,
    'b':2,
    'c':1,
    'd':4,
    'e':3
}
```

**both** *key_a* and *key_b*  produced the same string '[1, 2, 3, 3, 4]'

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -5,6 +5,7 @@ import os
 import random
 import uuid
 from collections import namedtuple
+from hashlib import sha256
 from hmac import HMAC
 
 from mongoengine import (BooleanField, DictField, Document, DynamicField, IntField,
@@ -631,7 +632,7 @@ class ContentUnit(AutoRetryDocument):
         """
         The unit key represented as a string ordered by unit key fields alphabetically
         """
-        return str(sorted([getattr(self, key) for key in self.unit_key_fields]))
+        return self.unit_key_as_digest()
 
     @property
     def unit_key_as_named_tuple(self):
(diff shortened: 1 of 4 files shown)


server/test/unit/server/content/test_storage.py
```diff
@@ -62,11 +62,14 @@ def test_exit(self):
 
 class TestFileStorage(TestCase):
 
+    @patch('pulp.server.content.storage.sha256')
```
mhrivnak: FWIW I don't think there's much value in mocking out standard library calls unless they do something expensive like hit disk, network, etc.

And in this case, since we're depending on the string representation of a list being consistent among python implementations, it would actually be valuable to hard-code the expected real hash, and compare it against the calculated hash.

server/pulp/server/db/model/__init__.py
```diff
@@ -631,7 +631,7 @@ def unit_key_str(self):
         """
         The unit key represented as a string ordered by unit key fields alphabetically
         """
-        return str(sorted([getattr(self, key) for key in self.unit_key_fields]))
+        return str(sorted(self.unit_key.items()))
```
mhrivnak: I wonder if depending on the string representation of a list of tuples could turn out to be non-portable. What if it changes subtly in python 3.x, for example? Since this change is the first time we're expecting results of this to persist long-term, maybe we should consider how to make it more deterministic.

An easy solution is to choose a string that's unlikely to be part of a unit key value and use it as a separator. The hard part is choosing that string. Something like this might work, although I'm at least half-kidding about the example separator:

```
return '!@#$%^&*'.join(self.unit_key[key] for key in sorted(self.unit_key.keys()))
```

What do you think? Maybe there's a better way.

jortel: Good point.  The `__str__()` representation of a list of tuples is not guaranteed.
I will pursue your suggestion further before merging.

server/test/unit/server/content/test_storage.py
jortel: Another good point.  Will get rid of the mock.

server/pulp/server/db/model/__init__.py
jortel: Actually, on further consideration, I don't think it's possible to create a 100% unique (to the unit key) unit key string using string concatenation.  Submitting a *hashing* solution instead that is certain to be both deterministic and unique.

jortel:
@mhrivnak, mind having another look at this?

jortel:
ok test

PR #2386, opened 2016-02-03
Upgrade python-amqp dep to 1.4.9

pcreech:
Upgrade python-amqp dep to 1.4.9


pcreech:
Forgot to set source branch to a different branch

PR #2390, opened 2016-02-04
Bump kombu version

pcreech:
Upgrade python-kombu dep to 3.0.33


pcreech:
wrong upstream branch

PR #2391, opened 2016-02-04
Bump kombu version

pcreech:
Upgrade python-kombu dep to 3.0.33


deps/python-kombu/python-kombu.spec
```diff
@@ -180,6 +169,9 @@ popd
 %endif # with_python3
 
 %changelog
+* Wed Feb 03 2016 Patrick Creech <pcreech@redhat.com> 3.0.33-1.pulp
+- Upgrade python-kombu dep to 3.0.33 (pcreech@redhat.com)
```
bmbouter: @pcreech This all look really good. The only improvement I can see for next time is that the changelog should have another line indicating that a patch is being applied.

PR #2392, opened 2016-02-04
Adds unit-key uniqueness constraint for all unit types

dkliban:
https://pulp.plan.io/issues/1477
closes #1477

### DIFF
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -165,7 +165,17 @@ def ensure_database_indexes():
 
     # Load all the model classes that the server knows about and ensure their indexes as well
     plugin_manager = PluginManager()
-    for model_class in plugin_manager.unit_models.itervalues():
+    for unit_type, model_class in plugin_manager.unit_models.items():
+        unit_key_index = {'fields': model_class.unit_key_fields, 'unique': True}
+        for index in model_class._meta['indexes']:
+            if isinstance(index, dict) and 'fields' in index:
+                if list(index['fields']) == list(unit_key_index['fields']):
+                    raise ValueError("Content unit type '%s' explicitly defines an index for its "
+                                     "unit key. This is not allowed because the platform handles"
+                                     "it for you." % unit_type)
+        model_class._meta['indexes'].append(unit_key_index)
+        model_class._meta['index_specs'] = \
+            model_class._build_index_specs(model_class._meta['indexes'])
         model_class.ensure_indexes()
 
 
--- a/server/test/unit/server/db/test_manage.py
+++ b/server/test/unit/server/db/test_manage.py
@@ -141,9 +141,30 @@ class TestManageDB(MigrationTest):
         the appropriate platform models
         """
         test_model = MagicMock()
-        mock_plugin_manager.return_value.unit_models.itervalues.return_value = [test_model]
+        mock_plugin_manager.return_value.unit_models.items.return_value = [('test-unit',
+                                                                            test_model)]
         manage.ensure_database_indexes()
         test_model.ensure_indexes.assert_called_once_with()
+
+    @patch.object(manage, 'PluginManager')
+    @patch.object(manage, 'model')
+    def test_ensure_database_indexes_throws_exception(self, mock_model, mock_plugin_manager):
+        """
+        Make sure that the ensure_indexes method is called for all
+        the appropriate platform models
+        """
+        test_model = MagicMock()
+        test_model.unit_key_fields = ('1', '2', '3')
+        unit_key_index = {'fields': test_model.unit_key_fields, 'unique': True}
+        test_model._meta.__getitem__.side_effect = [[unit_key_index]]
+        mock_plugin_manager.return_value.unit_models.items.return_value = [('test-unit',
+                                                                            test_model)]
+        with self.assertRaises(ValueError) as context:
+            manage.ensure_database_indexes()
+        self.assertEqual(context.exception.message, "Content unit type 'test-unit' explicitly "
+                                                    "defines an index for its unit key. This is "
+                                                    "not allowed because the platform handlesit "
+                                                    "for you.")
 
     @patch.object(manage, 'ensure_database_indexes')
     @patch('logging.config.fileConfig')


server/pulp/server/db/manage.py
```diff
@@ -165,7 +165,16 @@ def ensure_database_indexes():
 
     # Load all the model classes that the server knows about and ensure their indexes as well
     plugin_manager = PluginManager()
-    for model_class in plugin_manager.unit_models.itervalues():
+    for unit_type, model_class in plugin_manager.unit_models.items():
+        unit_key_index = {'fields': model_class.unit_key_fields, 'unique': True}
+        for index in model_class._meta['indexes']:
+            if unit_key_index == index:
+                raise Exception(_("Content unit type '%s' explicitly defines an index for its unit"
```
seandst: I know that I suggested adding this, but I think I want a second opinion on whether or not this is even necessary. Right now I could take it or leave it.

Maybe the right solution is adding information about this to the docs for adding a new type, but I believe those are all so out of date that adding it here would be a ton of work for you.

seandst:
Any thoughts about adding a test for this? It looks like you can probably mock out ensure_indexes and keep things out of mongo.

server/pulp/server/db/manage.py
bmbouter: The docs are epicly out of date, but this could be a platform release note for plugin writers. There is usually a section for them in the release notes. +1 to putting it there

bmbouter:
@dkliban Since this has the `closes` tag it will put the issue to MODIFIED before it should be. Each plugin needs the unique indexes they manually define removed. Those Prs need to be made and likely everything needs to be merged at once.

bmbouter:
+1 to writing a test for this code. It would be easy.

server/pulp/server/db/manage.py
bmbouter: This raise behavior helps plugin writers avoid duplicate index definition so I think it's worth keeping. Consider making an explicit coded exception for this instead of a general `Exception`. I think that would be good.

server/pulp/server/db/manage.py
```diff
@@ -165,7 +165,16 @@ def ensure_database_indexes():
 
     # Load all the model classes that the server knows about and ensure their indexes as well
     plugin_manager = PluginManager()
-    for model_class in plugin_manager.unit_models.itervalues():
+    for unit_type, model_class in plugin_manager.unit_models.items():
+        unit_key_index = {'fields': model_class.unit_key_fields, 'unique': True}
+        for index in model_class._meta['indexes']:
+            if unit_key_index == index:
```
bmbouter: I haven't tested it, but does this work as expected on Python 2.6 also? Somehow I remember dictionary equality comparison being a Python 2.7 thing.

seandst: dict equality should be `sorted(dict1.keys()) == sorted(dict2.keys()) and sorted(dict1.values()) == sorted(dict2.values())`[0], at least as far back as python 2.4[1], but a unit test proving it would be ideal.

(look for the mappings bullet point in both of the referenced links)
[0]: https://docs.python.org/2/reference/expressions.html#not-in
[1]: https://docs.python.org/2.4/ref/comparisons.html

dkliban:
ok test

common/pulp/common/error_codes.py
```diff
@@ -122,6 +122,9 @@
 PLP0042 = Error("PLP0042", _("This request is forbidden."), [])
 PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
                              "Refer to /etc/pulp/server.conf for proper use."), [])
+PLP0044 = Error("PLP0044", _("Content unit type '%(content_unit)s' explicitly defines an index for"
+                             "its unit key. This is not allowed because the platform handles it for"
+                             "you."), ['content_unit'])
```
rbarlow: I know @bmbouter suggested making this a coded exception, but I think that's heavy handed. Coded exceptions are for users, not developers. And no user should ever see this error because it will cause failure before we release and the developer will be informed. I think just plain Exception was a better approach because it's less code.

rbarlow: Actually, a ValueError kind of makes the most sense, because it's an incorrect value in the index dictionary.

bmbouter: @dkliban Incorporate @rbarlow 's feedback how you would like. He makes a good point. It should be noted that there already exist coded exceptions that are develop focused, so we're already in the state he suggest we shouldn't be.

rbarlow: A good reason not to put developer errors in the "coded exceptions" is that the coded exceptions need to be well documented and communicated to the users. As these users include our API consumers who may wish to respond to our error code, this will cause them to spend time thinking about how to handle a problem they will never experience. Let's save them the time by just raising a ValueError. Also, it's probably not necessary to use gettext since working on our code already implies that developers know English enough to read/write our docblocks and comments and variable/function names ☺

server/test/unit/server/db/test_manage.py
```diff
@@ -141,10 +143,31 @@ def test_ensure_database_indexes(self, mock_model, mock_plugin_manager):
...
+        Make sure that the ensure_indexes method is called for all
+        the appropriate platform models
+        """
+        test_model = MagicMock()
+        test_model.unit_key_fields = ('1', '2', '3')
+        unit_key_index = {'fields': test_model.unit_key_fields, 'unique': True}
+        test_model._meta.__getitem__.side_effect = [[unit_key_index]]
+        mock_plugin_manager.return_value.unit_models.items.return_value = [('test-unit',
+                                                                            test_model)]
+        try:
```
mhrivnak: It's usually simpler and more readable to use ``self.assertRaises`` as a context manager (which does require the use of ``pulp.common.compat.unittest``) Off the top of my head, it's something like

```
with self.assertRaises(PulpCodedException) as context:
  manage.ensure_database_indexes()
self.assertEqual(context.exception.error_code, error_codes.PLP0044)

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

PR #1512, opened 2015-01-12
165403 - deprecated cacert and cakey settings from server.conf

skarmark:
https://bugzilla.redhat.com/show_bug.cgi?id=1165403

PR #2394, opened 2016-02-05
Create a simple directory index for the content.wsgi app.

jeremycline:
In the past we were able to use Apache's directory index feature.
However, for lazy content we pass the request through to the content
WSGI application and are unable to use this feature. This introduces a
simply template that lists the directories and files within a requested
directory, as well as clear HTTP 404 and HTTP 500 error pages.

closes #1512

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -2,12 +2,17 @@ import mimetypes
 import os
 
 from django.http import \
-    HttpResponse, HttpResponseRedirect, HttpResponseNotFound, HttpResponseForbidden
+    HttpResponse, HttpResponseRedirect, HttpResponseForbidden, Http404
+from django.shortcuts import render_to_response
 from django.views.generic import View
 
 from pulp.repoauth.wsgi import allow_access
 from pulp.server.lazy import URL, Key
 from pulp.server.config import config as pulp_conf
+
+
+# Make sure all requested paths fall under this directory.
+PUBLISH_DIR = '/var/lib/pulp/published'
 
 
 class ContentView(View):
@@ -136,12 +141,36 @@ class ContentView(View):
             # Not Authorized
             return HttpResponseForbidden()
 
+        if not path.startswith(PUBLISH_DIR):
+            # Someone is requesting something they shouldn't.
+            return HttpResponseForbidden()
+
         # Immediately 404 if the symbolic link doesn't even exist
         if not os.path.lexists(request.path_info):
-            return HttpResponseNotFound(request.path_info)
+            raise Http404
+
+        if os.path.isdir(path):
+            return self.directory_index(path)
 
         # Already downloaded
         if os.path.exists(path):
             return self.x_send(path)
 
         return self.redirect(request, self.key)
+
+    @staticmethod
+    def directory_index(path):
+        """
+        Render the given path to a directory index.
+
+        :param path: Absolute path to the directory to list.
+        :type  path: str
+
+        :return: HttpResponse
+        """
+        listing = os.listdir(path)
+        context = {
+            'dirs': sorted([f for f in listing if os.path.isdir(os.path.join(path, f))]),
+            'files': sorted([f for f in listing if os.path.isfile(os.path.join(path, f))]),
+        }
+        return render_to_response('directory_index.html', context)
--- a/server/setup.py
+++ b/server/setup.py
@@ -28,5 +28,5 @@ setup(
     install_requires=[
         'blinker', 'celery >=3.1.0, <3.2.0', DJANGO_REQUIRES, 'httplib2', 'iniparse',
         'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.10.0', 'oauth2>=1.5.211', 'pymongo>=3.0.0',
-        'semantic_version>=2.2.0', 'setuptools']
+        'semantic_version>=2.2.0', 'setuptools'],
 )
--- a/server/usr/share/pulp/templates/directory_index.html
+++ b/server/usr/share/pulp/templates/directory_index.html
@@ -1,0 +1,21 @@
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Pulp Repository Index</title>
+</head>
+<body>
+    <h1>Pulp Repository Content</h1>
+    <a href="../">Parent Directory</a>
+
+    <ul style='list-style: none outside none; font-family: monospace'>
+        {% for d in dirs %}
+            <li><a href="{{ d }}/">{{ d }}/</a></li>
+        {% endfor %}
+        {% for f in files %}
+            <li><a href="{{ f }}">{{ f }}</a></li>
+        {% endfor %}
+    </ul>
+</body>
+</html>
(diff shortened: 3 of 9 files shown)


jeremycline:
ok test

rbarlow:
can we change it to "brok test"?

rbarlow:
In your commit message, s/simply/simple/

server/pulp/server/content/web/views.py
```diff
@@ -144,4 +148,13 @@ def get(self, request):
             return self.redirect(request, self.key)
 
         # NotFound
-        return HttpResponseNotFound(request.path_info)
+        raise Http404
+
+    @staticmethod
+    def directory_index(request):
+        listing = os.listdir(request.path_info)
```
rbarlow: I think we should be careful here - someone could use this to listdir ```../../../etc/pki/`.

rbarlow: Actually, couldn't they also just pass ```/etc/whatever```?

server/setup.py
```diff
@@ -28,5 +28,16 @@
...
-        'semantic_version>=2.2.0', 'setuptools']
+        'semantic_version>=2.2.0', 'setuptools'],
+    data_files=[
+        (
+            '/usr/share/pulp/templates/', 
+            [
+                'usr/share/pulp/templates/404.html',
+                'usr/share/pulp/templates/500.html',
+                'usr/share/pulp/templates/directory_index.html',
+
```
rbarlow: Maybe drop the blank line?

server/setup.py
```diff
@@ -28,5 +28,16 @@
     install_requires=[
         'blinker', 'celery >=3.1.0, <3.2.0', DJANGO_REQUIRES, 'httplib2', 'iniparse',
         'isodate>=0.5.0', 'm2crypto', 'mongoengine>=0.10.0', 'oauth2>=1.5.211', 'pymongo>=3.0.0',
-        'semantic_version>=2.2.0', 'setuptools']
+        'semantic_version>=2.2.0', 'setuptools'],
+    data_files=[
```
rbarlow: +1 for using this instead of relying the spec file.

server/usr/share/pulp/templates/directory_index.html
```diff
@@ -0,0 +1,31 @@
+
```
rbarlow: There isn't enough JavaScript in here. Please add minified JavaScript that nobody can tell what it does.

server/usr/share/pulp/templates/directory_index.html
```diff
@@ -0,0 +1,31 @@
...
+    {% if dirs %}
+        <h2>Directories</h2>
+        <ul>
+            {% for d in dirs %}
+            <li><a href="{{ d }}/">{{ d }}</a></li>
+            {% endfor %}
+        </ul>
+        {% endif %}
+
+        {% if files %}
```
rbarlow: Can there ever be files without dirs?

rbarlow:
I talked to @jeremycline in person about this. He suggested that httpd protects us against the ../ or /etc attacks. Independently verifying this would be good, but I am also trying to think about deployment scenarios that don't involve httpd (gunicorn, nginx, etc.)

server/pulp/server/content/web/views.py
rbarlow: @seandst suggested that we should compare os.normpath with os.path_info to see if someone is up to something. I am up to something. Right now.

server/setup.py
```diff
@@ -28,5 +28,16 @@
...
+        (
+            '/usr/share/pulp/templates/', 
+            [
+                'usr/share/pulp/templates/404.html',
```
mhrivnak: Doesn't django have default pages for 404 and 500? Is there a reason we can't or don't want to use them?

jeremycline: The way I read the Django 1.4 documentation, it didn't seem like it. It certainly wasn't serving anything visually useful before I did this, but maybe that's because we were returning HttpResponseNotFound instead of raising Http404.

server/usr/share/pulp/templates/directory_index.html
jeremycline: Yup. This sorts the content of a repository into directories and files and puts all the directories at the top (so you don't have to go hunting for `repodata` somewhere in 1000s of files). The contents of `repodata` would just be files, though.

jeremycline:
@rbarlow should know that my judgment about the security of anything is poor if that question is posed in a bar. Although Apache deals well with nefarious requests in conjunction with the mod_rewrite directives we use, mod_wsgi follows the specification of WSGI[0] (shocking that that is the case, I know) so a request directly to the WSGI app (for our configuration that's `/pulp/content/`) will let you avoid the mod_rewrite rule.

tl;dr: I didn't think this out. I think checking to make sure that `os.path.realpath(request.path_info)` is a sub-directory of `/var/lib/pulp/published/

[0] https://www.python.org/dev/peps/pep-0333/#environ-variables

rbarlow:
@seandst are you in agreement that my previous concerns are now assuaged?

jeremycline:
@rbarlow very nice use of 'assuaged' :+1:

server/pulp/server/content/web/views.py
```diff
@@ -144,4 +156,13 @@ def get(self, request):
...
         # NotFound
-        return HttpResponseNotFound(request.path_info)
+        raise Http404
+
+    @staticmethod
+    def directory_index(path):
+        listing = os.listdir(path)
+        context = {
+            'dirs': [f for f in listing if os.path.isdir(os.path.join(path, f))],
+            'files': [f for f in listing if os.path.isfile(os.path.join(path, f))]
```
seandst: iterating over a list and doing the same work twice makes me sad, but I can be mollified by mixing in a builtin...
```python
listing = [os.path.join(path, f) for f in listing]
# ...
'dirs': filter(os.path.isdir, listing),
'files': filter(os.path.isfile, listing),
# snooty trailing comma on  final  item ^ :D
```

server/setup.py
seandst: Silly question, but...does this actually work? I seem to recall needing some MANIFEST.in shenanigans for this to actually work. Running a `setup.py sdist` makes a source tarball that doesn't contain any html files. Is that expected?

seandst:
Is it worth putting something in the docs about this listing? I think it is, but welcome being talked out of it. :)

server/pulp/server/content/web/views.py
jeremycline: It also made me sad and I  briefly wondered if there was a better way (that was also readable).

seandst:
> @seandst are you in agreement that my previous concerns are now assuaged?

Looking at it, yes. I'm trying to get it up in a browser now to see some 403s.

seandst:
Security-wise, lgtm. The apache config appears to be mangling the URL appropriately whenever I try to do nefarious things, so I don't think I ever tripped the forbidden response as a result of passing a bad path.

I'm still concerned about the data_files in setup.py.

I think the directory index page can be greatly simplified (e.g. make it look like a normal apache/nginx directory index), but don't bother if you don't want to since it's a cosmetic thing.

server/setup.py
jeremycline: It did work when I ran `setup.py install`, but wasn't working when I tried `setup.py develop`. If there's a better way to do this, I'd definitely like to know.

My main goal was to not force the spec file to use `install`

jeremycline:
Sadly it looks like the `data_files` bit doesn't work in most cases and setuputils won't let you install your data files outside of your package directories (by design). I've gone ahead and made the change to the spec file and removed the use of `data_files`.

PR #2395, opened 2016-02-05
Set Apache to allow unlimited requests per TCP connection.

jeremycline:
By default, Apache allows 100 HTTP requests per TCP connection before
resetting it. Since it is not uncommon for Pulp to serve thousands or
tens of thousands of requests from a single client, it makes sense to
not use the default. This is especially noticeable when performing
Pulp-to-Pulp syncs.

This is related to issue #1629.

### DIFF
--- a/server/etc/httpd/conf.d/pulp_apache_22.conf
+++ b/server/etc/httpd/conf.d/pulp_apache_22.conf
@@ -61,6 +61,21 @@ Alias /pulp/static /var/lib/pulp/static
 </Location>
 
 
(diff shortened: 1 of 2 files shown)


mhrivnak:
Please bring this directly to katello's attention. This is the kind of change they could easily overlook.

jeremycline:
@mhrivnak FYI I updated the default to be 10,000 (admittedly still quite large) rather than unlimited, and added an inline comment about tweaking it based on DoS concerns.

PR #2396, opened 2016-02-05
Attempt to guess the MIME type of content served with xsendfile.

jeremycline:
This is mostly for the benefit of browsers, since the default MIME
type Django uses is text/html which causes the browser to attempt to
render RPMs and such instead of downloading them.

closes #1627

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -1,3 +1,4 @@
+import mimetypes
 import os
 
 from django.http import \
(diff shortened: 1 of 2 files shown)


server/pulp/server/content/web/views.py
```diff
@@ -68,7 +69,13 @@ def x_send(path):
         :rtype: django.http.HttpResponse
         """
         if os.access(path, os.R_OK):
-            reply = HttpResponse()
+            content_type = mimetypes.guess_type(path)
```
jeremycline: We could use a library like libmagic to determine the file type (which will be much more accurate), but I was hesitant to introduce a new dependency. The downside to this method is that if the file suffix is wrong/missing (like the `listing` files which are text of some variety) the 'application/octet-stream' is used.

rbarlow: FWIW, there is a Python binding to libmagic that seems to "come with" Fedora last I checked. I'm not sure about ELs though. Also, it's not documented, and I spent quite a lot of time trying to figure out how to use its API.

Oh, and there are no less than three Python libraries that are bindings to libmagic that share the same name but do not have the same API. If we depend on one of them, it will be difficult to express which one we need in our setup.py which could be bad if we ever want to work in Debian, which by the way uses a different one than Fedora uses.

PR #2401, opened 2016-02-09
Add tree to the core role.

rbarlow:


rbarlow:
I think we can ignore the Jenkies failure.

rbarlow:
ok test

PR #2403, opened 2016-02-09
Return gracefully when DownloadStep.cancel() is called before init.

rbarlow:
DownloadStep.cancel() had unconditionally called
self.downloader.cancel(), which caused an AttributeError if
cancel() was called before intialize(). This would happen, for
example, if a pulp-docker sync were canceled during metadata
fetching.

https://pulp.plan.io/issues/1645

fixes #1645


rbarlow:
ok test

PR #2405, opened 2016-02-09
Rework deferred download guide to include Varnish.

jeremycline:
We do not actually require Squid to be used as the caching proxy. Many
users may be interested in using a different caching proxy for numerous
reasons (performance, familiarity, ease of configuration etc). This just 
makes it clear that the services are interchangeable with other services.

As a side note, Varnish looks very nice. It'd be worth performance testing it side-by-side with Squid.

### DIFF
--- a/docs/user-guide/deferred-download.rst
+++ b/docs/user-guide/deferred-download.rst
@@ -26,21 +26,33 @@ with the Pulp core services.
 
 Deferred downloading relies on three services:
 
-* ``httpd`` - The Apache web server acts as a TLS termination point and reverse proxy.
-  This service handles the incoming client requests and forwards them to Squid.
+* A reverse proxy that terminates the TLS connection. In this guide Apache httpd is configured
+  to act as the reverse proxy, but any reverse proxy that is capable of running a WSGI
+  application should work. This service proxies the requests to the next service, a caching
+  proxy server.
 
-* ``squid`` - The Squid process caches content and de-duplicates client requests so that
-  the content is only downloaded one time.
+* A caching proxy server. In this guide Squid is used, but a simple Varnish configuration is
+  also provided. This service de-duplicates client requests and caches content for Pulp to
+  eventually save to permanent storage. It proxies requests on to the last service, the
+  ``pulp_streamer``
 
-* ``pulp_streamer`` - The pulp_streamer process interacts with Pulp's core services to determine
-  where the content is located and how to download it. It streams the content back to the client
-  through Squid and Apache as it is downloaded.
+* The ``pulp_streamer`` is a streaming proxy service that translates the files in Pulp
+  repositories to their locations in upstream repositories. This service interacts with
+  Pulp's core services to determine where the content is located and how to download it.
+  It streams the content back to the client through Squid and Apache httpd as it is downloaded.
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/deferred-download.rst
```diff
@@ -26,21 +26,33 @@ with the Pulp core services.
 
 Deferred downloading relies on three services:
 
-* ``httpd`` - The Apache web server acts as a TLS termination point and reverse proxy.
-  This service handles the incoming client requests and forwards them to Squid.
+* A reverse proxy that terminates the TLS connection. In this guide Apache is configured
```
rbarlow: You might s/Apache/httpd/g or s/Apache/Apache httpd/g or s/Apache/Apache HTTP Server/g throughout this page. I think most people do know what "Apache" means, but it could be good to use the real name. It's a nitpick, feel free to ignore.

docs/user-guide/deferred-download.rst
```diff
@@ -53,19 +65,87 @@ The packages necessary for deferred downloading can be installed with the follow
...
+make some configuration adjustments so that Pulp redirects clients to the correct host.
+All configuration options are documented inline. Once the Pulp server is configured, the
+Apache reverse proxy, Squid, and the Pulp streamer require configuration.
+
+Reverse Proxy
+^^^^^^^^^^^^^
+
+A default configuration for Apache is provided by the ``python-pulp-streamer`` package
+and is installed to ``/etc/httpd/conf.d/pulp_streamer.conf``, but it is also reproduced
+below for reference::
```
rbarlow: I think there is a way to get Sphinx to automatically inject the file here so that it stays up to date as it changes in the future. Could be worth looking in to.

docs/user-guide/deferred-download.rst
```diff
@@ -193,6 +273,24 @@ following is a basic configuration with inline documentation::
 
 For more information about a configuration option, please consult the Squid documentation.
 
+If you choose to use Varnish instead of Squid, there are two configuration files to look at.
```
rbarlow: You might want to add headings for Squid and Varnish so the user can navigate more easily to the one they want to use.

docs/user-guide/deferred-download.rst
```diff
@@ -193,6 +273,24 @@ following is a basic configuration with inline documentation::
...
+
+  # Default backend definition. Set this to point to the Pulp streamer, which is port 8751
+  # by default.
+  backend default {
+      .host = "127.0.0.1";
+      .port = "8751";
+  }
+
+The second is ``/etc/varnish/varnish.params`` where, among other things, the cache size
+and location is set.
```
rbarlow: Do you know what configs are needed to get Varnish to handle the range headers correctly? If so, can you include those here?

rbarlow:
Very well written, nice work!

docs/user-guide/deferred-download.rst
jeremycline: From my cursory test yesterday (I didn't tcpdump and inspect the request as it bounced from proxy to proxy), it does the right thing and pre-fetches the file. This was with Varnish 4.1.0.

One of Nectar's flaws is that it doesn't let us do Range requests, but if we fixed that (or stopped using Nectar), we wouldn't need that setting.

PR #2407, opened 2016-02-09
Query needs to match the catalog unique index.

jortel:
The query needs to match the catalog index when determining the next revision number.  When a  unit already exists and a plugin is generating catalog entries for a new that I **intends** to add causes duplicate key errors.

Needed for: https://pulp.plan.io/issues/1643


jeremycline:
For what it's worth, this looks good to me.

PR #2408, opened 2016-02-10
Revert merged changes for kombu-3.0.33-3.pulp

pcreech:
Revert merged changes for kombu-3.0.33-3.pulp


pcreech:
Fixed in pulp build scripts

PR #2411, opened 2016-02-10
Switches SELinux netlink_route_socket to use a Refpol macro

bmbouter:
The old statement whitelisted explicit permissions and was not
fully complete. It worked on most distributions but not all.
This Refpol version will use a superset which is maintained in
Refpol and is appropriate for use anywhere Refpol is available.

closes #1484
https://pulp.plan.io/issues/1484

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -36,7 +36,7 @@ require {
 # tcp_socket { shutdown } is needed for CentOS 6 `sudo service pulp_workers restart`
 #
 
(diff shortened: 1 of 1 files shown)


beav:
:penguin: :lock:

bmbouter:
I compiled this myself on EL6, EL7, F22, and F23 so this change should be valid on all desired distros. I also hand-tested the recompiled module on an affected machine and it resolved the issue.

dkliban:
LGTM

PR #2412, opened 2016-02-10
Allow xsendfile to serve /var/lib/pulp/content.

jeremycline:
When I initially tested the directory index, I used lazy and links
resolved to /var/lib/pulp/published/.../. However, if the content
exists, os.path.realpath follows it into /var/lib/pulp/content/.

This is one of the issues blocking issue 1654.

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -11,8 +11,9 @@ from pulp.server.lazy import URL, Key
 from pulp.server.config import config as pulp_conf
 
 
-# Make sure all requested paths fall under this directory.
+# Make sure all requested paths fall under these directories.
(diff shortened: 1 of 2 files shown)


server/pulp/server/content/web/views.py
```diff
@@ -11,8 +11,9 @@
 from pulp.server.config import config as pulp_conf
 
 
-# Make sure all requested paths fall under this directory.
+# Make sure all requested paths fall under these directory.
```
ipanova: I guess it should be 'under this directory' or 'under these directories' If my English knowledge does not betray me :D

jeremycline: Your English knowledge has not betrayed you

PR #2413, opened 2016-02-10
Use the local CLI transport for pulp-smash.

rbarlow:


rbarlow:
I am confident that this Jenkies error is not related to this PR.

PR #2416, opened 2016-02-12
Use SSL_CLIENT_CERT and fall back to mod_ssl.var_lookup.

jeremycline:
Depending on how a request arrives and who is performing the
authentication, `mod_ssl.var_lookup` may not be present and
`SSL_CLIENT_CERT` is. This commit causes the authenticator to look in
both locations for the certificate and if neither can be found, it
returns `False`. Previously, it would cause an HTTP 500 by raising a
KeyError.

closes #1667

### DIFF
--- a/oid_validation/pulp/oid_validation/oid_validation.py
+++ b/oid_validation/pulp/oid_validation/oid_validation.py
@@ -13,7 +13,8 @@ The OID structure follows the Red Hat model. Download URLs are found at:
 The * represents the product ID and is not used as part of this calculation.
 '''
 
-from ConfigParser import NoOptionError, SafeConfigParser
+from gettext import gettext as _
+from ConfigParser import NoOptionError, SafeConfigParser, NoSectionError
 
 from rhsm import certificate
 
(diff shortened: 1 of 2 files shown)


jeremycline:
ok test

oid_validation/pulp/oid_validation/oid_validation.py
```diff
@@ -29,7 +29,15 @@ def authenticate(environ, config=None):
...
-    cert_pem = environ["mod_ssl.var_lookup"]("SSL_CLIENT_CERT")
+
+    # Attempt to retrieve the client certificate, and if it isn't available, reject.
+    cert_pem = ''
+    if 'SSL_CLIENT_CERT' in environ:
+        cert_pem = environ['SSL_CLIENT_CERT']
+    if not cert_pem and 'mod_ssl.var_lookup' in environ:
+        cert_pem = environ["mod_ssl.var_lookup"]("SSL_CLIENT_CERT")
+    if not cert_pem:
+        return False
```
beav: Can debug logs be added to this? It would be difficult to see what is happening without them.

jeremycline: Yeah, that's definitely a good idea.

jeremycline:
@beav I've updated this PR to include some logging. Does it look reasonable to you?

beav:
@jeremycline i would default to false for logging it since its more of a debug thing, but looks good otherwise

oid_validation/pulp/oid_validation/oid_validation.py
```diff
@@ -29,14 +30,33 @@ def authenticate(environ, config=None):
...
+    wsgi_error_logger = environ["wsgi.errors"].write
 
     if config is None:
         config = _config()
 
+    # Attempt to retrieve the client certificate, and if it isn't available, reject.
+    cert_pem = ''
+    if 'SSL_CLIENT_CERT' in environ:
+        cert_pem = environ['SSL_CLIENT_CERT']
+    if not cert_pem and 'mod_ssl.var_lookup' in environ:
```
rbarlow: why not ```elif 'mod_ssl.var_lookup' in environ:```?

rbarlow: I guess you would also need to adjust the first if to be ```if 'SSL_CLIENT_CERT' in environ and environ['SSL_CLIENT_CERT']``` which I guess doesn't save you much typing.

oid_validation/pulp/oid_validation/oid_validation.py
```diff
@@ -29,14 +30,33 @@ def authenticate(environ, config=None):
     '''
     Framework hook method.
     '''
-    cert_pem = environ["mod_ssl.var_lookup"]("SSL_CLIENT_CERT")
+    # An output stream (file-like object) to which error output can be written.
+    # For many servers, this will output to the main error log. For more
+    # information, see PEP333.
+    wsgi_error_logger = environ["wsgi.errors"].write
```
rbarlow: I think this may get written to the httpd logs. Is that right? If so, is that the behavior we want? I guess this is a separate process from Pulp, so it doesn't have the Pulp logger set up. Maybe we could move the logging code to pulp.common if it would be useful to other bits than just pulp.server…? Or this is fine too…

rbarlow:
I think I'm going to have to reject this since you didn't use m2crypto to do any of this.

Just kidding, my buddy John said nice work:

~[Nice work.](http://massappeal.com/wp-content/uploads/2013/04/damn_u_johnclaude.jpg)

oid_validation/pulp/oid_validation/oid_validation.py
jeremycline: It does get written to Apache's error logs.

PR #2417, opened 2016-02-12
Only perform repo auth on HTTPS requests.

jeremycline:
This commit ensures the content.wsgi application only performs repo auth
when the request arrived via HTTPS. Client certificates won't arrive
over HTTP and this was causing access denied errors on unprotected
repositories.

closes #1666

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -137,10 +137,13 @@ class ContentView(View):
         host = request.get_host()
         path = os.path.realpath(request.path_info)
 
-        # Authorization
-        if not allow_access(request.environ, host):
-            # Not Authorized
-            return HttpResponseForbidden()
+        # Check authorization if http isn't being used. This environ variable must
+        # be available in all implementations so it is not dependant on Apache httpd:
+        # https://www.python.org/dev/peps/pep-0333/#environ-variables
+        if request.environ['wsgi.url_scheme'] != 'http':
+            if not allow_access(request.environ, host):
+                # Not Authorized
+                return HttpResponseForbidden()
 
         if not path.startswith(PUBLISH_DIR) and not path.startswith(CONTENT_DIR):
             # Someone is requesting something they shouldn't.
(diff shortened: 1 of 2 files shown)


jeremycline:
ok test

server/pulp/server/content/web/views.py
```diff
@@ -137,10 +137,13 @@ def get(self, request):
...
         path = os.path.realpath(request.path_info)
 
-        # Authorization
-        if not allow_access(request.environ, host):
-            # Not Authorized
-            return HttpResponseForbidden()
+        # Check authorization if http isn't being used. This environ variable must
+        # be available in all implementations so it is not dependant on Apache httpd:
+        # https://www.python.org/dev/peps/pep-0333/#environ-variables
+        if request.environ['wsgi.url_scheme'] != 'http':
```
rbarlow: What will stop users from accessing https content by substituting http:// on the URL? It doesn't seem that self.redirect() checks whether the content should be https protected or not. Does something else in the chain check this?

jeremycline: This isn't great, but it's this Apache snippet in the plugin (just `pulp_rpm` right now):
```
<Location /pulp/repos/>
  RewriteEngine On
  RewriteCond %{HTTPS} on
  RewriteRule (.+/pulp/repos/)(.*) /pulp/content/var/www/pub/yum/https/repos/$2 [DPI]
  RewriteCond %{HTTPS} off
  RewriteRule (.+/pulp/repos/)(.*) /pulp/content/var/www/pub/yum/http/repos/$2 [DPI]
</Location>
```
When we check the existence of `/var/www/pub/yum/http/repos/...` (which is the value of `path_info`) it won't exist unless you've published a protected repo to HTTP in which case wat?

I don't like that the security of this is held together by shoestrings and gum, but so it goes.

rbarlow: Yeah I'm not a fan of that either. Perhaps we can revisit all of this with Pulp 3 when we potentially drop the http/https "feature".

rbarlow:
Nice work!

Nice work!

PR #2422, opened 2016-02-15
Refine playpen/all_27.sh script

Ichimonji10:
The `all_27.sh` script walks through each repository in `~/devel/pulp*`, checks
out an appropriate branch, and either installs or uninstalls the code in that
branch using the `pulp-dev.py` script. This works so long as all of the
repositories found are, in fact, either Pulp or one of its plugins. pulp-smash
is neither. Refine the script to make it more precisely find correct
repositories.

### DIFF
--- a/playpen/all_27.sh
+++ b/playpen/all_27.sh
@@ -17,7 +17,7 @@ declare -rA repos_branches=(
 
 # Reinstall plugins we know how to handle, and uninstall all others.
 pushd ~/devel
-for repo in pulp*; do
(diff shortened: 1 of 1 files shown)


rbarlow:
ok test

rbarlow:
Since this change is entirely in playpen, I'm not gonna wait for Jenkies.

rbarlow:
JCvD says "nice catch!"

nice catch

Ichimonji10:
:smile:

PR #2423, opened 2016-02-15
Update the documented Squid configuration.

jeremycline:
RHEL6 and its derivatives ship with Squid 3.1. This version of Squid
does not define the access control lists used in the documented
configuration by default. This commit adds the necessary definitions
commented out by default with a note about when to uncomment them.

See the default values for each version:
3.1: http://www.squid-cache.org/Versions/v3/3.1/cfgman/acl.html
3.2: http://www.squid-cache.org/Versions/v3/3.2/cfgman/acl.html

### DIFF
--- a/docs/user-guide/deferred-download.rst
+++ b/docs/user-guide/deferred-download.rst
@@ -103,10 +103,18 @@ Squid
 +++++
 
 Squid requires significantly more configuration which is up to the user. To configure squid,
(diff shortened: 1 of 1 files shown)


rbarlow:
Now even 3.1 users are prepared for what lies ahead of them, like JCvD:

be prepared

PR #2426, opened 2016-02-16
Use load() instead of resolve()

pcreech:
Use load() method instead of resovle()


mhrivnak:
I've already fixed this in my branch and made some other changes to the same file, so I suggest just closing this PR. The fix will come shortly along with the dockerization of this script.

PR #2436, opened 2016-02-18
Dockerizing the 2.8 upgrade validation checker

mhrivnak:
https://pulp.plan.io/issues/1416

re #1416

This can be tested now. I'm building from this branch into the correct docker hub repository, so you can follow the instructions in the README exactly.

### DIFF
--- a/playpen/mongoengine/README
+++ b/playpen/mongoengine/README
@@ -1,0 +1,48 @@
+The run.sh script will test your data from pulp < 2.8 to ensure that the
+upgrade process will succeed. It requires you to creata a database dump. It
+does not modify data in your running pulp server, nor does it modify the
+contents of the database dump you provide.
+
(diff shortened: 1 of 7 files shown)


pcreech:
Worked well! 

I did want to note that I had some issues with the mongodump command itself, which were:

- Not installed (i had to install mongo-tools)
- Didn't connect until I specified "-host=127.0.0.1", which appears to be a bug in that version of mongodump

Not sure if you think that would be important at all, since this is probably specific to the fedora environment I ran it on (the vagrant environment, local mongo server), and out of scope of this PR.  but it might be something a user or two has trouble with later.

mhrivnak:
Good point. I'll add a hint about which package contains that command. Mongo packaging can be a bit confusing. On el7, I think mongodump is in the package called "mongodb", which is different than "mongodb-server".

PR #2441, opened 2016-02-19
Make 'working_dir' configurable.

ipanova:


server/pulp/plugins/importer.py
```diff
@@ -32,7 +32,7 @@ def get_downloader(config, url, **options):
         :raise ValueError: when the URL scheme is not supported.
         """
         url = urlparse(url)
-        nectar_config = importer_config_to_nectar_config(config.flatten())
+        nectar_config = importer_config_to_nectar_config(config.flatten(), working_dir='/tmp')
```
bmbouter: I had thought the working_dir would not be specified here, and instead the streamer's use of importers specifically would set `working_dir=/tmp`

bmbouter:
The commit message should have:

```
re #1705
https://pulp.plan.io/issues/1705
```

bmbouter:
Adding my commit onto this and posted back in another PR. This is mostly your code @ipanova . 

https://github.com/pulp/pulp/pull/2442

PR #2442, opened 2016-02-19
1705 - Streamer code explicitly sets /tmp as the temp folder

bmbouter:
I added to a commit from @ipanova.


jeremycline:
You'll need to make adjustments to the unit tests for both the streamer and the importer (some of which are modified in this PR). Other than the unit test changes, this looks good to me. I'm happy, so now you just need to make Jenkies happy.

bmbouter:
ok test

PR #2443, opened 2016-02-19
Fixed a simple argument mismatch bug in content sources code.

mhrivnak:
https://pulp.plan.io/issues/1692
The bug was introduced here: https://github.com/pulp/pulp/pull/2116/files#diff-4e345a77fb83e6a3dbcb29e9e12b7808R65
... without a corresponding change to the calling code.

fixes #1692


mhrivnak:
The test failure appears solidly unrelated, so I'm merging.

PR #2448, opened 2016-02-22
Move the log function definition up in the init script.

jeremycline:
This change is required because there is a log statement above the
function definition.


jeremycline:
ok test

PR #2450, opened 2016-02-22
Fix performance regression in associated unit queries.

jortel:
Returning dictionaries instead of document objects reduces the elapsed time by 30x.  For example: An RPM repository containing 12k units - the time for `get_associated_unit_ids()` is reduced from ~3 seconds to ~100 milliseconds.

### DIFF
--- a/server/test/unit/server/controllers/test_repository.py
+++ b/server/test/unit/server/controllers/test_repository.py
@@ -32,26 +32,26 @@ class DemoModel(model.ContentUnit):
 class TestGetAssociatedUnitIDs(unittest.TestCase):
     def setUp(self):
         self.associations = [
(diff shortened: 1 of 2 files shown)


server/test/unit/server/controllers/test_repository.py
```diff
@@ -37,21 +37,21 @@ def setUp(self):
         ]
 
     def test_returns_ids(self, mock_objects):
-        mock_objects.return_value.only.return_value = self.associations
+        mock_objects.return_value.only.return_value.as_pymongo.return_value = self.associations
```
mhrivnak: I think you should change ``self.associations`` to have the correct new data type. Apparently model instances work correctly with dictionary-style attribute access, but that's not representative of the data that will actually flow through this function.

jortel: Ah, good call.  Fixed.

PR #2453, opened 2016-02-23
Add fedora-easy-karma to the dev environment.

rbarlow:
This CLI tool automatically inspects your box for packages that were installed from the updates-testing repository, and then prompts you for Bodhi
feedback. It's a handy way for us to participate in the updates that are coming our way.


rbarlow:
I am confident that this change did not cause the Jenkins failure, since Jenkins does not use the dev Ansible code.

PR #2455, opened 2016-02-23
psmash returns a useful exit code.

rbarlow:


rbarlow:
I am confident that the test failures are not due to this change, as this only changes the development environment .bashrc which is not used by Jenkies.

PR #1711, opened 2015-03-09
add thank you note for new contributors

beav:
to find the contributors:

```bash
diff <(git shortlog origin/2.5-release origin/2.6-testing -s | cut -f2) \
     <( git shortlog origin/2.5-release -s | cut -f2) \
     --unchanged-group-format='' --changed-group-format='%<'
```

PR #2457, opened 2016-02-24
Add dev_read_urand to the pulp_streamer_t policy.

rbarlow:
This commit adds an SELinux rule to allow the streamer to access
/dev/urandom.

https://pulp.plan.io/issues/1711

fixes #1711

### DIFF
--- a/server/selinux/server/pulp-streamer.te
+++ b/server/selinux/server/pulp-streamer.te
@@ -11,6 +11,7 @@ require {
         type celery_exec_t;
         type squid_t;
         type rpm_exec_t;
+        type pulp_var_run_t;
 }
 
 
@@ -27,6 +28,9 @@ allow streamer_t rpm_exec_t:file getattr_file_perms;
 
 ## Read from /proc/meminfo
 kernel_read_system_state(streamer_t);
+
+## Read from pseudo random number generator devices (e.g., /dev/urandom).
+dev_read_urand(streamer_t)
 
 
 ##### Network #####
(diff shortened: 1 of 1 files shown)


rbarlow:
I have tested this on Fedora 22. I will now test it on RHEL 6.

rbarlow:
The test failure is some kind of qpid EOF error. I don't believe it is related to this change.

rbarlow:
I have tested that the streamer starts without SELinux errors with this policy on EL 6 and Fedora 22.

server/selinux/server/pulp-streamer.te
```diff
@@ -76,3 +80,22 @@ ifdef(`distro_fedora23', `
     fs_rw_inherited_tmpfs_files(squid_t);
     fs_read_tmpfs_files(squid_t);
 ')
+
+
+##### Pulp stuff #####
+
+## Allow access to /var/run/pulp for EL 6
+ifdef(`distro_rhel6', `
+    allow streamer_t pulp_var_run_t:file manage_file_perms;
```
bmbouter: L#89 - L#91 could be turned into a Refpol statement, but I don't off hand know which one it is. We already have the same code in the `pulp-celery` SELinux policy [here](https://github.com/pulp/pulp/blob/a473ddffb18bab5ed224a40198bf4c7cfaed30cf/server/selinux/server/pulp-celery.te#L57-L60) so I say leave it as-is and we can improve both at a later time.

server/selinux/server/pulp-streamer.te
```diff
@@ -76,3 +80,22 @@ ifdef(`distro_fedora23', `
...
+    allow streamer_t pulp_var_run_t:dir manage_dir_perms;
+    files_pid_filetrans(streamer_t, pulp_var_run_t, file)
+')
+
+
+##### Miscellaneous #####
+
+## Allow process to read localization information on EL 6
+ifdef(`distro_rhel6', `
+    miscfiles_read_localization(streamer_t)
```
bmbouter: I'm surprised this was only needed on EL6. Perhaps it's upstart related?

bmbouter:
I compiled and tested it on my yum updated EL6 box. Everything worked with no SELinux denials which I did see before checking out your PR. I did not compile it on the other distros, but this all makes sense so I expect it will compile. Also you compiled it on F22 so we know it compiles there. If something doesn't work Jenkins will tell us like it did for the issue being resolved. Good job!

server/selinux/server/pulp-streamer.te
rbarlow: Yeah I pretty much just stole it from the other one. Let's improve it later as you suggest.

server/selinux/server/pulp-streamer.te
rbarlow: I did not determine why this was the case, but that is one possible explanation. It could also be due to something our init script itself is doing.

rbarlow:
Thanks for the review @bmbouter!

PR #2459, opened 2016-02-25
Set up a CA and a cert for Apache in Vagrant.

jeremycline:
For too long we have lived with `verify_ssl: False` and `curl -k`. No longer. This sets up a CA and certificate for Apache as part of the Ansible setup.

This is just a first pass. A good next step would be to set up all our other services with certificates from this CA.

### DIFF
--- a/playpen/ansible/dev-playbook.yml
+++ b/playpen/ansible/dev-playbook.yml
@@ -8,3 +8,4 @@
     - qpidd
     - dev
     - lazy
+    - pulp_ca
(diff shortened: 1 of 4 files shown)


seandst:
> A good next step would be to add some subjectAltName stuff

Is the issue description out of date?
https://github.com/pulp/pulp/pull/2459/files#diff-aa41559fe4e3c42b79e57bb115dcc083R20

jeremycline:
@seandst The description is now more accurate.

seandst:
YOU LIAR.

jeremycline:
I am ashamed.

seandst:
I've had mixed experiences with lineinfile, where re-running the playbook inserts the line multiple times. I think that re-running the playbook is not supported for our usage of ansible here and this is not a problem, but if it bites us in the future we can deal with it then.

Great work, btw

PR #2460, opened 2016-02-29
Run pulp-smash with python3 and py.test because we can

rbarlow:


rbarlow:
Uh, apparently more is needed on this so hold off on reviewing.

rbarlow:
This is blocked on https://github.com/PulpQE/pulp-smash/issues/146

rbarlow:
I'm removing the LGTM as per my first comment.

rbarlow:
This is no longer blocked. I worked around https://github.com/PulpQE/pulp-smash/issues/146 by switching psmash to use py.test instead of unittest2.

rbarlow:
I am confident that the Jenkies failure is not due to this change.

asmacdo:
@rbarlow, how confident are you that py.test will work exactly the same? Seems like this is creating something that we will need to undo later, and at first glance I don't think that using python 3 now is enough reason to do that.

rbarlow:
Austin wrote:
> @rbarlow <https://github.com/rbarlow>, how confident are you that
> py.test will work exactly the same? Seems like this is creating
> something that we will need to undo later, and at first glance I don't
> think that using python 3 now is enough reason to do that.

I am confident that it does not work exactly the same, since this one
works and unittest2 does not work ☺ However it runs the tests and is
"officially" supported by pulp-smash. I don't have any intention of
removing it later.

PR #1733, opened 2015-03-20
PEP-8 several modules in pulp.bindings.

rbarlow:
This commit PEP-8's the following modules and their corresponding tests:
    pulp.bindings.consumer
    pulp.bindings.exceptions
    pulp.bindings.repo_groups
    pulp.bindings.upload

PR #2467, opened 2016-03-02
Pulp search call raising serialization issues

ipanova:
re #1733
https://pulp.plan.io/issues/1733

All credits to @goosemania as she found the spot.


ipanova:
ok test

PR #2468, opened 2016-03-02
Move `list_files()` from FileContentUnit to ContentUnit.

jeremycline:
Move the `list_files()` method definition to ContentUnit. This is
convenient when receiving a list of ContentUnit derivatives that may or
may not have inherited from FileContentUnit. The default implementation
on ContentUnit returns an empty list unless `_storage_path` is defined,
so a user can call `list_files` on any ContentUnit subclass when dealing
with content files.

closes #1735


asmacdo:
@jeremycline, did you intend this to go into 2.8.1? If so, LGTM!

PR #2470, opened 2016-03-03
Create mongoengine indexes only after migrations

asmacdo:
Fixes a bug that broke upgrades due to the fact that indexes were being
generated for newly renamed fields before the migrations were run
populating the indexes with 'null' and breaking upgrades if there are
uniqueness contstraints.

closes https://pulp.plan.io/issues/1738

### DIFF
--- a/server/pulp/plugins/loader/api.py
+++ b/server/pulp/plugins/loader/api.py
@@ -432,15 +432,18 @@ def load_content_types(types_dir=_TYPES_DIR, dry_run=False, drop_indices=False):
 
     # to handle node.json only
     descriptors = _load_type_descriptors(types_dir)
-    definitions = parser.parse(descriptors)
+    pre_mongoengine_definitions = parser.parse(descriptors)
 
     # get information about content unit types from entry points
-    definitions += _generate_plugin_definitions()
+    mongoengine_definitions = _generate_plugin_definitions()
 
     if dry_run:
-        return _check_content_definitions(definitions)
+        return _check_content_definitions(pre_mongoengine_definitions + mongoengine_definitions)
     else:
-        database.update_database(definitions, drop_indices=drop_indices)
+        database.update_database(pre_mongoengine_definitions, drop_indices=drop_indices,
+                                 create_indexes=True)
+        database.update_database(mongoengine_definitions, drop_indices=drop_indices,
+                                 create_indexes=False)
 
 # initialization methods -------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


server/pulp/plugins/types/database.py
```diff
@@ -48,18 +48,21 @@ def __str__(self):
         return 'MissingDefinitions [%s]' % ', '.join(self.missing_type_ids)
 
 
-def update_database(definitions, error_on_missing_definitions=False, drop_indices=False):
+def update_database(definitions, error_on_missing_definitions=False, drop_indices=False,
+                    create_indexes=False):
```
bmbouter: I'll suggest this should this default to `True` since that was the previous behavior.

bmbouter: Are there any other usages of `update_database` besides those two calls in `api.py`?

server/pulp/plugins/loader/api.py
```diff
@@ -432,15 +432,16 @@ def load_content_types(types_dir=_TYPES_DIR, dry_run=False, drop_indices=False):
 
     # to handle node.json only
     descriptors = _load_type_descriptors(types_dir)
-    definitions = parser.parse(descriptors)
+    old_definitions = parser.parse(descriptors)
```
bmbouter: let's make this variable name better.   s/old_definitions/non_mongoengine_definitions/

server/pulp/plugins/loader/api.py
```diff
@@ -432,15 +432,16 @@ def load_content_types(types_dir=_TYPES_DIR, dry_run=False, drop_indices=False):
...
 
     # get information about content unit types from entry points
-    definitions += _generate_plugin_definitions()
+    new_definitions = _generate_plugin_definitions()
```
bmbouter: let's make this variable name better.   s/new_definitions/mongoengine_definitions/

bmbouter:
Thanks for fixing this @asmacdo A few suggestions for improvements. I would like to see them made, but you can handle as you like.

server/pulp/plugins/types/database.py
asmacdo: @bmbouter, only in tests. But I agree, still safer.

rbarlow:
Seems fine to me, as it seems that the mongoengine indices do get created after the migrations are run:

https://github.com/pulp/pulp/blob/53b871f0ce6257fb506812114a54b60e856285e2/server/pulp/server/db/manage.py#L142

asmacdo:
Since jenkies isn't hooked up to this branch, running the tests here: https://github.com/pulp/pulp/pull/2471

PR #1738, opened 2015-03-23
Introducing Content sources view

ipanova:
`GET /pulp/api/v2/content/sources/` http://fpaste.org/201489/27123424/

`GET /pulp/api/v2/content/sources/source-id/ 200 ` http://fpaste.org/201490/71234781/ fixes [802] (https://pulp.plan.io/issues/802)
`GET /pulp/api/v2/content/sources/source-id/ 404 ` http://fpaste.org/201492/14271235/

`POST /pulp/api/v2/content/sources/action/refresh/ 202` http://fpaste.org/201495/14271236/
`POST /pulp/api/v2/content/sources/action/refresh/source-id 202` http://fpaste.org/201498/12378014/
`POST /pulp/api/v2/content/sources/action/refresh/source-id 404` http://fpaste.org/201496/42712374/

PR #2471, opened 2016-03-03
[for jenkies]

asmacdo:
Fixes a bug that broke upgrades due to the fact that indexes were being
generated for newly renamed fields before the migrations were run
populating the indexes with 'null' and breaking upgrades if there are
uniqueness contstraints.

closes #1738


asmacdo:
PR used to run tests. Tests pass, so closing.

PR #2473, opened 2016-03-04
Enable logging for the content.wsgi application.

jeremycline:
This commit configures loggers for the content.wsgi application. This
ensures unhandled exceptions are logged to syslog and that logging is
configured so messages can be sent from the views.

closes #1739

### DIFF
--- a/server/pulp/server/content/web/settings.py
+++ b/server/pulp/server/content/web/settings.py
@@ -1,4 +1,47 @@
+import ConfigParser
+import logging
 import os
+
+from pulp.server import logs as pulp_logs
+from pulp.server import config
+
+
+try:
+    log_level = config.config.get('server', 'log_level')
+    log_level = getattr(logging, log_level.upper())
+except (ConfigParser.NoOptionError, AttributeError):
+    # If the user didn't provide a log level, or if they provided an invalid one, let's use the
+    # default log level
+    log_level = pulp_logs.DEFAULT_LOG_LEVEL
+
+# Rather than using `pulp.server.logs.start_logging`, which dates back to the
+# pre-Django days, use Django to set up the logging. This is intended to mimic
+# the functionality of `start_logging`.
+LOGGING = {
+    'version': 1,
+    'formatters': {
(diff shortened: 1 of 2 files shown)


server/pulp/server/content/web/settings.py
```diff
@@ -1,5 +1,45 @@
...
+
+try:
+    log_level = config.config.get('server', 'log_level')
+    log_level = getattr(logging, log_level.upper())
+except (ConfigParser.NoOptionError, AttributeError):
+    # If the user didn't provide a log level, or if they provided an invalid one, let's use the
+    # default log level
+    log_level = pulp_logs.DEFAULT_LOG_LEVEL
+
+LOGGING = {
```
seandst: This appears to be duplicating the work in `pulp.server.logs.start_logging`; does django provide special handling of this LOGGER attr that you're using to emulate `start_logging`'s behavior? If so, could a comment be added to indicate why start_logging is used, and, if needed, if `start_logging` should be used as a reference for configuring django logging?

server/pulp/server/content/web/views.py
```diff
@@ -143,23 +147,30 @@ def get(self, request):
         if request.environ['wsgi.url_scheme'] != 'http':
             if not allow_access(request.environ, host):
                 # Not Authorized
+                logger.debug(_('Denying {host} access to {path}.').format(host=host, path=path))
```
seandst: If possible, it might be good to say why the access is denied (e.g. because of ACLs, or whatever). Also, this might be better at loglevel INFO.

server/pulp/server/content/web/settings.py
jeremycline: It does mimic the behaviour of `start_logging`. The reason I chose to use Django's facilities is that it handles actually setting up the logs, where-as if I wanted to use `start_logging` as far as I could tell I would need to make use of `pulp.server.webservices.application` or create something similar, and things got more complicated from there.

My gut told me using Django to do this was the "right" way to do it, but I'm far from an expert on this.

PR #2475, opened 2016-03-07
switching from nightlies to beta for the 2.8 upgrade check script

mhrivnak:


mhrivnak:
Since I'm the only person working tonight (I assume), I'm going out on a limb and merging this without review since it's just a playpen item, and we need to get docker hub building this image ASAP.

PR #2476, opened 2016-03-11
Source a second bash `run commands` file

asmacdo:
Allows a developer to add their own bash commands in a ~/devel/.localrc
if it exists.

### DIFF
--- a/playpen/ansible/roles/dev/files/bashrc
+++ b/playpen/ansible/roles/dev/files/bashrc
@@ -5,6 +5,11 @@ if [ -f /etc/bashrc ]; then
     . /etc/bashrc
 fi
 
+# Source independant bashrc additions if they exist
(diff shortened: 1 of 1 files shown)


rbarlow:
You can already do customization with your Vagrantfile that's more general and more powerful than this. I don't think we should make the .bashrc more complex to add another way to do something that you can already do.

rbarlow:
The Vagrantfile is intended for each dev to customize their dev box however they want and that's why it's not in source control. That's the official way to do this. Just put this or similar in your Vagrantfile:

```
dev.vm.provision "shell", inline: "echo \"source ~/devel/austin.sh\" > ~/.bashrc"
```

asmacdo:
@rbarlow, yeah, works for me

PR #2477, opened 2016-03-11
Allow the Celery process to resolve domain names.

rbarlow:
Fedora Rawhide restricts access to /etc/resolv.conf. This commit
adds the sysnet_dns_name_resolve policy to the celery_t context
which allows it to resolve hostnames.

https://pulp.plan.io/issues/1764

fixes #1764

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -71,9 +71,7 @@ ifndef(`distro_rhel6', `
 # pulp_ostree sync call on F23 calls out to gpg
 # temporary workaround for https://pulp.plan.io/issues/1731
 
-ifdef(`distro_fedora23', `
-    allow celery_t self:process { setrlimit };
-')
+allow celery_t self:process { setrlimit };
 
 ######################################
 


rbarlow:
I have only tested startup with this policy on Fedora Rawhide so far, so it is not ready to be merged. I will run pulp-smash against Rawhide with this in place. If it passes, I will test EL 6 next.

bmbouter:
We're sure this doesn't need to go into 2.8.0? I sounds like it doesn't, but I wanted to ask. If it does need to go in then the PR should not target master.

As long as this will compile on EL6, I think it's exactly the right resolution to the issue. Once you test that it compiles there go ahead and merge it!

rbarlow:
@bmbouter it does need to go into 2.8.0 for Fedora Rawhide and almost certainly 24, but not for our other OSes. Therefore, I'm putting it into 2.8.1 intentionally and patching it downstream.

bmbouter:
@rbarlow very good

rbarlow:
@bmbouter After running pulp-smash on Rawhide, I see one more recommendation in audit2allow:

```allow celery_t self:process setrlimit;```

I don't see a refpol that made sense to me for this, but it seems that something in our worker is calling setrlimit during pulp-smash:

http://linux.die.net/man/2/setrlimit

Unless you can find a refpol that makes sense, I think I'll just add the recommended statement.

bmbouter:
Interesting! I had to add this statement already, but it's guarded to only make it enabled for F23. I think we should enable it for F23+ which can be done with an `ifndef` that checks against EL6, EL7, and F22.

https://github.com/pulp/pulp/blob/master/server/selinux/server/pulp-celery.te#L75

That was a bugfix for .https://pulp.plan.io/issues/1731

PR #1747, opened 2015-03-25
Updated doc to reflect fedora21 is supported

Jmainguy:

PR #2482, opened 2016-03-14
Raise an exception if upload report indicates failure

goosemania:
https://pulp.plan.io/issues/1747
closes #1747

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -94,6 +94,7 @@ PLP0031 = Error("PLP0031", _("Content source %(id)s could not be found at %(url)
 PLP0032 = Error(
     "PLP0032", _("Task %(task_id)s encountered one or more failures during execution."),
     ['task_id'])
+PLP0033 = Error("PLP0033", _("The target importer does not support the types from the source"), [])
 PLP0034 = Error("PLP0034", _("The distributor %(distributor_id)s indicated a failed response when "
                              "publishing repository %(repository_id)s."),
                 ['distributor_id', 'repository_id'])
@@ -112,7 +113,9 @@ PLP0041 = Error("PLP0041", _("Database 'replica_set' config must be specified wh
 PLP0042 = Error("PLP0042", _("This request is forbidden."), [])
 PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
                              "Refer to /etc/pulp/server.conf for proper use."), [])
-PLP0044 = Error("PLP0033", _("The target importer does not support the types from the source"), [])
+PLP0044 = Error("PLP0044", _("The importer %(importer_id)s indicated a failed response when "
+                             "uploading %(unit_type)s unit to repository %(repository_id)s."),
+                ['importer_id', 'unit_type', 'repository_id'])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
--- a/server/pulp/server/managers/content/upload.py
+++ b/server/pulp/server/managers/content/upload.py
@@ -7,6 +7,7 @@ from uuid import uuid4
 
 from celery import task
 
+from pulp.common import error_codes
 from pulp.plugins.conduits.upload import UploadConduit
 from pulp.plugins.config import PluginCallConfiguration
 from pulp.plugins.loader import api as plugin_api, exceptions as plugin_exceptions
@@ -14,7 +15,7 @@ from pulp.server import config as pulp_config
 from pulp.server.async.tasks import Task
 from pulp.server.db import model
 from pulp.server.exceptions import (PulpDataException, MissingResource, PulpExecutionException,
-                                    PulpException)
+                                    PulpException, PulpCodedException)
 from pulp.server.controllers import repository as repo_controller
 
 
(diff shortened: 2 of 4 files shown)


common/pulp/common/error_codes.py
```diff
@@ -113,6 +113,9 @@
 PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
                              "Refer to /etc/pulp/server.conf for proper use."), [])
 PLP0044 = Error("PLP0033", _("The target importer does not support the types from the source"), [])
+PLP0045 = Error("PLP0045", _("The importer %(importer_id)s indicated a failed response when "
+                             "uploading %(unit_type)s unit to repository %(repository_id)s."),
+                ['importer_id', 'unit_type', 'repository_id'])
```
rbarlow: The line above this seems to be strange - the error is 0044 but the text is 0033. Furthermore, there does not seem to be an 0033 above. I wonder if this means that the one above this was really supposed to be 0033, which would mean your new one should be 0044. What do you think?

rbarlow: Well, I guess nothing says it's wrong to have a skip in the number schemes, but since the user-facing number for 0044 is actually 0033, it might make sense to fix that here.

server/test/unit/server/managers/content/test_upload.py
```diff
@@ -223,6 +223,17 @@ def test_import_uploaded_unit_importer_error_reraise_pulp_exception(self, mock_r
...
+    @mock.patch('pulp.server.controllers.importer.model.Repository.objects')
+    def test_import_uploaded_unit_success_flag_false(self, mock_repo_qs):
+        """Test that exception is raised if upload report indicates failure."""
+        importer_controller.set_importer('repo-u', 'mock-importer', {})
+        importer_return_report = {'success_flag': False, 'summary': '', 'details': {}}
+        mock_plugins.MOCK_IMPORTER.upload_unit.side_effect = None
+        mock_plugins.MOCK_IMPORTER.upload_unit.return_value = importer_return_report
+        upload_id = self.upload_manager.initialize_upload()
+        self.assertRaises(PulpCodedException, self.upload_manager.import_uploaded_unit, 'repo-u',
+                          'mock-type', {}, {}, upload_id)
```
rbarlow: If you catch the Exception or use the with statement to grab it, you can assert that the raised Exception has the correct values inside it (code number, error message, etc.)

rbarlow:
Hello @goosemania!

Looks fine, but @mhrivnak is going to finish the review. I do think you should sort out the error number and add stronger assertions to the unit test before merging.

server/pulp/server/managers/content/upload.py
```diff
@@ -207,6 +208,13 @@ def import_uploaded_unit(repo_id, unit_type_id, unit_key, unit_metadata, upload_
         try:
             result = importer_instance.upload_unit(transfer_repo, unit_type_id, unit_key,
                                                    unit_metadata, file_path, conduit, call_config)
+            if not result['success_flag']:
```
mhrivnak: Unfortunately, I noticed that at least one plugin (``pulp_docker``) does not return a dictionary as it should. It return ``None``. That's been ok so far, because nothing has ever checked the value of ``result`` here or down the stack.

One option is to just fix pulp_docker, and check the other plugins to make sure they return the right thing.

Another option is to be defensive, and skip this check in case ``result`` is the wrong type or doesn't have that attribute.

server/test/unit/server/managers/content/test_upload.py
mhrivnak: Great point. I like using ``assertRaises`` as a context manager a lot for this purpose.

common/pulp/common/error_codes.py
goosemania: I noticed that as well, but I didn't fix it in this PR since it is not related to the issue. Do you think it is better to fix it here?

server/test/unit/server/managers/content/test_upload.py
goosemania: I definitely prefer with statement, but all the tests are rather old(?) in this file and they are written that way.
So I thought that maybe it would be easier to rewrite all these unittests at once someday and not to add one new-style test to it. What do you think?

common/pulp/common/error_codes.py
goosemania: ok, I see your last comment

server/pulp/server/managers/content/upload.py
goosemania: I am not sure if skipping wrong `result` is an option, because we won't be able to tell the user if we succeed in importing unit or not.

common/pulp/common/error_codes.py
rbarlow: I only think it makes sense to fix here since it would probably make sense for you to use number 0044. It's not a big deal, but it should be easy to fix so I think it's good to do in this change.

server/test/unit/server/managers/content/test_upload.py
rbarlow: I think it's better to add the assertions than to stick with the existing style in this module.

common/pulp/common/error_codes.py
rbarlow: One more thought - we often do one commit per pull request but that's not written in stone. In order to be more atomic, you could fix the incorrect number in one commit, and then rebase the rest of your change as a second commit. Then it's nice and clearly separated changes, and they can go into one pull request.

In my opinion, it's commits that need to be atomic, not pull requests. Pull requests should *usually* be atomic, but there are times it makes sense to bring two changes together and this is an example of one of those times.

It's up to you though, I think either path is fine.

goosemania:
ok test

dkliban:
I am closing this PR in preparation for making 2.8-dev branch. Please reopen this PR against 2.8-dev branch.

dkliban:
I reopened this PR against 2.8-dev branch https://github.com/pulp/pulp/pull/2522

PR #1734, opened 2015-03-20
Fix pulp-dev.py for CentOS 7.

rbarlow:

PR #2485, opened 2016-03-14
Ensure a proper HTTP status code in the streamer on failure

jeremycline:
If a non-HTTP error occurs in Nectar the streamer would fail to return a
reasonable status code. An example of this is if the domain name is
unresolvable or the host is not accepting connections. This resulted in
an HTTP 200 from the streamer. It is now a HTTP 503.

closes #1734

### DIFF
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -1,5 +1,5 @@
 from gettext import gettext as _
-from httplib import NOT_FOUND, INTERNAL_SERVER_ERROR
+from httplib import NOT_FOUND, INTERNAL_SERVER_ERROR, SERVICE_UNAVAILABLE
 from urlparse import urlparse
 import logging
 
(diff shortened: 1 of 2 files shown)


streamer/pulp/streamer/server.py
```diff
@@ -95,16 +95,15 @@ def download_failed(self, report):
         :param report: The download report for this request.
         :type  report: nectar.report.DownloadReport
         """
-        error_report = report.error_report.copy()
-        error_report['url'] = report.url
-        logger.info(_('Download of %(url)s failed: HTTP %(response_code)s '
-                      '%(response_msg)s' % error_report))
```
rbarlow: Should we retain this log statement, but move it into the if block below so we can be sure there is a response_code?

streamer/pulp/streamer/server.py
```diff
@@ -95,16 +95,15 @@ def download_failed(self, report):
...
-
         # Currently Nectar returns headers with a content-length even
         # when it doesn't download anything.
         self.request.setHeader('Content-Length', '0')
         if 'response_code' in report.error_report:
             self.request.setResponseCode(report.error_report['response_code'])
+        else:
+            # Nectar doesn't give us a good way to know exactly went wrong; return
+            # a generic HTTP 503 and hope Nectar logged enough to be useful
+            self.request.setResponseCode(SERVICE_UNAVAILABLE)
```
rbarlow: Maybe add a log statement here? Not sure what you would say though, so maybe not worth it. "Something went wrong, but we don't know what it is" isn't a great message.

streamer/pulp/streamer/server.py
jeremycline: I thought about keeping it, but did some manual testing and realized that Nectar will log pretty much the same thing (https://github.com/pulp/nectar/blob/master/nectar/downloaders/threaded.py#L225-L226) if it's an HTTP status that isn't 200, and urllib3 logs connection-related problems.

streamer/pulp/streamer/server.py
jeremycline: I couldn't find anything useful in the report that wasn't already logged by urllib3 or nectar

PR #2486, opened 2016-03-14
Do not use autospec with the mock MongoEngine objects

jeremycline:
This was causing code to execute that required a database connection,
but the tests themselves never used the database.

closes #1494


rbarlow:
I'd put that in prod.

rbarlow:
JCvD

PR #2487, opened 2016-03-14
Clean out the squid cache as part of pclean.

jeremycline:


playpen/ansible/roles/dev/files/bashrc
```diff
@@ -82,6 +82,16 @@ pclean() {
     done
     popd
 
+    # Scrub out the Squid cache
+    if [ -d /var/spool/squid ]; then
+        sudo systemctl stop squid
+        sudo rm -rf /var/spool/squid
+        sudo mkdir --context=system_u:object_r:squid_cache_t:s0 --mode=750 /var/spool/squid
```
rbarlow: Would it be easier to rm -rf /var/spool/squid/* so you don't need to mkdir it again? Of course, that will only work if regular users can ls that folder so that glob matches. If it doesn't, I bet you can sudo bash -c "rm -rf /var/spool/squid/*". Might make things a little simpler.

playpen/ansible/roles/dev/files/bashrc
```diff
@@ -82,6 +82,16 @@ pclean() {
...
+        sudo chown squid:squid /var/spool/squid
```
rbarlow: If you can get my above suggestion working you won't need this either.

playpen/ansible/roles/dev/files/bashrc
```diff
@@ -82,6 +82,16 @@ pclean() {
...
+        sudo squid -z
```
rbarlow: I think it's strange that squid has this flag. If you don't run it, does starting squid fail? I don't mind it being here, just remarking that it's weird to me.

rbarlow:
I recommend playing with globbing so you don't have to recreate those folders so carefully. It's up to you, if you like it better this way that's fine too.

playpen/ansible/roles/dev/files/bashrc
jeremycline: It is indeed not readable by regular users which is why I resorted to removing the directory, but this is better.

playpen/ansible/roles/dev/files/bashrc
jeremycline: I think it might have failed on older platforms, but I'm not longer sure since it's been so long. It definitely doesn't fail on F23 (which as we all know is old news, it's all about F24 these days) so I will remove it.

rbarlow: On Mon, Mar 21, 2016 at 11:56:48AM -0700, Jeremy Cline wrote:
> It definitely doesn't fail on F23 (which as we all know is old news, it's all about F24 these
> days)…

Jeremy, what we all know is that Fedora 24 is old news. It's all about
F25 bro!

-- 
Randy Barlow
irc:   bowlofeggs

PR #2495, opened 2016-03-17
Allow the Celery process to resolve domain names.

rbarlow:
Fedora Rawhide restricts access to /etc/resolv.conf. This commit
adds the auth_use_nss policy to the celery_t context
which allows it to resolve hostnames.

https://pulp.plan.io/issues/1764

fixes #1764


jeremycline:
I spent a long time digging through the selinux-policy-doc pages before I could find the statement in question, so you have my deep thanks for taking on this work.

PR #2498, opened 2016-03-23
Adds upgrade note about python-semantic-version

dkliban:


docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -76,6 +76,18 @@ https://raw.githubusercontent.com/pulp/pulp/pulp-2.8.0-1/playpen/mongoengine/REA
...
+    When upgrading on CentOS7, it is possible that your system has 'python-semantic-version'
+    package installed. As result you will experience a problem when updating Pulp packages. If this
+    package is present on your system you should remove it::
+
+        sudo yum remove python-semantic-version
+
+    This will remove a number of pulp packages also. You will install the new packages::
+
+        sudo yum update
+        sudo yum groupinstall pulp-server-qpid
```
rbarlow: I think this is a fine proposal, but I want to suggest an alternative for your consideration. I think we might be able to add an Obsoletes statement to our spec file to remove python-semantic-version with the update, which won't require users to remove and reinstall Pulp. Admittedly, it's a pretty strange way to use Obsoletes, but I think it will make the upgrade more "normal" for users.

What you you think?

rbarlow: To be clear, either way is fine with me. I just wanted to throw another (crazy) idea out there.

rbarlow: One more thought - perhaps we could put this in for now, and then have another release ready to go that solves it some technical way.

I'll LGTM it since it's affecting current users now. We can take the time to solve it another way next if we want to.

bmbouter: +1 to doing these docs now and fixing this right in 2.8.2

docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -64,6 +64,12 @@ Bugs
 Known Issues
 ------------
 
+* CentOS 7 users may experience a problem when upgrading. Please refer to note in upgrade
+  instructions for workaround.
+
+* Users that have `/var/lib/pulp` directory symlinked will experience a
```
bmbouter: s/that have/that have the/
It also needs to say, "or one of it's subdirectories"
Also there is a sphinx rule that allows for linking by ID only instead of having the entire URL hardcoded in the docs

docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -64,6 +64,12 @@ Bugs
...
+  `problem <https://pulp.plan.io/issues/1791>`_. Bind mounting this directory will resolve the issue.
```
bmbouter: s/Bind mounting this directory/Replacing the symlink with an equivalent bind mount"

docs/user-guide/release-notes/2.8.x.rst
ipanova: @bmbouter @rbarlow The spec file already has  `Obsoletes` https://github.com/pulp/pulp/blob/master/deps/python-semantic_version/python-semantic_version.spec#L13

rbarlow:
@ipanova My suggestion was to put the obsolete statement in the pulp.spec file so that Pulp itself would obsolete the sem-ver package. The sem-ver package's obsoletes statement won't help upgrading users because it's in the package that needs to be removed.

However, your observation does raise a question in my mind: what will happen if Pulp requires sem_ver, obsoletes sem-ver, and sem-ver (which is already installed for upgrading users) obsoletes sem_ver (the package that users now need)? I'm not sure what would happen, but it does seem complicated.

PR #2499, opened 2016-03-28
Streamer download requests use permanently stored pem files.

rbarlow:
This commit begins a shift towards storing repository PEM data (CA
certificates, client certificates, and client keys) on the
filesystem, and passing paths to those certs to Nectar rather than
allowing nectar to write them in temporary files.

This commit provides the infrastructure for this change, but only
converts the streamer to using it. All other downloads in Pulp
still take the old, now deprecated path.

https://pulp.plan.io/issues/1771

fixes #1771

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -454,14 +454,17 @@ def queue_delete(repo_id):
 
 def get_importer_by_id(object_id):
     """
-    Get a plugin and call configuration using the document ID
+    Get a plugin, call configuration, and Importer document object using the document ID
     of the repository-importer association document.
 
     :param object_id: The document ID.
-    :type object_id: str
+    :type  object_id: str
+
     :return: A tuple of:
-        (pulp.plugins.importer.Importer, pulp.plugins.config.PluginCallConfiguration)
-    :rtype: tuple
+        (pulp.plugins.importer.Importer, pulp.plugins.config.PluginCallConfiguration,
+         pulp.server.db.model.Importer)
+    :rtype:  tuple
+
     :raise pulp.plugins.loader.exceptions.PluginNotFound: not found.
     """
     try:
@@ -474,7 +477,7 @@ def get_importer_by_id(object_id):
         raise plugin_exceptions.PluginNotFound()
     plugin, cfg = plugin_api.get_importer_by_id(document.importer_type_id)
     call_conf = PluginCallConfiguration(cfg, document.config)
-    return plugin, call_conf
+    return plugin, call_conf, document
 
 
 @celery.task(base=Task, name='pulp.server.tasks.repository.delete')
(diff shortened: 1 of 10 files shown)


rbarlow:
This pull request is not ready for review, but I wanted to file it so that some interested parties could use it to have a working patch if they didn't want to wait for me to finish.

server/pulp/server/controllers/repository.py
```diff
@@ -454,14 +454,17 @@ def queue_delete(repo_id):
 
 def get_importer_by_id(object_id):
     """
-    Get a plugin and call configuration using the document ID
+    Get a plugin, call configuration, and Importer document object using the document ID
```
jeremycline: Oxford comma :+1:

server/test/unit/server/db/test_model.py
```diff
@@ -935,6 +944,225 @@ def test_pre_delete(self, mock_lazy):
...
+    def test_delete_with_certs(self):
+        """
+        Ensure that the delete() method cleans up the Importer's local storage.
+        """
+        config = {
+            importer_constants.KEY_SSL_CA_CERT: 'CA Cert',
+            importer_constants.KEY_SSL_CLIENT_CERT: 'Client Cert',
+            importer_constants.KEY_SSL_CLIENT_KEY: 'Client Key'}
+        importer = model.Importer(repo_id='coolcars', importer_type_id='supercar', config=config)
+        temp_path = tempfile.mkdtemp()
```
jeremycline: Somebody told me about https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory recently. Too bad we can't use Python 3.

jeremycline:
Looks good to me. I got worried for a second with the ``get_importer_by_id`` signature change, but it seems we added that especially for lazy. Who knew?

Anyway, nice job. I'm not sure why Jenkies is unhappy, but it doesn't look to be related to this PR.

server/test/unit/server/db/test_model.py
rbarlow: Haha, wasn't that me, yesterday?

rbarlow:
Yeah that change also made me nervous, but I did grep for other uses and found that only the streamer uses it. Unfortunately, there is another function with the same name in pulp.plugins.something so the grep was slightly tricky.

server/test/unit/server/db/test_model.py
jeremycline: Yeah, I was trying to pour salt into the wound, but I guess my text tone was just a little off :smile:

server/pulp/server/controllers/repository.py
rbarlow: I knew you would like that @jeremycline.

http://www.theonion.com/article/4-copy-editors-killed-in-ongoing-ap-style-chicago--30806

server/test/unit/server/db/test_model.py
rbarlow: I really wish we could use Python 3 though ☹

PR #2501, opened 2016-03-30
Ensure streamer logging is clear about database lookup failures

jeremycline:
The commit ensures that failures to find a catalog entry and failures to
find the content unit referenced by a catalog entry are logged much more
clearly.

I ran across a case where the content unit referenced by a catalog entry was missing from the database. That's bad, and I still don't know how it happened, but this ensures the logging is much clearer. Previously it would complain that the catalog entry was missing even when it was, in fact, the content unit. Both these cases shouldn't happen unless the database is very wrong, so I've made them both error-level.


asmacdo:
ok test

PR #2504, opened 2016-03-31
Modify the content web view to account for symbolic links

jeremycline:
Users sometimes make /var/lib/pulp a symbolic link. This worked fine
until we introduced the content wsgi application for lazy syncing. This
modifies it to account for symbolic links when deciding whether or not
it's safe to serve a file. It also documents the necessary Apache
changes that need to take place to make symbolic links work.

closes #1791

### DIFF
--- a/server/pulp/server/content/web/views.py
+++ b/server/pulp/server/content/web/views.py
@@ -9,15 +9,13 @@ from django.shortcuts import render_to_response
 from django.views.generic import View
 
 from pulp.repoauth.wsgi import allow_access
-from pulp.server.lazy import URL, Key
 from pulp.server.config import config as pulp_conf
+from pulp.server.lazy import URL, Key
 
 
 logger = logging.getLogger(__name__)
 
-# Make sure all requested paths fall under these directories.
-PUBLISH_DIR = '/var/lib/pulp/published'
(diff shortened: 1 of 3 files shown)


server/pulp/server/content/web/views.py
```diff
@@ -128,6 +127,13 @@ def redirect(request, key):
     def __init__(self, **kwargs):
         super(ContentView, self).__init__(**kwargs)
         self.key = Key.load(pulp_conf.get('authentication', 'rsa_key'))
+        # Make sure all requested paths fall under these sub-directories, otherwise
+        # we might find ourselves serving private keys to all and sundry.
+        self.safe_serving_paths = [os.path.realpath(os.path.join(LOCAL_STORAGE, subdir))
```
jortel: Should use `config.get('server', 'storage_dir')` instead of `constants.LOCAL_STORAGE`.  In fact, `constants.LOCAL_STORAGE` should not exist but that's outside the scope of this PR.

jeremycline: Is this setting excluded from our configuration[0] on purpose? I can't imagine why it would be excluded, but it's not in our default configuration file. It is in the default settings dictionary, though.

[0] https://github.com/pulp/pulp/blob/master/server/etc/pulp/server.conf#L78

PR #2505, opened 2016-03-31
Add a migration that writes Importer TLS files to local storage.

rbarlow:
https://pulp.plan.io/issues/1771

fixes #1771

### DIFF
--- a/server/pulp/server/db/migrations/0023_importer_tls_storage.py
+++ b/server/pulp/server/db/migrations/0023_importer_tls_storage.py
@@ -1,0 +1,79 @@
+"""
+This migration inspects all the Importers and writes any TLS certificates or keys they have to local
+storage, as required by the streamer.
+"""
+import errno
+import os
+
+from pulp.server.db import connection
+
+
+LOCAL_STORAGE = os.path.join('/', 'var', 'lib', 'pulp')
+
+
+def migrate():
+    """
+    Write the TLS certificates and keys for each Importer to local storage.
+    """
+    for importer in connection._DATABASE.repo_importers.find():
+        pki_path = _pki_path(importer['repo_id'], importer['importer_type_id'])
+        pem_key_paths = (
+            ('ssl_ca_cert', 'ca.crt'),
+            ('ssl_client_cert', 'client.crt'),
+            ('ssl_client_key', 'client.key'))
+        for key, filename in pem_key_paths:
+            _write_pem_file(pki_path, importer['config'], key, filename)
+
+
+def _mkdir(path):
+    """
+    Create the specified directory, ignoring if the path exists.
+
+    :param path: The absolute path to the directory.
+    :type  path: str
+    """
+    try:
+        os.makedirs(path)
+    except OSError, e:
+        if e.errno != errno.EEXIST:
+            raise
+
+
+def _pki_path(repo_id, importer_type_id):
+    """
+    Return the path where an importer's PKI data should be written to disk.
+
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/migrations/0023_importer_tls_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        pki_path = _pki_path(importer['repo_id'], importer['importer_type_id'])
+        pem_key_paths = (
+            ('ssl_ca_cert', 'ca.crt'),
+            ('ssl_client_cert', 'client.crt'),
+            ('ssl_client_key', 'client.key'))
+        for key, filename in pem_key_paths:
+            _write_pem_file(pki_path, importer['config'], key, filename)
+
+
+def _mkdir(path):
```
asmacdo: suggest rename to `_makedirs` to be explicit about use of `os.makedirs` and `os.mkdir`.

asmacdo: suggest rename to `_makedirs` to be explicit about use of `os.makedirs` and `os.mkdir`.

server/pulp/server/db/migrations/0023_importer_tls_storage.py
```diff
@@ -0,0 +1,79 @@
...
+    :param pki_path:   The base path where the pem file should be written.
+    :type  pki_path:   basestring
+    :param config:     The dictionary found in an importer's config.
+    :type  config:     dict
+    :param config_key: The key corresponding to a value in self.config to write to path.
+    :type  config_key: basestring
+    :param filename:   The filename to write the PEM data to.
+    :type  filename:   basestring
+    """
+    if config_key in config and config[config_key]:
```
asmacdo: `if config.get(config_key)` does the same thing.

asmacdo:
LGTM, but I think someone with more experience with TLS should also review.

server/pulp/server/db/migrations/0023_importer_tls_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        if not os.path.exists(pki_path):
+            _mkdir(os.path.dirname(pki_path))
+            os.mkdir(pki_path, 0700)
```
jeremycline: Won't this always raise an ``OSError``? It looks like ``_mkdir`` uses ``os.makedirs`` to make ``pki_path``, and ``os.mkdir`` documentation says it will raise an OSError if the directory already exists (which it should if ``os.makedirs`` works). Maybe ``os.chmod`` is what you want?

server/pulp/server/db/migrations/0023_importer_tls_storage.py
rbarlow: This is a copy of the function that the importer uses of the same name, so I'd like the name to remain _mkdir() for that reason.

server/pulp/server/db/migrations/0023_importer_tls_storage.py
rbarlow: True, but I like that this is explicit.

server/pulp/server/db/migrations/0023_importer_tls_storage.py
rbarlow: @jeremycline ☺

jeremycline: I am not a smart man.

PR #2508, opened 2016-04-04
Fix string format for python 2.6 compat

asmacdo:
closes https://pulp.plan.io/issues/1809

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -318,7 +318,7 @@ class Importer(AutoRetryDocument):
         """
         return os.path.join(
             LOCAL_STORAGE, 'importers',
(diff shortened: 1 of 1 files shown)


mhrivnak:
How about a quick unit test to make sure this does what it says it does? Then the PR automation will verify for us that it works on all OSs. It looks like there shouldn't be much state to setup for the test.

asmacdo:
@mhrivnak, I would have expected that this unit test would fail on python 2.6:
https://github.com/pulp/pulp/blob/1503f2772aee59d30621b58582ea997feb28fb41/server/test/unit/server/db/test_model.py#L1051

It hasn't been around long, I am guessing that this test failure has been masked by another one.

mhrivnak:
Oh, good point. It looks like it did fail on the original PR, but may have gone unnoticed in the noise of unrelated failing tests.

PR #2512, opened 2016-04-05
Support standard storage path migration used by plugins.

jortel:
https://pulp.plan.io/issues/1815

The goals:
1. the platform provides base migration logic that is leveraged by each plugin.
2. the plugin is responsible to adding the actual migration (as seen by the migration system) that uses the common logic but is responsible to providing the content unit type (model) specifics.
3. the migration is idempotent (can be killed and restarted).
4. the migration achieves eventual correctness.
5. the migration balances performance and bounded memory usage.
6. the migration runs as fast as possible while consuming a reasonable about of system resources.
7. the result must be flawless.

I decided to batch the migration to ensure the memory usage is not unbounded.  The batch size is 500k units.  Each batch uses ~500 mB of memory.  Even very large installations will complete each unit type using 1 batch but it seemed irresponsible not to batch.

Each plugin creates a Migration and adds a *plan* for each content type to be migrated.  All of the units planned to be migrated are chained (plans are iterated) together and grouped into batches.   When a batch reaches capacity, it is executed.  On each batch execution:
1. walk the published tree and re-link symlinks.
2. move the bits to the new storage location.
3. update the db

The reason behind combining unit types into batches is to limit the number of batches and thus limit the number of times we walk the published tree to update symlinks (even though this is very fast).  Even with batching/chaining, the migration of each pulp installation will execute 1 batch per plugin project at a minimum.

I have bench-marked the migration at 1.5 minutes per 100k units.  In the test environment, each unit had 3 symlinks to the published tree contained 300k symlinks.

### DIFF
--- a/server/pulp/plugins/migration/standard_storage_path.py
+++ b/server/pulp/plugins/migration/standard_storage_path.py
@@ -1,0 +1,374 @@
+
+"""
+Provides common classes used to migrate content unit storage to
+the standard path format introduced in pulp 2.8. Each plugin needs
+to add a migration using these classes.
+For example:
+
+from pulp.plugins.migration.standard_storage_path import Migration, Plan
+from pulp.server.db import connection
+
+
+def migrate(*args, **kwargs):
+    key_fields = (.,..,.)
+    collection = connection.get_collection('units_rpm')
+    migration = Migration()
+    plan = Plan(collection, key_fields)
+    migration.add(plan)
+    migration()
+"""
+import os
+import shutil
+
+from collections import namedtuple
+from hashlib import sha256
+from itertools import chain
+
+from pulp.plugins.util.misc import mkdir
+from pulp.server.config import config
+
+
+# stored in the Batch
+Item = namedtuple(
+    'Item',
+    [
+        'plan',
+        'unit_id',
+        'storage_path',
+        'new_path'
+    ]
+)
+
+
+class Batch(object):
+    """
+    A batch of units to be migrated.
+
+    :ivar items: A dictionary of Item keyed by storage path.
+    :type items: dict
+    """
+
+    # The number of items in each batch.
+    # Using 500k based on memory profile of 100 mb / 100k.
+    LIMIT = 500000
+
+    def __init__(self):
+        self.items = {}
+
+    def add(self, unit):
+        """
+        Add the specified unit to the batch.
+
+        :param unit: A unit to add.
+        :type unit: Unit
+        """
+        item = Item(unit.plan, unit.id, unit.storage_path, unit.new_path)
+        self.items[unit.storage_path] = item
+
+    def reset(self):
+        """
+        Reset the batch by clearing added units.
+        """
+        self.items = {}
+
+    def _relink(self):
+        """
+        Update (relink) symlinks to reflect the storage path of migrated units.
+        Foreach symlink found, find the *new* path using the plan.  Then,
+        re-create the symlink with the updated target.
+        """
+        root = Migration.publish_dir()
+        for path, directories, files in os.walk(root):
+            for name in chain(files, directories):
+                abs_path = os.path.join(path, name)
+                if not os.path.islink(abs_path):
+                    continue
+                target = os.readlink(abs_path)
+                unit = self.items.get(target)
+                if not unit:
+                    continue
+                os.unlink(abs_path)
+                os.symlink(unit.new_path, abs_path)
+
+    def _migrate(self):
+        """
+        Migrate the units referenced in the batch.
+          1. Move the content files.
+          2. Update the unit in the DB.
+        """
+        for item in self.items.values():
+            Unit.migrate(item.plan, item.unit_id, item.storage_path, item.new_path)
+
+    def __call__(self):
+        """
+        Execute the batch.
+          1. Update published links to reference the new path.
+          2. Move the content to the new path.
+          2. Update _storage_path on the unit and saved.
+        """
+        if self.items:
+            self._relink()
+            self._migrate()
+            self.reset()
+
+    def __len__(self):
+        return len(self.items)
+
+
+class Plan(object):
+    """
+    The Unit migration plan contains all that is necessary
+    to migrate a specific type of content unit.  This includes the collection,
+    the list of unit keys (needed to calculate the new storage path) and an
+    indication of whether the storage path references a file or directory.
+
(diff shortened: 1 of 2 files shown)


jortel:
ok test

jortel:
ok test

jortel:
ok test

jortel:
RPM Plugin example: https://github.com/pulp/pulp_rpm/pull/834

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def _relink(self):
+        """
+        Update (relink) symlinks to reflect the storage path of migrated units.
+        Foreach symlink found, find the *new* path using the plan.  Then,
+        re-create the symlink with the updated target.
+        """
+        root = Migration.publish_dir()
+        for path, directories, files in os.walk(root):
+            for name in chain(files, directories):
```
mhrivnak: It's interesting that a symlink to a directory will show up in the ``directories`` list here. I wasn't sure which way it would go, but I did a quick test that confirms what you also concluded.

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+                abs_path = os.path.join(path, name)
+                if not os.path.islink(abs_path):
```
mhrivnak: You could eliminate this read operation from the loop if you call ``os.readlink`` in a try...except, and ``continue`` if there's an exception raised. I like the readability of this better, but removing one read operation per link could make a big difference.

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+                os.unlink(abs_path)
+                os.symlink(unit.new_path, abs_path)
+
+    def _migrate(self):
+        """
+        Migrate the units referenced in the batch.
+          1. Move the content files.
+          2. Update the unit in the DB.
+        """
+        for item in self.items.values():
```
mhrivnak: ``values()`` creates a new list. Since ``self.items`` could be big, ``itervalues()`` may be a more memory-efficient choice.

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+    :type key_fields: tuple
+    :ivar join_leaf: Join the existing path leaf (filename) to the new path.
+    :type join_leaf: bool
+    """
+
+    def __init__(self, collection, key_fields, join_leaf=True):
+        """
+        :param collection: A DB collection.
+        :type collection: pymongo.collection.Collection
+        :param key_fields: List of fields that make up the unit key.
```
mhrivnak: s/List/Tuple to match the line below

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+        :type key_fields: tuple
+        :param join_leaf: Join the existing path leaf (filename) to the new path.
+        :type join_leaf: bool
+        """
+        self.collection = collection
+        self.key_fields = key_fields
+        self.join_leaf = join_leaf
+
+    @property
+    def base_fields(self):
```
mhrivnak: Would this be better as an actual attribute of the Class? I'm not sure it merits having a function.

```
class Plan(object):
    """Fancy docs"""
    base_fields = ('_storage_path, '_content_type_id')
```

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+
+    @staticmethod
+    def _prune():
+        """
+        Delete empty directories in the content storage tree.
+        The migration will create empty directories.
+        """
+        root = Migration.content_dir()
+        for path, directories, files in os.walk(root, topdown=False):
+            if not os.listdir(path):
```
mhrivnak: Is this logically different than ``if not (directories or files)`` ?

server/pulp/plugins/migration/standard_storage_path.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        :param collection: A DB collection.
+        :type collection: pymongo.collection.Collection
+        :param key_fields: List of fields that make up the unit key.
+        :type key_fields: tuple
+        :param join_leaf: Join the existing path leaf (filename) to the new path.
+        :type join_leaf: bool
+        """
+        self.collection = collection
+        self.key_fields = key_fields
```
mhrivnak: I commented on this in the rpm migration PR. It looks like those subclasses are defining a class attribute of this name, passing it into this function, which just overshadows the class attribute with an instance attribute of the same name and value.

jortel: See updated: https://github.com/pulp/pulp_rpm/pull/834

dkliban:
I am closing this PR in preparation for making 2.8-dev branch. Please reopen this PR against 2.8-dev branch.

dkliban:
I reopened this PR against 2.8-dev https://github.com/pulp/pulp/pull/2521

server/pulp/plugins/migration/standard_storage_path.py
jortel: Great suggestion.

server/pulp/plugins/migration/standard_storage_path.py
jortel: nope.  good suggestion.

server/pulp/plugins/migration/standard_storage_path.py
jortel: This is less obvious because it's decorated as a property.  The reason for making this a property was so that sub-classes could include additional fields to be pulled.  I ran into this when doing the migration for the `python` plugin.  Now that you have pointed this out, I'm not crazy about using a property for this.  It would be safer to provide a way for the sub-class specify additional fields instead of redefining the entire list (with the override).  I'll revise this.

server/pulp/plugins/migration/standard_storage_path.py
jortel: Oh, wait, now I remember.  The walk is yielding a snapshot of the tree.  During the bottom-up walk of the tree - if a leaf directory is deleted, it still show up in the `directories` list on the parent even though we just deleted it and the parent is now empty.

PR #2514, opened 2016-04-08
Translate _id only if it is in remapped fields

asmacdo:
Do not attempt to objectidify `_id`s if `_id` is not a remapped field.


mccun934:
Tested this patch on my system, worked fine for me

seandst:
> objectidify

O.o

PR #2520, opened 2016-04-11
added distributor methods to ensure all units in a repo have been downloaded

mhrivnak:
A follow-up PR in pulp_rpm will call the new methods from actual distributors.

https://pulp.plan.io/issues/1835

re #1835

### DIFF
--- a/common/pulp/common/error_codes.py
+++ b/common/pulp/common/error_codes.py
@@ -113,6 +113,10 @@ PLP0042 = Error("PLP0042", _("This request is forbidden."), [])
 PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
                              "Refer to /etc/pulp/server.conf for proper use."), [])
 PLP0044 = Error("PLP0044", _("The target importer does not support the types from the source"), [])
+PLP0045 = Error("PLP0045", _("The repository cannot be exported because some units are "
+                             "not downloaded."), [])
+PLP0046 = Error("PLP0046", _("The repository group cannot be exported because these repos have "
+                             "units that are not downloaded: %(repos)s"), ['repos'])
 
 # Create a section for general validation errors (PLP1000 - PLP2999)
 # Validation problems should be reported with a general PLP1000 error with a more specific
--- a/server/pulp/plugins/distributor.py
+++ b/server/pulp/plugins/distributor.py
@@ -13,6 +13,10 @@
 
 import sys
 
+from pulp.common import error_codes
+from pulp.server import exceptions
+from pulp.server.controllers import repository as repo_controller
+
 
 class Distributor(object):
     """
@@ -187,6 +191,20 @@ class Distributor(object):
         """
         return {}
 
+    @staticmethod
+    def ensure_all_units_downloaded(repo_id):
+        """
+        Checks the database to make sure all units in the repo have been downloaded. If not, raises
+        an exception.
+
+        :param repo_id: repository id
+        :type  repo_id: basestring
+
+        :raises exceptions.PulpCodedException:  if any unit in the repo is un-downloaded
+        """
+        if not repo_controller.has_all_units_downloaded(repo_id):
+            raise exceptions.PulpCodedException(error_code=error_codes.PLP0045)
+
 
 class GroupDistributor(object):
     """
@@ -337,3 +355,20 @@ class GroupDistributor(object):
         :raise SystemExit: raised through a call to sys.exit()
         """
         sys.exit()
+
+    @staticmethod
+    def ensure_all_units_downloaded(repo_group):
+        """
+        Checks the database to make sure all units in the repos have been downloaded. If not, raises
+        an exception.
+
+        :param repo_group:  RepositoryGroup instance that should be checked
+        :type  repo_group:  pulp.plugins.model.RepositoryGroup
+
+        :raises exceptions.PulpCodedException:  if any unit in the group is un-downloaded
+        """
+        bad_repo_ids = filter(lambda x: not repo_controller.has_all_units_downloaded(x),
+                              repo_group.repo_ids or [])
+        if bad_repo_ids:
+            raise exceptions.PulpCodedException(error_code=error_codes.PLP0046,
+                                                repos=', '.join(bad_repo_ids))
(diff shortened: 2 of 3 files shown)


common/pulp/common/error_codes.py
```diff
@@ -113,6 +113,10 @@
 PLP0043 = Error("PLP0043", _("Database 'write_concern' config can only be 'majority' or 'all'. "
                              "Refer to /etc/pulp/server.conf for proper use."), [])
 PLP0044 = Error("PLP0044", _("The target importer does not support the types from the source"), [])
+PLP0045 = Error("PLP0045", _("The repository cannot be exported because some units have "
```
bmbouter: active voice rewrite "have not yet been downloaded" -> "are not yet downloaded" or just "are not downloaded"

common/pulp/common/error_codes.py
```diff
@@ -113,6 +113,10 @@
...
+                             "not yet been downloaded."), [])
+PLP0046 = Error("PLP0046", _("The repository group cannot be exported because these repos have "
+                             "units that have not yet been downloaded: %(repos)s"), ['repos'])
```
bmbouter: s/have not yet been downloaded/are not yet downloaded/   or   s/have not yet been downloaded/are not downloaded/

server/pulp/plugins/distributor.py
```diff
@@ -13,6 +13,10 @@
 
 import sys
 
+from pulp.common import error_codes
+from pulp.server.controllers import repository as repo_controller
```
bmbouter: Maybe swap lines 17 and 18 for alphabetical reasons?

server/pulp/plugins/distributor.py
```diff
@@ -187,6 +191,20 @@ def create_consumer_payload(self, repo, config, binding_config):
...
+    @staticmethod
+    def ensure_all_units_downloaded(repo_id):
+        """
+        Checks the database to make sure all units in the repo have been downloaded. If not, raises
+        an exception.
+
+        :param repo_id: repository id
+        :type  repo_id: basestring
+
+        :raises exceptions.PulpCodedException:  if any unit in the repo has un-downloaded files
```
bmbouter: Is it undownloaded files or undownloaded units? or both?

server/pulp/plugins/distributor.py
```diff
@@ -187,6 +191,20 @@ def create_consumer_payload(self, repo, config, binding_config):
...
+        """
+        if not repo_controller.has_all_units_downloaded(repo_id):
+            raise exceptions.PulpCodedException(error_code=error_codes.PLP0045)
```
bmbouter: I like how this raises an Exception rather than using a return code. +1

server/pulp/plugins/distributor.py
```diff
@@ -337,3 +355,20 @@ def cancel_publish_group(self, call_request, call_report):
...
+        Checks the database to make sure all units in the repos have been downloaded. If not, raises
+        an exception.
+
+        :param repo_group:  RepositoryGroup instance that should be checked
+        :type  repo_group:  pulp.plugins.model.RepositoryGroup
+
+        :raises exceptions.PulpCodedException:  if any unit in the group has un-downloaded files
+        """
+        bad_repo_ids = filter(lambda x: not repo_controller.has_all_units_downloaded(x),
+                              repo_group.repo_ids or [])
```
bmbouter: What about leaving off the `or []`?

`RepositoryGroup` has that attribute as a required parameter to its `__init__` so I expect it to always be [] or do you know of a place where it is `None` or some other non-list value?

server/pulp/plugins/distributor.py
mhrivnak: good call

server/pulp/plugins/distributor.py
mhrivnak: Both I guess? I think a unit is "undownloaded" if it has one or more files that are undownloaded. Do you think there's a way to make the comment here more clear?

server/pulp/plugins/distributor.py
mhrivnak: I looked at the model and a bit of how it's used, and I couldn't reach a confident conclusion that it will always be a list. Beyond current code, we have all the code from the past that could have written this document with a repo_ids attribute of ``None``.

It's unlikely I admit, but we've been bit by that kind of bug many times in the past. I don't think there's anything incorrect that could come from the ``or []``, but it does give us a line of defense in case such a repo group makes its way through this code.

So I concluded it's better to take the defensive option, especially since it's so easy, than risk even a small chance of a user or customer getting a traceback here.

server/pulp/plugins/distributor.py
bmbouter: s/files/units/ would resolve it for me. The function call used in this method uses units.

server/pulp/plugins/distributor.py
```diff
@@ -337,3 +355,20 @@ def cancel_publish_group(self, call_request, call_report):
...
+    @staticmethod
+    def ensure_all_units_downloaded(repo_group):
+        """
+        Checks the database to make sure all units in the repos have been downloaded. If not, raises
+        an exception.
+
+        :param repo_group:  RepositoryGroup instance that should be checked
+        :type  repo_group:  pulp.plugins.model.RepositoryGroup
+
+        :raises exceptions.PulpCodedException:  if any unit in the group has un-downloaded files
```
bmbouter: s/files/units/ here too

server/pulp/plugins/distributor.py
bmbouter: I also looked for some usage to find non list instances, and I couldn't find any. If there are usages that are non list, this statement would mask that problem in our code even further. Generally I think experiencing a problem and fixing it is better than maintaining code that deals with inconsistent data types. Besides pulp-smash should find it before our users do.

bmbouter:
I don't need to re-review so I'm putting the LGTM on it. Handle the comments as you see fit.

server/pulp/plugins/distributor.py
mhrivnak: I generally feel the same way. In this case, any potential problems stem from the lack of schema and lack of enforcement at the model level. Since we have a plan in place to take care of that, I think it's better in this case to wait for that process to resolve it, rather than hope smash finds any problems, and otherwise risk letting our users find the problems.

PR #2521, opened 2016-04-11
Support standard storage path migration used by plugins.

dkliban:
closes #1815

### DIFF
--- a/server/test/unit/plugins/migration/test_standard_storage_path.py
+++ b/server/test/unit/plugins/migration/test_standard_storage_path.py
@@ -1,0 +1,444 @@
+import os
+
+from unittest import TestCase
+from itertools import izip, chain
+
+from mock import patch, Mock, call
+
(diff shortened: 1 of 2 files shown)


server/test/unit/plugins/migration/test_standard_storage_path.py
```diff
@@ -0,0 +1,444 @@
...
+from pulp.plugins.migration.standard_storage_path import Batch, Plan, Migration, Unit, Item
+
+
+MODULE = 'pulp.plugins.migration.standard_storage_path'
+
+
+class TestBatch(TestCase):
+
+    def test_add(self):
+        unit = Mock(plan=1, id=2, storage_path=3, new_path=4)
```
mhrivnak: Personally, I think it's better to use the real data structures whenever possible, and only mock out behaviors as necessary. I'm not opposed to this, but I don't see that using a Mock instance is better than using a Unit instance.

jortel: Agreed.

PR #2530, opened 2016-04-13
pulp.spec now generate RSA keys with umask 077 (CVE-2016-3111)

jeremycline:
During installation, the RSA key pairs used to validate messages between
the pulp server and pulp consumers were generated in a directory that is
world-readable with a umask of 002. After it was written, the permissions
were modified to protect the key. For a brief moment, the RSA keys were
world-readable. This commit explicitly sets the umask in the %post
scriptlet to be 077 so it is only readable to the owner.

https://pulp.plan.io/issues/1837

fixes #1837

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -476,8 +476,12 @@ KEY_PATH="$KEY_DIR/rsa.key"
 KEY_PATH_PUB="$KEY_DIR/rsa_pub.key"
 if [ ! -f $KEY_PATH ]
 then
+  # Ensure the key generated is only readable by the owner.
+  OLD_UMASK=$(umask)
+  umask 077
   openssl genrsa -out $KEY_PATH 2048 &> /dev/null
   openssl rsa -in $KEY_PATH -pubout > $KEY_PATH_PUB 2> /dev/null
+  umask $OLD_UMASK
 fi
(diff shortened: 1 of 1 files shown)


pulp.spec
```diff
@@ -476,6 +476,8 @@ KEY_PATH="$KEY_DIR/rsa.key"
 KEY_PATH_PUB="$KEY_DIR/rsa_pub.key"
 if [ ! -f $KEY_PATH ]
 then
+  # Ensure the key generated is only readable by the owner.
+  umask 077
```
seandst: The original umask doesn't get set back. Could this introduce unintended consequences, or would it be good to stash the current umask and reset it after generating this key?

jeremycline: The way I thought scriptlets worked was it set up a new shell for each one... But I should be up front and say I don't really know much about the internals of RPM with respect to scriptlets. 

Given that I don't know with certainty how they work, I think it would indeed be best to stash the umask and change it back later. That way I don't have to spend lots of time reading man pages just so I can be "safe" at the expense of clarity.

seandst: Cool. I definitely don't even know what a scriptlet is, so I'm happy with however you want to deal with this. Thanks!

PR #2532, opened 2016-04-13
Safely create the dir in which the CA is created (CVE-2016-3106).

rbarlow:
Prior to this commit, the pulp-gen-ca-certificate script created
the Pulp CA certificate and key in /tmp/$RANDOM. This led to about
32,768 possible directories. Florian Weimer and Sander Bos
notified the Pulp team that if a user happened to own a directory
that the script chose, the user would be able to read the
certificate authority certificate. Additionally, both security
researchers concluded that there is a race condition between
creating the secure folder and the later chmod, during which an
attacker could read and hold open the inode for the $TMP
directory. This would allow an attacker to read the key that is
later written.

Sander Bos additionally concluded that an attacker could create a
DoS attack two ways: 0) By creating a symlink within the $TMP
directory that points at an important system resource, such as
/etc/passwd, or 1) By creating $TMP itself as a symlink to /,
which would cause / to be chmod'd to 0700 later in the script.

This commit adjusts the script to use mktemp -d to ensure that a
unique and safe directory is used to create the Pulp CA
certificate. Additionally, the script uses set -e so that it will
halt if there are errors.

Thanks to Florian Weimer and to Sander Bos for independently
notifying Pulp of the issue and for suggesting the needed changes
to fix it. Thanks to Adam Mariš for advising the Pulp team through
the fix. The Pulp team is thankful to the security community for
their thoughtful analysis, and for taking the time to report these
issues.

https://pulp.plan.io/issues/1827

fixes #1827


rbarlow:
ok test

rbarlow:
ok test

seandst:
ok test

rbarlow:
ok test

rbarlow:
ok test

seandst:
ok test

seandst:
ok test

rbarlow:
@seandst due to EL 6 lacking the -Z flag on mv, I had to refactor this pull request. I decided to set the umask in a similar manner to @jeremycline's recent fixes so we don't have to rely as much on the secure folder. This allows the files to be created with the proper permissions to begin with, including the SELinux label. If you wouldn't mind re-reviewing, I believe it is ready now.

rbarlow:
ok test

rbarlow:
ok test

rbarlow:
ok test

seandst:
test failures unrelated, lgtm

PR #2538, opened 2016-04-15
Adds known issue about Qpid repo location switch

dkliban:


docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -91,6 +91,10 @@ Known Issues
   experience a `problem <https://pulp.plan.io/issues/1791>`_. Replacing the symlink with a bind
   mount will resolve the issue.
 
+* CentOS 6 users need to be aware that the Qpid repository has changed locations. The most recent
```
bmbouter: s/CentOS 6/RHEL 6 and CentOS 6 who use Qpid as their broker/

bmbouter:
I'm suggesting we don't merge this but if you want to I'm OK with that. Some things I think about:

If people can't read pulp-list I'm not sure why they would read the release notes.
What about users who upgrade from 2.7 -> 2.9 and skip the 2.8 release?
It's a slippery slope if we start documenting changes is software other than Pulp.

dkliban:
I am willing to stand on this slippery slope.

rbarlow:
On Fri, Apr 15, 2016 at 08:11:03AM -0700, bmbouter wrote:
> If people can't read pulp-list I'm not sure why they would read the release notes.

I am not disagreeing with your conclusion (I don't think it makes sense
to document other software projects), but I do think it's important to
note that our mailing list isn't the most effective way to communicate
with users. I read release notes for several software projects whose
mailing lists I do not subscribe to, and I think that's a more typical
use pattern. People don't like subscribing to lots of mailing lists, but
they do use lots of software.

> It's a slippery slope if we start documenting changes is software other than Pulp.

+MAXINT - let's not document how to use software that isn't Pulp. We
haven't even done a thorough job of documenting our own software, so
let's put our documentation effort there instead.

-- 
Randy Barlow
irc:   bowlofeggs

rbarlow:
On Fri, Apr 15, 2016 at 08:13:33AM -0700, Dennis Kliban wrote:
> I am willing to stand on this slippery slope. 

lol

-- 
Randy Barlow
irc:   bowlofeggs

bmbouter:
@dkliban does it get added to 2.7 then too? Or 2.6? 2.9? 2.10? 3.0?

dkliban:
@bmbouter We should only it add it to 2.7 when we do another release of it. At this point it would not be an issue for someone on 2.7. Depending on when 2.9 is released, it might be wise to have it there also. Same goes for 2.10.

bmbouter:
@dkliban oh good point it wouldn't affect 2.7. I'm convinced enough that if you want to merge it you can.

The downside is that the issue a user would experience by not having the right Qpid repo is not obvious that this note is the info they need to solve it. I'm not sure how to improve that though.

dkliban:
@bmbouter Should I elaborate on the problem a user may experience as result of having older packages?

bmbouter:
Yes that would be good. Showing that one line with the session error
missing attribute in a quoted section would be good

docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -84,13 +84,20 @@ Bugs
...
+  upgrade instructions for workaround.
 
 * Users that have the `/var/lib/pulp` directory or one of it's subdirectories symlinked will
   experience a `problem <https://pulp.plan.io/issues/1791>`_. Replacing the symlink with a bind
   mount will resolve the issue.
 
+* RHEL 6 and CentOS 6 users who use Qpid as their broker need to be aware that the Qpid repository
+  has changed locations. The most recent repository definition file can be obtained from `Qpid
+  packaging docs <http://qpid.apache.org/packages.html#epel>`_. Using an older version of
+  python-qpid package will produce the following error in the logs::
```
bmbouter: escape python-qpid

docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -84,13 +84,20 @@ Bugs
...
-  instructions for workaround.
+* RHEL 7 and CentOS 7 users may experience a problem when upgrading. Please refer to note in
+  upgrade instructions for workaround.
 
 * Users that have the `/var/lib/pulp` directory or one of it's subdirectories symlinked will
   experience a `problem <https://pulp.plan.io/issues/1791>`_. Replacing the symlink with a bind
   mount will resolve the issue.
 
+* RHEL 6 and CentOS 6 users who use Qpid as their broker need to be aware that the Qpid repository
+  has changed locations. The most recent repository definition file can be obtained from `Qpid
```
bmbouter: s/repository/Qpid repository/

bmbouter:
2 small suggestions, LGTM

PR #2539, opened 2016-04-15
Remove inaccurate param from docs

asmacdo:
This looks like an accidental copy/paste and does not belong.


docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -88,7 +88,6 @@ is centered around updating only that metadata.
 * :param:`?delta,object,object containing keys with values that should be updated on the repository`
 * :param:`?importer_config,object,object containing keys with values that should be updated on the repository's importer config`
 * :param:`?distributor_configs,object,object containing keys that are distributor ids, and values that are objects containing plugin specific keys/value pairs`
-* :param:`?delta,object,object containing keys with values that will update the distributor object, currently only supports "auto_publish"`
```
rbarlow: Are you sure this isn't supposed to be here? Isn't that what this line is for?

https://github.com/pulp/pulp/blob/pulp-2.8.2-1/server/pulp/server/webservices/views/repositories.py#L209

asmacdo:
Note that there is another delta 3 lines up

rbarlow:
@asmacdo ah yes, I had not noticed that. LGTM.

PR #1830, opened 2015-04-28
webpy search and destroy

ipanova:
closes #730

PR #2545, opened 2016-04-18
Add release notes for the upcoming 2.8.3.

rbarlow:
re #1827
re #1830
re #1833
re #1834
re #1837

### DIFF
--- a/docs/user-guide/release-notes/2.8.x.rst
+++ b/docs/user-guide/release-notes/2.8.x.rst
@@ -2,6 +2,165 @@
 Pulp 2.8 Release Notes
 ======================
 
+
+Pulp 2.8.3
+==========
+
+.. warning::
+
+    User action is required to address the CVEs associated with this upgrade! Read the upgrade
+    instructions below.
+
+2.8.3 is a security and bugfix release. Included in the list of :fixedbugs:`2.8.3`
+are a number of CVEs:
+
+    * `CVE-2016-3106 <https://pulp.plan.io/issues/1827>`_: Insecure creation of temporary directory
+      when generating new CA key
+    * `CVE-2016-3107 <https://pulp.plan.io/issues/1833>`_: Node certificate containing private key
+      stored in world-readable file
+    * `CVE-2016-3108 <https://pulp.plan.io/issues/1830>`_: Insecure temporary file used when
+      generating certificate for Pulp Nodes
+    * `CVE-2016-3111 <https://pulp.plan.io/issues/1837>`_: pulp.spec generates its RSA keys for
+      message signing insecurely
+    * `CVE-2016-3112 <https://pulp.plan.io/issues/1834>`_: Pulp consumer private keys are
+      world-readable
+
+Additionally, `CVE-2013-7450 <https://bugzilla.redhat.com/show_bug.cgi?id=1003326>`_ was
+announced during this release cycle, even though it was fixed in Pulp 2.3.0. *Users who have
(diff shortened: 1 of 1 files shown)


rbarlow:
This PR is not ready for review yet. I wanted to make a place for @jeremycline to put notes for me to include in this commit related to the CVEs he fixed.

jeremycline:
CVE-2016-3112
----------------------
The client certificate for consumers, ``/etc/pki/pulp/consumer/consumer-cert.pem``, was installed world-readable. This issue, CVE-2016-3112, has been fixed for new certificates issued to consumers, but it does not modify the permissions of old certificates. It is recommended that users regenerate the certificates by unregistering and re-registering all consumers::

    $ sudo pulp-consumer unregister
    $ sudo pulp-consumer -u <admin-username> register --consumer-id=<consumer-id>

If you choose to address CVE-2016-3095 and CVE-2016-3106 by regenerating Pulp's internal certificate authority, you should perform that after you've unregistered all your consumers, but before you re-register them.

CVE-2016-3111
----------------------
Both the RSA key pair for the Pulp server and RSA key pair for each Pulp consumer was generated during installation in an insecure directory. This vulnerability, CVE-2016-3111, allowed a local attacker to read the private key portion of the key pair. These keys are used for message authentication between the Pulp server and the Pulp consumers. If you are concerned that a local attacker was able to read these keys, you can regenerate them. We do not ship a script to perform this, but the process is straight-forward. For the Pulp server, do the following as root::

    $ cd /etc/pki/pulp/
    $ rm rsa.key rsa_pub.key
    $ umask 077
    $ openssl genrsa -out rsa.key <bits>  # <bits> should be at least 2048
    $ openssl rsa -in rsa.key -pubout > rsa_pub.key
    $ chgrp apache rsa.key rsa_pub.key
    $ chmod 640 rsa.key  # Apache must be able to read the private key
    $ chmod 644 rsa_pub.key  # The public key is world-readable as it is served via Apache

The Pulp consumer key is similar::

    $ cd /etc/pki/pulp/consumer/
    $ rm rsa.key rsa_pub.key
    $ umask 077
    $ openssl genrsa -out rsa.key <bits>  # <bits> should be at least 2048
    $ openssl rsa -in rsa.key -pubout > rsa_pub.key

docs/user-guide/release-notes/2.8.x.rst
```diff
@@ -2,6 +2,165 @@
...
+    $ chmod 640 rsa.key  # Apache must be able to read the private key
+    $ chmod 644 rsa_pub.key  # The public key is world-readable as it is served via Apache
+
+The Pulp consumer key is similar::
+
+    $ cd /etc/pki/pulp/consumer/
+    $ rm rsa.key rsa_pub.key
+    $ umask 077
+    $ openssl genrsa -out rsa.key <bits>  # <bits> should be at least 2048
+    $ openssl rsa -in rsa.key -pubout > rsa_pub.key
```
rbarlow: @jortel and/or @jeremycline: Can one or both of you confirm whether need the above three lines? Will the pulp-consumer register statement below will take care of this for us?

jeremycline: The key is generated in the %post install scriptlet[0]. The unregister command does not remove it.

Honestly, I find it weird that it's generated when the package is installed rather than when it starts to register, but it doesn't seem unsafe to me.

[0] https://github.com/pulp/pulp/blob/master/pulp.spec#L896-L910

seandst:
I thought this got merged before the beta cut-off, so I need to take care of merging this.

PR #2547, opened 2016-04-19
Add --force-full flag to publish command

goosemania:
re #1158
https://pulp.plan.io/issues/1158


rbarlow:
I think this PR should be against master since it is a new feature.

PR #1724, opened 2015-03-16
Fix: 758.  Ensure gofer connector is added.

jortel:
https://pulp.plan.io/issues/758

Ensures that the gofer lib connector is configured before each RMI call.

Minor changes to the agent queue delete task to support query and cancel of this task when things go wrong.  Currently, no way to cancel the task.

Other minor changes to be consistent with contemporary gofer API.

PR #2548, opened 2016-04-19
Publish should be a no-op if nothing changed since the last publish

ipanova:
closes #1724
https://pulp.plan.io/issues/1724

### DIFF
--- a/docs/dev-guide/integration/rest-api/repo/cud.rst
+++ b/docs/dev-guide/integration/rest-api/repo/cud.rst
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
 configuration values are contingent on the type of distributor being added; each
 distributor type will support a different set of values relevant to how it functions.
 
+.. note::
+        There is an optional distributor configuration parameter that is usable
+        on all distributor plugin types called ``force_full``. If this parameter is set,
+        every publish of the repo using this distributor will be done from scratch.
+        More details are `here <http://pulp.readthedocs.org/en/latest/user-guide/release-notes/2.9.x.html#pulp-2-9-0>`_.
+
 Multiple distributors may be associated with a repository at a given time. There
 may be more than one distributor with the same type. The only restriction is
 that the distributor ID must be unique across all distributors for a given repository.
--- a/docs/user-guide/release-notes/2.9.x.rst
+++ b/docs/user-guide/release-notes/2.9.x.rst
@@ -1,0 +1,15 @@
+======================
+Pulp 2.9 Release Notes
+======================
+
+Pulp 2.9.0
+==========
+
+New Features
+------------
+
+* Re-publish of the repository will be a no operation by default if since the last publish, no units
+  were added/removed and there were no changes made to the distributor settings. A publish which
+  causes no operation has a skipped result and therefore does not re-create repodata files.
+  To force a re-publish, which does publish from sratch, use `force_full` option via API or
+  `--force-full` via CLI.
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -968,8 +968,67 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
     # Fire events describing the publish state
     fire_manager = manager_factory.event_fire_manager()
     fire_manager.fire_repo_publish_started(repo_id, dist_id)
-    result = _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config)
+    result = check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config)
     fire_manager.fire_repo_publish_finished(result)
+    return result
+
+
+def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config):
+    """
+    Check if the publish should be a no operation and therefore skipped.
+
+    :param repo_obj: repository object
+    :type  repo_obj: pulp.server.db.model.Repository
+    :param dist_id: identifies the distributor
+    :type  dist_id: str
+    :param dist_inst: instance of the distributor
+    :type  dist_inst: dict
+    :param transfer_repo: dict representation of a repo for the plugins to use
+    :type  transfer_repo: pulp.plugins.model.Repository
+    :param conduit: allows the plugin to interact with core pulp
+    :type  conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+    :param call_config: allows the plugin to retrieve values
+    :type  call_config: pulp.plugins.config.PluginCallConfiguration
+
+    :return: publish result containing information about the publish
+    :rtype:  pulp.server.db.model.repository.RepoPublishResult
+    """
+
+    force_full = call_config.get('force_full', False)
+    last_published = conduit.last_publish()
+    last_unit_removed = repo_obj.last_unit_removed
+    if last_published:
+        the_timestamp = dateutils.format_iso8601_datetime(last_published)
+        last_updated = model.RepositoryContentUnit.objects(repo_id=repo_obj.repo_id,
+                                                           updated__gte=the_timestamp).count()
+        dist = model.Distributor.objects.get_or_404(repo_id=repo_obj.repo_id,
+                                                    distributor_id=dist_id)
+
+        units_removed = last_unit_removed is not None and last_unit_removed > last_published
+        dist_updated = dist.last_updated > last_published
+
+    if last_published and not force_full and not last_updated and not units_removed and \
+            not dist_updated:
+
+        publish_result_coll = RepoPublishResult.get_collection()
+        publish_start_timestamp = _now_timestamp()
+        publish_end_timestamp = _now_timestamp(string=False)
+
+        # Use raw pymongo not to fire the signal hander
+        model.Distributor.objects(
+            repo_id=repo_obj.repo_id).update(set__last_publish=publish_end_timestamp)
+
+        result_code = RepoPublishResult.RESULT_SKIPPED
+        _logger.debug('publish skipped for repo [%s] with distributor ID [%s]' % (
+                      repo_obj.repo_id, dist_id))
+        result = RepoPublishResult.skipped_result(
+            repo_obj.repo_id, dist.distributor_id, dist.distributor_type_id,
+            publish_start_timestamp, publish_end_timestamp, result_code)
+        publish_result_coll.save(result)
+
+    else:
+        result = _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config)
+
     return result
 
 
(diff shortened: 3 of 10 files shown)


ipanova:
ok test

server/pulp/server/controllers/repository.py
```diff
@@ -968,10 +968,63 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
+    last_published = conduit.last_publish()
+    last_unit_removed = repo_obj.last_unit_removed
+    if last_published:
+        the_timestamp = dateutils.format_iso8601_datetime(last_published)
+        last_updated = model.RepositoryContentUnit.objects(repo_id=repo_obj.repo_id, 
+                                                           updated__gte=the_timestamp).count()
+    dist = model.Distributor.objects.get_or_404(repo_id=repo_obj.repo_id, distributor_id=dist_id)
+
+    if last_published and not force_full and not last_updated and \
+        (not last_unit_removed or last_unit_removed < last_published) and \
```
bmbouter: Can this expression be simplified so that it's not evaluated inside of the if statement. It would make it much more readable.

server/pulp/server/controllers/repository.py
```diff
@@ -968,10 +968,63 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
+        (not dist.last_updated or dist.last_updated < last_published):
```
bmbouter: Can this expression be simplified so that it's not evaluated inside of the if statement. It would make it much more readable.

server/pulp/server/controllers/repository.py
```diff
@@ -968,10 +968,63 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
     fire_manager = manager_factory.event_fire_manager()
     fire_manager.fire_repo_publish_started(repo_id, dist_id)
-    result = _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config)
+    result = check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config)
     fire_manager.fire_repo_publish_finished(result)
     return result
 
+def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config):
+    """
+    Check if the publish should be a no operation, and therefore skipped.
```
bmbouter: no comma after operation since the second part is not a complete sentence on its own.

server/pulp/server/controllers/repository.py
ipanova: I am not sure because we need to cover different cases like - since last publish there were some units removed or there were no units removed since ever, same stands for last updated distributor.

ipanova: so `last_unit_removed` can be `None` if there were no units removed at all since the first publish or it can have a date.

server/pulp/server/controllers/distributor.py
```diff
@@ -237,6 +237,8 @@ def update(repo_id, dist_id, config=None, delta=None):
 
     if not valid_config:
         raise exceptions.PulpDataException(message)
+
+    distributor.last_updated = dateutils.now_utc_datetime_with_tzinfo()
```
bmbouter: How is this being saved? The distributor object doesn't have a last_updated field. I see it's model definition here:

https://github.com/pulp/pulp/blob/51b36d4a5e19e68308f02adfbce1c4bf08144b5f/server/pulp/server/db/model/__init__.py#L1211-L1221

bmbouter:
We need docs to be written with this documenting this behavior. Please add before merging. It would be good to get those reviewed by someone (not necessarily me, but I can help). I'm done with the review otherwise.

bmbouter:
Can this come with a pulp smash test asserting that it is skipped when expected and run when expected for the different cases?

Oh also this [pulp smash this test](https://github.com/PulpQE/pulp-smash/issues/127) is related. Can it enable that test?

server/pulp/server/controllers/repository.py
bmbouter: What I'm suggesting is that instead of `(not last_unit_removed or last_unit_removed < last_published) and \` the result of the expression could be broken apart into smaller, more readable lines, with the intermediate results saved into variable names that are meaningful. For example prior to the `if` statement you could write:

```
publish_due_to_unit_removal = last_unit_removed is None or last_unit_removed < last_published
```

Then `publish_due_to_unit_removal` could be used directly in the `if` statement. This would be more readable I think. Similar for the other line.

bmbouter:
@ipanova The release note looks good, but this needs docs in addition to the release note. The distributor option also needs a presence in the docs itself indicating the distributor option is usable on all distributor types. This would be a new area of documentation in pulp's docs since I don't think there are global distributor settings yet.

docs/user-guide/release-notes/2.9.x.rst
```diff
@@ -0,0 +1,15 @@
...
+Pulp 2.9 Release Notes
+======================
+
+Pulp 2.9.0
+==========
+
+New Features
+------------
+
+* Re-publish of the repository will be a no operation by default if since last publish, no units
```
bmbouter: s/since/since the/

server/pulp/server/controllers/repository.py
```diff
@@ -968,11 +968,67 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
+    if last_published:
+        the_timestamp = dateutils.format_iso8601_datetime(last_published)
+        last_updated = model.RepositoryContentUnit.objects(repo_id=repo_obj.repo_id,
+                                                           updated__gte=the_timestamp).count()
+    dist = model.Distributor.objects.get_or_404(repo_id=repo_obj.repo_id, distributor_id=dist_id)
+
+    units_removed = last_unit_removed is None or last_unit_removed < last_published
+    dist_updated = dist.last_updated is None or dist.last_updated < last_published
+
+    if last_published and not force_full and not last_updated and units_removed and dist_updated:
```
mhrivnak: This might be more readable as

```
if all([last_published, not force_full, not last_updated, units_removed, dist_updated]):
```

but perhaps others prefer to see the word "and"? You can decide which you think is better.

server/pulp/server/controllers/repository.py
```diff
@@ -968,11 +968,67 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
+def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_config):
+    """
+    Check if the publish should be a no operation and therefore skipped.
+
+    :param repo_obj: repository object
+    :type  repo_obj: pulp.server.db.model.Repository
+    :param dist_id: identifies the distributor
+    :type  dist_id: str
+    :param dist_inst: instance of the distributor
+    :type  dist_inst: dict
```
mhrivnak: I would have expected ``pulp.server.db.model.Distributor`` based on the line above. Is this correct?

ipanova: it is correct, because `_get_distributor_instance_and_config()` returns distributor instance and its config.

docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
 configuration values are contingent on the type of distributor being added; each
 distributor type will support a different set of values relevant to how it functions.
 
+.. note::
+        There is a new optional parameter in the distributor configuration that is usable
```
bmbouter: s/a new/an/
s/parameter in the distributor configuration/distributor configuration parameter/

docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
...
+        on all distributor plugin types called ``force_full``. If this parameter will be set,
```
bmbouter: s/will be/is/

docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
...
+        every publish of the repo will be done from scratch.
```
bmbouter: s/repo/repo using this distributor/

docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
 configuration values are contingent on the type of distributor being added; each
 distributor type will support a different set of values relevant to how it functions.
 
+.. note::
+        There is an optional distributor configuration parameter that is usable
+        on all distributor plugin types called ``force_full``. If this parameter is be set,
```
bmbouter: s/is be/is/

server/pulp/server/controllers/distributor.py
bmbouter: Rather than setting this here, can a pre-save signal be created for the Distributor model which will set this automatically. This is how it is done with ContentUnit for example: https://github.com/pulp/pulp/blob/master/server/pulp/server/db/model/__init__.py#L681-L693

Doing it with a signal has the typical object oriented design advantages. It couples behaviors with data. In this case the behavior is the maintenance of that field. By having it maintained by the design itself instead of in this controller if people don't use the controller it will still work as expected so more consistency with fewer bugs.

Not using the controllers will likely become more typical over time since so much of controller functionality deals with validation and setting of attributes both of which the mongoengine definitions can handle all on their own.

bmbouter:
I think it would be good to write a migration for this so that the last_updated field will be present on all Distributor documents in the database. Also the code can be adjusted so that it never assumes that field is None since it will always be set.

The reasoning is that since we're moving to postgres, having a data model that is consistence has immediate value. Making that transition is going to be hard enough as-is so anything we can do to make it easier is good.

docs/dev-guide/integration/rest-api/repo/cud.rst
```diff
@@ -233,6 +233,12 @@ dictate how the distributor will function when it publishes content. The possibl
 configuration values are contingent on the type of distributor being added; each
 distributor type will support a different set of values relevant to how it functions.
 
+.. note::
+        There is an optional distributor configuration parameter that is usable
```
bmbouter: Great note. :palm_tree:

server/pulp/server/db/model/repository.py
```diff
@@ -243,6 +244,38 @@ def error_result(cls, repo_id, distributor_id, distributor_type_id, started, com
...
+        @param completed: iso8601 formatted timestamp when the publish completed
+        @type  completed: str
+
+        @param result_code: one of the RESULT_* constants in this class
+        @type  result_code: str
+        """
+
+        r = cls(repo_id, distributor_id, distributor_type_id, started, completed,
+                cls.RESULT_SKIPPED)
+        message = 'Skipped. Nothing changed since last publish'
```
bmbouter: Maybe internationalize this with `gettext`?

server/pulp/server/db/model/__init__.py
```diff
@@ -1237,6 +1238,20 @@ def resource_tag(self):
         """
         return 'pulp:distributor:{0}:{1}'.format(self.repo_id, self.distributor_id)
 
+    @classmethod
+    def pre_save_signal(cls, sender, document, **kwargs):
```
bmbouter: The signal looks good.

server/pulp/server/controllers/repository.py
```diff
@@ -968,11 +968,67 @@ def publish(repo_id, dist_id, publish_config_override=None, scheduled_call_id=No
...
+
+        publish_result_coll = RepoPublishResult.get_collection()
+        publish_start_timestamp = _now_timestamp()
+        publish_end_timestamp = _now_timestamp(string=False)
+
+        dist.last_publish = publish_end_timestamp
+        dist.save()
+
+        result_code = RepoPublishResult.RESULT_SKIPPED
+        _logger.debug('publish skipped for repo [%s] with distributor ID [%s]' % (
```
bmbouter: I think we leave debug logs in English so maybe don't internationalize this one?

bmbouter:
I left a few comments, but you can handle them how you want. This looks really good. Thanks for the great work @ipanova !

PR #2550, opened 2016-04-20
Add --force-full flag to publish command

goosemania:
re #1158
https://pulp.plan.io/issues/1158

### DIFF
--- a/client_lib/pulp/client/commands/repo/sync_publish.py
+++ b/client_lib/pulp/client/commands/repo/sync_publish.py
@@ -25,7 +25,10 @@ DESC_VERIFY_ALL_FLAG = _('check all units in the repository for corrupted or '
                          'missing files and re-download files as necessary '
                          'rather than just downloading files that are known '
                          'to be missing.')
+DESC_FORCE_FULL_FLAG = _('forces to publish from scratch, even if no changes were '
+                         'introduced since last publish')
 FLAG_VERIFY_ALL = PulpCliFlag('--verify-all', DESC_VERIFY_ALL_FLAG, aliases=['-a'])
+FLAG_FORCE_FULL = PulpCliFlag('--force-full', DESC_FORCE_FULL_FLAG)
 
 
 class StatusRenderer(object):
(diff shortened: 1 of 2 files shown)


client_lib/pulp/client/commands/repo/sync_publish.py
```diff
@@ -190,13 +193,14 @@ def __init__(self, context, renderer, distributor_id, name='run', description=DE
...
                                         or PulpCliFlag instance
-        :type override_config_options: list
+        :type override_config_options: tuple
         """
         super(RunPublishRepositoryCommand, self).__init__(name, description, method, context,
                                                           renderer)
 
         self.distributor_id = distributor_id
         self.override_config_keywords = []
+        override_config_options += (FLAG_FORCE_FULL, )
```
mhrivnak: Will this make the flag appear for every plugin that uses this base class? Since only ``pulp_rpm`` is going to support it (for now at least), it would be better to only have it show up for that plugin.

bmbouter: +1 to this

goosemania: @mhrivnak , @bmbouter ,
@ipanova does no-op on a platform #2548  and checks this flag for any plugin.
If no changes were made since last publish and user wants to publish repository of any type from scratch, he/she has to use the --force-full option.
Am I missing something?

mhrivnak: Oh I see, I didn't realize this was was being paired with that feature. Seems reasonable to me.

ipanova: @mhrivnak ,so are we fine to merge these 3 related PRs once i will add docs as mentioned @bmbouter here?https://github.com/pulp/pulp/pull/2548#issuecomment-212457808

mhrivnak: Works for me.

goosemania:
ok test

goosemania:
ok test

goosemania:
Jenkins failure is unrelated to the PR

PR #2551, opened 2016-04-20
Fix misleading comment about last_publish date.

goosemania:
The value of last publish is set upon success https://github.com/pulp/pulp/blob/master/server/pulp/server/controllers/repository.py#L1053


goosemania:
ok test

PR #2552, opened 2016-04-22
Updating validation script for FieldDoesNotExist

pcreech:
Updating validation script to handle the FieldDoesNotExist case

### DIFF
--- a/playpen/mongoengine/dockerimage/validation_check.py
+++ b/playpen/mongoengine/dockerimage/validation_check.py
@@ -11,7 +11,7 @@ import sys
 from subprocess import call
 
 from bson.json_util import dumps
-from mongoengine import ValidationError
+from mongoengine import ValidationError, FieldDoesNotExist
 import pkg_resources
 
 from pulp.server.db import model as platform_models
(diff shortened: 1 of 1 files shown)


playpen/mongoengine/dockerimage/validation_check.py
```diff
@@ -61,12 +61,15 @@ def handle_exception(self, item, ex):
         self._write_newline()
         self.error_file.write(ex.message)
         self._write_newline()
-        self.error_file.write(json.dumps(ex.to_dict()))
-        self._write_newline()
+        try:
+            self.error_file.write(json.dumps(ex.to_dict()))
+            self._write_newline()
+        except:
```
bmbouter: What exception does this catch?

bmbouter:
This looks really good, but I want to think about one other improvement.  :mouse: :cookie: 

Can we adjust the error count per collection from 50 to 10?
Also can we reduce the amount of data that is being written to error_file. I think just knowing the `ex.message` and then the raw dump of the json is all we need. We can have a line space separating each entry and have no `-------------------` lines. This is motivated by [this comment](https://pulp.plan.io/issues/1749?pn=1#change-10230) which indicated that he couldn't upload it because it is too large.

pcreech:
@bmbouter While those are WONDERFUL suggestions (and I plan on working to implement them), i think they are out of scope of this PR, which is to address our inability to catch the FieldDoesNotExist exception

bmbouter:
sounds good

PR #2553, opened 2016-04-24
moving "deleting" context manager into platform (from pulp_rpm)

mhrivnak:
https://github.com/pulp/pulp_rpm/blob/pulp-rpm-2.8.2-1/plugins/pulp_rpm/plugins/importers/yum/listener.py#L172

Many plugins could benefit from using this. I have a use for it right now in
pulp_puppet. One this is merged, a follow-up PR will remove it from pulp_rpm.

### DIFF
--- a/server/pulp/server/util.py
+++ b/server/pulp/server/util.py
@@ -1,12 +1,17 @@
 """
 This module contains utility code to be used by pulp.server.
 """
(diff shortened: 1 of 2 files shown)


mhrivnak:
The 2 rhel6 failures appear to be unrelated to this PR.

PR #2556, opened 2016-04-26
using 2.8 betas for the 2.8 upgrade test docker image

mhrivnak:
I tested by building and running it locally.

### DIFF
--- a/playpen/mongoengine/dockerimage/rhel-pulp.repo
+++ b/playpen/mongoengine/dockerimage/rhel-pulp.repo
@@ -4,6 +4,15 @@
 [pulp-v2-stable]
 name=Pulp 2.8 Stable Builds
 baseurl=http://repos.fedorapeople.org/repos/pulp/pulp/stable/2.8/7Server/$basearch/
+enabled=0
+skip_if_unavailable=0
+gpgcheck=1
+gpgkey=https://repos.fedorapeople.org/repos/pulp/pulp/GPG-RPM-KEY-pulp-2
+
+# Version 2.8 Beta Builds
+[pulp-v2-beta]
+name=Pulp 2.8 Beta Builds
+baseurl=http://repos.fedorapeople.org/repos/pulp/pulp/beta/2.8/7Server/$basearch/
 enabled=1
 skip_if_unavailable=0
 gpgcheck=1


playpen/mongoengine/dockerimage/rhel-pulp.repo
```diff
@@ -4,6 +4,15 @@
...
 baseurl=http://repos.fedorapeople.org/repos/pulp/pulp/stable/2.8/7Server/$basearch/
+enabled=0
+skip_if_unavailable=0
+gpgcheck=1
+gpgkey=https://repos.fedorapeople.org/repos/pulp/pulp/GPG-RPM-KEY-pulp-2
+
+# Version 2.8 Beta Builds
+[pulp-v2-beta]
+name=Pulp 2.8 Beta Builds
+baseurl=http://repos.fedorapeople.org/repos/pulp/pulp/beta/2.8/7Server/$basearch/
```
bmbouter: I forget, did we determine that pcreech's changes were contained in the current published beta?

bmbouter: Also the script of interest is in playpen, how does it end up in this docker environment?

mhrivnak: I did not verify if the desired changes are in the beta. I thought someone else did?

The build on docker hub is configured to watch this Dockerfile. It can live at any arbitrary path. Does that answer your question?

pcreech: I just went though the commit log for 2.8.3-0.1.beta tag and my commits are not in there.  They do exist in the 2.8-dev branch.  It looks like this didn't make it in time for the tagging of the first 2.8.3 beta

bmbouter: After looking into this more, I think I know the answer to [my question](https://github.com/pulp/pulp/pull/2556/files/a4c3bcfa5c8da4bffa4bd7fc6a26eab02a533337#r61090071). The docker environment doesn't bundle or include the script in any way specifically. The value of rebuilding the docker update is that the test environment where a user runs a script will contain the latest model definitions and migrations. Given that as long as docker rebuilds when we GA a pulp release then I think we already have what we want. If ^ is true we should close this PR, and I regret asking you to update this when it wasn't needed.

Does this make sense to you?

mhrivnak: The docker image does include the script and an installation of pulp code. You can see an ``ADD`` statement for it in the ``Dockerfile``.

To update the script itself, it must be changed in git, and then the docker image just needs to be re-built on docker hub. Docker hub builds the image from our master branch.

To change how the image gets built, such as which packages get installed, requires a similar process. A PR modifies the Dockerfile, or any related files (in this case the repo file), and then the image must be re-built on docker hub.

bmbouter: With an `ADD` statement it sounds like the `run.sh` is bundled with the docker image which is good.

The `run.sh` is merged forward to master already so wouldn't a rebuild of the docker image include the up to date `run.sh` even though the version in `2.8.3-0.1.beta` is not up to date?

If ^ is true, then all we need to do is have docker rebuild the image after the `run.sh` was merged forward to master. Perhaps we should keep the enabled repo used by the Dockerfile to be the GA release, what do you think?

mhrivnak: Ah, that's a good point. If the changes we need from @pcreech are only within the playpen directory, then they are already available on master and would be included in the current image build. I agree we could close this PR then and stick with GA.

bmbouter:
@mhrivnak Yes all the changes we need from @pcreech are in the playpen directory.

PR #1775, opened 2015-04-09
Modified - Add a SearchView and convert TaskSearch and UserSearch to use it

asmacdo:
Hijacked from #1725 so we can merge while @rbarlow is away.

There are a couple of open questions to discuss with @rbarlow, but these are not important enough to hold up the merge (this PR is blocking the remaining Django work). 

1. https://github.com/pulp/pulp/pull/1725/files#r27686346
2. I changed the variable name here, but we also talked about moving `staticmethod` into the parent class. I tried to do this a few different ways and none of them worked immediately. Since this is not a very big deal, I elected to move on and refactor later if desired. https://github.com/pulp/pulp/pull/1725/files#r27433295

PR #2559, opened 2016-05-02
Make last_sync date available from sync conduit

goosemania:
re #1775
https://pulp.plan.io/issues/1775

### DIFF
--- a/server/pulp/plugins/conduits/repo_sync.py
+++ b/server/pulp/plugins/conduits/repo_sync.py
@@ -35,6 +35,7 @@ from pulp.plugins.conduits.mixins import (
     ImporterScratchPadMixin, SingleRepoUnitsMixin, StatusMixin,
     SearchUnitsMixin)
 from pulp.plugins.model import SyncReport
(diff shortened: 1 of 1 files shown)


server/pulp/plugins/conduits/repo_sync.py
```diff
@@ -189,3 +190,21 @@ def build_cancel_report(self, summary, details):
...
+
+        :raises ImporterConduitException: if any errors occur
+        """
+        try:
+            importer = model.Importer.objects.only('last_sync').get_or_404(
+                id=self.importer_object_id)
+            return importer.last_sync
+        except Exception as e:
+            _logger.exception('Error getting last sync time for repo [%s]' % self.repo_id)
+            raise ImporterConduitException(e), None, sys.exc_info()[2]
```
rbarlow: I know this is the pattern found elsewhere in this file, but one of the painful things about Pulp is the many places that it obfuscates Exceptions like this. I think it would be better to just let the Exception raise (i.e., just drop the try/except).

PR #2561, opened 2016-05-02
Add release notes for the CVEs included with Pulp 2.8.4.

rbarlow:
re #1854
re #1858

### DIFF
--- a/docs/user-guide/release-notes/2.8.x.rst
+++ b/docs/user-guide/release-notes/2.8.x.rst
@@ -3,6 +3,45 @@ Pulp 2.8 Release Notes
 ======================
 
 
+Pulp 2.8.4
+==========
+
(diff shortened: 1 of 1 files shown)


rbarlow:
@jortel what will users need to do for consumers and/or nodes if they replace the qpid certificates by running ```pulp-qpid-ssl-cfg``` again?

jortel:
@rbarlow, consumers won't need to do anything if admins run `pulp-qpid-ssl-cfg` so long as the same CA is used.  Consumers use the consumer certificate to authenticate to qpid.

rbarlow:
I merged this PR with the git CLI, but for some reason github's web UI doesn't think that commit is in 2.8-dev even though a fresh git clone does show it.

PR #2562, opened 2016-05-03
Fix repository search by attributes in dict fields

alanoe:


mhrivnak:
ok test

asmacdo:
https://pulp.plan.io/issues/1865

seandst:
If possible, a unit test covering the dict field case would be great.

We have some magical redmine automation that will update the issue status if you reference it in your commit. We normally add a link to the issue, followed by something like "fixes #issue". In this case, it would look like:

```
https://pulp.plan.io/issues/1865
fixes #1865
```

server/pulp/server/webservices/views/search.py
```diff
@@ -226,6 +226,16 @@ def _trim_results(model, results, only):
     min_fields = set(['_id', 'id', '_href'])
     required_fields = set([field for field, val in model._fields.items() if val.required])
     return_fields = set(only) | min_fields | required_fields
+    # remove subfield from return fields names
+    return_fields_list = list(return_fields)
+    for i in range(len(return_fields_list)):
```
asmacdo: This would be a really good use case for python's `enumerate` function.

    for i, field in enumerate(return_fields_list):
        if "." in field:
            return_fields_list[i] = field.split(".")[0]

asmacdo: Also, enumerate works on sets too, no need to cast to a list.

server/pulp/server/webservices/views/search.py
```diff
@@ -226,6 +226,16 @@ def _trim_results(model, results, only):
...
+        f = return_fields_list[i]
+        if "." in f:
+            return_fields_list[i] = f.split(".")[0]
+    return_fields = set(return_fields_list)
```
asmacdo: `in` works for sets as well as lists, no need to return to a set here. There is the possibility that if "notes" and "notes.subfield" was passed, that notes will exist in the list twice, but that doesn't cause a problem.

asmacdo:
Looks good. I'll hand test once this case is covered by a unit test.

dkliban:
ok test

dkliban:
ok test

dkliban:
I am using the PR to do some testing with our Jenkins integration.

dkliban:
ok test

dralley:
@alanoe, could you let us know what your status is w/r/t being able to respond to this feedback?  If you do not have time, that's fine, just let us know :)

dralley:
@alanoe , should one of us take over this PR?

dralley:
ok test

dralley:
@alanoe I'm closing this until the unit test can be added and the feedback addressed

PR #2568, opened 2016-05-17
Add usage of gprof2dot which can make profiling graphs

emilyzheng:


dkliban:
Thanks for the contribution! I will test it and merge it if all works as you describe :)

dkliban:
You should probably add that `graphviz` package should be installed using yum or dnf. When I installed using `pip`, the `dot` command was not available.

emilyzheng:
Please check whether the modified description is clear.

dkliban:
Thanks you for the contribution!

bmbouter:
@emilyzheng Thank you very much for this!

emilyzheng:
My pleasure!

PR #2572, opened 2016-05-17
Fixes RTD pages which don't build plantuml images at build time.

bmbouter:
These images were commited before in the repo with commit
ab1e5c833d3e928252eef0b4b6ff05d06b1a504c. They are being brought
back so that read the docs can have static images to use.

git already knows about these images so there is no additional
storage space used by this commit.

This will be merged to 2.8-release -> 2.8-dev -> master

The future build machinery is introduced with 7fa1527d419a2c8f20979c145a8e0559c7587218
and will build all .dot diagrams freshly with each docs build.
Once we no longer use Read the Docs we will remove these two
images from _static and revert the `.. image::` to use the
images available at build time in the `_diagrams` directory.

re #950
https://pulp.plan.io/issues/950


ipanova:
--->`This will be merged to 2.8-dev -> 2.8-dev -> master`
I guess you meant 2.8-release -> 2.8-dev -> master. LGTM

bmbouter:
@ipanova oh yes the commit message isn't right, I just pushed a new one before merging. Thanks for the review.

PR #1716, opened 2015-03-11
740 - Added warning to release notes for 2.6

dkliban:

PR #2573, opened 2016-05-18
Add SAX writer to generate XML without etree

goosemania:
re #1716
https://pulp.plan.io/issues/1716

### DIFF
--- a/server/pulp/plugins/util/saxwriter.py
+++ b/server/pulp/plugins/util/saxwriter.py
@@ -1,0 +1,150 @@
+from xml.sax.handler import ContentHandler
+from xml.sax.saxutils import escape, quoteattr
+
+
+class XMLWriter(ContentHandler):
+    """
+    XML writer similar to xml.sax.saxutils.XMLGenerator.
+    The result is indented XML, which is written sequentially to the stream.
+    xml.sax.saxutils.XMLGenerator is not used as is because of the lack of the
+    following functionality:
+     - short_empty_elements flag which is backported from Python 3.5.1 xml.sax.saxutils.XMLGenerator
+     - ability to generate an indented XML
+
+    :ivar _pending_start_element: indicates that element was started but it is not clear yet if it
+                                  should be closed right away or there will be some content. Needed
+                                  for generation of the empty elements in a short form. Backported
+                                  from Python 3.5.1 xml.sax.saxutils.XMLGenerator.
+    :type _pending_start_element: bool
+    :ivar _indent_lvl: current level of indentation. Needed to generate proper indentation.
+    :type _indent_lvl: int
+    :ivar _indent_sep: indentation separator. Needed to generate proper indentation.
+    :type _indent_sep: str
+    :ivar _start_element: indicates that element was started and not ended yet. Needed to generate
+                          proper indentation.
+    :type _start_element: bool
+    """
+
+    def __init__(self, stream, encoding='utf-8', short_empty_elements=False):
+        """
+        :param stream: a stream to write XML to
+        :type  stream: file-like object
+        :param encoding: encoding of the generated XML
+        :type  encoding: str
+        :param short_empty_elements: indicates that the empty elements should be generated in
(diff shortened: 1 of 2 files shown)


server/pulp/plugins/util/saxwriter.py
```diff
@@ -0,0 +1,94 @@
+from xml.sax.handler import ContentHandler
+from xml.sax.saxutils import escape, quoteattr
+
+
+class XMLWriter(ContentHandler):
+    """
+    XML writer similar to xml.sax.saxutils.XMLGenerator.
+    The result is indented XML, which is written sequentially to the stream.
```
mhrivnak: Since you also added the ``short_empty_elements`` feature, it's definitely worth mentioning here.

server/pulp/plugins/util/saxwriter.py
```diff
@@ -0,0 +1,94 @@
+from xml.sax.handler import ContentHandler
+from xml.sax.saxutils import escape, quoteattr
+
+
+class XMLWriter(ContentHandler):
```
mhrivnak: Is some or all of this backported from python 3.y? If so, it would be a good idea to state that in comments both on the class, and on individual methods.

server/pulp/plugins/util/saxwriter.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    XML writer similar to xml.sax.saxutils.XMLGenerator.
+    The result is indented XML, which is written sequentially to the stream.
+    """
+
+    def __init__(self, stream, encoding='utf-8', short_empty_elements=False):
```
mhrivnak: A doc block would be very helpful to make it clear what the new ``short_empty_elements`` argument does.

Generally, doc blocks for all of the methods in this class would be helpful.

server/pulp/plugins/util/saxwriter.py
```diff
@@ -0,0 +1,94 @@
...
+        ContentHandler.__init__(self)
+        self._write = stream.write
+        self._flush = stream.flush
+        self._encoding = encoding
+        self._short_empty_elements = short_empty_elements
+        self._pending_start_element = False
+        self._indent_lvl = 0
+        self._indent_sep = '  '
+        self._start_element = False
```
mhrivnak: A comment would be helpful about what exactly this boolean means. Looking at the code below, I'm trying to reverse-engineer it, but a comment would be easier.

server/pulp/plugins/util/saxwriter.py
```diff
@@ -0,0 +1,94 @@
...
+    The result is indented XML, which is written sequentially to the stream.
+    """
+
+    def __init__(self, stream, encoding='utf-8', short_empty_elements=False):
+        ContentHandler.__init__(self)
+        self._write = stream.write
+        self._flush = stream.flush
+        self._encoding = encoding
+        self._short_empty_elements = short_empty_elements
+        self._pending_start_element = False
```
mhrivnak: A comment would be helpful about what exactly this boolean means. Looking at the code below, I'm trying to reverse-engineer it, but a comment would be easier.

mhrivnak:
Lots more in-line comments would be helpful, but otherwise I think this looks reasonable.

What are your thoughts on unit tests? This strikes me as a case where testing individual methods may be unproductive, but unit testing some use cases would be valuable and doable.

Perhaps you could parse some contrived XML and verify that the output from this class meets expectations? Even if you have the entire expected XML output as a string, and just assert that the written output equals the expected string, that would be relatively easy to implement and effective.

goosemania:
I think that testing individual methods is probably not very valuable.
Maybe we can test that the sequences of calls of different methods of this class produces the expected XML.

goosemania:
And I will add more comments, for sure. Thanks!

mhrivnak:
Ok. Just do whatever testing you can, then go ahead and merge.

goosemania:
ok test

PR #1929, opened 2015-06-16
PEP-8 pulp.server.

rbarlow:

PR #2576, opened 2016-05-19
Change migration 23 to not assume that Importers have configs.

rbarlow:
https://pulp.plan.io/issues/1929

fixes #1929


kfiresmith:
Thanks for this fix @rbarlow!  It got me much farther but still dies.

PR #2577, opened 2016-05-22
Adds a function to calculate checksums of multiple types at once.

mhrivnak:
Calculation of checksums was moved out of the "verification"
module, because it is useful in many more cases than just in
the process of verification. The only plugin using the moved
code is pulp_rpm, and corresponding changes to that plugin
will be in a separate PR.

re #1618
https://pulp.plan.io/issues/1618

### DIFF
--- a/server/pulp/server/util.py
+++ b/server/pulp/server/util.py
@@ -2,16 +2,44 @@
 This module contains utility code to be used by pulp.server.
 """
 from contextlib import contextmanager
+from gettext import gettext as _
+import hashlib
 import logging
 import os
 from shutil import copy, Error
 
-from gettext import gettext as _
+from pulp.common import error_codes
 
-from pulp.server.exceptions import PulpExecutionException
+from pulp.server.exceptions import PulpCodedException, PulpExecutionException
 
 
 _logger = logging.getLogger(__name__)
+
+
+# Number of bytes to read into RAM at a time when validating the checksum
+CHECKSUM_CHUNK_SIZE = 8 * 1024 * 1024
+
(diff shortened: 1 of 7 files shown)


server/pulp/server/util.py
```diff
@@ -2,16 +2,42 @@
...
 from contextlib import contextmanager
+from gettext import gettext as _
+import hashlib
 import logging
 import os
 from shutil import copy, Error
 
-from gettext import gettext as _
+from pulp.common import error_codes
+from pulp.server.exceptions import PulpExecutionException, PulpCodedException
```
bmbouter: Consider moving this to it's own import block one line below since `pulp.server` is the local package this module is in.

bmbouter: Also consider switching the order of `PulpExecutionException` and `PulpCodedException` so they are alphabetical. These are such small things it's even silly to write it.

server/pulp/server/util.py
```diff
@@ -264,3 +290,62 @@ def deleting(path):
...
+    for checksum_type in checksum_types:
+        try:
+            hashers[checksum_type] = CHECKSUM_FUNCTIONS[checksum_type]()
+        except KeyError:
+            raise InvalidChecksumType('Unknown checksum type [%s]' % checksum_type)
+
+    file_object.seek(0)
+    bits = file_object.read(CHECKSUM_CHUNK_SIZE)
+    while bits:
+        for hasher in hashers.values():
```
bmbouter: nice :+1:

bmbouter:
This looks good to me, nice job!

PR #2578, opened 2016-05-22
Enabling use of the django template engine.

mhrivnak:
This change is required in pulp_rpm to templatize repo metadata.

re #1618
https://pulp.plan.io/issues/1618

### DIFF
--- a/server/pulp/server/webservices/settings.py
+++ b/server/pulp/server/webservices/settings.py
@@ -10,6 +10,12 @@ DEBUG = False
 
 TEMPLATE_DEBUG = False
 
+TEMPLATES = [
+    {
+        'BACKEND': 'django.template.backends.django.DjangoTemplates',
+    },
+]
+
 ALLOWED_HOSTS = ['*']
 
 


server/pulp/server/webservices/settings.py
```diff
@@ -10,6 +10,12 @@
 
 TEMPLATE_DEBUG = False
 
+TEMPLATES = [
```
bmbouter: I don't think this will work for older versions of Django. The Django docs says this setting was [introduced on 1.8](https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-TEMPLATES).

We need to support as far back as 1.4 for EL6:  https://apps.fedoraproject.org/packages/Django14

bmbouter:
I'm done w/ this review, it needs some reworking for compatibility.

server/pulp/server/webservices/settings.py
mhrivnak: Thanks for catching that. I'll read up on it.

mhrivnak:
I simplified the change in pulp_rpm to not use django's template "engine" feature, which is what really would have failed in django 1.4. With that change, this setting is no longer required.

PR #1928, opened 2015-06-16
PEP-8 pulp.devel.

rbarlow:

PR #2584, opened 2016-05-26
Publish should be operational if override config values were specified

ipanova:
closes #1928
https://pulp.plan.io/issues/1928


server/pulp/server/db/migrations/0024_distributor_last_updated.py
```diff
@@ -4,9 +4,10 @@
 
 def migrate(*args, **kwargs):
     """
-    Add last_updated to the distributor collection.
+    Add last_updated and last_config updated to the distributor collection.
     """
     key = 'last_updated'
+    key2 = 'last_config_updated'
```
goosemania: Should it be the `last_override_config`?

server/pulp/server/db/migrations/0024_distributor_last_updated.py
```diff
@@ -15,4 +16,9 @@ def migrate(*args, **kwargs):
             distributor[key] = distributor['last_publish']
         else:
             distributor[key] = dateutils.now_utc_datetime_with_tzinfo()
+
+        if key2 in distributor.keys():
+            continue
+        else:
+            distributor[key2] = {}
```
goosemania: How about this approach?
```
if key2 not in distributor:
    distributor[key2] = {}
```

server/pulp/server/db/migrations/0024_distributor_last_updated.py
```diff
@@ -15,4 +16,9 @@ def migrate(*args, **kwargs):
             distributor[key] = distributor['last_publish']
```
goosemania: This first `continue` will prevent you to check the `key2`, you will just move on to the next distributor.
My suggestion is not to use `continue` at all in this case. How about this:
```
if 'last_publish' in distributor:
    distributor[key] = distributor['last_publish']
elif key not in distributor:
    distributor[key] = dateutils.now_utc_datetime_with_tzinfo()
```
Does it make sense or maybe I am missing something?

ipanova: if i omit continue statement and the key with its value will be already present in the collection, then instead of be left untouched it will be changed, which is what we don't want.

PR #2585, opened 2016-05-26
Fixes #1948: Prevent RPM install error when services fail

ehelms:
Sometimes on EL6 boxes, the services can fail to stop in time which
during an upgrade can lead to the packages not getting updated.

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -458,13 +458,13 @@ Pulp provides replication, access, and accounting for software repositories.
 # If we are upgrading
 if [ $1 -gt 1 ] ; then
     %if %{pulp_systemd} == 1
-        /bin/systemctl stop pulp_workers > /dev/null 2>&1
-        /bin/systemctl stop pulp_celerybeat > /dev/null 2>&1
-        /bin/systemctl stop pulp_resource_manager > /dev/null 2>&1
+        /bin/systemctl stop pulp_workers > /dev/null 2>&1 || :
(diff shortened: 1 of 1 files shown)


pulp.spec
```diff
@@ -458,13 +458,13 @@ Pulp provides replication, access, and accounting for software repositories.
 # If we are upgrading
 if [ $1 -gt 1 ] ; then
     %if %{pulp_systemd} == 1
-        /bin/systemctl stop pulp_workers > /dev/null 2>&1
-        /bin/systemctl stop pulp_celerybeat > /dev/null 2>&1
-        /bin/systemctl stop pulp_resource_manager > /dev/null 2>&1
+        /bin/systemctl stop pulp_workers > /dev/null 2>&1 || :
```
jeremycline: Since this is only seen on RHEL6 (presumably because systemd is more robust than our init scripts and just does the right thing) do we need this on the systemctl versions?

ehelms: I suppose not, but it would break against any future breakages that this step could cause during the RPM installation.

jeremycline: Hmm. On the one hand, I'd like to know if our services become unstoppable, but on the other hand, having a scriptlet fail is a big no-no. I'm sure others will weigh in, but I think I'm okay with leaving them.

jeremycline:
@ehelms Is this something you want in the next 2.8.z build or is 2.9+ acceptable?

ehelms:
2.9 is fine with me, thanks.

PR #2586, opened 2016-05-26
Update Pulp Smash settings file

Ichimonji10:
As of Pulp Smash version 2016.05.05, the various
`pulp_smash.config.ServerConfig` methods default to working with a
configuration file section named "pulp" In earlier versions, the
defaulted to working with a configuration file section named "default."
This change was made so that Pulp Smash can more sanely adapt to the
case where not all Pulp components are on the same host. If the various
components are on different hosts, then the configuration file might
have sections named "pulp," "pulp agent," etc.

No breakage has occurred because Pulp Smash has logic for backwards
compatibility. It's a good idea to update, regardless.


bmbouter:
Nice backwards compatibility.  :+1:

PR #2587, opened 2016-05-26
Document how to install Pulp on Fedora 24+.

rbarlow:


bmbouter:
This needs conflicts resolved.

docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS           = -W -n
+SPHINXOPTS           =
```
bmbouter: Sadly this is what we must do

bmbouter:
I built the docs and there were no errors.

docs/user-guide/installation/f24+.rst
```diff
@@ -0,0 +1,282 @@
...
+   $ sudo systemctl enable rabbitmq-server
+   $ sudo systemctl start rabbitmq-server
+
+
+Pulp Server
+^^^^^^^^^^^
+
+Now we are ready to install and configure the Pulp server!
+
+#. Install the Pulp server, task workers, and dependencies. You can get the packages you need by
```
bmbouter: Should this be broken into two steps with the addition of `sudo dnf install pulp-server` being one of them? Also for it to work with `qpidd` as the broken it will not start without the installation of `python-qpid` and `qpid-tools`. This mimics an aspect of the [`pulp-server-qpid` definition in comps.xml](https://github.com/pulp/pulp/blob/73f178121522563fd9e8809ea824be409825f13a/comps.xml#L31-L32).

The `qpid-tools` is used by `goferd` but also required by `pulp-server`.

Am I thinking about this right?

rbarlow:
@bmbouter I resolved the conflicts.

docs/user-guide/installation/f24+.rst
rbarlow: You don't need to install pulp-server - it will come along when you install the plugins you want. As for qpidd, see the steps that come later.

docs/user-guide/installation/f24+.rst
```diff
@@ -0,0 +1,282 @@
...
+   simply installing the plugins you want to use and they will pull in the needed dependencies. The
+   following example installs all currently supported plugins, but feel free to season to taste:
+
+   ::
+
+      $ sudo dnf install pulp-rpm-plugins pulp-docker-plugins pulp-ostree-plugins pulp-puppet-plugins pulp-python-plugins
+
+#. Install the gofer adapter you need for the type of message broker you chose. For qpidd::
+
+      $ sudo dnf install python-gofer-qpid qpid-tools
```
bmbouter: Does Pulp server really require this or is it just `goferd` which does. I thought it was just `goferd`. @jortel would know.

docs/user-guide/installation/f24+.rst
bmbouter: Oh nice that makes sense.  And does `python-qpid` get brought in as a dep on either `qpid-tools` or `python-gofer-qpid`?

docs/user-guide/installation/f24+.rst
```diff
@@ -0,0 +1,282 @@
...
+script must be run as root to permit access to add references to the Pulp server's
+repositories.
+
+#. Install the Gofer bindings for the message broker you are using. For qpidd::
+
+      $ sudo dnf install python-gofer-qpid
+
+   For RabbitMQ::
+
+      $ sudo dnf install python-gofer-amqp
```
bmbouter: I love the equal treatment of rabbitMQ. choice++

docs/user-guide/installation/index.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Server
+  This is the main application server that stores data and distributes content.
+
+Agent
+  This component runs on consumers and communicates with the server to provide remote content
+  management.
+
+Client
+  This is a command line component that comes as two pieces: admin-client,
```
bmbouter: This could be more clearly stated that it is a management CLI. That could be stated more up front.

docs/user-guide/installation/f24+.rst
```diff
@@ -0,0 +1,282 @@
...
+   following the steps defined in the :ref:`Qpid SSL Configuration Guide <qpid-ssl-configuration>`.
+   By default, the agent will connect using a plain TCP connection.
+
+#. Set the agent to start at boot::
+
+      $ sudo systemctl enable goferd
+      $ sudo systemctl start goferd
+
+
+Extra Configuration
```
bmbouter: What about adding a similar note to `f23-.rst` as L#279-L#282. I think that would be good and so easy.

docs/user-guide/installation/index.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Pulp
+^^^^
+
+Pulp stores its content in ``/var/lib/pulp``. The size requirements of this
+directory vary depending on how much content you wish to download.
+
+.. note::
+   Making ``/var/lib/pulp`` a symbolic link to a different directory is possible,
+   but it is recommended that you use a bind mount instead. As of Pulp 2.8.0, using
```
bmbouter: Nice inclusion of this issue.

bmbouter:
Consider not having the renamed file `f23-.rst` but instead name it `el67.rst` or something EL related. It's going to become that in 6ish months anyway.

bmbouter:
I'm done w/ my review. I left some comments. You can handle them how you want. Thanks for these great docs. Really they are great. Thanks! Adding LGTM.

docs/user-guide/installation/f24+.rst
rbarlow: Yeah python-qpid definitely makes its way in - I have hand tested these install docs with both qpidd and RabbitMQ and everything works.

docs/user-guide/installation/f24+.rst
rbarlow: I believe that pulp-server uses the gofer library to communicate with the consumers.

docs/user-guide/installation/f24+.rst
rbarlow: Yeah that was one of my main goals with this. There are those who want Pulp to support Redis as well…

docs/user-guide/installation/index.rst
rbarlow: If you read the next line, it says "which manages the server". I didn't write this text, it's just moved from the original single file that this all came from.

docs/user-guide/installation/f24+.rst
rbarlow: I had meant to do that and forgot, so good catch! Fixed.

docs/user-guide/installation/index.rst
rbarlow: I didn't actually write this text either because it came from the previous installation.rst, but thanks ☺

rbarlow:
I think I'll leave the filename as it is for now, since it does cover Fedora 23. It can be renamed when F25 comes out.

It's ready to merge!

docs/user-guide/installation/f24+.rst
bmbouter: Oh yes @jortel verbally confirmed this earlier today.

docs/user-guide/installation/f24+.rst
bmbouter: Very excellent.

bmbouter:
Merging!

PR #2589, opened 2016-05-27
Issue #1113: If an instance of pulp_celerybeat dies unexpectedly, Pulp incorrectly tries to "cancel all tasks in its queue"

dralley:
Fixes a bug wherein duplicate celerybeat instances will see one or the other going offline as a worker going offline and treats them as such.  Hence, it will try to delete the scheduler's work queue and prints a misleading error message.

https://pulp.plan.io/issues/1113

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 This is an alphabetical (by last name) list of the authors of the Pulp project. If you submit a pull request
 to Pulp and your name is not on this list, please add yourself.
 
+Daniel Alley (dalley@redhat.com)
 Randy Barlow (rbarlow@redhat.com)
 Shubham Bhawsinka(sbhawsin@redhat.com)
 Sander Bos (www.sbosnet.nl)
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -9,8 +9,7 @@ import time
 from celery import beat
 import mongoengine
 
-from pulp.common.constants import (CELERYBEAT_WAIT_SECONDS, RESOURCE_MANAGER_WORKER_NAME,
-                                   SCHEDULER_WORKER_NAME, TICK_SECONDS)
+from pulp.common import constants
 from pulp.common.dateutils import ensure_tz
 from pulp.server.async import worker_watcher
 from pulp.server.async.celery_instance import celery as app
@@ -86,13 +85,6 @@ class CeleryProcessTimeoutMonitor(threading.Thread):
 
     Once a Celery process is determined to be missing it logs and handles cleanup appropriately.
     """
-
-    # The amount of time in seconds before a Celery process is considered missing
-    CELERY_TIMEOUT_SECONDS = 300
-
-    # The frequency in seconds with which this thread should look for missing Celery processes.
-    FREQUENCY = 60
-
     def run(self):
         """
         The thread entry point. Sleep for FREQUENCY seconds, then call check_celery_processes()
@@ -102,7 +94,7 @@ class CeleryProcessTimeoutMonitor(threading.Thread):
         """
         _logger.info(_('Worker Timeout Monitor Started'))
         while True:
-            time.sleep(self.FREQUENCY)
+            time.sleep(constants.CELERY_CHECK_INTERVAL)
             try:
                 self.check_celery_processes()
             except Exception as e:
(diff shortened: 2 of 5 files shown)


bmbouter:
ok test

server/pulp/server/async/scheduler.py
```diff
@@ -9,8 +9,10 @@
 from celery import beat
 import mongoengine
 
-from pulp.common.constants import (CELERYBEAT_WAIT_SECONDS, RESOURCE_MANAGER_WORKER_NAME,
-                                   SCHEDULER_WORKER_NAME, TICK_SECONDS)
+from pulp.common.constants import (CELERYBEAT_LOCK_MAX_AGE, RESOURCE_MANAGER_WORKER_NAME,
+                                   SCHEDULER_WORKER_NAME, TICK_DEFAULT_WAIT_TIME,
+                                   CELERY_CHECK_FREQUENCY, CELERY_TIMEOUT_SECONDS,
```
bmbouter: Can these be reordered so they are alphebatized

server/pulp/server/async/scheduler.py
```diff
@@ -122,29 +117,40 @@ def check_celery_processes(self):
...
         scheduler_count = 0
+
         for worker in worker_list:
             if worker.last_heartbeat < oldest_heartbeat_time:
-                msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
-                _logger.error(msg)
-                _delete_worker(worker.name)
+                if worker.name.startswith(SCHEDULER_WORKER_NAME):
+                    # CeleryBeat worker
+                    msg = _("pulp_celerybeat '%s' has gone missing, removing from list "
```
bmbouter: It would be useful to have this be `pulp_celerybeat@<hostname>` here. If we redefined the constant to be `pulp_celerybeat` instead of `scheduler` that would do it I think. Then this logging statement and the `_logger.error(msg)` could be moved above the `if`/`else` statments.

server/pulp/server/async/scheduler.py
```diff
@@ -121,30 +113,39 @@ def check_celery_processes(self):
...
         scheduler_count = 0
+
         for worker in worker_list:
             if worker.last_heartbeat < oldest_heartbeat_time:
                 msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
                 _logger.error(msg)
-                _delete_worker(worker.name)
-            elif worker.name.startswith(SCHEDULER_WORKER_NAME):
+
+                # Check if worker is a scheduler, and clean up appropriately
```
bmbouter: Years ago a friend of mine convinced me not to put comments of intent like this because over time they don't get maintained and then they become wrong. Instead he encouraged me to write "crazy readable code" which was as precise as code but as readable as text. This is like an implicit comment.

For example in this case, `worker.delete()` could be renamed to `worker.delete_db_entry()` and `_delete_worker(worker.name)` could be `_delete_worker_db_cancel_tasks_cleanup_reservations(worker.name)`

You don't necessarily have to do all this, I'm just giving an example about readable code.

bmbouter: As you mentioned on IRC, `delete()` is mongo defined so let's leave that one alone. [_delete_worker](https://github.com/dralley/pulp/blob/72b05d94ad312821c6ddd04805c79ae429774aea/server/pulp/server/async/tasks.py#L227) is something we define though.

AUTHORS
```diff
@@ -39,3 +39,4 @@ Chris St. Pierre (chris.a.st.pierre@gmail.com)
 Florian Weimer (fw@deneb.enyo.de)
 Adam Young (ayoung@redhat.com)
 Lukas Zapletal (lzap+git@redhat.com)
+Daniel Alley (dalley@redhat.com)
```
mhrivnak: Woo! But note that the list is alphabetical by last name, so move this up about 40 lines.

server/pulp/server/async/scheduler.py
```diff
@@ -121,30 +113,38 @@ def check_celery_processes(self):
...
         scheduler_count = 0
+
         for worker in worker_list:
             if worker.last_heartbeat < oldest_heartbeat_time:
                 msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
                 _logger.error(msg)
-                _delete_worker(worker.name)
-            elif worker.name.startswith(SCHEDULER_WORKER_NAME):
+
+                if not worker.name.startswith(constants.SCHEDULER_WORKER_NAME):
```
mhrivnak: This would be slightly more tidy if you drop the ``not`` and swap lines 132 and 134. That makes a nice flow for readability:

If this is the special case involving the scheduler:
  Do the scheduler-specific thing.
Otherwise:
  Do the normal thing.

Note that the two scheduler-related lines are together this way.

server/test/unit/server/async/test_tasks.py
```diff
@@ -525,6 +527,7 @@ class EInfo(object):
             as one of the attributes. It stores string representation of traceback
             in it's traceback instance variable. This is a stub to imitate that behavior.
             """
+
```
mhrivnak: What's with all of these added blank lines? Did your IDE perhaps do that without you meaning for it to happen?

dralley: Yeah, it was IDE autoformatting.  Want me to remove them (and blacklist that rule)?

bmbouter: Yes, I think removing the extra whitespace would be good.

server/pulp/server/async/scheduler.py
```diff
@@ -121,30 +113,38 @@ def check_celery_processes(self):
...
         resource_manager_count = 0
         scheduler_count = 0
+
         for worker in worker_list:
             if worker.last_heartbeat < oldest_heartbeat_time:
                 msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
                 _logger.error(msg)
-                _delete_worker(worker.name)
-            elif worker.name.startswith(SCHEDULER_WORKER_NAME):
+
```
bmbouter: The whitespace here could also be removed.

server/pulp/server/async/scheduler.py
```diff
@@ -121,30 +113,38 @@ def check_celery_processes(self):
...
+                if worker.name.startswith(constants.SCHEDULER_WORKER_NAME):
+                    worker.delete()
+                else:
+                    _delete_worker(worker.name)
+
```
bmbouter: The whitespace here could also be removed.

dkliban:
ok test

dkliban:
ok test

PR #2591, opened 2016-06-01
adding a new error code for missing unit file

mhrivnak:
I'm using this error code in ``pulp_rpm`` for https://pulp.plan.io/issues/1618


bmbouter:
This looks good, but the commit won't be associated with 1618 because it doesn't have `re #1618`. All commits being merged related to 1618 should be associated with https://pulp.plan.io/issues/1618

bmbouter:
LGTM, pending fixing commit message

PR #2593, opened 2016-06-01
Show repoview content if available

goosemania:
re #189
https://pulp.plan.io/issues/189

### DIFF
--- a/server/usr/share/pulp/wsgi/content.wsgi
+++ b/server/usr/share/pulp/wsgi/content.wsgi
@@ -1,4 +1,32 @@
+import os
 from pulp.server.content.web import wsgi
 
 
-application = wsgi.application
+def application(environ, start_response):
+    """
+    WSGI application which shows repoview if it exists.
+
+    :param environ: WSGI environment
+    :type  environ: dict
+    :param start_response: callable that starts HTTP response
+    :type  start_response: callable
+    :return: message body
+    :rtype: iterable
+    """
+    index_html_relative_path = 'repoview/index.html'
+    index_html_absolute_path = os.path.join(environ['PATH_INFO'], index_html_relative_path)
+    if os.path.exists(index_html_absolute_path):
+        scheme = environ['wsgi.url_scheme']
+        port = ''
+        is_custom_http_port = scheme == 'http' and environ['SERVER_PORT'] != 80
+        is_custom_https_port = scheme == 'https' and environ['SERVER_PORT'] != 443
+        if is_custom_http_port or is_custom_https_port:
+            port = ':' + environ['SERVER_PORT']
+        hostname = environ.get('HTTP_HOST', environ['SERVER_NAME'])
+        index_html_path = os.path.join(environ['REQUEST_URI'], index_html_relative_path)
+        index_html_url = '%(scheme)s://%(hostname)s%(port)s%(index_html_path)s' % \
+            {'scheme': scheme, 'port': port, 'hostname': hostname,
+             'index_html_path': index_html_path}
+        start_response('302 Moved Temporarily', [('Location', index_html_url)])
+        return []
+    return wsgi.application(environ, start_response)


server/usr/share/pulp/wsgi/content.wsgi
```diff
@@ -1,4 +1,32 @@
...
+    if os.path.exists(index_html_absolute_path):
+        scheme = environ['wsgi.url_scheme']
+        port = ''
+        is_custom_http_port = scheme == 'http' and environ['SERVER_PORT'] != 80
+        is_custom_https_port = scheme == 'https' and environ['SERVER_PORT'] != 443
+        if is_custom_http_port or is_custom_https_port:
+            port = ':' + environ['SERVER_PORT']
+        hostname = environ.get('HTTP_HOST', environ['SERVER_NAME'])
+        index_html_path = os.path.join(environ['REQUEST_URI'], index_html_relative_path)
+        index_html_url = '%(scheme)s://%(hostname)s%(port)s%(index_html_path)s' % \
```
jeremycline: I think this code be simplified by using https://docs.python.org/2.7/library/urlparse.html#urlparse.urlunsplit (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit in Python 3, and it'd be nice to be compatible)

jeremycline:
I'm wonder if it would be better to handle this redirect in the Apache configuration rather than in the WSGI app. It may be that we can get by with one or two lines of Apache configuration using ``mod_rewrite``.

jeremycline:
I think something along the lines of
```
RewriteEngine on
RewriteCond "%{DOCUMENT_ROOT}/%{REQUEST_URI}/repoview/index.html" -f
RewriteRule "^(.+)" "%{DOCUMENT_ROOT}/%{REQUEST_URI}/repoview/index.html" [L]
```
might do it.

goosemania:
@jeremycline, thank you for the ideas!
What if someone would like to use other web server rather than Apache... Maybe it will be easier for them not to port those rewrites but to have this logic in the WSGI app. What do you think?

jeremycline:
I'm definitely torn on which is better. On the one hand, having Apache handle it is appealing because it's only another line or to of configuration and it will probably perform better since there's no 302, just an internal redirect. On the other hand it would increase the complexity of our already-painfully-complex Apache configs and we probably don't care much about performance in this case.

I think I lean slightly more towards the Apache config option because if it's not present everything still works, the user just needs to navigate to repoview/index.html manually.

jeremycline:
@goosemania the more I think about it, the more I think we should do the Apache configuration (and it sounded like you thought that as well when we talked). The big advantage I see is people can turn it off if they don't want it.

Based on the configuration you sent me, I think the only thing that needs to change is to add the [R] flag and make a note about why an external redirect is used:
```
diff --git a/server/etc/httpd/conf.d/pulp_content.conf b/server/etc/httpd/conf.d/pulp_content.conf
index 822b116..830dc1f 100755
--- a/server/etc/httpd/conf.d/pulp_content.conf
+++ b/server/etc/httpd/conf.d/pulp_content.conf
@@ -20,4 +20,9 @@ WSGIImportScript /usr/share/pulp/wsgi/content.wsgi process-group=pulp-content ap
     XSendFile on
     XSendFilePath /var/lib/pulp/content
     XSendFilePath /var/lib/pulp/published
+    RewriteEngine on
+    # Redirect to the repoview if it exists; an external redirect is used to ensure the
+    # relative links in the html document don't break.
+    RewriteCond "%{PATH_INFO}repoview/index.html" -f
+    RewriteRule "^(.+)" "%{REQUEST_URI}repoview/index.html" [R]
 </Location>
```

goosemania:
@jeremycline, thanks for the discussion!
I did more experiments today and found the other not good detail about this rule - the final URI that user will see does not look good, something like `/pulp/content/var/www/pub/yum/https/repos/sat61rhel7/repoview/index.html`. I tried different ways to make it work better, for example, to pass through [PT] this request so it will be processed one more time via all the scripts and so on, but I have not succeeded.
So at the end I wrote the similar rule(s) in pulp_rpm config, it is in pulp_rpm PR pulp/pulp_rpm#892. Please take a look at it, is there a way to simplify it?
If we will agree on the config in pulp_rpm PR , I will close this one then.

goosemania:
This configuration moved to the plugin PR pulp/pulp_rpm#892

PR #2595, opened 2016-06-05
Support specialized unit classes in storage-path migration

jortel:
To support unit types like *distributions*, the migration in the platform needs to support specialized units that can specify the paths to files stored within the directory referenced by `_storage_path`.  This is needed to migrate all of the possible symlinks created during publishing.

Need for: https://pulp.plan.io/issues/1972


ipanova:
@jortel it looks good, would you mind fixing flake8 issues before merging. I guess it is because SideEffect imported but unused. https://github.com/pulp/pulp/pull/2595/files#diff-c86969faab72b9e6d6e7bdeda50f5978R8

PR #2597, opened 2016-06-06
[WIP] Issue #898 - Prevent two instances of pulp_resource_manager from operating in parallel

dralley:
Adds a mechanism to enforce that only one resource manager can be executing at a time.  Duplicate copies of the resource manager will wait until the original one dies off.

closes #898
https://pulp.plan.io/issues/898


server/pulp/server/async/app.py
```diff
@@ -49,6 +61,22 @@ def initialize_worker(sender, instance, **kwargs):
     """
     initialization.initialize()
 
+    if sender.startswith(constants.RESOURCE_MANAGER_WORKER_NAME):
+        msg = _("An instance of pulp_resource_manager is already running. "
+                "Halting execution of the dpulicate pulp_resource_monitor "
+                "until the original dies.")
+        _logger.info(msg)
```
dkliban: This is logged too early.

dkliban:
You should rebase against the latest changes on 2.8-dev branch. You should also find all references to the resource manager in the documentation and make sure that we no longer tell users that the resource manager is a singleton. Add a release note for 2.8.x stream.

server/pulp/server/async/app.py
dkliban: The message should probably be more along the lines of "An instance of pulp_resource_manager is already active. Halting execution until the currently active resource_manager becomes unavailable."

server/pulp/server/async/app.py
```diff
@@ -49,6 +61,27 @@ def initialize_worker(sender, instance, **kwargs):
...
+        name = constants.RESOURCE_MANAGER_WORKER_NAME + "@" + platform.node()
+        lock = ResourceManagerLock(name=name)
+
+        while True:
+            try:
+                lock.save()
+                break
+            except mongoengine.NotUniqueError:
+                if not _logged:
+                    msg = _("An instance of pulp_resource_manager is already running. "
```
dkliban: The message should probably be more along the lines of "An instance of pulp_resource_manager is already active. Halting execution until the currently active resource_manager becomes unavailable."

server/pulp/server/async/app.py
```diff
@@ -49,6 +61,27 @@ def initialize_worker(sender, instance, **kwargs):
...
+    _logged = False
+
+    if sender.startswith(constants.RESOURCE_MANAGER_WORKER_NAME):
+
+        name = constants.RESOURCE_MANAGER_WORKER_NAME + "@" + platform.node()
+        lock = ResourceManagerLock(name=name)
+
+        while True:
+            try:
+                lock.save()
```
dkliban: Probably should add a log statement about this instance of resource_manager becoming active.

PR #2607, opened 2016-06-15
Serve index.json for directory requests

asmacdo:
Python API requires serving a json file for a call to a directory. Using
index.html would return the incorrect `Content-Type`


asmacdo:
I was able to do this in the pulp_python apache config instead.

PR #2610, opened 2016-06-20
dev env: add support for Vagrant docker provider

rohanpm:
Let the developer choose between libvirt or docker for the Vagrant provider.

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -3,11 +3,25 @@
 
 VAGRANTFILE_API_VERSION = "2"
 
+# Synced folders are declared here,
+# host => guest
+VAGRANT_SYNCED_FOLDERS = {
+  # This code
+  '.' => '/vagrant',
+
+  # For access to sibling repos (e.g. pulp_rpm, pulp_docker, ...)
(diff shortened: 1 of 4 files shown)


Vagrantfile.example
```diff
@@ -64,13 +83,66 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+        # Uncomment this to resize the root partition and filesystem to fill the base box disk
+        # This script is only guaranteed to work with the default official fedora image, and is
+        # only needed it you changed machine_virtual_size above.
+        # For other boxen, use at your own risk
+        # override.vm.provision "shell", path: "playpen/vagrant-resize-disk.sh"
+    end
+
+    dev.vm.provider :docker do |d, override|
+        override.vm.box = nil
+        d.image = 'rohanpm/fedora-vagrant:23'
```
seandst: I'd rather use the "official" fedora image directly, but looking at the dockerfile here that doesn't seem reasonable.

seandst:
Works great for me, I'll get this merged today.

PR #2611, opened 2016-06-20
Prevent more than one resource manager from operating in parallel - #898

dralley:
Adds a mechanism to enforce that only one resource manager can be executing at a time. Duplicate copies of the resource manager will wait until the original one dies off.

closes #898
https://pulp.plan.io/issues/898

### DIFF
--- a/docs/dev-guide/integration/rest-api/status.rst
+++ b/docs/dev-guide/integration/rest-api/status.rst
@@ -18,12 +18,11 @@ and ``known_workers`` to get more detailed status information.
     of a Pulp deployment, and is not meant to replace monitoring of Pulp
     components in a production environment.
 
-A healthy Pulp installation will contain exactly one record for
-"resource_manager" and "scheduler" in the worker list, and one or more
-"reserved_resource_worker" records. It will also have
-``messaging_connection`` and ``database_connection`` entries that contain ``{connected: True}``.
-Note that if the scheduler is not running, other workers may be running but not
-updating their last heartbeat record.
+A healthy Pulp installation will contain one or more records for "resource_manager",
+one or more records for "scheduler", and one or more records for "reserved_resource_worker"
+in the worker list. It will also have ``messaging_connection`` and ``database_connection``
+entries that contain ``{connected: True}``. Note that if the scheduler is not running,
+other workers may be running but not updating their last heartbeat record.
 
 The version of Pulp is also returned via ``platform_version`` in the
 ``versions`` object. This field is calculated from the "pulp-server" python
--- a/docs/user-guide/installation/f23-.rst
+++ b/docs/user-guide/installation/f23-.rst
@@ -220,14 +220,13 @@ Server
       $ sudo systemctl enable pulp_celerybeat
       $ sudo systemctl start pulp_celerybeat
 
-   .. warning::
+   Lastly, a ``pulp_resource_manager`` process must be running in the installation. This process
+   acts as a task router, deciding which worker should perform certain types of tasks. As with
+   ``pulp_celerybeat``, multiple instances of ``pulp_resource_manager`` may be run concurrently on
+   separate hosts to increase fault tolerance, however, only one instance will ever be active at a
+   time. Should the active instance become unavailable, another instance will take over after some
+   delay.
 
-      ``pulp_resource_manager`` must be singleton, so be sure that you
-      only enable this on one host if you are Pulp's clustered deployment.
-
-   Lastly, one ``pulp_resource_manager`` process must be running in the installation. This process
-   acts as a task router, deciding which worker should perform certain types of tasks. Apologies
-   for the repetitive message, but it is important that this process only be enabled on one host.
    Edit ``/etc/default/pulp_resource_manager`` to your liking. Then, for upstart::
 
       $ sudo chkconfig pulp_resource_manager on
@@ -319,7 +318,7 @@ specific consumer dependencies with one command by running:
    By default, the agent will connect using a plain TCP connection.
 
 
-4. Set the agent to start at boot.  For upstart::
+4. Set the agent to start at boot. For upstart::
 
       $ sudo chkconfig goferd on
       $ sudo service goferd start
--- a/docs/user-guide/installation/f24+.rst
+++ b/docs/user-guide/installation/f24+.rst
@@ -178,18 +178,17 @@ Now we are ready to install and configure the Pulp server!
       $ sudo systemctl enable pulp_celerybeat
       $ sudo systemctl start pulp_celerybeat
 
-   Lastly, exactly one ``pulp_resource_manager`` process must be running in the installation. This
-   process acts as a task router, deciding which worker should perform certain types of tasks.
-   Apologies for the repetitive message, but it is important that this process only be enabled on
-   one host. Edit ``/etc/default/pulp_resource_manager`` to your liking. Then::
+   Lastly, a ``pulp_resource_manager`` process must be running in the installation. This process
+   acts as a task router, deciding which worker should perform certain types of tasks. As with
+   ``pulp_celerybeat``, multiple instances of ``pulp_resource_manager`` may be run concurrently on
+   separate hosts to increase fault tolerance, however, only one instance will ever be active at a
+   time. Should the active instance become unavailable, another instance will take over after some
+   delay.
+
+   Edit ``/etc/default/pulp_resource_manager`` to your liking. Then::
 
       $ sudo systemctl enable pulp_resource_manager
       $ sudo systemctl start pulp_resource_manager
-
-   .. warning::
-
-      ``pulp_resource_manager`` must be singleton, so be sure that you
-      only enable this on one host if you are deploying Pulp in a clustered environment.
 
 
 Admin Client
@@ -250,7 +249,7 @@ repositories.
    and Puppet plugins support the Pulp Agent.
 
    ::
- 
+
       $ sudo dnf install pulp-puppet-consumer-extensions pulp-rpm-consumer-extensions
 
 
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -3,18 +3,35 @@ This module is the Pulp Celery App. It is passed to the workers from the command
 will see the "celery" module attribute and use it. This module also initializes the Pulp app after
 Celery setup finishes.
 """
+
+import contextlib
+import logging
+import platform
+import sys
+import signal
+import time
+from datetime import datetime
+from gettext import gettext as _
+
+import mongoengine
 from celery.signals import celeryd_after_setup
+
+from pulp.common import constants
+from pulp.server import initialization
+from pulp.server.async import tasks
+from pulp.server.db.model import ResourceManagerLock, Worker
+from pulp.server.managers.repo import _common as common_utils
 
 # This import will load our configs
 from pulp.server import config  # noqa
-from pulp.server import initialization
 # We need this import so that the Celery setup_logging signal gets registered
 from pulp.server import logs  # noqa
-from pulp.server.async import tasks
 # This import is here so that Celery will find our application instance
 from pulp.server.async.celery_instance import celery  # noqa
-from pulp.server.managers.repo import _common as common_utils
+
 import pulp.server.tasks  # noqa
+
+_logger = logging.getLogger(__name__)
 
 
 @celeryd_after_setup.connect
@@ -22,7 +39,7 @@ def initialize_worker(sender, instance, **kwargs):
     """
     This function performs all the necessary initialization of the Celery worker.
 
-    It starts by cleaning up old state if this worker was previously running, but died unexpectedly.
+    We clean up old state in case this worker was previously running, but died unexpectedly.
     In such cases, any Pulp tasks that were running or waiting on this worker will show incorrect
     state. Any reserved_resource reservations associated with the previous worker will also be
     removed along with the worker entry in the database itself. The working directory specified in
@@ -38,6 +55,15 @@ def initialize_worker(sender, instance, **kwargs):
     It uses the celeryd_after_setup signal[0] so that it gets called by Celery after logging is
     initialized, but before Celery starts to run tasks.
 
+    If the worker is a resource manager, it tries to acquire a lock stored within the database.
+    If the lock cannot be acquired immediately, it will wait until the currently active instance
+    becomes unavailable, at which point the worker cleanup routine will clear the lock for us to
+    acquire. While the worker remains in this waiting state, it is not connected to the broker and
+    will not attempt to do any work. A side effect of this is that, if terminated while in this
+    state, the process will not send the "worker-offline" signal used by the EventMonitor to
+    immediately clean up terminated workers. Therefore, we override the SIGTERM signal handler
+    while in this state so that cleanup is done properly.
+
     [0] http://celery.readthedocs.org/en/latest/userguide/signals.html#celeryd-after-setup
 
     :param sender:   The hostname of the worker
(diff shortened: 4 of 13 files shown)


docs/dev-guide/integration/rest-api/status.rst
```diff
@@ -18,12 +18,11 @@ and ``known_workers`` to get more detailed status information.
...
-A healthy Pulp installation will contain exactly one record for
-"resource_manager" and "scheduler" in the worker list, and one or more
-"reserved_resource_worker" records. It will also have
-``messaging_connection`` and ``database_connection`` entries that contain ``{connected: True}``.
-Note that if the scheduler is not running, other workers may be running but not
-updating their last heartbeat record.
+A healthy Pulp installation will contain one or more records for "resource_manager",
+one or more records for "scheduler", and one or more records for "reserved_resource_worker"
+in the worker list. It will also have ``messaging_connection`` and ``database_connection``
+entries that contain ``{connected: True}``.  Note that if the scheduler is not running,
```
mhrivnak: One space between sentences please.

docs/user-guide/installation/f23-.rst
```diff
@@ -220,14 +220,13 @@ Server
...
 
-   .. warning::
-
-      ``pulp_resource_manager`` must be singleton, so be sure that you
-      only enable this on one host if you are Pulp's clustered deployment.
+   Lastly, a ``pulp_resource_manager`` process must be running in the installation. This process
+   acts as a task router, deciding which worker should perform certain types of tasks. As with
+   ``pulp_celerybeat``, multiple instances of ``pulp_resource_manager`` may be run concurrently on
+   separate hosts to increase fault tolerance, however, only one instance will ever be active at a
+   time.  Should the active instance become unavailable, another instance will take over after some
```
mhrivnak: One space between sentences please.

docs/user-guide/installation/f24+.rst
```diff
@@ -178,19 +178,18 @@ Now we are ready to install and configure the Pulp server!
...
 
-   Lastly, exactly one ``pulp_resource_manager`` process must be running in the installation. This
-   process acts as a task router, deciding which worker should perform certain types of tasks.
-   Apologies for the repetitive message, but it is important that this process only be enabled on
-   one host. Edit ``/etc/default/pulp_resource_manager`` to your liking. Then::
+   Lastly, a ``pulp_resource_manager`` process must be running in the installation. This process
+   acts as a task router, deciding which worker should perform certain types of tasks. As with
+   ``pulp_celerybeat``, multiple instances of ``pulp_resource_manager`` may be run concurrently on
+   separate hosts to increase fault tolerance, however, only one instance will ever be active at a
+   time.  Should the active instance become unavailable, another instance will take over after some
```
mhrivnak: One space between sentences please.

server/pulp/server/async/app.py
```diff
@@ -3,26 +3,41 @@
 will see the "celery" module attribute and use it. This module also initializes the Pulp app after
 Celery setup finishes.
 """
+import sys
```
mhrivnak: We try to keep each import group sorted alphabetically.

server/pulp/server/async/app.py
```diff
@@ -3,26 +3,41 @@
...
+import time
+import signal
+import logging
+import platform
+import mongoengine
```
mhrivnak: PEP8 includes this bit:

```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

Since this is a third-party library, it goes in the second group of imports.

server/pulp/server/db/model/__init__.py
```diff
@@ -983,6 +983,24 @@ class CeleryBeatLock(AutoRetryDocument):
...
+    :ivar lock: A unique key set to "locked" when lock is acquired.
+    :type lock: basestring
+    :ivar _ns: (Deprecated), Contains the name of the collection this model represents
+    :type _ns: mongoengine.StringField
+    """
+    name = StringField(required=True)
+    lock = StringField(required=True, default="locked", unique=True)
+
+    # For backward compatibility
+    _ns = StringField(default='resource_manager_lock')
```
mhrivnak: I don't think we need this field on new models. I think it's only needed on models that existed before we started using mongoengine. It's worth querying the team just to be sure.

server/test/unit/server/async/test_app.py
```diff
@@ -1,11 +1,16 @@
 """
 This module contains tests for the pulp.server.async.app module.
 """
-import unittest
-
 import mock
+import unittest
```
mhrivnak: Since these two imports are part of the standard library, they should be in the first group of imports. ``mock`` should be alone in the second group.

server/test/unit/server/async/test_app.py
```diff
@@ -1,11 +1,16 @@
...
+import platform
 
 from pulp.server.async import app
+from pulp.common.constants import RESOURCE_MANAGER_WORKER_NAME
```
mhrivnak: For organization, it helps to keep these sorted. So this should go before any ``pulp.server`` imports.

server/pulp/server/async/app.py
```diff
@@ -49,8 +74,54 @@ def initialize_worker(sender, instance, **kwargs):
     """
     initialization.initialize()
 
+    # Delete any potential old state
     tasks._delete_worker(sender, normal_shutdown=True)
 
     # Create a new working directory for worker that is starting now
     common_utils.delete_worker_working_directory(sender)
     common_utils.create_worker_working_directory(sender)
+
```
mhrivnak: Everything you have added below is distinct from what happens above. It's a large body of code that operates independent of everything else in the function, so it's a great candidate for moving into a separate function.

One of the indicators I noticed that made me think about this was the stack of 7 mocks on one of the unit tests. That's often a sign that some re-organization would be beneficial, usually to separate parts of code that don't need to be together. Testing becomes much easier when the "units" get smaller.

server/pulp/server/db/model/__init__.py
dralley: Don't think there's much point to determining whether it's still necessary when the Postgres transition is only a few months away.  Even if it isn't necessary, it won't hurt anything, and it's consistent with the rest of the models.  Or, if we make that determination, we can remove it from all of the models in one pass separately.

server/pulp/server/async/app.py
```diff
@@ -49,8 +74,54 @@ def initialize_worker(sender, instance, **kwargs):
...
+                        tasks._delete_worker(name, normal_shutdown=True)
+                        sys.exit(0)
+
+                    signal.signal(signal.SIGTERM, sigterm_handler)
+
+                    msg = _("An instance of pulp_resource_manager is already active. "
+                            "Halting '%s' until the currently active "
+                            "resource_manager becomes unavailable." % name)
+                    _logger.info(msg)
+                    _cycled = True
```
mhrivnak: The replacement of the sigterm handler looks good and correct. It's worth considering a slightly different approach that may help readability and testability.

Consider making a context manager or decorator that adds your handler, then replaces the original. Something like what's below would work, although I'd also consider making ``custom_sigterm_handler`` a decorator, and just decorate the ``get_lock`` function. Either way would be fine, and it might just depend on whether you need control about where in the function the replacement happens.

```
@contextlib.contextmanager
def custom_sigterm_handler(name):
    def sigterm_handler(_signo, _stack_frame):
        msg = _("Worker '%s' shutdown" % name)
        _logger.info(msg)
        tasks._delete_worker(name, normal_shutdown=True)
        sys.exit(0)

    signal.signal(signal.SIGTERM, sigterm_handler)

    yield

    signal.signal(signal.SIGTERM, signal.SIG_DFL)


def get_lock():
    # Whether this instance has had to wait for availability
    _cycled = False

    # If the worker is a resource manager, try to acquire the lock, or wait until it
    # can be acquired
    if sender.startswith(constants.RESOURCE_MANAGER_WORKER_NAME):

        name = constants.RESOURCE_MANAGER_WORKER_NAME + "@" + platform.node()
        lock = ResourceManagerLock(name=name)

        with custom_context_manager():
            # I'm going off memory for this, but you can do something like this to know if the save worked
            while lock.pk is None:

                # Create a worker record so the user can tell that we're running
                Worker.objects(name=sender).update_one(set__last_heartbeat=datetime.utcnow(),
                                                       upsert=True)
                try:
                    lock.save()

                    msg = _("A new instance of pulp_resource_manager '%s' has become "
                            "active.") % sender
                    _logger.info(msg)
                except mongoengine.NotUniqueError:
                    msg = _("An instance of pulp_resource_manager is already active. "
                            "Halting '%s' until the currently active "
                            "resource_manager becomes unavailable." % name)
                    _logger.info(msg)

                    time.sleep(constants.CELERY_CHECK_INTERVAL)
```

server/pulp/server/db/model/__init__.py
bmbouter: We need to continue modeling this because it gets injected onto every document. This is configured at the PyMongo layer [here](https://github.com/pulp/pulp/blob/644b607528c406fddc69b2c07fa491c2497235a4/server/pulp/server/db/connection.py#L159). We will remove the son_injector and add a migration to remove the field from each collection with [issue 625](https://pulp.plan.io/issues/625) as part of Pulp 3.

server/pulp/server/async/app.py
```diff
@@ -49,8 +75,72 @@ def initialize_worker(sender, instance, **kwargs):
...
+
+
+def get_resource_manager_lock(sender):
+    """
+    Tries to acquire the resource manager lock. If the lock cannot be acquired immediately, it
+    will wait until the currently active instance becomes unavailable, at which point the worker
+    cleanup routine will clear the lock for us to acquire. A worker record will be created so that
+    the waiting resource manager will appear in the Status API. We override the SIGTERM signal
+    handler so that that the worker record will be immediately cleaned up if the process is killed
+    while in this states.
```
mhrivnak: Needs to define param ``sender``.

server/pulp/server/async/app.py
```diff
@@ -49,8 +75,72 @@ def initialize_worker(sender, instance, **kwargs):
...
+                    _first_check = False
+
+                time.sleep(constants.CELERY_CHECK_INTERVAL)
+
+
+@contextlib.contextmanager
+def custom_sigterm_handler(name):
+    """
+    Temporarily installs a custom SIGTERM handler that performs cleanup of the worker record with
+    the provided name. Resets the signal handler to the default one after leaving.
```
mhrivnak: Needs to define param ``name``.

server/pulp/server/async/app.py
```diff
@@ -49,8 +75,72 @@ def initialize_worker(sender, instance, **kwargs):
...
+                if _first_check:
+                    msg = _("Resource manager '%s' attempted to acquire the the resource manager "
+                            "lock but was unable to do so. It will retry every %d seconds until "
+                            "the lock can be acquired." % (name, name,
+                                                           constants.CELERY_CHECK_INTERVAL))
+                    _logger.info(msg)
+                    _first_check = False
+
+                time.sleep(constants.CELERY_CHECK_INTERVAL)
+
```
mhrivnak: I assume that if a connection error happens, that would bubble up as unhandled, and the process would exit? If that's the case, it would be a great and easy enhancement to also handle a connection error here, log that the connection failed, sleep for the check interval, and try again.

mhrivnak: Depending on what the previous behavior was, and how this changes it, there could be a regression. I *think* all of our services are supposed to be tolerant now of waiting for the DB to become available.

bmbouter: @dralley @mhrivnak Yes all Pulp services must wait for database services to become available. If they are not available they are to wait and retry. This is true for all `pulp_*` services and `httpd`.

server/test/unit/server/async/test_app.py
```diff
@@ -16,17 +23,36 @@ class InitializeWorkerTestCase(unittest.TestCase):
...
-    def test_initialize_worker(self, _delete_worker, initialize, create_worker_working_directory,
+    def test_initialize_worker(self, _delete_worker, initialize,
+                               create_worker_working_directory,
                                delete_worker_working_directory):
         """
         Assert that initialize_worker() calls Pulp's initialization code and the appropriate worker
-        monitoring code.
+        monitoring code for a non-resource mananger worker.
         """
         sender = mock.MagicMock()
```
mhrivnak: Is ``sender`` just a string in normal operation? If so, I suggest making it a string here for simplicity, and make it actually start with the value you want. It seems silly that it was ever a ``MagicMock`` to begin with.

Then you can test both branches of the ``if`` statement by doing two tests with two different values of ``sender``.

server/pulp/server/async/app.py
```diff
@@ -49,8 +75,72 @@ def initialize_worker(sender, instance, **kwargs):
...
+
+@contextlib.contextmanager
+def custom_sigterm_handler(name):
```
mhrivnak: What do you think about a basic unit test or two for this function?

bmbouter:
@dralley Also can a release not be added to the 2.10 release notes? If they don't exist yet can you start them?

server/pulp/server/async/app.py
dralley: @mhrivnak @bmbouter All of that is handled by the call to initialization.initialize(), which calls to connection.initialize().

Since that is the first step here (same as it was before) there should be no regression.

mhrivnak: Great. We should also handle the case where the database becomes temporarily unavailable after initialization. That should be easy to handle. We can do so in a separate story if you like. Do you mind filing that story?

bmbouter: @mhrivnak We already have a set of behaviors which are implicit everywhere in Pulp for just such a situation. The user can choose to have our DB auto retry by enabling an option, or if they don't an exception will be raised and all processes will exit (which is expected). Why would we need more than what we already have?

mhrivnak: @bmbouter talked me out of this. ;)

server/test/unit/server/async/test_app.py
```diff
@@ -16,17 +24,87 @@ class InitializeWorkerTestCase(unittest.TestCase):
...
+        handler = signal.getsignal(signal.SIGTERM)
+        self.assertEquals(handler, signal.SIG_DFL)
+
+        name = RESOURCE_MANAGER_WORKER_NAME + '@' + platform.node()
+
+        with app.custom_sigterm_handler(name):
+            handler = signal.getsignal(signal.SIGTERM)
+            self.assertNotEquals(handler, signal.SIG_DFL)
+
+            handler(None, None)
```
mhrivnak: This would be good as a separate test. A goal for a unit test should be that if it fails, the developer knows exactly what broke. This test is currently testing two different things. 1) did the context manager swap the signal handler in and out, and 2) is the custom signal handler doing its job.

Even though the setup for both tests will be similar or identical, and it seems convenient to assert both things in the same test, the test suite will be more useful if we separate them.

server/test/unit/server/async/test_app.py
```diff
@@ -16,17 +24,87 @@ class InitializeWorkerTestCase(unittest.TestCase):
...
         initialize.assert_called_once_with()
         _delete_worker.assert_called_once_with(sender, normal_shutdown=True)
         create_worker_working_directory.assert_called_once_with(sender)
         delete_worker_working_directory.assert_called_once_with(sender)
+        mock_get_resource_manager_lock.assert_called_once_with(sender)
+
+    @mock.patch('pulp.server.async.app.datetime')
+    @mock.patch('pulp.server.async.app.Worker')
+    @mock.patch('pulp.server.async.app.ResourceManagerLock')
+    def test_get_resource_manager_lock(self, mock_rm_lock, mock_worker, mock_datetime):
```
mhrivnak: One last testing request: it would be particularly valuable to have a test that exercises the block of code under the ``except`` statement. Especially since the normal path for typical deployments will never hit that block, basic test coverage will be a good idea.

You can mock ``sleep()`` so it doesn't actually sleep, and make ``mock_rm_lock.objects().save.side_effects`` be a list so it raises an exception the first time, and not the second time.

mhrivnak:
This is looking great. I *think* these last two unit test requests are all I'll need. Thanks!

server/test/unit/server/async/test_app.py
```diff
@@ -16,17 +24,98 @@ class InitializeWorkerTestCase(unittest.TestCase):
...
+    def test_custom_sigterm_handler(self, _delete_worker, mock_sys):
+        """
+        Assert that the signal handler installed by the custom_sigterm_handler context manager
+        calls the delete_worker cleanup routine with the correct worker name and then exits.
+        """
+        name = RESOURCE_MANAGER_WORKER_NAME + '@' + platform.node()
+
+        with app.custom_sigterm_handler(name):
+            handler = signal.getsignal(signal.SIGTERM)
+            self.assertNotEquals(handler, signal.SIG_DFL)
```
mhrivnak: I think this assertion belongs in the test below. It's purpose is to make sure the context manager did its job of switching the handler in and out.

If you want to keep it here as a point of sanity so the test doesn't accidentally call the wrong handler, which could have unfortunate consequences, that could be fine. I'd personally keep the assertions as focused as possible though on the purpose of the test.

server/test/unit/server/async/test_app.py
```diff
@@ -16,17 +24,98 @@ class InitializeWorkerTestCase(unittest.TestCase):
...
+        Assert that the custom_sigterm_handler context manager properly sets and restores the
+        SIGTERM signal handler upon entry and exit.
+        """
+        handler = signal.getsignal(signal.SIGTERM)
+        self.assertEquals(handler, signal.SIG_DFL)
+
+        name = RESOURCE_MANAGER_WORKER_NAME + '@' + platform.node()
+
+        with app.custom_sigterm_handler(name):
+            handler(None, None)
```
mhrivnak: I don't think this test should call the handler, especially if it's not making any assertions about the execution of it. This would be a great place to instead call ``getsignal`` and assert that the current handler is not the default one. That keeps the assertions focused on the purpose of the test, which is to swap the custom handler in and out.

mhrivnak:
Well done!

PR #2612, opened 2016-06-20
trying to make flake8 happy by fixing an import statement

mhrivnak:
I don't know why the jenkins flake8 runs are complaining about this while my
vagrant environment is not, but it does appear to be a valid complaint.


mhrivnak:
No flake8 complaints!

PR #2613, opened 2016-06-21
Revise use of the TargetRelease field

asmacdo:


seandst:
LGTM in general, but for consistency it would be good to use the same field name as in Redmine: "Target Platform Release".

PR #2614, opened 2016-06-22
Handle queries containing base64 padding gracefully

jeremycline:
The lazy URL parser does not work with url-safe base64-encoded queries containing
padding ('='). It's worth investigating why the standard library parser
isn't used, but this is an easy fix for the immediate future.

closes #2031


bmbouter:
All tests pass except RHEL5 which is a known failure. @jeremycline Thanks for the fix.

PR #2615, opened 2016-06-23
Update Vagrant and Ansible to work with Fedora 24

jeremycline:
Since all our dependencies are in F24 we don't need to enable external repositories to get a working development environment.


asmacdo:
vagrant ups as expected, without error.

PR #2616, opened 2016-06-23
Importer config is not validated during the update.

ipanova:
closes #2030
https://pulp.plan.io/issues/2030


dkliban:
LGTM once tests pass

PR #2617, opened 2016-06-27
Merge the Apache 2.2 and 2.4 configurations

jeremycline:
In preparation for a clean-up of our Apache configurations[0], merging
the Apache configurations seems wise. ``mod_version``[1] appears in
Apache 2.0 onward, so we can use it to do if/else blocks in the configs.
There are very few differences between the two configurations so it
makes sense to maintain just one.

[0] https://pulp.plan.io/issues/1257
[1] https://httpd.apache.org/docs/2.4/mod/mod_version.html

I do not expect this to land in 2.9 and as such have not made changes to the spec files. Before this is merged, a change to the spec file for Pulp in https://github.com/pulp/packaging should be made (unless we haven't switched by then, in which case a change will need to be made to the spec file in this repository).


server/etc/httpd/conf.d/pulp.conf
```diff
@@ -51,7 +53,13 @@ WSGIImportScript /usr/share/pulp/wsgi/webservices.wsgi process-group=pulp applic
 </Files>
 
 <VirtualHost *:80>
-    IncludeOptional /etc/pulp/vhosts80/*.conf
+    <IfVersion >= 2.4>
+        # Starting in 2.4 if there are 0 matches this is a configuration error
+        IncludeOptional /etc/pulp/vhosts80/*.conf
+    </IfVersion>
+    <IfVersion >= 2.4>
```
mhrivnak: I suspect this should be ``< 2.4``

server/etc/httpd/conf.d/pulp.conf
```diff
@@ -51,7 +53,13 @@ WSGIImportScript /usr/share/pulp/wsgi/webservices.wsgi process-group=pulp applic
 </Files>
 
 <VirtualHost *:80>
-    IncludeOptional /etc/pulp/vhosts80/*.conf
+    <IfVersion >= 2.4>
+        # Starting in 2.4 if there are 0 matches this is a configuration error
```
mhrivnak: I read this as describing the behavior prior to 2.4. Before ``IncludeOptional`` existed, we could only use ``Include``, which considers 0 matches to be an error.

For this block, my understanding is that 0 matches would not be an error.

One way or another, I think it would be helpful to clarify this comment.

server/etc/httpd/conf.d/pulp.conf
```diff
@@ -35,9 +35,11 @@ WSGISocketPrefix run/wsgi
 WSGIScriptAlias /pulp/api /usr/share/pulp/wsgi/webservices.wsgi
 WSGIImportScript /usr/share/pulp/wsgi/webservices.wsgi process-group=pulp application-group=pulp
 
-<Directory /usr/share/pulp/wsgi>
-    Require all granted
-</Directory>
+<IfVersion >= 2.4>
+    <Directory /usr/share/pulp/wsgi>
```
mhrivnak: Not directly related to this review: any idea why there wasn't a corresponding block in the 2.2 config? Obviously the WSGI app works on 2.2, so maybe this block isn't required at all? Or maybe there's some new behavior in 2.4 that requires it? A comment would be helpful if we can figure out why this is here.

mhrivnak:
Looks good, pending the spec file changes.

server/etc/httpd/conf.d/pulp.conf
jeremycline: Woops

server/etc/httpd/conf.d/pulp.conf
jeremycline: Ah okay. I'll clarify it with something like
```
# `Include /etc/pulp/vhosts80/*.conf` results in a configuration error in 2.4+ if the glob fails to match anything.
# The new `IncludeOptional` directive behaves as `Include` did in 2.2 and below.
```

server/etc/httpd/conf.d/pulp.conf
jeremycline: No, I didn't look into the saneness (or lack there of) of any of the directives. I want to tackle that separately for issue 1257.

jeremycline:
Given that our packaging is sort of in limbo and that we're dropping EL6 support (likely) so we only need to provide an Apache configuration for 2.4+, I think it's best to just close this.

PR #2034, opened 2015-09-10
a repo unit association query now works for types defined with mongoengine

mhrivnak:
https://pulp.plan.io/issues/1239

This bug affected the "get_units" method of at least two conduit mixins.

fixes #1239

PR #2620, opened 2016-06-28
Removes intersphinx, updates urls, adds doc guide

bmbouter:
This removes intersphinx and removes all content which used
intersphinx.

URLs are also updated to use docs.pulpproject.org instead of
readthedocs.io

Finally, the docs guide is updated to match the current state
of things for contributors.

Also, one really tiny formatting fix I saw while I was editing.

https://pulp.plan.io/issues/2034
fixes #2034

### DIFF
--- a/docs/dev-guide/contributing/documenting.rst
+++ b/docs/dev-guide/contributing/documenting.rst
@@ -18,7 +18,6 @@ platform guide and instead focus on these two topics:
    and interesting operations. For example, create a repository, sync it, and
    publish it. Show more advanced stories as "recipes".
 
-
 Command Line User Guide
 -----------------------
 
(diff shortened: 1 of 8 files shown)


docs/dev-guide/contributing/documenting.rst
```diff
@@ -40,52 +39,36 @@ must be run as root should be shown using ``sudo``.
...
--------------
-
-Pulp's documentation is hosted on `Read the Docs <http://readthedocs.org>`_.
-Links to all current documentation can be found at
-`http://www.pulpproject.org/docs <http://www.pulpproject.org/docs>`_.
+Doc Hosting
+-----------
 
+Pulp's documentation is hosted on `OpenShift <https://www.openshift.com/>`_.
+and is available at `https://docs.pulpproject.org/ <https://docs.pulpproject.org/>`_.
```
asmacdo: Should these be `http://`?

bmbouter: I'm trying to do everything with `https`. docs.pulpproject.org has SSL configured with a cert to match.

asmacdo: Not sure what the deal is with the ssl, but that is not a blocker for this PR.

PR #2622, opened 2016-06-29
Fix importer validation logic.

ipanova:
Main issue was in the validation of importer config in the update call. In validate_importer_config() was passed and checked just the current specified values of the update call and values of importer config stored in db where not considered.

https://pulp.plan.io/issues/2039

### DIFF
--- a/server/pulp/server/controllers/importer.py
+++ b/server/pulp/server/controllers/importer.py
@@ -69,7 +69,10 @@ def set_importer(repo_id, importer_type_id, repo_plugin_config):
     :raises exceptions.InvalidValue: if the values passed to create the importer are invalid
     """
     repo = model.Repository.objects.get_repo_or_missing_resource(repo_id)
-    validate_importer_config(repo.repo_id, importer_type_id, repo_plugin_config)
+
+    if not plugin_api.is_valid_importer(importer_type_id):
+        raise exceptions.PulpCodedValidationException(error_code=error_codes.PLP1008,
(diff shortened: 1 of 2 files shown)


bmbouter:
@ipanova consider putting a `#closes 2039` into the commit message so that this gets moved to `MODIFIED` automatically. Also consider putting more details onto the commit message itself.

server/pulp/server/controllers/importer.py
```diff
@@ -69,14 +69,18 @@ def set_importer(repo_id, importer_type_id, repo_plugin_config):
     :raises exceptions.InvalidValue: if the values passed to create the importer are invalid
     """
     repo = model.Repository.objects.get_repo_or_missing_resource(repo_id)
-    validate_importer_config(repo.repo_id, importer_type_id, repo_plugin_config)
+
+    if not plugin_api.is_valid_importer(importer_type_id):
```
bmbouter: You didn't write the `plugin_api.is_valid_importer()`, but I'll write my suggestion here anyway. Overall rather than having a caller check the return code of something like `is_valid_importer()` consider having `is_valid_importer()` raise the `PulpCodedValidationException` itself, if it's not valid. This would be a much more Pythonic way of doing things and allow the Exception handling to work naturally. You didn't write this, and maybe now isn't the time to change that.

bmbouter:
This looks like the right improvement. I left some comments, but you can handle them however you like. I'm adding the LGTM now. Thanks @ipanova for the bugfix.

PR #2624, opened 2016-07-06
Adds rsync distributor publish steps to platform

dkliban:
This adds shared rsync distributor code. The classes provided need to be extended in each plugin that implements the rsync distributor.

### DIFF
--- a/server/pulp/plugins/rsync/configuration.py
+++ b/server/pulp/plugins/rsync/configuration.py
@@ -1,0 +1,261 @@
+from gettext import gettext as _
+import logging
+import os
+
+_LOG = logging.getLogger(__name__)
+
+
+class OneOfValidation(object):
+    """
+    Validates that the the value is one of possible values
+    """
+    def __init__(self, values):
+        """
+        :param values: list of valid values
+        """
+        self.values = values
+
+    def __call__(self, value, config):
+        """
+        :param value: value to validate
+        :type value: any
+        :param config: distributor config
+        :type config: PulpCallConfig object
+
+        :return: tuple indicating whether config value validates and error message or None
+        :rtype: (bool, str) or (bool, None)
+        """
+        if value in self.values:
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        """
+        :param value: value that did not pass validation
+        :type value: any
+
+        :return: error message
+        :rtype: str
+        """
+        params = {'value': value, 'allowed_values': ", ".join(self.values)}
+        return _("%(value)s is not in allowed values: %(allowed_values)s") % params
+
+
+class NonEmptyValidation(object):
+    """
+    Validates that the value is not None
+    """
+    def __call__(self, value, config):
+        """
+        :param value: value to validate
+        :type value: any
+        :param config: distributor config
+        :type config: PulpCallConfig object
+
+        :return: tuple indicating whether config value validates and error message or None
+        :rtype: (bool, str) or (bool, None)
+        """
+        if value:
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        """
+        :param value: value that did not pass validation
+        :type value: any
+
+        :return: error message
+        :rtype: str
+        """
+        return _("attribute cannot be empty")
+
+
+class TypeValidation(object):
+    """
+    Validates that the value is one of allowed types
+    """
+    def __init__(self, allowed_types):
+        """
+        :param allowed_types: list of valid types
+        :type allowed_types: list of types
+        """
+        self.allowed_types = allowed_types
+
+    def __call__(self, value, config):
+        """
+        :param value: value to validate
+        :type value: any
+        :param config: distributor config
+        :type config: PulpCallConfig object
+
+        :return: tuple indicating whether config value validates and error message or None
+        :rtype: (bool, str) or (bool, None)
+        """
+        if all(issubclass(type(value), _type) for _type in self.allowed_types):
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        """
+        :param value: value that did not pass validation
+        :type value: any
+
+        :return: error message
+        :rtype: str
+        """
+        params = {'type': type(value), 'allowed_types': ", ".join(self.allowed_types)}
+        return _("%(type)s type is not one of allowed types: %(allowed_types)s") % params
+
+
+class RequireOptionalIf(object):
+    """
+    Validates that if a particular config is present, other configs that are needed for it are also
+    present.
+    """
+    def __init__(self, required_for_attr, condition):
+        """
+        :param required_for_attr: a list of required config keys or dictionary with keys in the
+                                  config and values are lists of configs that need to be present
+                                  in the value of the key in config.
+        :type required_for_attr: list or dict
+        :param condition: callable that takes a config name and returns a boolean
+        :type condition: callable
+        """
+        self.required_for_attr = required_for_attr
+        self.condition = condition
+
+    def __call__(self, value, config):
+        """
+        :param value: config name that is being validated
+        :type value: str
+        :param config: configuration instance to validate
+        :type  config: pulp.plugins.config.PluginCallConfiguration
+
+        :return: tuple indicating whether config value validates and error message or None
+        :rtype: (bool, str) or (bool, None)
+        """
+
+        subconfig = config
+        path = []
+        if not self.condition(value):
+            return (True, None)
+
+        if isinstance(self.required_for_attr, list):
+            fifo = [(x, subconfig, path) for x in self.required_for_attr]
+        elif isinstance(self.required_for_attr, dict):
+            fifo = [(val, subconfig.get(key), [key])
+                    for key, val in self.required_for_attr.iteritems()]
+        while fifo:
+            (required_for_attr, subconfig, path) = fifo.pop(0)
+            if isinstance(required_for_attr, list):
+                for x in required_for_attr:
+                    fifo.insert(0, (x, subconfig, path))
+            elif isinstance(required_for_attr, dict):
+                for key, val in required_for_attr.iteritems():
+                    fifo.insert(0, (val, subconfig.get(key)), path + [key])
+            elif isinstance(required_for_attr, basestring):
+                if required_for_attr not in subconfig:
+                    return (False, self._err(path + [required_for_attr]))
+        return (True, None)
+
+    def _err(self, path):
+        """
+        :param path: list of value that were missing
+        :type path: list
+
+        :return: error message
+        :rtype: str
+        """
+        return _("%(attribute)s attribute is required") % {'attribute': "::".join(path)}
+
+
+class RelativePathValidation(object):
+    """
+    Validates that a path does not start with a forward slash.
+    """
+    def __call__(self, value, *args):
+        """
+        :param value: path to validate
+        :param args: any extra args that get passed in but are not needed for validation
+
+        :return: tuple indicating whether config value validates and error message or None
+        :rtype: (bool, str) or (bool, None)
+        """
+        if os.path.isabs(value):
+            return (False, self._err(value))
+        else:
+            return (True, None)
+
+    def _err(self, value):
+        """
+        :param value: value that did not pass validation
+        :type value: any
+
+        :return: error message
+        :rtype: str
+        """
+        return _("attribute cannot start with a /")
+
+REMOTE_MANDATORY_KEYS = {
+    "auth_type": [OneOfValidation(["publickey", "password"]),
+                  RequireOptionalIf({"remote": ["ssh_password", "ssh_user"]},
+                                    lambda x: x == "password"),
+                  RequireOptionalIf({"remote": ["ssh_identity_file", "ssh_user"]},
+                                    lambda x: x == "publickey")],
+    "host": [TypeValidation([basestring]), NonEmptyValidation()],
+    "root": [TypeValidation([basestring]), NonEmptyValidation()]
+}
+
+REMOTE_OPTIONAL_KEYS = {
+    "remote_units_path": [TypeValidation([basestring]), RelativePathValidation()],
+    "ssh_identity_file": [TypeValidation([basestring]), NonEmptyValidation()],
+    "ssh_user": [TypeValidation([basestring]), NonEmptyValidation()],
+    "ssh_password": [TypeValidation([basestring]), NonEmptyValidation()]
+}
+
+
+def validate_config(repo, config, config_conduit):
+    """
+    Validate the prospective configuration instance for the given repository.
+
+    :param repo: repository to validate the config for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance to validate
+    :type  config: pulp.plugins.config.PluginCallConfiguration
+    :param config_conduit: conduit providing access to relevant Pulp functionality
+    :type  config_conduit: pulp.plugins.conduits.repo_config.RepoConfigConduit
+
+    :return: tuple of (bool, str) stating that the configuration is valid or not and why
+    :rtype:  tuple of (bool, str) or (bool, None)
+    """
+    _config = config.flatten()  # now config is {}
+    err_list = []
+    if "remote" not in _config or ("remote" in _config and not isinstance(_config["remote"], dict)):
+        err_list.append("'remote' dict missing in distributor's configuration")
+    else:
+        missing_attr_tmpl = _("'%(attribute)s' is missing in 'remote' section of distributor's '"
+                              "configuration")
+        for attr, validations in REMOTE_MANDATORY_KEYS.iteritems():
+            if attr not in _config["remote"]:
+                err_list.append(missing_attr_tmpl % {'attribute': attr})
+                continue
+            for validation in validations:
+                succeed, _err = validation(_config["remote"][attr], _config)
+                if not succeed:
+                    err_list.append(_("%(attribute)s : %(error)s") % {'attribute': attr,
+                                                                      'error': _err})
+        for attr, validations in REMOTE_OPTIONAL_KEYS.iteritems():
+            if attr not in _config["remote"]:
+                continue
+            for validation in validations:
+                succeed, _err = validation(_config["remote"][attr], _config)
+                if not succeed:
+                    err_list.append(_("%(attribute)s : %(error)s") % {'attribute': attr,
+                                                                      'error': _err})
+    if err_list:
+        return (False, "\n".join(err_list))
+    else:
+        return (True, None)
--- a/server/pulp/plugins/rsync/publish.py
+++ b/server/pulp/plugins/rsync/publish.py
@@ -1,0 +1,463 @@
+from gettext import gettext as _
+import logging
+import os
+import time
+import uuid
+import shutil
+
+import kobo.shortcuts
+import mongoengine
+
+from pulp.common import dateutils
+from pulp.plugins.util.publish_step import PublishStep
+from pulp.server.config import config as pulp_config
+from pulp.server.exceptions import PulpCodedException
+from pulp.server.db.model.repository import RepoPublishResult
+from pulp.server.db.model import Distributor
+
+
+ASSOCIATED_UNIT_DATE_KEYWORD = "created"
+START_DATE_KEYWORD = 'start_date'
+END_DATE_KEYWORD = 'end_date'
+
+_logger = logging.getLogger(__name__)
+
+
+class RSyncPublishStep(PublishStep):
+
+    _CMD = "rsync"
+    # recursive, symlinks, timestamps, keep dir links, omit dir times, compress, itemize
+    _AZQ = "-rtKOzi"
+
+    def __init__(self, step_type, file_list, src_directory, dest_directory, config=None,
+                 exclude=[], delete=False, links=False):
+        """
+        Copies files using rsync to a remote server and filesystem location defined in the
+        distributor config.
+
+        :param step_type: Name of step
+        :type step_type: str
+        :param file_list: list of paths relative to src_directory
+        :type file_list: list
+        :param src_directory: absolute path to directory which contains all items in file_list
+        :type src_directory: str
+        :param dest_directory: path to directory relative to remote root where files should be
+               rsynced
+        :type dest_directory: str
+        :param config: distributor configuration
+        :type config: pulp.plugins.config.PluginCallConfiguration
+        :param exclude: list of files/directories to skip
+        :type exclude: list of str
+        :param delete: determines if --delete is passed to rsync
+        :type delete: bool
+        :param links: if true, --links is passed to rsync
+        :type links: bool
+        """
+        super(PublishStep, self).__init__(step_type, config=config)
+        self.description = _('Rsync files to remote destination')
+        self.file_list = file_list
+        self.exclude = exclude
+        self.delete = delete
+        self.src_directory = src_directory
+        self.dest_directory = dest_directory
+        self.links = links
+
+    def remote_mkdir(self, path):
+        """
+        Creates path on remote server. The path is rooted in distributor's remote_root directory.
+
+        :param path: path to create on remote server
+        :type path: str
+
+        :return: (return code from rsync, string made up of stdout and stderr from running rsync
+                  to create remote directory)
+        :rtype: tuple
+        """
+        tmpdir = os.path.join(self.get_working_dir(), '.tmp')
+        os.makedirs(os.path.join(tmpdir, path.lstrip("/")))
+        args = ['rsync', '-avr', '-f+ */']
+        args.extend(self.make_authentication())
+        args.append("%s/" % tmpdir)
+        args.append(self.make_destination(path).replace(str(path), ""))
+        _logger.debug("remote mkdir: %s" % args)
+        is_ok, output = self.call(args, include_args_in_output=True)
+        _logger.debug("remote mkdir out: %s" % output)
+        shutil.rmtree(tmpdir)
+        return is_ok, output
+
+    def make_ssh_cmd(self, args=None):
+        """
+        Returns a list of arguments needed to form an ssh command for connecting to remote server.
+
+        :param args:list of extra args to append to the standard ssh command
+        :type args: list
+
+        :return: list of arguments for ssh portion of command
+        :rtype: list
+        """
+        user = self.get_config().flatten()["remote"]['ssh_user']
+        auth_type = self.get_config().flatten()["remote"]['auth_type']
+
+        # -e 'ssh -l ssh_user -i ssh_identity_file'
+        # use shared ssh connection for other threads
+        cmd = ['ssh', '-l', user]
+        if auth_type == "publickey":
+            key = self.get_config().flatten()["remote"]['ssh_identity_file']
+            cmd += ['-i', key]
+        cmd += ['-o', 'StrictHostKeyChecking no',
+                '-o', 'UserKnownHostsFile /dev/null',
+                '-S', '/tmp/rsync_distributor-%r@%h:%p',
+                '-o', 'ControlMaster auto',
+                '-o', 'ControlPersist 10']
+        if self.get_config().flatten()["remote"]['auth_type'] == 'password':
+            password_file = os.path.join(self.get_working_dir(), str(uuid.uuid4()))
+            open(password_file, 'w').write(self.get_config().flatten()['remote']['ssh_password'])
+            cmd = ['sshpass', '-f', password_file] + cmd
+        if args:
+            cmd += args
+        return cmd
+
+    def make_authentication(self):
+        """
+        Returns a list of strings representing args for command for authenticating against a remote
+        server.
+
+        :return: list of arguments for auth. e.g., ['-e',  'ssh, '-l', 'ssh_user', '-i',
+                                                    '/ssh_identity_file', 'hostname']
+        :rtype: list
+        """
+        if self.get_config().flatten()["remote"]['auth_type'] == 'local':
+            return []
+        ssh_parts = []
+        for arg in self.make_ssh_cmd():
+            if " " in arg:
+                ssh_parts.append('"%s"' % arg)
+            else:
+                ssh_parts.append(arg)
+
+        return ['-e', " ".join(ssh_parts)]
+
+    def make_full_path(self, relative_path):
+        """
+        Returns absolute path of a relative path on the remote server.
+
+        :param relative_path: relative path
+        :type relative_path: str
+
+        :return: absolute path to the relative path
+        :rtype: str
+        """
+        return os.path.join(self.get_config().flatten()['remote']['root'], relative_path)
+
+    def make_destination(self, relative_path):
+        """
+        Parse from self.config information to make up a hostname and remote path used for rsync
+        command.
+
+        :param relative_path: path relative to the root configured in distributor
+        :type relative_path: str
+
+        :return: str of the combination of user, host, and dir.
+        :rtype:  str
+        """
+        if self.get_config().flatten()["remote"]['auth_type'] == 'local':
+            return self.get_config().flatten()["remote"]['root']
+
+        user = self.get_config().flatten()["remote"]['ssh_user']
+        host = self.get_config().flatten()["remote"]['host']
+        remote_root = self.get_config().flatten()["remote"]['root']
+        return '%s@%s:%s' % (user, host, os.path.join(remote_root, relative_path))
+
+    def call(self, args, include_args_in_output=True):
+        """
+        A wrapper around kobo.shortcuts.run. If ssh_exchange_identification or
+        max-concurrent-connections exceptions are thrown by ssh, up to 10 retries follows.
+
+        :param args: list of args for rsync
+        :type args: list
+        :param include_args_in_output: include the rsync arguments in output or not
+        :type include_args_in_output: bool
+
+        :return: (boolean indicating success or failure, output from rsync command)
+        :rtype: tuple of boolean and string
+        """
+        for t in xrange(10):
+            rv, out = kobo.shortcuts.run(cmd=args, can_fail=True)
+            possible_known_exceptions = \
+                ("ssh_exchange_identification:" in out) or ("max-concurrent-connections=25" in out)
+            if not (rv and possible_known_exceptions):
+                break
+            _logger.info(_("Connections limit reached, trying once again in thirty seconds."))
+            time.sleep(30)
+        if include_args_in_output:
+            message = "%s\n%s" % (args, out)
+        else:
+            message = out
+        return (rv == 0, message)
+
+    def make_rsync_args(self, files_from, source_prefix, dest_prefix, exclude=None):
+        """
+        Creates a list of arguments for the rsync command
+
+        :param files_from: absolute path to a file with list of file paths to rsync
+        :type files_from: str
+        :param source_prefix: path to directory that contains all paths in files_from
+        :type source_prefix: str
+        :param dest_prefix: path to directory on remote server where files should be rsynced
+        :type dest_prefix: str
+        :param exclude: list of file/directory paths to exclude
+        :type exclude: list
+
+        :return: list of arguments for rsync command
+        :rtype: list of strings
+        """
+        args = [self._CMD, self._AZQ]
+        if not self.delete:
+            args.extend(["--files-from", files_from, "--relative"])
+
+        if exclude:
+            for x in exclude:
+                args.extend(["--exclude", x])
+        args.extend(self.make_authentication())
+        if self.delete:
+            args.append("--delete")
+        if self.links:
+            args.append("--links")
+        else:
+            args.append("--copy-links")
+        args.append(source_prefix)
+        args.append(self.make_destination(dest_prefix))
+        return args
+
+    def rsync(self):
+        """
+        This method formulates the rsync command based on parameters passed in to the __init__ and
+        then executes it.
+
+        :return: (boolean indicating success or failure, str made up of stdout and stderr
+                  generated by rsync command)
+        :rtype: tuple
+        """
+        if not self.file_list:
+            return (True, _("Nothing to sync"))
+        output = ""
+        list_of_files = os.path.join(self.get_working_dir(), str(uuid.uuid4()))
+        open(list_of_files, 'w').write("\n".join(sorted(self.file_list)))
+
+        # copy files here, not symlinks
+        (is_successful, this_output) = self.remote_mkdir(self.dest_directory)
+        if not is_successful:
+            params = {'directory': self.dest_directory, 'output': this_output}
+            _logger.error(_("Cannot create directory %s: %s") % params)
+            return (is_successful, this_output)
+        output += this_output
+        rsync_args = self.make_rsync_args(list_of_files, self.src_directory,
+                                          self.dest_directory, self.exclude)
+        (is_successful, this_output) = self.call(rsync_args)
+        _logger.info(this_output)
+        if not is_successful:
+            _logger.error(this_output)
+            return (is_successful, this_output)
+        output += this_output
+        return (is_successful, output)
+
+    def process_main(self):
+        """
+        This method is the main method executed when the step system executes a step.
+        """
+        (successful, output) = self.rsync()
+        if not successful:
+            raise PulpCodedException(message=output)
+
+
+class Publisher(PublishStep):
+    """
+    RSync publisher class that provides the common code for publishing to remote server. Each
+    plugin needs to extend this class and add necesary logic to the _add_necesary_steps method.
+    """
+    def __init__(self, repo, publish_conduit, config, distributor_type):
+        """
+        :param repo: Pulp managed Yum repository
+        :type  repo: pulp.plugins.model.Repository
+        :param publish_conduit: Conduit providing access to relative Pulp functionality
+        :type  publish_conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+        :param config: Pulp configuration for the distributor
+        :type  config: pulp.plugins.config.PluginCallConfiguration
+        :param distributor_type: The type of the distributor that is being published
+        :type distributor_type: str
+
+        :ivar last_published: last time this distributor published the repo
+        :ivar last_delete: last time a unit was removed from this repository
+        :ivar repo: repository being operated on
+        :ivar predistributor: distributor object that is associated with this distributor. It's
+                              publish history affects the type of publish is performed
+        :ivar symlink_list: list of symlinks to rsync
+        :ivar content_unit_file_list: list of content units to rsync
+        :ivar symlink_src: path to directory containing all symlinks
+        """
+
+        super(Publisher, self).__init__("Repository publish", repo,
+                                        publish_conduit, config,
+                                        distributor_type=distributor_type)
+
+        distributor = Distributor.objects.get_or_404(repo_id=self.repo.id,
+                                                     distributor_id=publish_conduit.distributor_id)
+        self.last_published = distributor["last_publish"]
+        self.last_deleted = repo.last_unit_removed
+        self.repo = repo
+        self.predistributor = self._get_predistributor()
+
+        if self.last_published:
+            string_date = dateutils.format_iso8601_datetime(self.last_published)
+        else:
+            string_date = None
+        if self.predistributor:
+            search_params = {'repo_id': repo.id,
+                             'distributor_id': self.predistributor["id"],
+                             'started': {"$gte": string_date}}
+            self.predist_history = RepoPublishResult.get_collection().find(search_params)
+        else:
+            self.predist_history = []
+
+        self.remote_path = self.get_remote_repo_path()
+
+        if self.is_fastforward():
+            start_date = self.last_published
+            end_date = None
+            if self.predistributor:
+                end_date = self.predistributor["last_publish"]
+            date_filter = self.create_date_range_filter(start_date=start_date, end_date=end_date)
+        else:
+            date_filter = None
+
+        self.symlink_list = []
+        self.content_unit_file_list = []
+        self.symlink_src = os.path.join(self.get_working_dir(), '.relative/')
+
+        self._add_necesary_steps(date_filter=date_filter, config=config)
+
+    def is_fastforward(self):
+        """
+        This method checks whether this publish should be a fastforward publish.
+
+        :return: Whether or not this publish should be in fast forward mode
+        :rtype: bool
+        """
+        skip_fast_forward = False
+        for entry in self.predist_history:
+            predistributor_skip_ff = entry.get("distributor_config", {}).get("skip_fast_forward",
+                                                                             False)
+            skip_fast_forward |= predistributor_skip_ff
+            if entry.get("result", "error") == "error":
+                skip_fast_forward = True
+        if self.last_published:
+            last_published = self.last_published.replace(tzinfo=None)
+        else:
+            last_published = None
+        if self.last_deleted:
+            last_deleted = self.last_deleted.replace(tzinfo=None)
+        else:
+            last_published = None
+
+        config_skip_ff = self.get_config().get("skip_fast_forward", False)
+        skip_fast_forward = skip_fast_forward | config_skip_ff
+        delete = self.get_config().get("delete", False)
+
+        return last_published and ((last_deleted and last_published > last_deleted) or
+                                   not last_deleted) and not skip_fast_forward and\
+            not delete
+
+    def create_date_range_filter(self, start_date=None, end_date=None):
+        """
+        Create a date filter based on start and end issue dates specified in the repo config.
+
+        :param start_date: start time for the filter
+        :type  start_date: datetime.datetime
+        :param end_date: end time for the filter
+        :type  end_date: datetime.datetime
+
+        :return: Q object with start and/or end dates, or None if start and end dates are not
+                 provided
+        :rtype:  mongoengine.Q or types.NoneType
+        """
+        if start_date:
+            start_date = dateutils.format_iso8601_datetime(start_date)
+        if end_date:
+            end_date = dateutils.format_iso8601_datetime(end_date)
+
+        if start_date and end_date:
+            return mongoengine.Q(created__gte=start_date, created__lte=end_date)
+        elif start_date:
+            return mongoengine.Q(created__gte=start_date)
+        elif end_date:
+            return mongoengine.Q(created__lte=end_date)
+
+    def get_remote_repo_path(self):
+        """
+        Returns the full path to the published repository on remote server.
+
+        :return: relative url for the repo on remote server
+        :rtype: str
+        """
+        if "relative_url" in self.repo.notes and self.repo.notes["relative_url"]:
+            return self.repo.notes["relative_url"]
+        else:
+            return self.repo.id
+
+    def get_master_directory(self):
+        """
+        Returns path to master directory of the predistributor.
+
+        :return: path to 'master' publish directory
+        :rtype: str
+        """
+        repo_relative_path = self.predistributor['config']['relative_url']
+        return os.path.realpath(os.path.join(self._get_root_publish_dir(), repo_relative_path))
+
+    def get_units_directory_dest_path(self):
+        """
+        Returns the path on the remote server where content units should be stored.
+
+        :return: relative path to the directory where content units remote server
+        :rtype: str
+        """
+
+        if self.get_config().get("remote_units_path"):
+            origin_dest_prefix = self.get_config().get("remote_units_path")
+        else:
+            origin_dest_prefix = os.path.join("content", "units")
+        return origin_dest_prefix
+
+    def get_units_src_path(self):
+        return os.path.join(pulp_config.get('server', 'storage_dir'), 'content', 'units')
+
+    def _add_necesary_steps(self, date_filter=None):
+        """
+        This method needs to be implemented in each plugin. This method should include calls to
+        self.add_child() with particular rsync steps that are needed to perform the full publish
+        to the remote server.
+
+        :param date_filter:  Q object with start and/or end dates, or None if start and end dates
+                             are not provided
+        :type date_filter:  mongoengine.Q or types.NoneType
+        """
+        raise NotImplementedError()
+
+    def _get_root_publish_dir(self):
+        """
+        Returns the publish directory path for the predistributor
+
+        :return: path to the publish directory of the predistirbutor
+        :rtype: str
+        """
+        raise NotImplementedError()
+
+    def _get_predistributor(self):
+        """
+        Returns distributor which repo has to be published in before
+        publish in rsync distributor, content generator.
+
+        :return: predistributor that was configured in rsyn distributor's config
+        :rtype: Distributor
+        """
+        raise NotImplementedError()
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -23,6 +23,7 @@ from pulp.server.controllers import units as units_controller
 from nectar import listener
 from nectar.downloaders.local import LocalFileDownloader
 from nectar.downloaders.threaded import HTTPThreadedDownloader
+from pulp.server.config import config as pulp_config
 import pulp.server.managers.factory as manager_factory
 from pulp.server.managers.repo import _common as common_utils
 from pulp.server.util import copytree
@@ -574,7 +575,7 @@ class UnitModelPluginStep(PluginStep):
     The QuerySetNoCache objects themselves do not cache results, as the name implies.
     """
     def __init__(self, step_type, model_classes, repo_content_unit_q=None, repo=None, conduit=None,
-                 config=None, working_dir=None, plugin_type=None, **kwargs):
+                 config=None, working_dir=None, plugin_type=None, unit_fields=None, **kwargs):
         """
         :param step_type: The id of the step this processes
         :type  step_type: str
@@ -593,12 +594,15 @@ class UnitModelPluginStep(PluginStep):
         :type  working_dir: str
         :param plugin_type: The type of the plugin
         :type  plugin_type: str
+        :param unit_fields: list of unit fields to retrieve from database, if None all are retrieved
+-       :type unit_fields: list of str
         """
         super(UnitModelPluginStep, self).__init__(step_type, repo, conduit, config, working_dir,
                                                   plugin_type, **kwargs)
 
         self.model_classes = model_classes
         self._repo_content_unit_q = repo_content_unit_q
+        self.unit_fields = unit_fields
 
         # the corresponding publicly-accessible values get cached here
         self._unit_querysets = None
@@ -632,6 +636,9 @@ class UnitModelPluginStep(PluginStep):
                                                                    model_class,
                                                                    self._repo_content_unit_q)
                 self._unit_querysets.extend(queries)
+
+        if self.unit_fields:
+            return (qs.only(*self.unit_fields) for qs in self._unit_querysets)
         return self._unit_querysets
 
     def get_total(self):
@@ -1231,3 +1238,110 @@ class GetLocalUnitsStep(SaveUnitsStep):
             for unit in units_group:
                 if hash(unit) not in units_we_already_had:
                     self.units_to_download.append(unit)
+
+
+class RSyncFastForwardUnitPublishStep(UnitModelPluginStep):
+
+    def __init__(self, step_type, model_classes, repo_content_unit_q=None, repo=None,
+                 config=None, remote_repo_path=None, published_unit_path=None, unit_fields=None):
+        """
+        Set the default parent, step_type and units_type for the the publish step.
+
+        :param step_type: The id of the step this processes
+        :type step_type: str
+        :param model_classes:   list of ContentUnit subclasses that should be queried
+        :type  model_classes:   list
+        :param repo_content_unit_q: optional Q object that will be applied to the queries performed
+                                    against RepoContentUnit model
+        :type  repo_content_unit_q: mongoengine.Q
+        :param repo: The repo being worked on
+        :type  repo: pulp.plugins.model.Repository
+        :param config: The publish configuration
+        :type  config: PluginCallConfiguration
+        :param remote_repo_path: relative path on remote server where published repo should live
+        :type remote_repo_path: str
+        :param published_unit_path: relative path to the content unit from root of published repo
+        :type published_unit_path: str
+        :param unit_fields: list of unit fields to retrieve from database
+-       :type unit_fields: list of str
+
+        """
+        self.description = _('Generating relative symlinks')
+        self.remote_repo_path = remote_repo_path
+        self.published_unit_path = published_unit_path
+
+        super(RSyncFastForwardUnitPublishStep,
+              self).__init__(step_type, model_classes, repo=repo, config=config,
+                             repo_content_unit_q=repo_content_unit_q, unit_fields=unit_fields)
+
+    def process_main(self, item=None):
+        """
+        Creates symlink for a unit and appends the symlink to a list used to perform rsync later.
+        This also generates the list of actual content units that need to be rsynced later.
+
+        :param unit: Content type unit
+        :type unit: ContentUnit
+        """
+        storage_dir = os.path.join(pulp_config.get('server', 'storage_dir'), 'content', 'units')
+        relative_content_unit_path = os.path.relpath(item.storage_path, storage_dir)
+        self.parent.content_unit_file_list.append(relative_content_unit_path)
+        symlink = self.make_link_unit(item, self.get_working_dir(), self.remote_repo_path,
+                                      self.get_config().get("remote")["root"],
+                                      self.published_unit_path)
+        self.parent.symlink_list.append(symlink)
+
+    def make_link_unit(self, unit, working_dir, remote_repo_path, remote_root,
+                       published_unit_path):
+        """
+        This method creates symlink in working directory, pointing to remote
+        content directory, so they can be sync to remote server
+
+        :param unit: Pulp content unit
+        :type unit: Unit
+        :param working_dir: Working directory
+        :type working_dir: str
+        :param remote_repo_path: relative repo destination path on remote server
+        :type remote_repo_path: str
+        :param remote_root: remote destination root directory
+        :type remote_root: str
+        :param published_unit_path: directory name inside repo that the symlink should be in
+        :type published_unit_path: list of str
+
+        """
+        filename = unit.get_symlink_name()
+        extra_src_path = ['.relative'] + published_unit_path
+        if not os.path.exists(os.path.join(working_dir, *extra_src_path)):
+            os.makedirs(os.path.join(working_dir, *extra_src_path))
+
+        origin_path = self.get_origin_rel_path(unit)
+
+        dest = os.path.normpath(os.path.join(working_dir,
+                                             *(extra_src_path + [filename])))
+        full_remote_path = os.path.normpath(os.path.join(remote_root, origin_path))
+        abs_remote_path = os.path.normpath(os.path.join(remote_root,
+                                                        remote_repo_path.lstrip("/"),
+                                                        *published_unit_path))
+
+        link_source = os.path.relpath(full_remote_path, abs_remote_path)
+
+        _logger.debug("LN %s -> %s " % (link_source, dest))
+        if os.path.islink(dest):
+            os.remove(dest)
+        os.symlink(link_source, dest)
+        return os.path.join(*(published_unit_path + [filename]))
+
+    def get_origin_rel_path(self, unit):
+        """
+        Returns path relative to the remote root for the content unit
+
+        :param unit: Pulp content unit
+        :type unit: Unit
+        """
+
+        storage_dir = pulp_config.get('server', 'storage_dir')
+        content_base = os.path.join(storage_dir, "content", "units", unit.type_id)
+        rel_unit_path = unit.storage_path.replace(content_base, '').lstrip("/")
+        remote_content_base = self.parent.get_units_directory_dest_path()
+        remote_content_type_dir = unit.type_id
+
+        return os.path.join(remote_content_base, remote_content_type_dir, rel_unit_path)
--- a/pulp.spec
+++ b/pulp.spec
@@ -379,6 +379,8 @@ Requires: mod_wsgi >= 3.4-1.pulp
 Requires: mod_xsendfile >= 0.12
 Requires: m2crypto
 Requires: genisoimage
+Requires: kobo
+Requires: sshpass
 # RHEL6 ONLY
 %if 0%{?rhel} == 6
 Requires: nss >= 3.12.9
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -914,6 +914,14 @@ class FileContentUnit(ContentUnit):
         orphan_manger = factory.content_orphan_manager()
         orphan_manger.delete_orphan_content_units_by_type(self._content_type_id, self.id)
 
+    def get_symlink_name(self):
+        """
+        Provides the name that should be used when creating a symlink.
+        :return: file name as it appears in a published repository
+        :rtype: str
+        """
+        raise NotImplemented()
+
 
 class SharedContentUnit(ContentUnit):
     """


bmbouter:
Does `kobo` work with the oldest version EL6. That version is 4.2.

https://apps.fedoraproject.org/packages/kobo

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,139 @@
+import logging
+
+_LOG = logging.getLogger(__name__)
+
+
+class OneOfValidation(object):
```
bmbouter: All of these objects need docstrings. Maybe not all of the methods, but at least the objects themselves.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+import inspect
+
+from pulp.common import dateutils
+from pulp.plugins.util import misc
+from pulp.plugins.util.publish_step import PublishStep
+from pulp.server.exceptions import PulpCodedException
+from pulp.server.db.model.repository import RepoPublishResult
+from pulp.server.db.model import Distributor
+
+import kobo.shortcuts
```
bmbouter: Usually 3rd part imports go in the second group.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        self.src_prefix = src_prefix
+        self.dest_prefix = dest_prefix
+        self.links = links
+
+    def remote_mkdir(self, path):
+        """
+        Creates path on remote server. The path is rooted in distributor's remote_root directory.
+        :param path: path to create on remote server
+        :type path: str
+        :return:
```
bmbouter: This needs a value for `:return:`.
Also please add a `:rtype:`

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        _logger.info("remote mkdir: %s" % args)
+        is_ok, output = self.call(args, include_args_in_output=True)
+        _logger.info("remote mkdir out: %s" % output)
+        shutil.rmtree(tmpdir)
+        return is_ok, output
+
+    def make_ssh_cmd(self, args=None):
+        """
+        Returns a list of arguments needed to form an ssh command for connecting to remote server.
+        :param args:
```
bmbouter: Consider separating the params section from the docblock description with a newline.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        :type args: list of extra args to append to the standard ssh command
+        :return:
+        """
+        user = self.get_config().flatten()["remote"]['ssh_login']
+        key = self.get_config().flatten()["remote"]['key_path']
+        auth_type = self.get_config().flatten()["remote"]['auth_type']
+
+        # -e 'ssh -l l_name -i key_path'
+        # use shared ssh connection for other threads
```
bmbouter: Yes a shared ssh connection. :+1:

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        cmd = ['ssh',
```
bmbouter: For readability this could become one line

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        :param relative_path: relative path
+        :type relative_path: str
+        """
+        return os.path.join(self.get_config().flatten()["remote"]['root'], relative_path)
+
+    def make_destination(self, relative_path):
+        """
+        Parse from self.config information to make up a hostname and remote path used for rsync
+        command.
+        :return: str of the combination of user, host, and dir.
```
bmbouter: newline between description

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+                ssh_parts.append('"%s"' % arg)
+            else:
+                ssh_parts.append(arg)
+
+        return ['-e', " ".join(ssh_parts)]
+
+    def make_full_path(self, relative_path):
+        """
+        Returns full path of a relative path on the remote server.
+        :param relative_path: relative path
```
bmbouter: newline between description

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        self.extra_path = self._get_extra_path()
+        self.original_lists = {}
+        self.link_lists = {}
+
+        self._add_necesary_steps(date_filter=date_filter, config=config)
+
+    def is_fastforward(self):
+        """
+        This method checks whether this publish should be a fastforward publish.
+        :return:
```
bmbouter: This needs a return info.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+            origin_dest_prefix = os.path.join("content", "origin", "units")
+        else:
+            origin_dest_prefix = os.path.join("content", "origin", "units",
+                                              self.get_config().get("content_types")[0])
+        return origin_dest_prefix
+
+    def get_original_list(self):
+        """
+        Returns a list of files that need to be rsynced to the content directory
+        :return:
```
bmbouter: This needs a return info.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        original_list = reduce(lambda x, y: x + y, self.original_lists.values())[:]
+        original_rel_prefix = misc.common_path(original_list)
+        for p in range(len(original_list)):
+            original_list[p] = os.path.relpath(original_list[p], original_rel_prefix)
+        return original_list
+
+    def get_original_prefix(self):
+        """
+        Returns the path to the directory in which Pulp stores the content units
+        :return:
```
bmbouter: This needs a return info.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        link_rel_prefix = misc.common_path(self.link_lists[master])
+        link_list = self.link_lists[master][:]
+        for p in range(len(link_list)):
+            link_list[p] = os.path.relpath(link_list[p], link_rel_prefix)
+        return link_list
+
+    def get_link_prefix(self, master_dir):
+        """
+        Returns relative path inside the working directory to where the symlinks are stored
+        :param master_dir:
```
bmbouter: This needs a description and a `:type:` added.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        raise NotImplementedError()
+
+    def _get_root_publish_dir(self):
+        """
+        Returns the publish directory path for the predistributor
+        """
+        raise NotImplementedError()
+
+    def _get_predistributor(self):
+        """ Returns distributor which repo has to be published in before
```
bmbouter: This docstring could be better formatted.

mhrivnak:
When using password auth, every rsync command that got logged included my password. It would be much better to use the --password-file option of rsync. By having the password as an argument to the process, it is also visible in the process table to any user, and possibly other places.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def call(self, args, include_args_in_output=True):
+        """
+        A wrapper around kobo.shortcuts.run. If ssh_exchange_identification or
+        max-concurrent-connections exceptions are thrown by ssh, up to 10 retries follows.
+        """
+        for t in xrange(10):
+            rv, out = kobo.shortcuts.run(cmd=args, can_fail=True)
+            possible_known_exceptions = \
+                ("ssh_exchange_identification:" or "max-concurrent-connections=25") in out
```
mhrivnak: I don't think this works how you're hoping. I expect the expression in parenthesis will be evaluated to just the first string, and the second string will not be searched for in ``out``. To illustrate the problem:

```
>>> (4 or 6) in range(5)
True
>>> (6 or 4) in range(5)
False
```

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+
+        user = self.get_config().flatten()["remote"]['ssh_login']
+        host = self.get_config().flatten()["remote"]['host']
+        remote_root = self.get_config().flatten()["remote"]['root']
+        return '%s@%s:%s' % (user, host, os.path.join(remote_root, relative_path))
+
+    def call(self, args, include_args_in_output=True):
+        """
+        A wrapper around kobo.shortcuts.run. If ssh_exchange_identification or
+        max-concurrent-connections exceptions are thrown by ssh, up to 10 retries follows.
```
mhrivnak: need return value defined

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        args.append(source_prefix)
+        args.append(self.make_destination(dest_prefix))
+        return args
+
+    def rsync(self):
+        """
+        This method formulates the rsync command based on parameters passed in to the __init__ and
+        then executes it.
+
+        :return: (return code, stdout + stderr)
```
mhrivnak: It appears that ``True`` is a valid return code, looking below. What else is a valid return code? Is it always a bool?

Since other parts of this class use the integer return value from processes, this could get confusing, since the meanings are inverted. In those cases, ``0`` is success. But here you seem to return ``True`` for success, which evaluates to ``1``.

This could (and should) all be alleviated by using exceptions instead of writing these functions like they are in C.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+    def get_repo_content_types(self):
+        """
+        This method needs to be implemented in each plugin. It returns a list of content types
+        supported by the plugin.
+        """
+        raise NotImplementedError()
+
+    def get_master_dirs(self):
+        """
+        Returns a list of paths to master directories of the predistributor.
```
mhrivnak: missing return value

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        dirs = []
+        for x in self.get_repo_content_types():
+            repo_relative_path = self.predistributor['config']['relative_url']
+            dirs.append(os.path.realpath(os.path.join(self._get_root_publish_dir(),
+                                                      repo_relative_path)))
+        return dirs
+
+    def get_origin_dest_prefix(self):
+        """
+        Returns the path on the remote server where content units should be stored.
```
mhrivnak: missing return value

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        return origin_dest_prefix
+
+    def get_original_list(self):
+        """
+        Returns a list of files that need to be rsynced to the content directory
+
+        :return: List of files from /var/lib/pulp/content that need to be rsynced
+        :rtype: list of str
+        """
+        original_list = reduce(lambda x, y: x + y, self.original_lists.values())[:]
```
mhrivnak: It's hard to know what's in ``self.original_lists``, but I assume the values are lists? That would be helpful to document somewhere.

Assuming that's true, the ``reduce`` call returns a list, and then this line immediately makes a new copy of it (by adding the ``[:]``) without preserving a reference to the original. Why is that copy being made?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        original_rel_prefix = misc.common_path(original_list)
+        for p in range(len(original_list)):
```
mhrivnak: See the built-in ``enumerate`` function for a way to simplify this.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TypeValidation(object):
+    """
+    Validates that the value is one of allowed types
+    """
+    def __init__(self, allowed_types):
+        self.allowed_types = allowed_types
+
+    def __call__(self, value, config):
+        if [True for _type in self.allowed_types if issubclass(type(value), _type)]:
```
Ichimonji10: Consider using `all`:

```python
all(issubclass(type(value), _type) for _type in allowed_types)
```

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        :return: path to symlinks relative to the working directory
+        :rtype: str
+        """
+        return misc.common_path(self.link_lists[master_dir])
+
+    def _add_necesary_steps(self, date_filter=None, config=None):
+        """
+        This method needs to be implemented in each plugin. This method should include calls to
+        self.add_child() with particular rsync steps that are needed to perform the full publish
+        to the remote server.
```
mhrivnak: need to document parameters and any return value

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        raise NotImplementedError()
+
+    def _get_root_publish_dir(self):
+        """
+        Returns the publish directory path for the predistributor
```
mhrivnak: need to doc the return value

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        raise NotImplementedError()
+
+    def _get_predistributor(self):
+        """
+        Returns distributor which repo has to be published in before
+        publish in rsync distributor, content generator.
```
mhrivnak: need to doc the return value

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,106 @@ def clear_directory(path, skip_list=()):
...
+    :type working_dir: str
+    :param remote_repo_path: relative repo destination path on remote server
+    :type remote_repo_path: str
+    :param remote_root: remote destination root directory
+    :type remote_root: str
+    :param configuration: configuration of distributor
+    :type configuration: dict
+    """
+    out_path = ""
+    if unit.type_id == "rpm":
```
mhrivnak: This type-specific logic does not belong in the platform.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        _logger.info("remote mkdir out: %s" % output)
+        shutil.rmtree(tmpdir)
+        return is_ok, output
+
+    def make_ssh_cmd(self, args=None):
+        """
+        Returns a list of arguments needed to form an ssh command for connecting to remote server.
+
+        :param args:
+        :type args: list of extra args to append to the standard ssh command
```
mhrivnak: looks like this text should be for the ``param`` line, and just ``list`` should be this line.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def make_authentication(self):
+        """
+        Returns a string representation of command for authenticating against a remote server.
+
+        :return: str. e.g., '-e ssh -l ssh_login_user -i /key_path'
+        :rtype: str
+        """
+        if self.get_config().flatten()["remote"]['auth_type'] == 'local':
+            return ()
```
mhrivnak: This is not a ``str``

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        ssh_parts = []
+        for arg in self.make_ssh_cmd():
+            if " " in arg:
+                ssh_parts.append('"%s"' % arg)
+            else:
+                ssh_parts.append(arg)
+
+        return ['-e', " ".join(ssh_parts)]
```
mhrivnak: Neither is this. Perhaps the ``rtype`` is wrong in the doc block.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def make_full_path(self, relative_path):
+        """
+        Returns full path of a relative path on the remote server.
+
+        :param relative_path: relative path
+        :type relative_path: str
```
mhrivnak: missing ``return`` and ``rtype``

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        """
+        return os.path.join(self.get_config().flatten()["remote"]['root'], relative_path)
```
mhrivnak: FWIW there are tons of cases in this PR like this line that seem to arbitrarily mix single and double quotes. I hate to even bring it up given the other stuff that needs to be addressed, but normally I would ask to standardize them on single quotes. Feel free to make your own judgement call on that.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def make_destination(self, relative_path):
+        """
+        Parse from self.config information to make up a hostname and remote path used for rsync
+        command.
+
```
mhrivnak: missing ``param`` docs

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        :return: str of the combination of user, host, and dir.
+        :rtype:  str
+        """
+        if self.get_config().flatten()["remote"]['auth_type'] == 'local':
+            return self.get_config().flatten()["remote"]['root']
+        relative_path = relative_path.lstrip("/")
```
mhrivnak: If it has a leading ``/``, it's not a relative path. Hopefully that would fail validation above, but I'd also be inclined to raise a ``ValueError`` here instead of silently "fixing" it.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,151 @@
...
+                  RequireOptionalIf({"remote": ["ssh_password", "ssh_login"]},
+                                    lambda x: x == "password"),
+                  RequireOptionalIf({"remote": ["key_path", "ssh_login"]},
+                                    lambda x: x == "publickey")],
+    "host": [TypeValidation([basestring]), NonEmptyValidation()],
+    "root": [TypeValidation([basestring]), NonEmptyValidation()]
+}
+
+REMOTE_OPTIONAL_KEYS = {
+    "relative_dir": [TypeValidation([basestring]), NonEmptyValidation()],
```
mhrivnak: Should this be ``remote_rel_prefix`` to match what the code appears to be using?

Whatever it's called, given a comment below, it would be valuable to validate that it does not begin with a forward slash.

server/pulp/plugins/rsync/publish.py
mhrivnak: also need parameters defined

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            message = "%s\n%s" % (args, out)
+        else:
+            message = out
+        return (rv == 0, message)
+
+    def make_rsync_args(self, files_from, source_prefix, dest_prefix, exclude=None):
+        """
+        Creates a list of arguments for the rsync command
+
+        :param files_from: list of files to rsync
```
mhrivnak: File names I assume? Can they be paths? relative or absolute?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        :type files_from: list of strings
+        :param source_prefix: location of files to rsync
```
mhrivnak: I assume since "prefix" is in the name that this location is relative to something else? Same question for ``dest_prefix``.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        :param exclude: list of files/directories to exclude
+        :type exclude: list
+        :return: list of arguments for rsync command
+        :rtype: list
+        """
+        args = [self._CMD, self._AZQ]
+        if not self.delete:
+            args.extend(["--files-from", files_from, "--relative"])
+
+        if exclude:
```
mhrivnak: I don't think this line is necessary.

server/pulp/plugins/rsync/publish.py
Ichimonji10: https://pypi.python.org/pypi/flake8-quotes

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+    plugin needs to extend this class and add necesary logic to the _add_necesary_steps method.
+    """
+    def __init__(self, repo, publish_conduit, config, distributor_type):
+        """
+        :param repo: Pulp managed Yum repository
+        :type  repo: pulp.plugins.model.Repository
+        :param publish_conduit: Conduit providing access to relative Pulp functionality
+        :type  publish_conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+        :param config: Pulp configuration for the distributor
+        :type  config: pulp.plugins.config.PluginCallConfiguration
```
mhrivnak: missing ``distributor_type``

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+        super(Publisher, self).__init__("Repository publish", repo,
+                                        publish_conduit, config,
+                                        distributor_type=distributor_type,
+                                        working_dir="")
+
+        distributor = Distributor.objects.get_or_404(repo_id=self.repo.id,
+                                                     distributor_id=publish_conduit.distributor_id)
+        self.last_published = distributor["last_publish"]
+        self.last_deleted = repo.last_unit_removed
```
mhrivnak: Making this value available on the local object might have a tiny bit of value (I'm trusting that it does), but why rename it? That makes its origin and meaning less clear.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        self.repo_type = repo.notes["_repo-type"]
+        self.predistributor = self._get_predistributor()
+        if self.last_published and not isinstance(self.last_published, basestring):
```
mhrivnak: Is the type of this still uncertain? Or does the mongoengine work that was done make it possible to guarantee what type this will be?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        """
+
+        super(Publisher, self).__init__("Repository publish", repo,
+                                        publish_conduit, config,
+                                        distributor_type=distributor_type,
+                                        working_dir="")
+
+        distributor = Distributor.objects.get_or_404(repo_id=self.repo.id,
+                                                     distributor_id=publish_conduit.distributor_id)
+        self.last_published = distributor["last_publish"]
```
mhrivnak: Since it's a proper object, I think we'd prefer accessing with normal attribute notation. Minor suggestion, but it makes it more clear what we're dealing with.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+    def process_main(self):
+        (successful, output) = self.rsync()
+        if not successful:
+            raise PulpCodedException(message=output)
+
+
+class Publisher(PublishStep):
+    """
+    RSync publisher class that provides the common code for publishing to remote server. Each
+    plugin needs to extend this class and add necesary logic to the _add_necesary_steps method.
```
mhrivnak: There are a ton of instance variables created in ``__init__`` that should be documented somewhere, such as here.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+    def is_fastforward(self):
+        """
+        This method checks whether this publish should be a fastforward publish.
+
+        :return: Whether or not this publish should be in fast forward mode
+        :rtype: bool
+        """
+        skip_fast_forward = False
+        for entry in self.predist_history:
+            str_skip_ff = entry.get("distributor_config", {}).get("skip_fast_forward", False)
```
mhrivnak: Any idea why this could be a ``str`` instead of already being a ``bool``?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            skip_fast_forward |= misc.strbool2bool(str_skip_ff)
+            if entry.get("result", "error") == "error":
+                skip_fast_forward = True
+
+        if self.last_published and isinstance(self.last_published, basestring):
```
mhrivnak: Rather than continuing to check what type this is every time it gets used, let's do it just once in ``__init__``, or not at all if the model takes care of it.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        else:
+            self.predist_history = []
+
+        self.remote_path = self.get_remote_repo_path()
+
+        if not self.is_fastforward():
+            date_filter = None
+        else:
+            range_criteria = {START_DATE_KEYWORD: self.last_published,
+                              END_DATE_KEYWORD: self.predistributor["last_publish"]}
```
mhrivnak: Since ``self.predistributor`` is (I think) a proper object, I think we'd prefer accessing with normal attribute notation. Minor suggestion, but it makes it more clear what we're dealing with.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+                skip_fast_forward = True
+
+        if self.last_published and isinstance(self.last_published, basestring):
+            last_published = dateutils.parse_iso8601_datetime(self.last_published)
+        else:
+            last_published = self.last_published
+        if last_published:
+            last_published = last_published.replace(tzinfo=None)
+
+        if self.last_deleted and isinstance(self.last_deleted, basestring):
```
mhrivnak: Same. Let's get a guarantee of the type further up the stack.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        else:
+            last_deleted = self.last_deleted
+        if last_deleted:
+            last_deleted = last_deleted.replace(tzinfo=None)
+
+        config_skip_ff = self.get_config().get("skip_fast_forward", False)
+        skip_fast_forward = skip_fast_forward | misc.strbool2bool(config_skip_ff)
+        force_refresh = self.get_config().get("force_refresh", False)
+
+        return last_published and ((last_deleted and last_published > last_deleted) or
```
mhrivnak: This would be easier to read if we could pull some of it out into an independent evaluation. For example:

```
removal_since_last_publish = (last_deleted and last_published > last_deleted) or not last_deleted
```

That would make it easier to reason about whether the logic is correct.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            date_filter = {ASSOCIATED_UNIT_DATE_KEYWORD: {"$gte": start_date_str}}
+        elif end_date:
+            date_filter = {ASSOCIATED_UNIT_DATE_KEYWORD: {"$lte": end_date_str}}
+        return date_filter
+
+    def get_remote_repo_path(self):
+        """
+        Returns the full path to the published repository on remote server.
+        """
+        if "relative_url" in self.repo.notes and self.repo.notes["relative_url"]:
```
mhrivnak: Ideally this should come from the distributor config. The yum distributor does have the ability to look in the notes field to see if the importer left it a note about the feed's relative path, but even in that case, I think it saves the result in the distributor config and reads it from there. Reading the value from the notes field does not give the user a chance to change the value in the config.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            return self.repo.notes["relative_url"]
+        else:
+            return self.repo.id
+
+    def get_repo_content_types(self):
+        """
+        This method needs to be implemented in each plugin. It returns a list of content types
+        supported by the plugin.
```
mhrivnak: need to doc the return value

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            repo_relative_path = self.predistributor['config']['relative_url']
+            dirs.append(os.path.realpath(os.path.join(self._get_root_publish_dir(),
+                                                      repo_relative_path)))
+        return dirs
+
+    def get_units_directory_dest_prefix(self):
+        """
+        Returns the path on the remote server where content units should be stored.
+
+        :return: absolute path to /var/lib/pulp/
```
mhrivnak: This seems confusing. Perhaps there is some text missing?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+            origin_dest_prefix = os.path.join("content", "units",
+                                              self.get_config().get("content_types")[0])
+
+        return origin_dest_prefix
+
+    def get_original_list(self):
+        """
+        Returns a list of files that need to be rsynced to the content directory
+
+        :return: List of files from /var/lib/pulp/content that need to be rsynced
```
mhrivnak: File names only? Or relative paths? If the latter, relative to what? In any case, it should be specified. Saying just "files" could lead someone to think they are file objects. Instead, I think you want some portion of the paths to files.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+        :return: path to the content units
+        :rtype: str
+        """
+        original_list = reduce(lambda x, y: x + y, self.original_lists.values())
+        return misc.common_path(original_list)
+
+    def get_link_list(self, master, *args):
+        """
+        Returns list of symlinks that need to be coppied using rsync
```
mhrivnak: s/coppied/copied/

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
```
mhrivnak: need to doc args, such as ``master``

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+        link_list = self.link_lists[master][:]
+        for p in range(len(link_list)):
+            link_list[p] = os.path.relpath(link_list[p], link_rel_prefix)
+        return link_list
+
+    def get_link_prefix(self, master_dir):
+        """
+        Returns relative path inside the working directory to where the symlinks are stored
+
+        :param master_dir: path to the master publish directory for the repo
```
mhrivnak: missing ``:type``

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
 
         elif os.path.isfile(entry_path):
             os.unlink(entry_path)
+
+
+def common_path(dirs):
```
mhrivnak: need doc block. It would be helpful to explain what exactly this is doing beyond what ``os.path.commonprefix`` does and why.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+    return _common_path
+
+
+def make_link_unit(unit, extra_dst_path, working_dir,
+                   remote_repo_path, remote_root, configuration, extra_src_path, out_path):
+    """
+    This method creates symlink in working directory, pointing to remote
+    content directory, so they can be sync to remote server
+
+    :param repo_type: type of repository
```
mhrivnak: looks like this got removed

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+    :type repo_type: str
+    :param unit: Pulp content unit
+    :type unit: Unit
```
mhrivnak: need full python path unless it is imported into current namespace

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+    :param extra_dst_path: ending path for processes units
```
mhrivnak: I think I know what this is getting at, but it should probably be re-worded.for clarity.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+                          (for example 'Packages' for rpm)
+    :type extra_dst_path: str
+    :param working_dir: Working directory
+    :type working_dir: str
+    :param remote_repo_path: relative repo destination path on remote server
+    :type remote_repo_path: str
+    :param remote_root: remote destination root directory
+    :type remote_root: str
+    :param configuration: configuration of distributor
+    :type configuration: dict
```
mhrivnak: missing ``extra_src_path`` and ``out_path``

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+        os.makedirs(os.path.join(working_dir, extra_dst_path, *extra_src_path))
+
+    origin_path = get_origin_rel_path(unit, configuration)
+
+    dest = os.path.normpath(os.path.join(working_dir,
+                                         *([extra_dst_path] + extra_src_path + [filename])))
+    link_source = os.path.relpath(os.path.normpath(os.path.join(remote_root,
+                                                                os.path.dirname(origin_path))),
+                                  os.path.normpath(os.path.join(remote_root,
+                                                                remote_repo_path.lstrip("/"),
```
mhrivnak: As with somewhere else, if there is a leading "/", it's not a relative path. It should fail validation as part of user input, and here it should raise a ``ValueError`` or similar rather than trying to "fix" it.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+                                                                out_path, extra_dst_path)))
+
+    _log.info("LN %s -> %s " % (link_source, dest))
+    if os.path.islink(dest):
+        os.remove(dest)
+    os.symlink(link_source, dest)
+    return os.path.join(*([extra_dst_path] + extra_src_path + [filename]))
```
mhrivnak: Given the complexity, it would be better to build this path once and reference it twice. Then you could possibly even add a comment explaining what it's doing.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+
+    dest = os.path.normpath(os.path.join(working_dir,
+                                         *([extra_dst_path] + extra_src_path + [filename])))
+    link_source = os.path.relpath(os.path.normpath(os.path.join(remote_root,
+                                                                os.path.dirname(origin_path))),
+                                  os.path.normpath(os.path.join(remote_root,
+                                                                remote_repo_path.lstrip("/"),
+                                                                out_path, extra_dst_path)))
+
+    _log.info("LN %s -> %s " % (link_source, dest))
```
mhrivnak: Do users actually want to see this?

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+    os.symlink(link_source, dest)
+    return os.path.join(*([extra_dst_path] + extra_src_path + [filename]))
+
+
+def get_origin_rel_path(unit, config):
+    """
+    return relative path of content unit in remote fs
+
+    :param unit: Pulp content unit
+    :type unit: Unit
```
mhrivnak: need full python path unless it is imported into current namespace

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+
+    storage_dir = pulp_config.get('server', 'storage_dir')
+    content_base = os.path.join(storage_dir, "content", unit.metadata["_content_type_id"])
+    rel_unit_path = os.path.relpath(unit.storage_path, content_base)
+    remote_content_base = get_remote_content_base(config)
+    remote_content_type_dir = unit.metadata["_content_type_id"]
+    return os.path.join(remote_content_base, remote_content_type_dir, rel_unit_path)
+
+
+def get_remote_content_base(config):
```
mhrivnak: doc block

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+    remote_content_base = config.get("relative_url")
+    if not remote_content_base:
+        remote_content_base = os.path.join("content")
```
mhrivnak: It doesn't look like the call to ``os.path.join`` is adding anything.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+
+
+def common_path(dirs):
+    _common_path = os.path.commonprefix(dirs)
+    if not _common_path.endswith("/"):
+        _common_path = os.path.dirname(_common_path)
+    return _common_path
+
+
+def make_link_unit(unit, extra_dst_path, working_dir,
```
mhrivnak: This function and the next two look like they are probably specific to the rsync distributor. I think there is probably a better place to put them than a ``util.misc`` module.

server/pulp/plugins/util/misc.py
```diff
@@ -142,3 +143,89 @@ def clear_directory(path, skip_list=()):
...
+
+
+def get_remote_content_base(config):
+    remote_content_base = config.get("relative_url")
+    if not remote_content_base:
+        remote_content_base = os.path.join("content")
+    return remote_content_base
+
+
+def strbool2bool(_str):
```
mhrivnak: I'm hoping we can remove this. It looks like each use of it could or should not have any questions about what type it's dealing with.

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
             for unit in units_group:
                 if hash(unit) not in units_we_already_had:
                     self.units_to_download.append(unit)
+
+
+class RSyncFastForwardUnitPublishStep(UnitPublishStep):
+
+    def __init__(self, step_type, link_list, original_list, unit_types=None, repo_type=None,
+                 working_dir=None, distributor_type=None, config=None, association_filters=None,
```
mhrivnak: lots of these are missing in the doc block

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
...
+                 remote_repo_path=None, unit_fields=None, extra_src_path=None, out_path=None):
+        """
+        Set the default parent, step_type and units_type for the the
+        publish step.
+
+        :param step_type: The id of the step this processes
+        :type step_type: str
+        :param link_list: share list object across publish steps which
```
mhrivnak: s/share/shared/

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
...
+         holds information of file links that will be synced
+        :type link_list: list
+        :param original_list: share list object across publish steps which
```
mhrivnak: s/share/shared/

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
...
+        self.out_path = out_path
+
+        super(RSyncFastForwardUnitPublishStep,
+              self).__init__(step_type, association_filters=association_filters,
+                             unit_type=unit_types)
+
+    def process_unit(self, unit):
+        """
+        Creates symlink for a unit and appends the symlink to a list used to perform rsync later.
+        This also generates the list of actual content units that need to be rsynced later.
```
mhrivnak: need to doc ``unit`` parameter.

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
...
+        """
+        self.content_types.add(unit.metadata["_content_type_id"])
```
mhrivnak: Does this still work post-mongoengine?

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1231,73 @@ def process_main(self, item=None):
...
+                                                  self.get_config(),
+                                                  self.extra_src_path,
+                                                  self.out_path))
+        # add to list of content unit paths
+        self.original_list.append(unit.storage_path)
+
+    def _process_block(self, item=None):
+        """
+        This method is implemented here so we can adjust the remote_rel_prefix based on data in
+        the database.
```
mhrivnak: Can you explain more about what this is doing and why it's necessary?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def _get_predistributor(self):
+        """
+        Returns distributor which repo has to be published in before
+        publish in rsync distributor, content generator.
+        """
+        raise NotImplementedError()
+
+
+class Lazy(object):
```
mhrivnak: Based on IRC discussion, we can remove this class and just organize the data differently that gets shared between steps.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,456 @@
...
+        :return: (return code from rsync, stdout + stderr)
+        :rtype: tuple
+        """
+        tmpdir = os.path.join(self.get_working_dir(), '.tmp')
+        os.makedirs(os.path.join(tmpdir, path.lstrip("/")))
+        args = ['rsync', '-avr', '-f+ */']
+        args.extend(self.make_authentication())
+        args.append("%s/" % tmpdir)
+        args.append(self.make_destination(path).replace(str(path), ""))
+        _logger.info("remote mkdir: %s" % args)
```
mhrivnak: Perhaps these should be logged at ``debug``?

server/pulp/plugins/rsync/configuration.py
dkliban: I added doc strings for all these objects.

bmbouter: Thanks!

server/pulp/plugins/rsync/configuration.py
dkliban: I appreciate the suggestion. I am going to keep it as is though.

server/pulp/plugins/rsync/publish.py
dkliban: I updated the doc strings

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class TypeValidation(object):
+    """
+    Validates that the value is one of allowed types
+    """
+    def __init__(self, allowed_types):
+        self.allowed_types = allowed_types
+
+    def __call__(self, value, config):
```
bmbouter: It would be great to have a docstring with the return arguments explained.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+class RequireOptionalIf(object):
+    """
+    Validates that if a particular config is present, other configs that are needed for it are also
+    present.
+    """
+    def __init__(self, required_for_attr, condition):
+        self.required_for_attr = required_for_attr
+        self.condition = condition
+
+    def __call__(self, value, config):
```
bmbouter: It would be great to have a docstring with the return arguments explained.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _err(self, path):
+        return "%s attribute is required" % ("::".join(path))
+
+
+class RelativePathValidation(object):
+    """
+    Validates that a path does not start with a forward slash.
+    """
+    def __call__(self, value, config):
```
bmbouter: It would be great to have a docstring with the return arguments explained.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Validate the prospective configuration instance for the given repository.
+
+    :param repo: repository to validate the config for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance to validate
+    :type  config: pulp.plugins.config.PluginCallConfiguration
+    :param config_conduit: conduit providing access to relevant Pulp functionality
+    :type  config_conduit: pulp.plugins.conduits.repo_config.RepoConfigConduit
+    :return: tuple of (bool, str) stating that the configuration is valid or not and why
```
bmbouter: I usually separate arguments from return params with a line of whitespace. I think it's more readable when they are separated.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __call__(self, value, config):
+        if value in self.values:
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        values_str = ", ".join(self.values)
+        return "{value} is not in allowed values: {allowed}" % (value, values_str)
```
bmbouter: Please use gettext to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    Validates that the value is not None
+    """
+    def __call__(self, value, config):
+        if value:
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        return "attribute cannot be empty"
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __call__(self, value, config):
+        if [True for _type in self.allowed_types if issubclass(type(value), _type)]:
+            return (True, None)
+        else:
+            return (False, self._err(value))
+
+    def _err(self, value):
+        str_types = ", ".join(self.allowed_types)
+        return "type of atribute(%s) is not in allowed types: %s" % (type(value), str_types)
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+            elif isinstance(required_for_attr, dict):
+                for key, val in required_for_attr.iteritems():
+                    fifo.insert(0, (val, subconfig.get(key)), path + [key])
+            elif isinstance(required_for_attr, basestring):
+                if required_for_attr not in subconfig:
+                    return (False, self._err(path + [required_for_attr]))
+        return (True, None)
+
+    def _err(self, path):
+        return "%s attribute is required" % ("::".join(path))
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    Validates that a path does not start with a forward slash.
+    """
+    def __call__(self, value, config):
+        if os.path.isabs(value):
+            return (False, self._err(value))
+        else:
+            return (True, None)
+
+    def _err(self, value):
+        return "attribute cannot start with a /"
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    :param config_conduit: conduit providing access to relevant Pulp functionality
+    :type  config_conduit: pulp.plugins.conduits.repo_config.RepoConfigConduit
+    :return: tuple of (bool, str) stating that the configuration is valid or not and why
+    :rtype:  tuple of (bool, str or None)
+    """
+    _config = config.flatten()  # now config is {}
+    err_list = []
+    crit_fail = False
+    if "remote" not in _config or ("remote" in _config and not isinstance(_config["remote"], dict)):
+        err_list.append("'remote' dict missing in distributor's configuration")
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+        crit_fail = True
+    if not crit_fail:
```
bmbouter: I think the variable `crit_fail` is uneeded here. Instead the `if not crit_fail` could become an `else:`

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+        missing_attr_tmpl = "'%s' missing in 'remote' section of distributor's configuration"
```
bmbouter: Please use `gettext` to internationalize.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+        for attr, validations in REMOTE_MANDATORY_KEYS.iteritems():
+            if attr not in _config["remote"]:
+                err_list.append(missing_attr_tmpl % attr)
+                continue
+            for validation in validations:
+                succeed, _err = validation(_config["remote"][attr], _config)
+                if not succeed:
+
```
bmbouter: Strange whitespace here

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    _config = config.flatten()  # now config is {}
+    err_list = []
+    crit_fail = False
+    if "remote" not in _config or ("remote" in _config and not isinstance(_config["remote"], dict)):
+        err_list.append("'remote' dict missing in distributor's configuration")
+        crit_fail = True
+    if not crit_fail:
+        missing_attr_tmpl = "'%s' missing in 'remote' section of distributor's configuration"
+        for attr, validations in REMOTE_MANDATORY_KEYS.iteritems():
+            if attr not in _config["remote"]:
```
bmbouter: This doesn't have to be rewritten here, but for next time consider using the [easier-to-ask-forgiveness approach](http://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html) for readability.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,165 @@
...
+    Validate the prospective configuration instance for the given repository.
+
+    :param repo: repository to validate the config for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance to validate
+    :type  config: pulp.plugins.config.PluginCallConfiguration
+    :param config_conduit: conduit providing access to relevant Pulp functionality
+    :type  config_conduit: pulp.plugins.conduits.repo_config.RepoConfigConduit
+    :return: tuple of (bool, str) stating that the configuration is valid or not and why
+    :rtype:  tuple of (bool, str or None)
```
bmbouter: s/(bool, str or None)/(bool, str) or (bool, None)/

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
+from gettext import gettext as _
+import logging
+import os
+import time
+import uuid
+import shutil
+
+import kobo.shortcuts
+
+import mongoengine
```
bmbouter: These imports need to be regrouped. See the [grouping in PEP8](https://www.python.org/dev/peps/pep-0008/#imports). I think the bottom group should only contain `pulp.plugins` imports since that is this package.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+
+ASSOCIATED_UNIT_DATE_KEYWORD = "created"
+START_DATE_KEYWORD = 'start_date'
+END_DATE_KEYWORD = 'end_date'
+
+_logger = logging.getLogger(__name__)
+
+
+class RSyncPublishStep(PublishStep):
+    _CMD = "rsync"
```
bmbouter: I usually put a whitespace between class attributes and the class definition.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :param step_type: Name of step
+        :type step_type: str
+        :param file_list: list of paths relative to src_directory
+        :type file_list: list
+        :param src_directory: absolute path to directory which contains all items in file_list
+        :type src_directory: str
+        :param dest_directory: path to directory relative to remote root where files should be
+               rsynced
+        :type dest_directory: str
+        :param config: distributor configuration
```
bmbouter: This needs a `:type config:` section

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :param fast_forward:
```
bmbouter: `fast_forward` needs a description

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :type fast_forward: bool
+        :param exclude: list of files/directories to skip
+        :type exclude: list
+        :param delete: determines if --delete is passed to rsync
+        :type delete: boolean
```
bmbouter: Is it `bool` or `boolean`? Anyways the same style should be used everywhere

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :param src_directory: absolute path to directory which contains all items in file_list
+        :type src_directory: str
+        :param dest_directory: path to directory relative to remote root where files should be
+               rsynced
+        :type dest_directory: str
+        :param config: distributor configuration
+        :param fast_forward:
+        :type fast_forward: bool
+        :param exclude: list of files/directories to skip
+        :type exclude: list
```
bmbouter: What type is contained in the list?

bmbouter: For instance this could be `list of str`

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :param delete: determines if --delete is passed to rsync
+        :type delete: boolean
+        :param links: if true, --links is passed to rsync
+        :type links: bool
+        """
+        self.description = _('Rsync files to remote destination')
```
bmbouter: internationalization :+1:

bmbouter: This can probably be moved lower with the rest of the instance vars right?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        self.dest_directory = dest_directory
+        self.links = links
+
+    def remote_mkdir(self, path):
+        """
+        Creates path on remote server. The path is rooted in distributor's remote_root directory.
+
+        :param path: path to create on remote server
+        :type path: str
+        :return: (return code from rsync, stdout + stderr)
```
bmbouter: Similar to ^ I separate return docs from param docs with a line of whitespace. Similar throughout rest of PR. This is a style choice, not required.

bmbouter: What does `stdout + stderr` mean? Is that a string output?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        if args:
+            cmd += args
+        return cmd
+
+    def make_authentication(self):
+        """
+        Returns a list of strings representing args for command for authenticating against a remote
+        server.
+
+        :return: list of arguments for auth. e.g., ['-e',  ssh -l ssh_user -i /key_path'
```
bmbouter: This list example is invalid. The quoting isn't complete and there there isn't a closing list bracket at the end.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :rtype: str
+        """
+        return os.path.join(self.get_config().flatten()['remote']['root'], relative_path)
+
+    def make_destination(self, relative_path):
+        """
+        Parse from self.config information to make up a hostname and remote path used for rsync
+        command.
+
+        :param relative_path: path relative to the root configured in distributor
```
bmbouter: `relative_path` needs a `:type` section.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        """
+        if self.get_config().flatten()["remote"]['auth_type'] == 'local':
+            return self.get_config().flatten()["remote"]['root']
+
+        user = self.get_config().flatten()["remote"]['ssh_user']
+        host = self.get_config().flatten()["remote"]['host']
+        remote_root = self.get_config().flatten()["remote"]['root']
+        return '%s@%s:%s' % (user, host, os.path.join(remote_root, relative_path))
+
+    def call(self, args, include_args_in_output=True):
```
bmbouter: `include_args_in_output` needs `:param` docs in the docstring.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        :return: (boolean indicating success or failure and std_out + std_err)
+        :rtype: tuple of boolean and string
+        """
+        for t in xrange(10):
+            rv, out = kobo.shortcuts.run(cmd=args, can_fail=True)
+            possible_known_exceptions = \
+                ("ssh_exchange_identification:" in out) or ("max-concurrent-connections=25" in out)
+            if not (rv and possible_known_exceptions):
+                break
+            _logger.info("Connections limit reached, trying once again in thirty seconds.")
```
bmbouter: Please internationalize this with `gettext`.

server/pulp/plugins/rsync/publish.py
dkliban: resolved

server/pulp/plugins/rsync/publish.py
bmbouter: I think this is necessary because NoneType is not iterable so we have to guard against the default case of `None`.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+    def rsync(self):
+        """
+        This method formulates the rsync command based on parameters passed in to the __init__ and
+        then executes it.
+
+        :return: (boolean indicating success or failure, stdout + stderr)
+        :rtype: tuple
+        """
+        if not self.file_list:
+            return (True, "Nothing to sync")
```
bmbouter: Maybe i8ln?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        output = ""
+        list_of_files = os.path.join(self.get_working_dir(), str(uuid.uuid4()))
+        open(list_of_files, 'w').write("\n".join(sorted(self.file_list)))
+
+        # copy files here, not symlinks
+        (is_successful, this_output) = self.remote_mkdir(self.dest_directory)
+        if not is_successful:
+            _logger.error("Cannot create directory %s: %s" % (self.dest_directory, this_output))
```
bmbouter: Maybe i8ln?

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+            return (is_successful, this_output)
+        output += this_output
+        rsync_args = self.make_rsync_args(list_of_files, self.src_directory,
+                                          self.dest_directory, self.exclude)
+        (is_successful, this_output) = self.call(rsync_args)
```
bmbouter: There isn't agreement in the Python community about this, but I recommend [exceptions over return codes](http://nedbatchelder.com/text/exceptions-vs-status.html). Specifically this advice is for `call()`. This function checks the return codes many times and then returns, that could all be collapsed into one `except` block for example.

One reason possible not to use Exceptions is if multiple ones could occur and `call()` may need to continue past the first one. I don't think that is the case here, but I could be wrong.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        _logger.info(this_output)
+        if not is_successful:
+            _logger.error(this_output)
+            return (is_successful, this_output)
+        output += this_output
+        return (is_successful, output)
+
+    def process_main(self):
```
bmbouter: This needs a docblock.

server/pulp/plugins/rsync/configuration.py
dkliban: On second thought, I'll change it to use all(). Thanks!

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+            raise PulpCodedException(message=output)
+
+
+class Publisher(PublishStep):
+    """
+    RSync publisher class that provides the common code for publishing to remote server. Each
+    plugin needs to extend this class and add necesary logic to the _add_necesary_steps method.
+
+    :ivar symlink_list: list of symlinks to rsync
+    :ivar content_unit_file_list: list of content units to rsync
```
bmbouter: There are a lot more variables than this. Some non-public ones could have `_` put in front of them. Also this could be consolidated with the variables in the docstring for `__init__()`.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        distributor = Distributor.objects.get_or_404(repo_id=self.repo.id,
+                                                     distributor_id=publish_conduit.distributor_id)
+        self.last_published = distributor["last_publish"]
+        self.last_deleted = repo.last_unit_removed
+        self.repo = repo
+        self.predistributor = self._get_predistributor()
+
+        if self.last_published:
+            string_date = dateutils.format_iso8601_datetime(self.last_published)
+        elif self.last_published:
```
bmbouter: This branch will never be taken because if `self.last_published` is `True` then the earlier branch will always be taken and this being an `elif` will never run.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+            string_date = self.last_published
+        else:
+            string_date = None
+        if self.predistributor:
+            search_params = {'repo_id': repo.id,
+                             'distributor_id': self.predistributor["id"],
+                             'started': {"$gte": string_date}}
+            self.predist_history = RepoPublishResult.get_collection().find(search_params)
```
bmbouter: This should be converted to a mongoengine based query. This is using pymongo directly and it should not do that.

server/pulp/plugins/rsync/publish.py
```diff
@@ -0,0 +1,443 @@
...
+        else:
+            self.predist_history = []
+
+        self.remote_path = self.get_remote_repo_path()
+
+        if not self.is_fastforward():
```
bmbouter: This would be more readable if there wasn't a negative in the expression. The order could be swapped easily:

```
if self.is_fastforward():
    L#309 - L#313
else:
    L#307
```

server/pulp/plugins/util/publish_step.py
```diff
@@ -23,6 +24,7 @@
 from nectar import listener
 from nectar.downloaders.local import LocalFileDownloader
 from nectar.downloaders.threaded import HTTPThreadedDownloader
+from pulp.server.config import config as pulp_config
```
bmbouter: You didn't create this, but these import groupings are not right.

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
             for unit in units_group:
                 if hash(unit) not in units_we_already_had:
                     self.units_to_download.append(unit)
+
+
+class RSyncFastForwardUnitPublishStep(PluginStep):
+    def __init__(self, step_type, repo=None, config=None, unit_type=None, date_filter=None,
+                 remote_repo_path=None, unit_fields=None, published_unit_path=None,
+                 **kwargs):
```
bmbouter: Why is kwargs needed here?

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
...
+        :type unit_fields: list of str
+        :param date_filter: filter for unit association with repo
+        :type date_filter: mongoengine.Q or types.NoneType
+        :param remote_repo_path: relative path on remote server where published repo should live
+        :type remote_repo_path: str
+        :param published_unit_path: relative path to the content unit from root of published repo
+        :type published_unit_path: str
+
+        """
+        self.description = _('Generating relative symlinks')
```
bmbouter: i8ln :+1:

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
...
+
+    def get_iterator(self):
+        """
+        Returns a generator for iterating over content units that need to have symlinks generated
+
+        :return: a generator of ContentUnit objects
+        :rtype:  generator
+        """
+        if self.date_filter:
+            full_filter = self.date_filter & mongoengine.Q(unit_type_id=self.unit_type)
```
bmbouter: yay mongoengine

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
...
+        return repo_controller.find_repo_content_units(self.repo, repo_content_unit_q=full_filter,
+                                                       unit_fields=self.unit_fields,
+                                                       yield_content_unit=True)
+
+    def process_main(self, item=None):
+        """
+        Creates symlink for a unit and appends the symlink to a list used to perform rsync later.
+        This also generates the list of actual content units that need to be rsynced later.
+
+        :param unit: Content type unit
```
bmbouter: The param name here is `item`, not `unit`.

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
...
+        :type unit: Unit
+        :param working_dir: Working directory
+        :type working_dir: str
+        :param remote_repo_path: relative repo destination path on remote server
+        :type remote_repo_path: str
+        :param remote_root: remote destination root directory
+        :type remote_root: str
+        :param published_unit_path: directory name inside repo that the symlink should be in
+        :type published_unit_path: list of str
+
```
bmbouter: This whitespace line is uneccessary

server/pulp/plugins/util/publish_step.py
```diff
@@ -1231,3 +1233,126 @@ def process_main(self, item=None):
...
+        symlink = self.make_link_unit(item, self.get_working_dir(), self.remote_repo_path,
+                                      self.get_config().get("remote")["root"],
+                                      self.published_unit_path)
+        self.parent.symlink_list.append(symlink)
+
+    def make_link_unit(self, unit, working_dir, remote_repo_path, remote_root,
+                       published_unit_path):
+        """
+        This method creates symlink in working directory, pointing to remote
+        content directory, so they can be sync to remote server
```
bmbouter: A period is needed at the end here I think.

server/pulp/plugins/rsync/publish.py
dkliban: added

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: Good call. I am removing it.

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done. but this should probably be rewritten cause describing the data structure is difficult.

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: thanks. i did that throughout the rest of the PR now.

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: done

server/pulp/plugins/rsync/configuration.py
dkliban: removed it

server/pulp/plugins/rsync/configuration.py
dkliban: got it

server/pulp/plugins/rsync/configuration.py
dkliban: i fixed it and i did the same for the methods that actually generate this tuple

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: The instance variable that is set by `fast_forward` parameter is not used anywhere. I am removing it.

This step simply rsyncs files that have already been selected by a step that is aware of this being an incremental publish.

server/pulp/plugins/rsync/publish.py
dkliban: bool

server/pulp/plugins/rsync/publish.py
dkliban: sure can! and now it is

server/pulp/plugins/rsync/publish.py
dkliban: it's a string made up of stdout and stderr from rsync command used to create remote directories. i updated the doc string

server/pulp/plugins/rsync/publish.py
dkliban: fixed this.

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: added!

server/pulp/plugins/rsync/publish.py
dkliban: Do you want me to rewrite this?

bmbouter: I do think it would be better if it were rewritten. If you want to leave it as-is you can.

server/pulp/plugins/rsync/publish.py
dkliban: i moved it all to __init__() docstring. and i added the rest of the ivars.

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: good catch. i removed it

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/util/publish_step.py
dkliban: They are not needed. I'll remove em.

server/pulp/plugins/util/publish_step.py
dkliban: i rearranged em

server/pulp/plugins/util/publish_step.py
dkliban: fixed

server/pulp/plugins/util/publish_step.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: Which mongoengine model should I use?

bmbouter: Actually, the model you need doesn't exist yet. This means you'll have to use pymongo as you've written it for now. No adjustment needed.

server/pulp/plugins/rsync/publish.py
dkliban: done

server/pulp/plugins/rsync/publish.py
dkliban: done

dkliban:
I believe it's ready for final review.

server/pulp/plugins/rsync/configuration.py
```diff
@@ -0,0 +1,260 @@
+import gettext
+import logging
+import os
+
+_ = gettext.gettext
```
bmbouter: This should not be done. Instead use `from gettext import gettext as _`

bmbouter:
This was shows to be fully working when tested using the rsync and iso distributors! I'm adding the LGTM.

@dkliban You've done a really great job on this. I'm very proud of this work being added to Pulp!

PR #2065, opened 2015-09-24
ref #1195 - Implement the pulp-streamer.

jeremycline:
This commit adds the 'lazy' module inside of pulp.server.
It also adds streamer settings to Pulp's server.conf.

PR #2627, opened 2016-07-06
Fixes #2065 - Handles OSTree cert paths correctly

parthaa:
Fixed a bug that caused the wrong repo destination path to be verified on a
ostree pull.

Look at https://pulp.plan.io/issues/2065 for more info

### DIFF
--- a/oid_validation/pulp/oid_validation/oid_validation.py
+++ b/oid_validation/pulp/oid_validation/oid_validation.py
@@ -194,16 +194,17 @@ class OidValidator:
 
         valid = False
         for prefix in repo_url_prefixes:
-            # Extract the repo portion of the URL
-            repo_dest = dest[dest.find(prefix) + len(prefix):]
-            try:
-                valid = cert.check_path(repo_dest)
-            except AttributeError:
-                # not an entitlement certificate, so no entitlements
-                log_func('The provided client certificate is not an entitlement certificate.\n')
-            # if we have a valid url check, no need to continue
-            if valid:
-                break
+            if dest.find(prefix) > -1:
+                # Extract the repo portion of the URL
+                repo_dest = dest[dest.find(prefix) + len(prefix):]
+                try:
+                    valid = cert.check_path(repo_dest)
+                except AttributeError:
+                    # not an entitlement certificate, so no entitlements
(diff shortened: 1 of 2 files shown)


mhrivnak:
ok test

mhrivnak:
We won't be able to accept this patch unless you add your name to the AUTHORS file at the project root. ;)

parthaa:
@mhrivnak updated :)

oid_validation/pulp/oid_validation/oid_validation.py
```diff
@@ -194,16 +194,17 @@ def _check_extensions(self, cert_pem, dest, log_func, repo_url_prefixes):
...
-            repo_dest = dest[dest.find(prefix) + len(prefix):]
-            try:
-                valid = cert.check_path(repo_dest)
-            except AttributeError:
-                # not an entitlement certificate, so no entitlements
-                log_func('The provided client certificate is not an entitlement certificate.\n')
-            # if we have a valid url check, no need to continue
-            if valid:
-                break
+            if dest.find(prefix) > -1:
```
jeremycline: For the sake of clarity you could do ``if prefix not in dest:``, but it's up to you. It just saves someone who can't remember the exact behavior of ``find`` a few seconds looking at the docs.

jeremycline: I know you're only fixing this, but I think this whole block would be much more readable if ``os.path.commonprefix`` was used. Something like:
```
for prefix in repo_url_prefixes:
    if os.path.commonprefix([prefix, dest]) == prefix: 
        repo_dest = os.path.relpath(dest, prefix)
        repo_dest = '/' + repo_dest  # Looking at the current code I _think_ repo_dest starts with a /
    try:
        valid = cert.check_path(repo_dest)
....
```

parthaa: > I know you're only fixing this, but I think this whole block would be much more readable if os.path.commonprefix was used. Something like:

Shouldnt the "try" be inside the if block ?

jeremycline: Yeah I think so

parthaa:
Being addressed elsewhere..

PR #2629, opened 2016-07-08
Update the prefix-stripping method in oid_validator

jeremycline:
The prefix stripping method used in the oid_validator failed if the
prefix was _not_ found in the path. This fixes that particular issue,
improves readability, and adds a unit test.

fixes #2065

### DIFF
--- a/oid_validation/pulp/oid_validation/oid_validation.py
+++ b/oid_validation/pulp/oid_validation/oid_validation.py
@@ -13,6 +13,7 @@ The OID structure follows the Red Hat model. Download URLs are found at:
 The * represents the product ID and is not used as part of this calculation.
 '''
 
+import os
 from gettext import gettext as _
 from ConfigParser import NoOptionError, SafeConfigParser, NoSectionError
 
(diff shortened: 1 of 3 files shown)


oid_validation/pulp/oid_validation/oid_validation.py
```diff
@@ -194,16 +195,18 @@ def _check_extensions(self, cert_pem, dest, log_func, repo_url_prefixes):
...
-            except AttributeError:
-                # not an entitlement certificate, so no entitlements
-                log_func('The provided client certificate is not an entitlement certificate.\n')
-            # if we have a valid url check, no need to continue
-            if valid:
-                break
+            if dest.startswith(prefix):
+                # rhsm throws a ValueError if there's no leading /. Amusingly, it immediately
+                # strips it off.
+                repo_dest = '/' + os.path.relpath(dest, prefix)
```
jortel: I recommend using path manipulation functions whenever possible and avoid string manipulation.  `os.path.join('/', os.path.relpath(dest, prefix))`.  Not a big deal.

parthaa:
Tested on a sat. Worked well

PR #2630, opened 2016-07-08
1829 - Fixed repo list summary view

BrnoPCmaniak:
closes #1829
https://pulp.plan.io/issues/1829


client_lib/pulp/client/commands/repo/cudl.py
```diff
@@ -522,16 +522,22 @@ def _default_summary_view(repo_list, prompt):
 
     if isinstance(repo_list, dict) and repo_list != {}:
         id_value = repo_list['id'] + ' '
-        name_value = repo_list['display_name']
```
goosemania: Actually, what do you think about making a `name_value` an empty string and leave everything else as is to preserve the output syntax, just in case someone parses it?
Something like
`name_value = repo_list['display_name'] if repo_list['display_name'] is not None else ''`

client_lib/test/unit/client/commands/repo/test_cudl.py
```diff
@@ -517,6 +517,33 @@ def test_summary(self):
         self.assertEqual(self.recorder.lines[0], 'abcdef  ABCDEF\n')
         self.assertEqual(self.recorder.lines[1], 'xyz     XYZ\n')
 
+    def test_summary_empty_display_name(self):
```
goosemania: It will be good to add a docstring which will describe what this test does, like [here](https://github.com/pulp/pulp/blob/master/server/test/unit/server/controllers/test_importer.py), for example.

client_lib/test/unit/client/commands/repo/test_cudl.py
```diff
@@ -540,6 +567,29 @@ def test_one_repo_summary(self):
         self.assertEqual(self.command.get_repository.call_count, 1)
         self.assertEqual(self.recorder.lines[0], 'zoo-repo   zoo-repo\n')
 
+    def test_one_repo_summary_empty_display_name(self):
```
goosemania: It will be good to add a docstring which will describe what this test does.

client_lib/pulp/client/commands/repo/cudl.py
mhrivnak: Good idea. Perhaps an even simpler approach would be:

```
name_value = repo_list['display_name'] or ''
```

BrnoPCmaniak: Well I had it like this, but I thought avoiding a bunch of spaces is better. Maybe they don't really matter...

client_lib/test/unit/client/commands/repo/test_cudl.py
```diff
@@ -517,6 +517,36 @@ def test_summary(self):
         self.assertEqual(self.recorder.lines[0], 'abcdef  ABCDEF\n')
         self.assertEqual(self.recorder.lines[1], 'xyz     XYZ\n')
 
+    def test_summary_empty_display_name(self):
+        """
+            Test output of repo summary if display_name of one of the repos is not specified.
```
goosemania: There is no need for additional spaces at the beginning (take a look at other docstrings).

goosemania:
Good job! Let's wait for Jenkins to run the tests and then merge if the tests are ok

goosemania:
ok test

ipanova:
@BrnoPCmaniak just a note, once you merge a PR you should set the Target Release of where there the fix/feature will land. Please do so https://pulp.plan.io/issues/1829

PR #2631, opened 2016-07-09
missed mod_ssl for development environment

pronix:
it is required package for development environment

### DIFF
--- a/playpen/ansible/roles/dev/tasks/main.yml
+++ b/playpen/ansible/roles/dev/tasks/main.yml
@@ -68,6 +68,7 @@
       - tito
       - yum-utils
       - zlib-devel
+      - mod_ssl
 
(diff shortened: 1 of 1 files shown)


jeremycline:
Hi @pronix, thanks for the patch.

Is ``mod_ssl`` not pulled in as part of the "Install Pulp RPM dependencies" step (https://github.com/pulp/pulp/pull/2631/files#diff-0fc7fc8418befc643b0ef0ceaee17014L86)? It's an explicitly listed dependency in the Pulp platform spec file, so I would expect it to be installed at that point.

pronix:
hello @jeremycline 
you are right - it is working as expected. may be something was wrong yesterday with my internet.
and thanks for help.

jeremycline:
Happy to help, I'm glad things are working for you!

PR #2635, opened 2016-07-11
Allow at sign in user ids

asmacdo:
Adjust the regex to allow usernames to contain the at sign so that email
addresses can be valid usernames.

https://pulp.plan.io/issues/1728

### DIFF
--- a/client_lib/pulp/client/validators.py
+++ b/client_lib/pulp/client/validators.py
@@ -11,7 +11,7 @@ from pulp.common import dateutils
 from pulp.common.plugins import importer_constants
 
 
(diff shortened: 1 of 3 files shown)


client_lib/pulp/client/validators.py
```diff
@@ -115,8 +115,8 @@ def id_validator_allow_dots(x):
 
     for input_id in x:
         if ID_REGEX_ALLOW_DOTS.match(input_id) is None:
-            raise ValueError(_('value must contain only letters, numbers, underscores, periods and '
-                               'hyphens'))
+            raise ValueError(_('value must contain only letters, numbers, underscores, periods,'
+                               'hyphens, and @'))
```
ipanova: maybe just add s/and @/and @ sign

PR #1983, opened 2015-07-14
Adding documetation for /v2/plugins/* API

ipanova:
closes #1121
https://pulp.plan.io/issues/1121

PR #2637, opened 2016-07-11
1983 - Improve sync optimization to do a full sync after config file change or content removal

dralley:
Add full sync checks to the platform, where they can be used by all plugins.

closes #1983
https://pulp.plan.io/issues/1983

### DIFF
--- a/docs/user-guide/release-notes/2.11.x.rst
+++ b/docs/user-guide/release-notes/2.11.x.rst
@@ -1,0 +1,12 @@
+=======================
+Pulp 2.11 Release Notes
+=======================
+
+Pulp 2.11.0
+===========
+
+New Features
+------------
+
+* For RPM content, a full sync will be forced if the sync configuration has been changed or content
+  has been removed since the last sync.
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -798,6 +798,10 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
                                                                sync_start_timestamp, summary,
                                                                details, result_code,
                                                                before_sync_unit_count)
+        # Update the override config if it has changed
+        if check_override_config_change(repo_id, call_config):
+            model.Importer.objects(repo_id=repo_id).\
+                update(set__last_override_config=call_config.override_config)
         # Do an update instead of a save in case the importer has changed the scratchpad
         model.Importer.objects(repo_id=repo_obj.repo_id).update(set__last_sync=sync_end_timestamp)
         # Add a sync history entry for this run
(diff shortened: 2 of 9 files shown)


asmacdo:
@dralley, I'm going to assign you until you are ready for review.

dralley:
@asmacdo It is ready for review

docs/user-guide/release-notes/2.11.x.rst
```diff
@@ -0,0 +1,13 @@
...
+Pulp 2.11 Release Notes
+=======================
+
+Pulp 2.11.0
+===========
+
+New Features
+------------
+
+* Re-sync of the repository will be a no operation by default if since the last sync, no units
```
ipanova: @dralley re-sync of the repository cannot be no operational, therefore have skipped result based on your code and the feature requirements. The improvements that you did will do always an *operational* sync, the difference is - sometimes it will perform all sync steps without skipping any and sometimes it will take advantage of sync optimization logic and skip some of the steps. What you did is - you moved part of logic into the platform, so we first will look into platform at some conditions, if those conditions are true then we will disregard the rest of logic optimization that is done in each plugin. and do a so-called force sync.

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +816,149 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+    repo_importer = model.Importer.objects.get_or_404(repo_id=repo_id)
+
+    # check if the override config is different from the last override config,
+    # excluding the 'force_full' key.  otherwise using the --force-full flag
+    # would always trigger a full sync on the next sync too
+    prev_config = repo_importer.last_override_config.copy()
+    current_config = call_config.override_config.copy()
+
+    prev_config.pop('force_full', False)
+    current_config.pop('force_full', False)
```
ipanova: why do you remove this key?

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +816,149 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+
+    return prev_config != current_config
+
```
ipanova: if current override config is different from previous one, you need to update last_override_config otherwise with the next sync it will contain outdated info. like here https://github.com/pulp/pulp/blob/master/server/pulp/server/controllers/repository.py#L1020

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +816,149 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+    :return:  Whether a full sync needs to be performed
+    :rtype:  boolean
+    """
+    force_full = check_force_full(call_config)
+    not_synced = check_no_previous_sync(conduit)
+    content_removed = check_unit_removed_since_last_sync(conduit, repo_id)
+    config_changed = check_config_updated_since_last_sync(conduit, repo_id)
+    override_config_changed = check_override_config_change(repo_id, call_config)
+
+    return force_full or not_synced or content_removed or config_changed or override_config_changed
```
ipanova: any reason why you decided to put every check into a separate function?
as another approach you could do these checks all here in one place, and this way you will avoid repeating code.
so it will be something similar to https://github.com/pulp/pulp/blob/deabcbc05da114ea704f3c457093c78b1fb1de65/server/pulp/server/controllers/repository.py#L997-L1015

ipanova:
@dralley  back to you, i did not check the unittests though, i will do that once the functional code will be updated/ready.

ipanova:
@dralley one more suggestion about the commit message length.
The commit message doesn't have git recommended length which makes it hard to read. This is why it looks strange in the PR title/subject because that follows the same conventions as git and is populated with the commit info. The recommendations are:
First / subject line of max 50 characters;
Subsequent lines max 72 characters.

ipanova:
@dralley don't forget to provide the link to your PRs in the issue https://pulp.plan.io/issues/1983

server/pulp/server/controllers/repository.py
dralley: @ipanova I remove the key because of a corner case.  Consider using the "force_full" flag manually.  On the next sync, whether "force_full" is used or not, it will believe that the configuration has changed and force another full sync.

docs/user-guide/release-notes/2.11.x.rst
dralley: Would you say that since there is no change in behavior from a user perspective (logic has simply been moved rather than added, necessarily), it doesn't need a release note at all?

I agree that it's incorrect as it exists currently - I can fix it if there does need to be a note.

server/pulp/server/controllers/repository.py
dralley: Noted.  I will put the update in the platform sync() function in order to avoid side effects from a function that is meant for checking.

server/pulp/server/controllers/repository.py
dralley: @ipanova Putting the checks into one function made it much, much, much more difficult to test effectively.  I did initially start out with that approach, but I decided that the benefits of doing it that way were outweighed by the costs.  In this case I think that the little bit of repeated code avoids more headaches than it would potentially cause.  

Basically, because there are multiple checks that result in a single True/False return from the function, the only way to isolate each check was to copy and paste mocking code for every check other than the check being tested, for every test, and it made the tests much less robust in my opinion.  

Purely on a practical basis though, having the checks split up has some useful benefits, like in the case you pointed out where we need to update the last_override_config when it has changed.  That should happen outside of the actual checking function.

docs/user-guide/release-notes/2.11.x.rst
ipanova: it did not change from user perspective, but i think it still worth to mention that there is behaviour change, that if the content was removed from last sync or config was updated + other conditions that were moved to platform, then a force-sync will be triggered

server/pulp/server/controllers/repository.py
ipanova: yeah, but the configuration indeed has changed, because force_full is part of importer_config

server/pulp/server/controllers/repository.py
ipanova: i see your point, but i don't think that difficulty of testing the function should influence/dictate of how the actual function should be written/look like.
I would like hear also someone else opinion about this, because due to the fact you've slitted the checks into multiple functions they make repeated calls to the db.

asmacdo: From a general perspective, I usually prefer a function to be smaller (and more testable) because it is usually more readable.

That holds in this case with the exception of the  `check_force_full` and `check_no_previous_sync` since they are a single line. In this function, I think this would be more readable with less code by removing the two single line functions.

    force_full = call_config.get('force_full', False)
    not_synced = conduit.last_sync() is None

I would classify this as a personal style choice, and would be fine if you leave it as is since it is very clear what is happening.

server/pulp/server/controllers/repository.py
asmacdo: This makes sense to me. With these 2 lines removed, every time --force-full is passed, the *next 2* syncs would be full, an amusing edge case. Nice catch @dralley!

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +816,149 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+    :param repo_id:     identifies the repo to sync
+    :type  repo_id:     str
+    :param call_config: Plugin Call Configuration
+    :type  call_config: pulp.plugins.config.PluginCallConfiguration
+
+    :return:  Whether the override config is different from the override config on
+    the previous sync.
+    :rtype:   boolean
+    """
+    repo_importer = model.Importer.objects.get_or_404(repo_id=repo_id)
```
asmacdo: I am not too worried about repeated db calls since this occurs at the start of a sync (instead of say, for each package in a sync), but you *could* reduce the number of call by including the repo_importer as a param.

server/pulp/server/controllers/repository.py
mhrivnak: When a user does a sync with ``force_full``, then another sync without it, it is clear that the second time, the user does not want a full sync. I don't think it makes sense to do one then, so I agree with @dralley 's use of ``pop``.

docs/user-guide/release-notes/2.11.x.rst
dralley: I will fix it in my next push.

server/pulp/server/controllers/repository.py
ipanova: Agree.

docs/user-guide/release-notes/2.11.x.rst
```diff
@@ -0,0 +1,12 @@
...
+Pulp 2.11 Release Notes
+=======================
+
+Pulp 2.11.0
+===========
+
+New Features
+------------
+
+* For RPM content, a full sync will be forced if the sync configuration has been changed or content
```
ipanova: by the way, do we plan to extend this to other plugins (where it makes sense)?

server/pulp/server/controllers/repository.py
```diff
@@ -767,6 +767,11 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
         raise
 
     else:
+        # Update the override config if it has changed
+        if check_override_config_change(repo_id, call_config):
```
ipanova: maybe also add the explanation of why the raw pymongo is used here "# Use raw pymongo not to fire the signal hander"

ipanova:
@dralley i finished the review, there are couple of minor suggestions that do not require any re-review.
just take into account this https://github.com/pulp/pulp/pull/2637#issuecomment-238852194, fix the tests and lgtm! nice job!

docs/user-guide/release-notes/2.11.x.rst
dralley: I believe that was part of the rationale for doing this (although it isn't really explicitly stated on the redmine issue).  I won't be able to get to that since my internship ends tomorrow however.

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +820,124 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
     return TaskResult(sync_result, spawned_tasks=spawned_tasks)
 
 
+def check_unit_removed_since_last_sync(conduit, repo_id):
+    """
+    Checks whether the a content unit has been removed since the last_sync timestamp.
```
mhrivnak: s/the //

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +820,124 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+
+    :param conduit: allows the plugin to interact with core pulp
+    :type  conduit: pulp.plugins.conduits.repo_sync.RepoSyncConduit
+    :param repo_id: identifies the repo to sync
+    :type  repo_id: str
+
+    :return:  Whether a content unit has been removed since the last_sync timestamp
+    :rtype:   boolean
```
mhrivnak: s/boolean/bool/

server/pulp/server/controllers/repository.py
```diff
@@ -816,6 +820,124 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
...
+    """
+    Checks whether the config has been changed since the last sync occurred.
+
+    :param conduit: allows the plugin to interact with core pulp
+    :type  conduit: pulp.plugins.conduits.repo_sync.RepoSyncConduit
+    :param repo_id: identifies the repo to sync
+    :type  repo_id: str
+
+    :return:  Whether the config has been changed since the last sync occurred.
+    :rtype:   boolean
```
mhrivnak: s/boolean/bool/

PR #2648, opened 2016-07-20
As a user, units which don't pass the signature check are not imported

ipanova:
closes #1991


ipanova:
https://github.com/pulp/pulp_rpm/pull/938/

PR #2649, opened 2016-07-20
The pulp_streamer now uses the plugin-wide proxy config

jeremycline:
An importer has a configuration file in /etc/pulp/server/plugins.conf.d/
that can, among other things, configure the proxy used for the plugin.
Previously, the streamer was only using the importer configuration that
is present on the ``Importer`` module.

Although the most sensible thing to do is have the importer
configuration available on the Importer model object, this would require
a great deal of cleanup since importer/distributor models and
configuration are a mess of old and new, with the configuration spread
all over the place. This should be cleaned up as part plugin API work we
do for Pulp.

closes #2038


jeremycline:
ok test

mhrivnak:
Love the ratio of comments to code. ;)

PR #2650, opened 2016-07-21
Add a support policy section to the documentation

jeremycline:
I've placed this change in the introduction because I didn't see a better section and it seemed like a place most people would look, but I'm interested in other suggestions. 

I also considered adding something in there about the versions of Pulp itself we support (something like there's currently no LTS and the latest stable build is typically the only build to receive z streams). Does anyone have thoughts on that one way or the other?

### DIFF
--- a/docs/user-guide/introduction.rst
+++ b/docs/user-guide/introduction.rst
@@ -48,3 +48,35 @@ you are interested. You can find all of our documentation at `our docs page <htt
 
 Many examples require the use of a type, and for those we will use "rpm". However,
 examples in this guide will only cover features that are common across content types.
+
+.. _platform-support-policy:
+
+Platform Support Policy
+-----------------------
+
+CentOS/Red Hat Enterprise Linux (RHEL)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Pulp team aims to provide RPM builds for the last two major releases of RHEL. For
+example, if the latest version is RHEL 8, builds should be available for RHEL 8 and
+RHEL 7. However, due to the long life cycle of these platforms and the relatively rapid
+development cycle of both Pulp and its dependencies, we may need to stop supporting
+the older platform before the next major release of RHEL. If this does occur, announcements
+will be made well in advance.
+
+.. note::
+  It is likely that support for Red Hat Enterprise Linux 6 will be removed prior to
+  the release of Red Hat Enterprise Linux 8. This is due to the difficulty of
+  supporting both Python 2.6 and Python 3 at the same time.
+
+Currently we support Red Hat Enterprise Linux 6 and 7 as well as CentOS 6 and 7.
+
+
+Fedora
+^^^^^^
+
+Pulp follows the `Fedora release life cycle
+<https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle#Maintenance_Schedule>`_.
+This means that when a release of Fedora reaches its End of Life, Pulp will no longer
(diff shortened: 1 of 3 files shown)


docs/user-guide/introduction.rst
```diff
@@ -48,3 +48,21 @@ you are interested. You can find all of our documentation at `our docs page <htt
...
 Many examples require the use of a type, and for those we will use "rpm". However,
 examples in this guide will only cover features that are common across content types.
+
+Platform Support Policy
+-----------------------
+
+CentOS/Red Hat Enterprise Linux (RHEL)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Pulp team provides RPM builds for the last two major releases of RHEL. For
```
mhrivnak: I don't think we reached consensus on this, and I still don't think we need this policy. Before el7, we only deployed on el6 because el5 was too old and difficult to support. We are also seriously considering dropping el6 support soon. Foreman has already announced deprecation of el6, which will of course apply to katello. Adding this policy would prevent us from following suit.

https://theforeman.org/manuals/1.12/index.html#Releasenotesfor1.12

"EL6 is now deprecated - while fully supported in 1.12, please migrate to EL7 for future releases."

Retaining the flexibility to drop support for an EL* at whatever time makes the most sense is valuable, and it only happens every few years.

mhrivnak:
The Fedora part looks good to me, but I'd like to remove or substantially change the EL section. I'd be happy saying we support at least the latest EL X release, but I'm not sure how much value there is in declaring that officially. My inclination is to just not have a policy on X releases, say we only support the latest Y release for any X releases we do support, and see the docs for which X releases are currently supported.

seandst:
We mention supporting things like el5 and fedora 22 elsewhere in the docs. It would be good to see those references removed (or perhaps altered to point to this new support statement rather than name explicit versions) with this PR.

bmbouter:
Slightly unrelated to this but should this PR also include the removal of the EL5 and CentOS5 consumer support of the [Support Operating Systems](http://docs.pulpproject.org/user-guide/installation/index.html#supported-operating-systems) area?

bmbouter:
@seandst Yes maybe just have the [Supported Operating Systems](http://docs.pulpproject.org/user-guide/installation/index.html#supported-operating-systems) area say "Fedoras" or "Current Fedoras" and have that be a link to the Fedora policy you've written.

jeremycline:
@mhrivnak I've updated the RHEL verbiage, let me know what you think. 

@seandst Good catch, I'll go fix that now.

docs/user-guide/introduction.rst
```diff
@@ -48,3 +48,35 @@ you are interested. You can find all of our documentation at `our docs page <htt
...
+The Pulp team aims to provide RPM builds for the last two major releases of RHEL. For
+example, if the latest version is RHEL 8, builds should be available for RHEL 8 and
+RHEL 7. However, due to the long life cycle of these platforms and the relatively rapid
+development cycle of both Pulp and its dependencies, we may need to stop supporting
+the older platform before the next major release of RHEL. If this does occur, announcements
+will be made well in advance.
+
+.. note::
+  It is likely that support for Red Hat Enterprise Linux 6 will be removed prior to
+  the release of Red Hat Enterprise 8. This is due to the difficulty of supporting
```
mhrivnak: s/8/Linux 8/

mhrivnak:
This looks fine to me. Thanks.

jeremycline:
I've fixed the typo and squashed. Any other thoughts or comments?

PR #2652, opened 2016-07-21
Add dev setup instructions for SSHFS

jeremycline:
SSHFS is an alternative to NFS when sharing directories between the host
and guest. This documents the approach in the "Advanced" section since
the package is not yet in Fedora and requires a COPR.


pronix:
if nfs not fit, sshfs to slow : 9p is option :)

PR #2659, opened 2016-07-28
Update urls syntax for Django 1.10

asmacdo:
Also adds a compat_urls.py file to support Django versions older than
1.6.

closes #1766


asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

PR #2660, opened 2016-07-28
Adds logging for AMQP errors during /status/ API call

bmbouter:
Previously errors were caught and not logged. This makes
it very difficult to determine the root cause when an
error is experienced.

### DIFF
--- a/server/pulp/server/managers/status.py
+++ b/server/pulp/server/managers/status.py
@@ -1,12 +1,15 @@
 """
 Manager to return status information about a running Pulp instance
 """
-
+from logging import getLogger
 from pkg_resources import get_distribution
 
 from pulp.server.async.celery_instance import celery
 from pulp.server.db import connection
 from pulp.server.db.model import Worker
+
+
+_logger = getLogger(__name__)
 
 
 def get_version():
@@ -55,6 +58,7 @@ def get_broker_conn_status():
         conn.connect()
         conn.release()
         return {'connected': True}
-    except:
+    except Exception:
+        _logger.exception('Connection to broker failed during status check!')
         # if the above was not successful for any reason, return False
         return {'connected': False}


server/pulp/server/managers/status.py
```diff
@@ -55,6 +58,7 @@ def get_broker_conn_status():
         conn.connect()
         conn.release()
         return {'connected': True}
-    except:
+    except Exception, e:
+        _logger.exception(e)
```
seandst: My understanding of `logger.exception` is that it includes exc_info in addition to your message, so I think this might be redundant. Something like `_logger.exception('Connection to broker failed!')` might be better.

seandst: Re-reading these helped me out: 
https://docs.python.org/2/library/logging.html#logging.Logger.debug
https://docs.python.org/3/library/logging.html#logging.Logger.debug

In py3, there's a line reading "If an exception tuple (in the format returned by sys.exc_info()) _or an exception instance is provided_" (emphasis mine). In py2, that sentence reads line is "If an exception tuple (in the format returned by sys.exc_info()) is provided"; the "or an exception instance" bit is not there.

server/pulp/server/managers/status.py
```diff
@@ -55,6 +58,7 @@ def get_broker_conn_status():
         conn.connect()
         conn.release()
         return {'connected': True}
-    except:
+    except Exception, e:
```
mhrivnak: Just to make the py3 transition easier when it comes, it's good to start using the ``except Exception as e:`` syntax.

bmbouter:
I've adjusted the PR and will push a new version now. Here is an example of the logging it provides.

```
pulp.server.managers.status:ERROR: (6340-96384) Connection to broker failed during status check!
pulp.server.managers.status:ERROR: (6340-96384) Traceback (most recent call last):
pulp.server.managers.status:ERROR: (6340-96384)   File "/home/vagrant/devel/pulp/server/pulp/server/managers/status.py", line 60, in get_broker_conn_status
pulp.server.managers.status:ERROR: (6340-96384)     raise RuntimeError('This is a simulated error')
pulp.server.managers.status:ERROR: (6340-96384) RuntimeError: This is a simulated error
```

PR #2661, opened 2016-07-28
2077 - publish fails if Django Template Syntax in changelog

BrnoPCmaniak:
Added pulp.server.webservices.templatetags to include verbatim tag

re #2077
https://pulp.plan.io/issues/2077

### DIFF
--- a/server/pulp/server/webservices/templatetags/__init__.py
+++ b/server/pulp/server/webservices/templatetags/__init__.py
@@ -1,0 +1,13 @@
+"""Pulp backport of verbatim tag from Django 1.5.
+
+For more info take a look at pulp.server.webservices.templatetags.templatetags.
+
+This Package loads verbatim tag only if version is smaller than 1.5  and have version 1
+eg. 1.0 -> 1.4 - verbatim tag is designed for Django 1.4.
+On Django 1.3 - The tag does not support named verbatim blocks and therefore escaping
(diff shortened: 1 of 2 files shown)


server/pulp/server/webservices/templatetags/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+This Package loads verbatim tag only if version is smaller than 1.5  and have version 1
+eg. 1.0 -> 1.4 - verbatim tag is designed for Django 1.4.
+On Django 1.3 - The tag does not support named verbatim blocks and therefore escaping
+{% verbatim %} tags inside.
+"""
+import django
+
+if django.VERSION[0] == 1 and django.VERSION[1] < 5:
+    import templatetags  # NOQA
```
seandst: This makes it impossible to implement other template tags in the pulp 2 line, but that's probably fine since:
1. We aren't going to implement other template tags in the pulp 2 line
2. If I'm wrong, and we *are* going to implement other template tags, then we can move this conditional into the templatetags module as-needed to inject the verbatim tag support. Based on the docs and comments here, it should be immediately obvious to a developer adding new tags what needs to happen (in my opinion).

goosemania:
ok test

goosemania:
Re-open this PR against 2.9-dev

goosemania:
New PR #2673

PR #2662, opened 2016-07-29
Add consumer models.

jortel:
https://pulp.plan.io/issues/2085

The following fields were not carried forward to 3.0.

Consumer:
- **capability**  *supported Agent but never used*
- **rsa_pub**  *supported Agent*

Binding:
- **binding_config**  *supported Nodes*
- **notify_agent**  *supported Agent*
- **consumer_actions** *supported Agent*
- **deleted** *supported Agent*

The applicability models are not included in anticipation that a strong relational model will support calculation on demand.  If this proves to be incorrect, we can add them later.

Against the advice of @seandst, I decided to import classes from `django.db.models` instead of importing the `models` module.  His assertion was that is was django convention to import the module and prefix everything with `modules.`.  I find this really annoying, pointless and tedious to both type and read `models.` prefixing nearly everything.  Seeing it this way in documentation does not necessarily make it a standard and could just be a reflection of the writer's preference or done for illustration.  Sorry, I need to be convinced.  Thoughts?

On consumer, I initially wrote `bind()` and `unbind()` convenience methods.  Then, after further consideration, I removed them because each of them contained a single line of code that the caller could also do.  Also, It seemed inappropriate for a model method to operation on another model.  Thoughts?

The `ConsumerDistributor` association replaces the 2.x `Bind`.

I took a stab at convention for docstrings.  I'm not a fan of ReST markup (especially since we don't generate code documentation) but wanted to be consistent with our current practices.


platform/__init__.py
```diff
@@ -0,0 +1 @@
+
```
seandst: This file should not exist.

platform/pulp/platform/models/consumer.py
```diff
@@ -0,0 +1,67 @@
+"""
+Django models related to content consumers.
+"""
+
+from django.contrib.contenttypes.fields import GenericRelation
+from django.db.models import TextField, ForeignKey, ManyToManyField, CASCADE
```
seandst: By convention (Django convention, at least), this should be `from django.db import models`. I'd like this to be our convention in Pulp as well, unless there are strong objections. This convention can been seen throughout Django's documentation, and it would be good to have our models look as much like "normal" documented Django models as possible.

platform/pulp/platform/models/consumer.py
```diff
@@ -0,0 +1,67 @@
...
+
+from pulp.platform.models import Model, Notes
+from pulp.platform.models.repository import Distributor
```
seandst: For consistency, Models should always be imported from `pulp.platform.models`, or from their actual location. So, this should either be:
```python
from pulp.platform.models import Model, Notes, Distributor
```
or
```python
from pulp.platform.models.base import Model
from pulp.platform.models.generic import Notes
from pulp.platform.models.repository import Distributor
```

We should decide which we prefer, commit to it, and get it documented in the translation guide.

platform/pulp/platform/models/consumer.py
```diff
@@ -0,0 +1,67 @@
...
+    :type: TextField
+
+    Relations:
+
+    :cvar notes: Arbitrary information about the consumer.
+    :type notes: GenericRelation
+
+    :cvar distributors: Associated distributors.
+    :type distributors: ManyToManyField
+    """
```
seandst: You'd mentioned documenting cvars and ivars explicitly, is that still something you plan to do, or did it not work?

platform/__init__.py
jortel: You're right.  Not sure how that happened.  Pycharm magic?  Removing it.

platform/pulp/platform/models/consumer.py
jortel: Agreed.  I vote for importing from the package:
```
from pulp.platform.models import Model, Notes, Distributor
```

... just missed it when adding the new models.

platform/pulp/platform/models/consumer.py
pcreech: I'm favor using a "from django.db import models" syntax as well, but not because it's what django does.

The reasons I see for this are:

- 
less maintenance of the Import statement if we need to use another item from "django.db.models"
- 
we maintain context about what we are using.  In many places it can be frustrating to identify where something came from in code without having to scroll back to the import statements.  Having the name-spaced context in front of what is used can help. (simple example: ceil() vs math.ceil() )
 
While this introduce some more up-front cost and can be considered annoying, pointless and tedious, I find it to be of benefit in the future and worth the investment if someone not as familiar with this code can have an easier time understanding it and working with it.

PR #2645, opened 2016-07-18
Update the development guide to point users to pulp/devel

jeremycline:
All our development-related tools and scripts are moving to pulp/devel,
including things like our Vagrantfile. This updates the documentation to
reflect that.

It also removes the section on vagrant-cachier, which does not work
with dnf. Instead, it recommends the new method of caching packages.

PR #2664, opened 2016-07-29
Added ansible playbooks for install

BrnoPCmaniak:
Added ansible playbooks for installing server, admin and consumer + agent

### DIFF
--- a/docs/user-guide/installation/index.rst
+++ b/docs/user-guide/installation/index.rst
@@ -61,7 +61,6 @@ Prerequisites
   `serious limitations <http://docs.mongodb.org/manual/faq/fundamentals/#what-are-the-32-bit-limitations>`_
   on 32-bit architectures. It is strongly advised that you run MongoDB on a 64-bit architecture.
 
-
 Message Broker
 ^^^^^^^^^^^^^^
 
@@ -106,6 +105,9 @@ directory vary depending on how much content you wish to download.
 
 Instructions
 ------------
+
+Instalations below are production ready. That means if you want to only try pulp you can use
+our ansible playbook
 
 Please refer to the instructions below for your operating system.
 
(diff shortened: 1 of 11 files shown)


seandst:
Despite the lack of any documentation to support this, the playbooks in the pulp/pulp repo are deprecated. These changes should be submitted to the pulp/devel repo.

seandst:
I lied! The docs were updated in https://github.com/pulp/pulp/pull/2645

BrnoPCmaniak:
Oh I have not noticed. But these scripts does not need to be development only. I primary write them to make easier installation for those which want begin with pulp for my another story of making docs easier for newcomers So I'm not sure if devel repo is the right one for them.

seandst:
The devel repo exists in large part to deal with this problem (despite the name), which is that we have playbooks in the pulp repo and in the pulp_packaging repo, with a lot of common code. I haven't looked too closely, but chances are pretty good that there are duplicated tasks between what's here and what's in the pulp_packaging repo, for example (@elyezer would be able to say for sure).

The bottom line is that whether this gets merged or not we're `rm -rf`ing the ansible playbooks in playpen in this repo, so this should be closed.

elyezer:
We have similar playbook code on pulp_packaging, the playbook available there make sure to deploy a ready to be tested Pulp server with all optional bits that Pulp Smash uses. I have not taken time to take a close look to the playpen directory here, but as @seandst mentioned we should be duplicating a lot of playbook code.

I am wondering if makes sense to have the playbooks on a separated repo and maybe we can link here and pulp_package to it. The point is, we need some centralized place to have the roles code and then each playbook can make use of the roles it needs to do its job.

jeremycline:
I'm all for having a repository dedicated to Ansible roles. I started merging the ``pulp_packaging`` and ``devel`` roles before I went on PTO. I'd like to finish that up, it's just a question of whether we have a dedicated repository for it or put in ``devel``.

Is anyone _not_ in favor of a repository for Ansible roles?

PR #2665, opened 2016-07-31
Adds predistributor logic to the publish controller

dkliban:
If a distributor has a predistributor_id configured, the publish is skipped when the
predistributor has not published since the last publish by the distributor. This includes
situations where the predistributor has not published at all.

re #1887
https://pulp.plan.io/issues/1887

### DIFF
--- a/server/pulp/server/db/model/repository.py
+++ b/server/pulp/server/db/model/repository.py
@@ -1,3 +1,4 @@
+from gettext import gettext as _
 import traceback as traceback_module
 
 from pulp.server.db.model.base import Model
@@ -245,7 +246,7 @@ class RepoPublishResult(Model, ReaperMixin):
 
     @classmethod
     def skipped_result(cls, repo_id, distributor_id, distributor_type_id, started, completed,
-                       result_code):
+                       result_code, translated_message):
         """
         Creates a new history entry for a skipped publish.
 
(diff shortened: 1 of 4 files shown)


server/pulp/server/db/model/repository.py
```diff
@@ -266,11 +267,15 @@ def skipped_result(cls, repo_id, distributor_id, distributor_type_id, started, c
...
+
+        @param message: reason for skipping
+        @type message: str
+
         """
 
         r = cls(repo_id, distributor_id, distributor_type_id, started, completed,
                 cls.RESULT_SKIPPED)
-        message = 'Skipped. Nothing changed since last publish'
+        message = _('Skipped: %(reason)s' % {'reason': message})
```
bmbouter: The variable replacement is happening before the call to `_()`, but it needs to happen afterwards.

server/pulp/server/db/model/repository.py
```diff
@@ -266,11 +267,15 @@ def skipped_result(cls, repo_id, distributor_id, distributor_type_id, started, c
 
         @param result_code: one of the RESULT_* constants in this class
         @type  result_code: str
+
+        @param message: reason for skipping
```
bmbouter: This needs to indicate that it has already been internationalized. Maybe putting that explicitly here in the docstring and changing `message` to be `translated_message`.

dkliban:
ok test

dkliban:
ok test

bmbouter:
@dkliban I just did a round of testing against the latest and functionally it all worked. Since the test runner here is being strange I'm going to run the tests locally.

bmbouter:
All tests pass for me locally, and all comments have been addressed. I'm adding the LGTM.

PR #2666, opened 2016-08-01
Speeds up the unit file path migrations.

mhrivnak:
Removes the empty directory purge phase of the migrations, which was taking
some users many hours when done over NFS.

Introduces multi-threaded concurrency for the bulk of the migration's work.

https://pulp.plan.io/issues/2118
fixes #2118


mhrivnak:
Rebasing and resubmitting on master.

PR #2668, opened 2016-08-01
Removes ability to use password authentication when rsync-ing

dkliban:
re #1887
https://pulp.plan.io/issues/1887
re #1963
https://pulp.plan.io/issues/1963
re #1759
https://pulp.plan.io/issues/1759

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -380,7 +380,6 @@ Requires: mod_xsendfile >= 0.12
 Requires: m2crypto
 Requires: genisoimage
 Requires: kobo
-Requires: sshpass
 # RHEL6 ONLY
 %if 0%{?rhel} == 6
 Requires: nss >= 3.12.9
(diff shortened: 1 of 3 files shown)


bmbouter:
@dkliban Should we remove `sshpass` as a dependency? I forget if its `sshpass` or not, but I remember there was some dependency that was added for this. Can that also be removed?

dkliban:
We should!!!

bmbouter:
@dkliban Thanks I see the change.

bmbouter:
I did some hand testing and the configs validate as expected. Also some hand testing worked fine. There was one unrelated issue observed but we are handling that as a separate issue since it's unrelated to this PR.

bmbouter:
@dkliban Also the test failure looks unrelated so I'm going to add the LGTM now.

PR #2669, opened 2016-08-01
Add consumer models.

jortel:
https://pulp.plan.io/issues/2085

The following fields were not carried forward to 3.0.

Consumer:
- **capability**  *supported Agent but never used*
- **rsa_pub**  *supported Agent*

Binding:
- **binding_config**  *supported Nodes*
- **notify_agent**  *supported Agent*
- **consumer_actions** *supported Agent*
- **deleted** *supported Agent*

The applicability models are not included in anticipation that a strong relational model will support calculation on demand.  If this proves to be incorrect, we can add them later.

On consumer, I initially wrote `bind()` and `unbind()` convenience methods.  Then, after further consideration, I removed them because each of them contained a single line of code that the caller could also do.  Also, It seemed inappropriate for a model method to operation on another model.  Thoughts?

### DIFF
--- a/platform/pulp/platform/models/__init__.py
+++ b/platform/pulp/platform/models/__init__.py
@@ -1,3 +1,6 @@
 # https://docs.djangoproject.com/en/1.8/topics/db/models/#organizing-models-in-a-package
 from .base import Model, MasterModel  # NOQA
 from .generic import GenericRelationModel, GenericKeyValueStore, Config, Notes, Scratchpad  # NOQA
+
+from .consumer import Consumer, ConsumerContent  # NOQA
+from .repository import Distributor  # NOQA
--- a/platform/pulp/platform/models/consumer.py
+++ b/platform/pulp/platform/models/consumer.py
@@ -1,0 +1,50 @@
+"""
+Django models related to content consumers.
+"""
+
+from django.contrib.contenttypes import fields
+from django.db import models
+
+from pulp.platform.models import Model, Notes
+
+
+class Consumer(Model):
+    """
+    A content consumer.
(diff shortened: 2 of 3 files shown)


platform/pulp/platform/models/__init__.py
```diff
@@ -1,3 +1,5 @@
 # https://docs.djangoproject.com/en/1.8/topics/db/models/#organizing-models-in-a-package
 from .base import Model, MasterModel  # NOQA
+from .consumer import Consumer, ConsumerContent  # NOQA
 from .generic import GenericRelationModel, GenericKeyValueStore, Config, Notes, Scratchpad  # NOQA
+from .repository import Distributor  # NOQA
```
jortel: I don't think the `#NOQA` is needed here.

pcreech: Got this when removing it:

"platform/pulp/platform/models/__init__.py:5:1: F401 '.repository.Distributor' imported but unused"

It appears it is needed

pcreech:
> Also, It seemed inappropriate for a model method to operation on another model. Thoughts?

I don't recall the exact methods from before, but if the consumer model was modifying data on another model, then yes, that seems inappropriate.

So far, LGTM, but would like for @seandst give the final LGTM, as he had some lingering questions before the other PR got closed out.

platform/pulp/platform/models/__init__.py
jortel: Unfortunate that flake8 isn't smart enough to understand imports in __init__.py used to define a package interface :/

PR #2676, opened 2016-08-02
build compat templates for Django 1.4-1.10

asmacdo:
https://pulp.plan.io/issues/2049

### DIFF
--- a/server/pulp/server/content/web/settings.py
+++ b/server/pulp/server/content/web/settings.py
@@ -47,7 +47,6 @@ LOGGING = {
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 SECRET_KEY = 'I_am_a_secret_that_is_never_used_meaningfully_by_pulp'
 DEBUG = False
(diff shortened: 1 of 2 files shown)


seandst:
Out of curiosity, is the version-blocking necessary to prevent gnarly DeprecationWarnings (or something?). If not, is it reasonable to have `TEMPLATES` alongside `TEMPLATE_DIRS` and `TEMPLATE_DEBUG` in the same evaluated settings file unconditionally, and have the values defined in one place and referenced from the other?

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
@seandst, yes that works!

server/pulp/server/content/web/settings.py
```diff
@@ -2,6 +2,8 @@
 import logging
 import os
 
+import django
```
dkliban: why is this import needed?

asmacdo:
ok test

PR #2677, opened 2016-08-02
Fix Qpid Transport file descriptor leak in /status/

bmbouter:
For the Qpid transport this will release some file
descriptors earlier than when garbage collection will
occur. For other transports this should also be save
because the `__del__()` call occurs after the release()
call.

The upstream fix is still being discussed. Perhaps a
more comprehensive fix will land upstream. If one does
this can be adjusted then.

closes #2124
https://pulp.plan.io/issues/2124

### DIFF
--- a/server/pulp/server/managers/status.py
+++ b/server/pulp/server/managers/status.py
@@ -9,6 +9,9 @@ from pulp.server.db import connection
 from pulp.server.db.model import Worker
 
 
+conn = celery.connection()
+
+
 def get_version():
     """
     :returns:          Pulp platform version
@@ -51,9 +54,14 @@ def get_broker_conn_status():
     # explicit connection and then release it.
     # See https://github.com/celery/kombu/issues/432 for more detail.
     try:
-        conn = celery.connection()
         conn.connect()
+        transport = conn._transport
         conn.release()
+
+        # An explicit call to __del__ to release file descriptors
+        # https://pulp.plan.io/issues/2124#note-9
+        transport.__del__()
+
         return {'connected': True}
     except:
         # if the above was not successful for any reason, return False


server/pulp/server/managers/status.py
```diff
@@ -51,9 +54,14 @@ def get_broker_conn_status():
...
     # See https://github.com/celery/kombu/issues/432 for more detail.
     try:
-        conn = celery.connection()
         conn.connect()
+        transport = conn._transport
         conn.release()
+
+        # An explicit call to __del__ to release file descriptors
+        # https://pulp.plan.io/issues/2124#note-9
+        transport.__del__()
```
seandst: Does it make sense to throw this in a `finally:` block, or is there really no way ever that the connection will fail and we'll have file descriptors to clean up?

bmbouter: I was talking in `#python` and they explained that I should never expect the code in `__del__()` to be run. I think that is where we're going wrong with these file descriptor leaks. There is an improvement to be made here, but I think it's in kombu by adjusting the way the kombu.Connection.Transport closes itself [here](https://github.com/celery/kombu/blob/4401770d8b007ae164a833450d4b27252b58c3d6/kombu/connection.py#L309). This should have some sort of callback to release resources. For instance there is a `collect()` method [here](https://github.com/celery/kombu/blob/4401770d8b007ae164a833450d4b27252b58c3d6/kombu/connection.py#L313-L334), but I'm not sure if its appropriate for us to call that here instead or what that does for other cases of connection error.

Aside from this, observing the file descriptor count when there is a connection error would be illustrative also. Maybe I will try that too.

seandst: Yeah, the tl;dr of python for me on this has always been "just don't bother implementing `__del__`, and talk anyone you see using it out of it."

I think this is a tricky problem, and whatever you come up with to improve on this is probably fine. LGTM, but if you want to take a few minutes and chat about it I'll be available. :)

seandst: One thought that *just* occured to me, though, is that if upstream takes `__del__` away, then we'll `AttributeError` here. It is work guarding this call against the most likely upstream change, which is hopefully to stop doing anything that depends on `__del__` being called? :)

bmbouter:
Upon further testing I don't think this is a good fix for this issue. I'm going to be posting more info back on the bug about how this will be fixed which will include a new release for kombu.

PR #2679, opened 2016-08-03
Fixes Qpid file descriptor leak

bmbouter:
This fix is already in upstream Kombu so this ports
the important part of the fix to downstream

This commit bumps python-kombu to 3.0.33-6

https://pulp.plan.io/issues/2124
closes #2124

### DIFF
--- a/deps/external_deps.json
+++ b/deps/external_deps.json
@@ -41,8 +41,8 @@
     },
     {
         "name": "python-kombu",
-        "version": "3.0.33-5.pulp",
-        "platform": [ "el6", "el7", "fc22", "fc23"]
+        "version": "3.0.33-6.pulp",
+        "platform": [ "el6", "el7", "fc23"]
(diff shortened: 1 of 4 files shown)


seandst:
kombu's unit tests fail with this change when attempting a koji scratch build. I've sent more details to @bmbouter via email.

bmbouter:
I will fix the unit tests. Thanks @seandst !

bmbouter:
I pushed a new version which has fixed unit tests, but the hunk still doesn't apply correctly. I'll fix that next with another push.

bmbouter:
I resolved the patch apply issue. Here is a completed scratch build.

http://koji.katello.org/koji/taskinfo?taskID=508896

bmbouter:
non-scratch builds have been made for F23, EL6, and EL7 so this is now safe to merge.

PR #2684, opened 2016-08-05
Add repository models.

jortel:
https://pulp.plan.io/issues/2089

Not sold on `RepositoryContent.save()` and `RepositoryContent.delete()` managing the `last_content_added` and `last_content_removed`.  As a rule, we don't have one model update another.  *Maybe* it's okay in cases like this where the model is an *association* (join tables)?  Perhaps managing the `last_content_*` should be in the control layer?

Ignore the `Content` model (for now).

Anyone know if `RepositoryGroup.scratchpad` is need?  Used?

### DIFF
--- a/platform/pulp/platform/models/base.py
+++ b/platform/pulp/platform/models/base.py
@@ -51,7 +51,7 @@ class MasterModel(Model):
         subclass.
 
     """
-    type = models.CharField(max_length=63)
+    detail_model = models.CharField(max_length=63)
 
     objects = MasterModelManager()
 
@@ -64,8 +64,8 @@ class MasterModel(Model):
         # name. That name is what I'm using here to determine the value of
         # type. Storing type in a column on the MasterModel next to makes it trivial
         # to filter for specific detail model types across model relations.
-        if not self.type:
-            self.type = self._meta.model_name
+        if not self.detail_model:
+            self.detail_model = self._meta.model_name
         return super(MasterModel, self).save(*args, **kwargs)
 
     def cast(self):
@@ -74,16 +74,16 @@ class MasterModel(Model):
         If this model is already an instance of its detail type, it will return itself.
         """
         # If this instance's type matches the current model name, it is already cast. Return it.
-        if self.type == self._meta.model_name:
+        if self.detail_model == self._meta.model_name:
             return self
         else:
             try:
                 # Otherwise, return the cast model attribute for this instance
-                return getattr(self, self.type)
+                return getattr(self, self.detail_model)
             except AttributeError:
                 # Unknown content type. The generic content type is as specific as
                 # we can get here. This is a great place to throw a log message about
-                # encountering an unmodelled type, such as a type from an uninstalled
+                # encountering an un-modelled type, such as a type from an uninstalled
                 # plugin.
                 return self
 
--- a/platform/pulp/platform/models/repository.py
+++ b/platform/pulp/platform/models/repository.py
@@ -1,10 +1,330 @@
 """
-Django models related to repositories.
+Repository related Django models.
 """
+from django.db import models
+from django.contrib.contenttypes import fields
+from django.utils import timezone
 
-from pulp.platform.models import Model
+from pulp.platform.models import Model, Notes, Scratchpad, MasterModel
 
 
-class Distributor(Model):
+class Content(Model):
     # placeholder
-    pass
+    type = models.TextField()
+    created = models.DateTimeField(auto_now_add=True)
+
+
+class Repository(Model):
+    """
+    Collection of content.
+
+    Fields:
+
+    :cvar name: The repository name.
+    :type name: models.TextField
+
+    :cvar: description: An optional description.
+    :type: models.TextField
+
+    :cvar last_content_added: When content was last added.
+    :type last_content_added: models.DateTimeField
+
+    :cvar last_content_removed: When content was last removed.
+    :type last_content_removed: models.DateTimeField
+
+    Relations:
+
+    :cvar scratchpad: Arbitrary information stashed on the repository.
+    :type scratchpad: fields.GenericRelation
+
+    :cvar notes: Arbitrary repository properties.
+    :type notes: fields.GenericRelation
+
+    :cvar content: Associated content.
+    :type content: models.ManyToManyField
+    """
+    name = models.TextField(db_index=True, unique=True)
+    description = models.TextField(blank=True)
+
+    last_content_added = models.DateTimeField(blank=True, null=True)
+    last_content_removed = models.DateTimeField(blank=True, null=True)
+
+    scratchpad = fields.GenericRelation(Scratchpad)
+    notes = fields.GenericRelation(Notes)
+
+    content = models.ManyToManyField('Content', through='RepositoryContent')
+
+    @property
+    def content_summary(self):
+        """
+        The contained content summary.
+
+        :return: A dict of {<type>: <count>}
+        :rtype:  dict
+        """
+        mapping = self.content.values('type').annotate(count=models.Count('type'))
+        return {m['type']: m['count'] for m in mapping}
+
+
+class RepositoryGroup(Model):
+    """
+    A group of repositories.
+
+    Fields:
+
+    :cvar name: The group name.
+    :type name: models.TextField
+
+    :cvar: description: An optional description.
+    :type: models.TextField
+
+    Relations:
+    :cvar notes: Arbitrary group properties.
+    :type notes: fields.GenericRelation
+
+    :cvar members: Repositories associated with the group.
+    :type members: models.ManyToManyField
+    """
+    name = models.TextField(db_index=True, unique=True)
+    description = models.TextField(blank=True)
+
+    members = models.ManyToManyField('Repository')
+    scratchpad = fields.GenericRelation(Scratchpad)
+    notes = fields.GenericRelation(Notes)
+
+
+class Plugin(MasterModel):
+    """
+    An Abstract model for plugins.
+
+    Fields:
+
+    :cvar name: The plugin name.
+    :type type: models.TextField
+
+    :cvar type: The plugin type.
+    :type type: models.TextField
+
+    :cvar last_updated: When the plugin was last updated.
+    :type last_updated: fields.DateTimeField
+
+    Relations:
+
+    """
+    name = models.TextField(db_index=True)
+    type = models.TextField(blank=False, default=None)
+    last_updated = models.DateTimeField(auto_now=True, blank=True, null=True)
+
+    class Meta:
+        abstract = True
+
+
+class Importer(Plugin):
+    """
+    A abstract content importer.
+
+    Fields:
+
+    :cvar feed_url: The URL of an external content source.
+    :type feed_url: models.TextField
+
+    :cvar validate: Validate the imported context.
+    :type validate: models.BooleanField
+
+    :cvar ssl_ca_certificate: A PEM encoded CA certificate used to validate the server
+                              certificate presented by the external source.
+    :type ssl_ca_certificate: models.TextField
+
+    :cvar ssl_client_certificate: A PEM encoded client certificate used for authentication.
+    :type ssl_client_certificate: models.TextField
+
+    :cvar ssl_client_key: A PEM encoded private key used for authentication.
+    :type ssl_client_key: models.TextField
+
+    :cvar ssl_validation: Indicates whether SSL peer validation must be performed.
+    :type ssl_validation: models.BooleanField
+
+    :cvar proxy_url: The optional proxy URL. Format: scheme://user:password@host:port
+    :type proxy_url: models.ForeignKey
+
+    :cvar basic_auth_user: The user used in HTTP basic authentication.
+    :type basic_auth_user: models.TextField
+
+    :cvar basic_auth_password: The password used in HTTP basic authentication.
+    :type basic_auth_password: models.TextField
+
+    :cvar max_download_bandwidth: The max amount of bandwidth used per download (Bps).
+    :type max_download_bandwidth: models.IntegerField
+
+    :cvar max_concurrent_downloads: The number of concurrent downloads permitted.
+    :type max_concurrent_downloads: models.IntegerField
+
+    :cvar download_policy: The policy for downloading content.
+    :type download_policy: models.TextField
+
+    :cvar last_sync: When the last successful synchronization occurred.
+    :type last_sync: models.DateTimeField
+
+    Relations:
+
+    :cvar scratchpad: Arbitrary information stashed by the importer.
+    :type scratchpad: fields.GenericRelation
+    """
+
+    # Download Policies
+    IMMEDIATE = 'immediate'
+    ON_DEMAND = 'on_demand'
+    BACKGROUND = 'background'
+    DOWNLOAD_POLICIES = (
+        (IMMEDIATE, 'Download Immediately'),
+        (ON_DEMAND, 'Download On Demand'),
+        (BACKGROUND, 'Download In Background'))
+
+    feed_url = models.TextField()
+    validate = models.BooleanField(default=True)
+
+    ssl_ca_certificate = models.TextField(blank=True)
+    ssl_client_certificate = models.TextField(blank=True)
+    ssl_client_key = models.TextField(blank=True)
+    ssl_validation = models.BooleanField(default=True)
+
+    proxy_url = models.TextField(blank=True)
+
+    basic_auth_user = models.TextField(blank=True)
+    basic_auth_password = models.TextField(blank=True)
+
+    max_download_bandwidth = models.IntegerField(null=True)
+    max_concurrent_downloads = models.IntegerField(null=True)
+
+    download_policy = models.TextField(choices=DOWNLOAD_POLICIES)
+    last_sync = models.DateTimeField(blank=True, null=True)
+
+    scratchpad = fields.GenericRelation(Scratchpad)
+
+    class Meta:
+        abstract = True
+        unique_together = ('name', 'repository')
+
+
+class RepositoryImporter(Importer):
+    """
+    A content importer that is associated with a repository..
+
+    Fields:
+
+    Relations:
+
+    :cvar repository: The associated repository.
+    :type repository: models.ForeignKey
+    """
+
+    repository = models.ForeignKey(
+        Repository, related_name='importers', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('name', 'repository')
+
+
+class Distributor(Plugin):
+    """
+    An abstract content distributor.
+
+    Fields:
+
+    :cvar relative_path: The (relative) path component of the published url.
+    :type relative_path: models.TextField
+
+    :cvar last_published: When the last successful publish occurred.
+    :type last_published: models.DateTimeField
+
+    Relations:
+
+    """
+    relative_path = models.TextField(blank=True)
+    last_published = models.DateTimeField(blank=True, null=True)
+
+    class Meta:
+        abstract = True
+
+
+class RepositoryDistributor(Distributor):
+    """
+    A content distributor that is associated with a repository.
+
+    Fields:
+
+    :cvar auto_publish: Indicates that the distributor may publish automatically
+        when the associated repository's content has changed.
+    :type auto_publish: models.BooleanField
+
+    Relations:
+
+    :cvar repository: The associated repository.
+    :type repository: models.ForeignKey
+
+    """
+    auto_publish = models.BooleanField(default=True)
+
+    repository = models.ForeignKey(
+        Repository, related_name='distributors', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('name', 'repository')
+
+
+class GroupDistributor(Distributor):
+    """
+    A content distributor that is associated with a repository group.
+
+    Fields:
+
+    Relations:
+
+    :cvar group: The associated repository group.
+    :type group: models.ForeignKey
+
+    """
+    group = models.ForeignKey(
+        RepositoryGroup, related_name='distributors', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('name', 'group')
+
+
+class RepositoryContent(Model):
+    """
+    Association between a repository and its contained content.
+
+    Fields:
+
+    :cvar created: When the association was created.
+    :type created: fields.DateTimeField
+
+    Relations:
+
+    :cvar content: The associated content.
+    :type content: models.ForeignKey
+
+    :cvar repository: The associated repository.
+    :type repository: models.ForeignKey
+    """
+    created = models.DateTimeField(auto_now_add=True)
+
+    content = models.ForeignKey('Content', on_delete=models.CASCADE)
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
+
+    def save(self, *args, **kwargs):
+        """
+        Save the association.
+        """
+        self.repository.last_content_added = timezone.now()
+        self.repository.save()
+        super(RepositoryContent, self).save(*args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        """
+        Delete the association.
+        """
+        self.repository.last_content_removed = timezone.now()
+        self.repository.save()
+        super(RepositoryContent, self).delete(*args, **kwargs)
--- a/platform/pulp/platform/models/__init__.py
+++ b/platform/pulp/platform/models/__init__.py
@@ -3,4 +3,6 @@ from .base import Model, MasterModel  # NOQA
 from .generic import GenericRelationModel, GenericKeyValueStore, Config, Notes, Scratchpad  # NOQA
 
 from .consumer import Consumer, ConsumerContent  # NOQA
-from .repository import Distributor  # NOQA
+from .repository import (Repository, RepositoryGroup, Importer, Distributor,  # NOQA
+                         RepositoryImporter,  RepositoryDistributor, GroupDistributor,  # NOQA
+                         RepositoryContent)  # NOQA
--- a/platform/pulp/platform/models/consumer.py
+++ b/platform/pulp/platform/models/consumer.py
@@ -29,10 +29,10 @@ class Consumer(Model):
     :type distributors: models.ManyToManyField
     """
     name = models.TextField(db_index=True, unique=True)
-    description = models.TextField(blank=True, default='')
+    description = models.TextField(blank=True)
 
     notes = fields.GenericRelation(Notes)
-    distributors = models.ManyToManyField('Distributor')
+    distributors = models.ManyToManyField('RepositoryDistributor')
 
 
 class ConsumerContent(Model):
--- a/platform/pulp/platform/tests/models/test_repository.py
+++ b/platform/pulp/platform/tests/models/test_repository.py
@@ -1,0 +1,114 @@
+"""
+These tests are intended to be a HOW TO for working with the repository motel.
+"""
+
+from django.test import TestCase
+
+from pulp.platform.models import Repository, RepositoryImporter, RepositoryDistributor
+
+
+class TestRepository(TestCase):
+
+    NAME = 'zoo'
+
+    def _create_repository(self):
+        """
+        Create a repository with sample notes and scratchpad.
+        """
+        repository = Repository(name=TestRepository.NAME)
+        repository.notes.mapping['organization'] = 'Engineering'
+        repository.notes.mapping.update({'age': 10})
+        repository.scratchpad.mapping['hello'] = 'world'
+        repository.save()
+
+    def _delete_repository(self):
+        """
+        Delete the repository.
+        """
+        repository = Repository.objects.filter(name=TestRepository.NAME)
+        repository.delete()
+
+    def _add_importer(self):
+        """
+        Add an importer with feed URL and some standard settings.
+        """
+        repository = Repository.objects.get(name=TestRepository.NAME)
+        importer = RepositoryImporter(repository=repository)
+        importer.name = 'Upstream'
+        importer.type = 'YUM'
+        importer.feed_url = 'http://content-world/everyting/'
+        importer.ssl_validation = True
+        importer.ssl_ca_certificate = 'MY-CA'
+        importer.ssl_client_certificate = 'MY-CERTIFICATE'
+        importer.ssl_client_key = 'MY-KEY'
+        importer.proxy_url = 'http://elmer:fudd@warnerbrothers.com'
+        importer.basic_auth_user = 'Elmer'
+        importer.basic_auth_password = 'Fudd'
+        importer.save()
+
+    def _add_distributor(self):
+        """
+        Add a distributor with some standard settings.
+        """
+        repository = Repository.objects.get(name=TestRepository.NAME)
+        distributor = RepositoryDistributor(repository=repository)
+        distributor.name = 'Public'
+        distributor.type = 'YUM'
+        distributor.auto_publish = True
+        distributor.save()
+
+    def setUp(self):
+        self._create_repository()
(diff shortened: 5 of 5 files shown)


platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
 """
-Django models related to repositories.
+Repository related Django models.
 """
+from datetime import datetime
 
-from pulp.platform.models import Model
+from django.db import models
```
mhrivnak: \o/

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+
+    Fields:
+
+    :cvar name: The consumer name.
+    :type name: models.TextField
+
+    :cvar: description: An optional description.
+    :type: models.TextField
+
+    :cvar type: The repository type.
```
mhrivnak: What are the choices? How would someone discover those choices, besides asking on a PR? ;)

If this corresponds to the type on the Plugin model, maybe there should be a more concrete relationship?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    description = models.TextField(blank=True, default='')
+    type = models.TextField()
+
+    last_content_added = models.DateTimeField(blank=True, null=True)
+    last_content_removed = models.DateTimeField(blank=True, null=True)
+
+    scratchpad = fields.GenericRelation(Scratchpad)
+    notes = fields.GenericRelation(Notes)
+
+    content = models.ManyToManyField('Content', through='RepositoryContent')
```
mhrivnak: This makes me so happy.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    @property
+    def content_summary(self):
+        """
+        The contained content summary.
+
+        :return: A dict of {<type>: <count>}
+        :rtype:  dict
+        """
+        mapping = self.content.values('type').annotate(count=models.Count('type'))
+        return {m['type']: m['count'] for m in mapping}
```
mhrivnak: Wasting no time in dropping python 2.6!

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    notes = fields.GenericRelation(Notes)
+
+
+class Plugin(Model):
+    """
+    Abstract model for plugins.
+
+    Fields:
+
+    :cvar type: The plugin name.
```
mhrivnak: s/type/name/

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    :type type: models.TextField
+
+    :cvar type: The plugin type.
+    :type type: models.TextField
+
+    :cvar last_updated: When the plugin was last updated.
```
mhrivnak: Is it true that this is specifically scoped to the last time the *config* was updated? If so, let's doc that here.

mhrivnak: Perhaps this should explicitly be documented as being UTC?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    """
+    A content importer associated with a repository..
+
+    Fields:
+
+    :cvar scratchpad: Arbitrary information stashed by the importer.
+    :type scratchpad: fields.GenericRelation
+
+    :cvar last_sync: When the last successful synchronization occurred.
+    :type last_sync: fields.GenericRelation
```
mhrivnak: ``DateTimeField`` here and a couple more places below

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+
+    :cvar repository: The associated repository.
+    :type repository: models.ForeignKey
+    """
+    created = models.DateTimeField(auto_now_add=True)
+
+    content = models.ForeignKey('Content', on_delete=models.CASCADE)
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
+
+    def save(self, *args, **kwargs):
```
mhrivnak: Does this method definitely get called if someone does ``myrepo.content.add(myunit)`` ? Or is that even possible given that this through-table exists? Same question for ``delete()`` below.

mhrivnak:
What is your plan for "Include thorough documentation about how data should be mapped from the mongo collection to the django model"? Including that documentation in or with a PR like this would make the review much easier.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+from django.contrib.contenttypes import fields
 
+from pulp.platform.models import Model, Notes, Scratchpad, Config
 
-class Distributor(Model):
+
+class Content(Model):
     # placeholder
-    pass
+    type = models.TextField()
```
asmacdo: Could we name this something else since `type` is a python reserved word?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    created = models.DateTimeField(auto_now_add=True)
+
+
+class Repository(Model):
+    """
+    Collection of content.
+
+    Fields:
+
+    :cvar name: The consumer name.
```
asmacdo: consumer?

platform/pulp/platform/models/repository.py
jortel: The explicit *through=* (join table) causes methods such as`content.add()` to raise an exception.  Those methods cannot be used.

platform/pulp/platform/models/repository.py
jortel: Class and instance attributes are namespaced to the class (as I'm sure you know) and implicitly apply to the class or instance.  The `type()` builtin function (not a reserved word) unfortunately exists in the global namespace.  I'm opposed to redundantly qualifying it.  Naming it something like `repository_name` does not add clarity and leads to awkward semantics like: `repository.repository_type`.

https://docs.python.org/2/library/functions.html#type

platform/pulp/platform/models/repository.py
jortel: cut-n-paste error.  good catch.  thanks.

jortel:
@mhrivnak, The plan for documenting the migration is to create a related issue in redmine.  Eg: https://pulp.plan.io/issues/2117.  I see the value in referencing the migration issue in the PR so the reviewer has this information.  Would linking migration issue be sufficient?

platform/pulp/platform/models/repository.py
jortel: I was expecting that we'd document that all timestamps are UTC instead or documenting on every field.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+        when the associated repository's content has changed.
+    :type auto_publish: fields.BooleanField
+
+    :cvar last_publish: When the last successful publish occurred.
+    :type last_publish: models.DateTimeField
+
+    Relations:
+
+    """
+    auto_publish = models.BooleanField(default=False)
```
asmacdo: Perhaps now is the time to move this to the distributor config.

mhrivnak:
@jortel that seems reasonable, although there is another use case I have in mind. There are two situations when a developer needs to know both the mapping of old fields to new and have some background info to help understand the changes. One is when writing the pulp 2->3 migrator, and the other is when refactoring the code that uses the models. That second situation is a very important one, and it is valuable to make sure info about model changes is readily accessible.

For that reason, I was thinking it might make sense to put this info in the model class doc blocks themselves. The downside of that is that the info will become less relevant as time goes by. Another option is to just have a document in the repo that captures the info.

In either case, I want to emphasize that understanding why certain changes took place and exactly how to use certain new fields or models will be very important to developers trying to use the django models, so capturing that somewhere that's useful will be a big help.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+
+    :cvar last_updated: When the plugin was last updated.
+    :type last_updated: fields.DateTimeField
+
+    Relations:
+
+    :cvar config: Plugin configuration.
+    :type config: fields.GenericRelation
+    """
+    name = models.TextField()
```
asmacdo: should this be indexed as well?

asmacdo: If not indexed, should it at least be unique?

asmacdo:
Is there a reason you are using TextFields rather that CharFields?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,268 @@
...
+    :cvar scratchpad: Arbitrary information stashed on the repository.
+    :type scratchpad: fields.GenericRelation
+
+    :cvar notes: Arbitrary repository properties.
+    :type notes: fields.GenericRelation
+
+    :cvar content: Associated content.
+    :type content: models.ManyToManyField
+    """
+    name = models.TextField(db_index=True, unique=True)
```
asmacdo: Is there a reason there are no validations like we have in the old model?
https://github.com/pulp/pulp/blob/master/server/pulp/server/db/model/__init__.py#L118

There is a built in regex validator.
https://docs.djangoproject.com/en/1.8/ref/validators/#regexvalidator

platform/pulp/platform/models/repository.py
jortel: Good question about indexing and uniqueness.  I was thinking that a (for example) distributor (name, type, repository_id)  needed to be unique (see: RepositoryPlugin.unique_together) but my thinking might be colored by pulp 2.

jortel:
Using the `TextField` because I don't think we intend to limit the length.  In postgres, the performance is the same.  Seem reasonable?

https://www.postgresql.org/docs/9.0/static/datatype-character.html

platform/pulp/platform/models/repository.py
jortel: It's my understanding that the need for validation was imposed by mongod key restrictions.  Since the *name* is part of the natural key, I wonder if we should restrict characters that impact the REST API.  Like things that are not URL friendly.

platform/pulp/platform/models/repository.py
asmacdo: Ah, that makes sense. Though, I think we ought to be able to drop `type` from the list since repositories are typed now.

platform/pulp/platform/models/repository.py
asmacdo: The slug validator is designed for url friendly stuff:
https://docs.djangoproject.com/en/1.8/ref/validators/#validate-slug

asmacdo:
Ah yes, I was remembering something I read when using MySQL. TextField seems good!

platform/pulp/platform/models/repository.py
jortel: I added `index_together` to go with `unique_together`.  Dropping `type` is a good idea.  If users add several distributors to a repository, it seems reasonable they name them differently.

platform/pulp/platform/models/repository.py
jortel: Oh, yeah!  I'd be 100% in favor of that.

asmacdo:
Looks good to me, but I did this review mostly to familiarize myself with this type of work. I would prefer someone with Django ORM experience to add the flag.

Great work @jortel!

platform/pulp/platform/models/base.py
```diff
@@ -51,7 +51,7 @@ class MasterModel(Model):
         subclass.
 
     """
-    type = models.CharField(max_length=63)
+    real_model = models.CharField(max_length=63)
```
jortel: The ``type`` field is commonly used in subclasses.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,342 @@
...
+        (BACKGROUND, 'Download In Background'))
+
+    feed_url = models.TextField()
+    validate = models.BooleanField(default=False)
+
+    ssl_ca_certificate = models.TextField(blank=True, default='')
+    ssl_client_certificate = models.TextField(blank=True, default='')
+    ssl_client_key = models.TextField(blank=True, default='')
+    ssl_validation = models.BooleanField(default=False)
+
```
pcreech: Does it make sense to normalize this a little and change this into a 'ImporterSslConfig' Table and join on it with a 0..1 relationship??  So code would be:

(pseudocode)
if importer.sslconfig then
     # do stuff with importer.sslconfig.ssl_ca_certificate

That way we don't have a bunch of empty fields in the database, and can reuse the same configs across importers? 

Same for below for 'proxy_' and 'basic_auth' config contexts below.

jortel:
BTW: pretty sure I'm missing a bunch of _standard_ distributor settings.

platform/pulp/platform/models/base.py
seandst: I like the reasoning, but I think the name should reinforce the master/detail relationship. Instead of `real_model`, how about `detail_model`, or `detail_type`?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,342 @@
...
 
-from pulp.platform.models import Model
+from django.db import models
+from django.contrib.contenttypes import fields
 
+from pulp.platform.models import Model, Notes, Scratchpad, MasterModel
 
-class Distributor(Model):
+
+class Content(Model):
```
seandst: I love just calling this content.

platform/pulp/platform/models/repository.py
seandst: That's what I was thinking as well. The project timezone is explicitly declared to be UTC in [settings.py](https://github.com/pulp/pulp/blob/3.0-dev/platform/pulp/platform/settings.py#L134), so all that's missing is the project-wide docs that all datetime objects will be converted to UTC if they're tz-aware, and assumed to be UTC if they're naive (but really, we should avoid naive datetime objects to the point of never seeing them).

https://docs.djangoproject.com/en/1.8/topics/i18n/timezones/

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,342 @@
...
+    created = models.DateTimeField(auto_now_add=True)
+
+    content = models.ForeignKey('Content', on_delete=models.CASCADE)
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
+
+    def save(self, *args, **kwargs):
+        """
+        Save the association.
+        """
+        self.repository.last_content_added = datetime.utcnow()
```
seandst: This (and anywhere else datetime.utcnow() is used) create a naive datetime object. Instead:
```python
from django.utils import timezone
timezone.now()
```

This generates a tz-aware datetime object in the default configured timezone which, in our case, is UTC.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,342 @@
...
+    description = models.TextField(blank=True, default='')
+
+    members = models.ManyToManyField('Repository')
+    scratchpad = fields.GenericRelation(Scratchpad)
+    notes = fields.GenericRelation(Notes)
+
+
+class Plugin(MasterModel):
+    """
+    An Abstract model for plugins.
```
seandst: I think of a plugin to be something like the entirely of `pulp_rpm`, but that doesn't seem to be correct in this context. What exactly does "plugin" mean here?

platform/pulp/platform/models/repository.py
seandst: the slug validator "ensures a value consists of only letters, numbers, underscores or hyphens". If we want to be that restrictive for things like a repo name (formerly repo_id), then I agree. I'm concerned about the ability to migrate repos from pulp 2 that the slug validator wouldn't like, though, so we might want to open it up to whatever pulp 2 currently allows for repo_id. In particular, looking at some of the more interesting repo_ids that Katello comes up with would be good for guidance here.

platform/pulp/platform/models/repository.py
seandst: It's also worth pointing out that in an ideal (python 3) world, python has recognized the folly of polluting the global namespace with all of its builtin functions. If you really need the builtin type in this local namespace, `import builtins; builtins.type`. `six` provides a compatibility layer to get builtins from the correct place if we want to run on py2 or py3 (but please let's just shoot for ultimately being py3)

platform/pulp/platform/models/repository.py
jortel: Interesting idea.

Since there is nothing importer specific in these settings we could have generic tables to define TLS, proxy and basic auth settings. 

Something like:

```
class TLS(Model):
    ca_certificate = TextField()
    client_certificate = TextField()
    client_key = TextField()
    ...

class Proxy(Model):
    host = TextField()
    port = IntegerField()
    user = TextField()
    password = TextField()

class BasicAuth(Model):
    user = TextField()
    password = TextField()

class Importer(Plugin):
    ssl = ForeignKey(TLS, null=True)
    proxy = ForeignKey(Proxy, null=True)
    basic_auth = ForeignKey(BasicAuth, null=True)
```

Not sold on it ... just thinking out loud.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,342 @@
...
+    :cvar relative_url: The (relative) path component of the published url.
+    :type relative_url: models.TextField
+
+    :cvar last_published: When the last successful publish occurred.
+    :type last_published: models.DateTimeField
+
+    Relations:
+
+    """
+    relative_url = models.TextField(blank=True, default='')
```
mhrivnak: Minor point, but can we rename this ``relative_path``? It's not a URL.

platform/pulp/platform/models/base.py
jortel: :+1:  detail_model

platform/pulp/platform/models/repository.py
jortel: Ah.  Cool.

platform/pulp/platform/models/repository.py
jortel: It's an importer or distributor.  Sorry, could not think of a better name.

platform/pulp/platform/models/repository.py
jortel: Yes!  much better.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,341 @@
...
+    :cvar ssl_client_certificate: A PEM encoded client certificate used for authentication.
+    :type ssl_client_certificate: models.TextField
+
+    :cvar ssl_client_key: A PEM encoded private key used for authentication.
+    :type ssl_client_key: models.TextField
+
+    :cvar ssl_validation: Indicates whether SSL peer validation must be performed.
+    :type ssl_validation: models.BooleanField
+
+    :cvar proxy_host: The hostname of an SSL proxy to be used.
```
mhrivnak: I think this setting is actually a URL even though it's called host. This is a good opportunity to rename it to ``proxy_url``.

mhrivnak: From our docs of this setting: "Indicates the URL to use as a proxy server when synchronizing this repository."

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,341 @@
...
+    :cvar proxy_password: The password used to authenticate with an SSL proxy to be used.
+    :type proxy_password: models.TextField
+
+    :cvar basic_auth_user: The user used in HTTP basic authentication.
+    :type basic_auth_user: models.TextField
+
+    :cvar basic_auth_password: The password used in HTTP basic authentication.
+    :type basic_auth_password: models.TextField
+
+    :cvar max_download_bandwidth: The max amount of bandwidth used per download.
```
mhrivnak: in Bps? KBps?

platform/pulp/platform/models/repository.py
jortel: Could the URL be: `http://jeff:ortel@myhost.com:8080`?  If not, would naming this `proxy_url` be confusing.  If yes, wouldn't user, password and port conflict?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,341 @@
...
+    :cvar relative_path: The (relative) path component of the published url.
+    :type relative_path: models.TextField
+
+    :cvar last_published: When the last successful publish occurred.
+    :type last_published: models.DateTimeField
+
+    Relations:
+
+    """
+    relative_path = models.TextField(blank=True, default='')
```
asmacdo: Do we have a plan to ensure there are not collisions? I think this is currently this is done in the repo config conduit.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,341 @@
...
+    """
+    name = models.TextField(db_index=True)
+    type = models.TextField()
+    last_updated = models.DateTimeField(auto_now=True, blank=True, null=True)
+
+    class Meta:
+        abstract = True
+
+
+class Importer(Plugin):
```
asmacdo: Keeping Distributor abstract made sense because there are RepositoryDistributors and GroupDistributors, but I can't imagine a use for this class other than RepositoryImporter. Could/should they be combined?

platform/pulp/platform/models/repository.py
jortel: Any reason not to add a `unique` constraint on the model?

platform/pulp/platform/models/repository.py
jortel: They could be but I liked the symmetry/consistency with distributor hierarchy.

platform/pulp/platform/models/repository.py
asmacdo: @seandst good point. I guess that url-unfriendly stuff is encoded out.

platform/pulp/platform/models/repository.py
jortel: Actually, I'm wondering how useful the throttling settings really are.  In many ways, they're a lie.  They only apply to downloading done by a single importer (not across all importers) and probably does not throttle the streamer or background downloading.  Thinking me may want to remove them and find another way to throttle pulp downloading system-wide.  Thoughts?

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,341 @@
...
+    """
+    A abstract content importer.
+
+    Fields:
+
+    :cvar feed_url: The URL of an external content source.
+    :type feed_url: models.TextField
+
+    :cvar validate: Validate the imported context.
```
jeremycline: I'm sort of jumping into the middle of things here, but do we ever want to import content we haven't validated? I'm imagining having some sort of callable to perform validation, and if I _don't_ want to validate anything I just define a function that passes.

platform/pulp/platform/models/repository.py
jeremycline: My personal preference is a URL in the format ``<scheme>://<proxy_user>:<proxy_pass>@<host>:<port>`` which mirrors the requests style (http://docs.python-requests.org/en/master/user/advanced/#proxies) and dropping the ``proxy_<blah>`` fields.

platform/pulp/platform/models/repository.py
jeremycline: Can we _please_ not try to implement throttling. Like @jortel said, it's a lie. In my opinion, it's absolutely not our problem to solve. We should document throttling methods (saying "see ``man tc``" seems slightly cruel), but we shouldn't re-invent the wheel.

platform/pulp/platform/models/repository.py
jeremycline: I think it would be good to include some more network configuration with a repository. The two things that come to mind are timeout and retry values. They could be based off the urllib3 objects (maybe not as granular, though). Maybe that's something we want to do later, but it's worth keeping in mind.

http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.timeout
http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry

platform/pulp/platform/models/repository.py
jortel: The importer having the capability to perform validation and a user configuring an importer (instance) to do validation are two different things.  This field indicates whether the user want the importer to validate.

platform/pulp/platform/models/repository.py
jortel: The slug validator seems too restrictive.  Users want to use fqdn for consumers like `jortel-lt.redhat.com` and users may want to use email address for user names etc.

platform/pulp/platform/models/repository.py
mhrivnak: I think the current collision logic also prevents publishing one repo within the path of another. That can't be covered by a uniqueness constraint.

platform/pulp/platform/models/repository.py
mhrivnak: Let's check in with katello and Satellite. This was a very high priority feature when we added it.

I would expect the setting to be used by the streamer and by the background download task.

That said, a more robust option is to use a proxy to throttle all traffic into pulp (or be selective), but so far we've been required to also support it independently.

platform/pulp/platform/models/repository.py
jeremycline: Is it maybe worth just mirroring the requests configuration style that maps a scheme and optionally a host to a proxy (``{'https://cdn.redhat.com': 'http://10.10.10.10:3128'}``), so users have fine-grain control over proxy usage (for example, configuring all traffic destined for ``https://cdn.redhat.com`` to go through proxy X, or http traffic goes through one proxy and https traffic goes through a different proxy)? Perhaps at an instance level it doesn't make a ton of sense to configure proxies this way (unless alternate content sources gets its network configuration from the importer as well? Maybe it should?).

I guess what I want here is a single, clear style for proxy configuration that is the same at the instance level, the plugin level, and the system level. If we go with the ``scheme/host->proxy`` map everywhere, it'll be easy to merge together. If we don't, what's the protocol if we want to support ``scheme/host->proxy`` at a plugin/platform level and then we need to merge it with a importer configuration? 

Maybe I'm making mountains out of molehills and we can deal with all that later.

platform/pulp/platform/models/repository.py
jortel: After modeling [1] my _thinking out loud_ ^^, I don't think it's a good idea for 3.0.  The importer record in the DB defines an importer instance and it's association with a repository.  Except the foreign key (to repository), all of the other fields describe the importer (a.k.a. configuration).   We don't have use cases for sharing named configuration between importers so any modeling to support this is premature.  As for DB normalization, it's already in 3rd normal form because none of these fields appear in any other tables. 

[1] https://github.com/jortel/pulp/blob/dc0af19b4a1b4281f0120e5a20b86b5ff6a4212b/platform/pulp/platform/models/repository.py

pcreech: Sounds good to me.  Just an out-loud idea I had.  Thanks for looking into it!

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,329 @@
...
+    :type ssl_client_certificate: models.TextField
+
+    :cvar ssl_client_key: A PEM encoded private key used for authentication.
+    :type ssl_client_key: models.TextField
+
+    :cvar ssl_validation: Indicates whether SSL peer validation must be performed.
+    :type ssl_validation: models.BooleanField
+
+    :cvar proxy_url: The optional proxy URL. Format: scheme://user:password@host:port
+    :type proxy_url: models.ForeignKey
```
jortel: The `proxy_url` consolidates the removal of the individual `proxy_` fields.  The will necessitate a (good) change  in Nectar which currently builds a URL using the parts.  There is no value in storing the URL components instead of the URL.

platform/pulp/platform/models/generic.py
```diff
@@ -86,6 +86,20 @@ def items(self):
         # iterable, rather than making it __getitem__ for each key
         return ((kv.key, kv.value) for kv in self.manager.all())
 
+    def get(self, key, default=None):
+        """
```
jortel: This provides a nice convenience.  I found this very helpful when writing the `howto.py` and dealing with notes and scratchpad.

jortel:
I have noticed that using `TextField(blank=False)` does not make it required.  I also question the need for `default=''` since I *think* it's the default.  I have asked @seandst to look into both of these things.

jortel:
Discussed TextField on \#django and apparently the `blank` is only used for form validation.  I did some playing around and using `TextField(default=None)` does seem to produce the desired result which is for the field to try and store `NULL` in the `NOT-NULL` column which causes a constraint violation.

Still planning to discuss with @seandst (and others).  I'm thinking we need to adopt a consistent strategy on this.

platform/pulp/platform/models/repository.py
mhrivnak: The yum importer docs describe it: "The maximum download speed in bytes/sec for a task (such as a sync); defaults to None"

platform/pulp/platform/models/repository.py
mhrivnak: I suggest making a task in redmine to figure this out and not have it block this PR.

platform/pulp/platform/models/repository.py
mhrivnak: @jortel pointed out that there aren't (m?)any mutable fields that aren't config, so I agree there's no need to rename ``last_updated``

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,329 @@
...
+    IMMEDIATE = 'immediate'
+    ON_DEMAND = 'on_demand'
+    BACKGROUND = 'background'
+    DOWNLOAD_POLICIES = (
+        (IMMEDIATE, 'Download Immediately'),
+        (ON_DEMAND, 'Download On Demand'),
+        (BACKGROUND, 'Download In Background'))
+
+    feed_url = models.TextField()
+    validate = models.BooleanField(default=False)
```
mhrivnak: We should default this to ``True``

https://pulp.plan.io/issues/536

jortel: Done

mhrivnak:
LGTM, but I'll let @seandst give it a last once-over and add the label if he's happy.

mhrivnak:
This PR also resolves https://pulp.plan.io/issues/536

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,329 @@
...
+
+    Fields:
+
+    :cvar feed_url: The URL of an external content source.
+    :type feed_url: models.TextField
+
+    :cvar validate: Validate the imported context.
+    :type validate: models.BooleanField
+
+    :cvar ssl_ca_certificate: A PEM encoded CA certificate used for authentication.
```
jeremycline: This should probably read "A PEM encoded CA certificate used to verify the signature of the TLS certificate served by the ``feed_url``" or something like that or it might be confused with the CA certificate that signed the client certificates.

platform/pulp/platform/models/repository.py
```diff
@@ -1,10 +1,329 @@
...
+        (ON_DEMAND, 'Download On Demand'),
+        (BACKGROUND, 'Download In Background'))
+
+    feed_url = models.TextField()
+    validate = models.BooleanField(default=True)
+
+    ssl_ca_certificate = models.TextField(blank=True)
+    ssl_client_certificate = models.TextField(blank=True)
+    ssl_client_key = models.TextField(blank=True)
+    ssl_validation = models.BooleanField(default=False)
```
jeremycline: Is there a good reason for defaulting to ``False`` rather than ``True``?

jortel: Good question.  Probably should be `True`.  I'll get consensus on this.

PR #2685, opened 2016-08-08
Fixes selinux-policy Requires statement

bmbouter:
This is the same fix as 8c152a68 except it correctly
escapes the %{version} and %{release} fields using
%%.

Now when you run:

`rpmspec -q pulp.spec --requires|grep selinux-policy` I get
the following output:

selinux-policy >= 3.13.1-158.21.fc23

https://pulp.plan.io/issues/2121
closes #2121

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -16,7 +16,7 @@
 %if %{pulp_server}
 #SELinux
 %define selinux_variants mls strict targeted
-%define selinux_policyver %(sed -e 's,.*selinux-policy-\\([^/]*\\)/.*,\\1,' /usr/share/selinux/devel/policyhelp 2> /dev/null)
+%define selinux_policyver %(rpm --qf "%%{version}-%%{release}" -q selinux-policy)
 %define moduletype apps
 %endif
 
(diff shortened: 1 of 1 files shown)


jeremycline:
@bmbouter Hi, I saw this fly by when I git pulled. It looks like this problem is also present in the Fedora specfile and in the specfile in ``pulp/packaging``. I'm happy to apply this patch to both places, but in future I might not notice the specfile changes so it'd be very helpful if anyone who makes a specfile change checks both these places.

bmbouter:
@jeremycline Yes @seandst and I agree completely and talked about this on IRC last week. The plan was going to port the changes as part of his normal sync-up between `pulp/pulp` and `pulp/packaging` spec file changes at release time. That being said I made the change and I could also make the necessary PRs.

@seandst do you still want to port this newest change or should I do it? I'm happy to just let me know.

jeremycline:
Okay. I'll just pull it into Fedora 26/25/24, then.

seandst:
Yeah, we need to document all the different specfiles and how updates among them should propagate. Until that's done, I don't think it's fair to ask anything more than a single PR to the github repos for specfile changes.

PR #2686, opened 2016-08-09
Add twine, pkginfo, and setuptools to our external deps

jeremycline:
pulp-python-2.0 requires twine, which depends on pkginfo and setuptools
which are newer than those available in EL6. We must carry all three of
these dependencies until we drop EL6 or newer versions (see the twine
specfile for exact requirements) are available.

Note: I have not built these in Katello's Koji, as I do not have access.
I did build and install them locally, however, so they should build without
trouble.


seandst:
Have these already been built in koji? I can help with that, if not.

jeremycline:
They have not been built in Koji yet.

seandst:
LGTM, but we'll need the koji builds done for el6

seandst:
Wait, I lied. All the other external deps have the release component of the version. Do these need that?

jeremycline:
Oh good catch. I assume they do, but I have absolutely no clue how external_deps.json is consumed :shipit:

jeremycline:
I've updated external_deps.json with the releases.

ipanova:
@jeremycline if you want to get the access here is some useful info whom you need to ping http://projects.theforeman.org/projects/foreman/wiki/Koji

jeremycline:
To clarify, do I need to build this in Koji before merging, or after merging?

ipanova:
@jeremycline i think this PR should be divided in 2, first would contain all the spec files what you would need to build the deps. Once the deps are build on koji you create and merge PR w/r to external_deps.json update.

jeremycline:
I can just build the deps in Koji right now and merge it all in one go. Is there any reason to not do that?

ipanova:
closing this Pr in favour of https://github.com/pulp/pulp/pull/2766

PR #2693, opened 2016-08-11
Second attempt at backporting the fix for requests issue #2863

jeremycline:
The problem with the first backport was that it was not entirely
compatible with requests-2.6 and urllib3-1.10, which is what el6 and el7
ship.

For this backport I started with requests-2.6.2, applied the proposed patch, and then resolved all the conflicts.

I'll admit I'm a bit nervous of this approach, but the alternative is to carry requests and urllib3 for el6-f24 which I really don't want to do.


jeremycline:
I accidentally opened this against the wrong branch.

PR #2695, opened 2016-08-11
Fix broken unit tests in storage path migration

jeremycline:
https://github.com/pulp/pulp/pull/2675 introduced several broken unit
tests by adding non-determinism to the call order.


jeremycline:
ok test

PR #2697, opened 2016-08-11
Pull basic auth info out of feed url

asmacdo:
https://pulp.plan.io/issues/1977

Previous versions of Pulp allowed users to include their basic auth
information in the url. This is obviously less ideal than passing
this in the config, but in order to remain backwards compatibile, this
removes the auth info from the url and adds it to the importer config.
For importer configurations that were created before this change, we pass
the configuration through the same cleaner function as we do during
importer creation, so a temporary importer configuration includes the
basic auth information.

**Important note about testing:**
To test this, there are two things that need to be checked. First, a new importer feed should contain basic auth in the url. In this case, inspection of the importer config should show a feed without the basic auth, and with credentials in the config. The second thing to test to create an importer with the same config, but before this change. Then, upgrade to include this commit, and sync.

### DIFF
--- a/server/pulp/server/controllers/importer.py
+++ b/server/pulp/server/controllers/importer.py
@@ -1,10 +1,12 @@
 import logging
 import sys
+import urlparse
 
 import celery
 from mongoengine import ValidationError
 
 from pulp.common import error_codes, tags
+from pulp.common.plugins import importer_constants
 from pulp.plugins.config import PluginCallConfiguration
 from pulp.plugins.loader import api as plugin_api
 from pulp.server import exceptions
@@ -35,7 +37,10 @@ def build_resource_tag(repo_id, importer_type_id):
 
 def clean_config_dict(config):
     """
-    Remove keys from a dict that have a value None.
+    Ensure that the config is ready to be passed to the plugin. There are two things that must
+    occur:
+        1. Remove keys from the dict that have a value None.
+        2. If they exist, move basic auth credentials from the feed_url to the config.
 
     :param config: configuration for plugin
     :type  config: dict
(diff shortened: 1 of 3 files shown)


asmacdo:
Given previous work on this I'd also like to add @pcreech to the list of potential reviewers.

pcreech:
Looks good, and feels like the appropriate place to do this.   Nice Work! LGTM

server/pulp/server/controllers/importer.py
```diff
@@ -35,16 +37,28 @@ def build_resource_tag(repo_id, importer_type_id):
...
     :rtype:  dict:
     """
-
     if config is not None:
+        feed_url = config.get('feed')
+        if feed_url:
+            parsed = urlparse.urlparse(feed_url)
+            if parsed.username and parsed.password:
+                at = parsed.netloc.find('@')
+                new_url = parsed._replace(netloc=parsed.netloc[at + 1:])
```
seandst: Any reason not to `(netloc=parsed.netloc.rsplit('@', 1)[1])`?

seandst: This implementation is broken, and the rfind approach (or something like it) needs to be used, since username and password can contain `@`. hostnames, cannot, so the rightmost `@` must be used to partition the hostname off from the user/pass components. Also, it's possible for only the username to be set (where netloc is `username@hostname`) and for only the password to be set (where netloc is `:password@hostname`), so the conditional on line 55 should be `if parsed.username or parsed.password`.

```python
In [1]: from urllib import parse

In [2]: parsed = parse.urlparse('scheme://user@name:pass@word@hostname/path')

In [3]: at = parsed.netloc.find('@')

In [4]: parsed._replace(netloc=parsed.netloc[at + 1:])
Out[4]: ParseResult(scheme='scheme', netloc='name:pass@word@hostname', path='/path', params='', query='', fragment='')

In [5]: parsed._replace(netloc=parsed.netloc.rsplit('@', 1)[1])
Out[5]: ParseResult(scheme='scheme', netloc='hostname', path='/path', params='', query='', fragment='')

In [6]: parsed.netloc, parsed.username, parsed.password
Out[6]: ('user@name:pass@word@hostname', 'user@name', 'pass@word')

```
Note that with the `at = ` approach, the netloc in the new url is wrong.

asmacdo: Updated, good catch! Please note that I had to make a new decision. If not present, `urlparse` sets the password to `None`, which will cause that key to be removed from the config in the next part of this function. If username is not present, `parsed.username` becomes the empty string.

PR #2699, opened 2016-08-13
Add task models

pcreech:
Add task models.

The tasking system is a pretty solid, well baked solution, so making too many changes at this time might not make sense.  I tried to make ones that made sense in relation to utilizing a relational database.

Utilized foreign keys where it appears to make sense, and converted some of the data types to utilize their natural data state in the database (mainly DateTimes)

There are four fields that will need more thought, but wanted to go ahead and put up the PR to start the conversation. These fields store complex data structures, that could potentially vary based on utilization.  

These are:
- Task.result
- Task.error
- ScheduledCalls.args
- ScheduledCalls.kwargs

My initial thought was to convert them into properties, so we can split up the internal representation (TextField) with an external implementation of some type of complex object.  This should be transparent for the user, as when they use "scheduled_call.args" it should execute the code in the property function.  Currently the implementation is a naive example implementation utilizing json.dumps, that will need to be fleshed out.  

With that, I'm also looking into how creating a custom Django field (Michael's suggestion) could provide the same functionality.  It appears to provide a more natural interface for converting between database state and python state.  Brian suggested utilizing django.picklefield.  This is a viable and enticing option, but would become another dependency we would have to manage.  We should probably weigh the cost of developing our own solution with the cost of introducing another dependency.

closes [#2087](https://pulp.plan.io/issues/2087)

### DIFF
--- a/platform/pulp/platform/fields.py
+++ b/platform/pulp/platform/fields.py
@@ -1,0 +1,78 @@
+"""
+Custom Django fields provided by pulp platform
+"""
+import json
+
+from django.db import models
+from django.utils.translation import six
+
+
+class JSONField(models.TextField):
+    """
+    A custom Django field to serialize data into a text field and vice versa
+
+    """
+    def from_db_value(self, value, *args, **kwargs):
+        """
+        Converts a value as returned by the database to a Python object
+
+        :param value: The value to convert to Python
+        :type value: object
+
+        :param args: unused positional arguments
+        :type args: list
+
+        :param kwargs: unused keyword arguments
+        :type kwargs: dict
+
+        :return: A Python representation of value
+        :rtype: object
+        """
+        if isinstance(value, six.string_types):
+            return self.to_python(value)
+        return value
+
+    def to_python(self, value):
+        """
+        Converts the value into the correct Python object
+
+        :param value: The value to convert to Python
+        :type value: object
+
+        :return: A Python representation of value
+        :rtype: object
+        """
+        return json.loads(value)
+
+    def get_db_prep_value(self, value, *args, **kwargs):
+        """
+        Converts value to a backend-specific value
+
+        :param value: The value to be converted
+        :type value: object
+
+        :param args: unused positional arguments
+        :type args: list
+
+        :param kwargs: unused keyword arguments
+        :type kwargs: dict
+
+        :return: json string representing the object
+        :rtype: str
+        """
+        if value is None:
+            return None
+        return json.dumps(value)
+
+    def value_to_string(self, obj):
+        """
+        Converts obj to a string. Used to serialize the value of the field
+
+        :param obj: The object to be converted
+        :type obj: object
+
+        :return: Serialized value
+        :rtype: str
+        """
+        value = self._get_val_from_obj(obj)
+        return self.get_db_prep_value(value, None)
--- a/platform/pulp/platform/models/__init__.py
+++ b/platform/pulp/platform/models/__init__.py
@@ -6,3 +6,5 @@ from .consumer import Consumer, ConsumerContent  # NOQA
 from .repository import (Repository, RepositoryGroup, Importer, Distributor,  # NOQA
                          RepositoryImporter,  RepositoryDistributor, GroupDistributor,  # NOQA
                          RepositoryContent)  # NOQA
+
+from .task import ReservedResource, Worker, Task, TaskTag, TaskLock, ScheduledCalls  # NOQA
--- a/platform/pulp/platform/models/task.py
+++ b/platform/pulp/platform/models/task.py
@@ -1,0 +1,213 @@
+"""
+Django models related to the Tasking system
+"""
+from django.db import models
+
+from pulp.platform.models import Model
+from pulp.platform.fields import JSONField
+
+
+class ReservedResource(Model):
+    """
+    Resources that have been reserved
+
+    Fields:
+
+    :cvar resource: The name of the resource reserved for the task.
+    :type resource: models.TextField
+
+    Relations:
+
+    :cvar task: The task associated with this reservation
+    :type task: models.ForeignKey
+
+    :cvar worker: The worker associated with this reservation
+    :type worker: models.ForeignKey
+    """
+    resource = models.TextField()
+
+    task = models.OneToOneField("Task")
+    worker = models.ForeignKey("Worker")
+
+
+class Worker(Model):
+    """
+    Represents a worker
+
+    Fields:
+
+    :cvar name: The name of the worker, in the format "worker_type@hostname"
+    :type name: models.TextField
+
+    :cvar last_heartbeat: A timestamp of this worker's last heartbeat
+    :type last_heartbeat: models.DateTimeField
+    """
+    name = models.TextField(db_index=True, unique=True)
+    last_heartbeat = models.DateTimeField()
+
+
+class TaskLock(Model):
+    """
+    Locking mechanism for services that utilize active/passive fail-over
+
+    Fields:
+
+    :cvar name: The name of the item that has the lock
+    :type name: models.TextField
+
+    :cvar timestamp: The time the lock was acquired
+    :type timestamp: models.DateTimeField
+
+    :cvar lock: The name of the lock acquired
+    :type lock: models.TextField
+
+    """
+    CELERY_BEAT = 'CeleryBeat'
+    RESOURCE_MANAGER = 'ResourceManager'
+    LOCK_STRINGS = (
+        (CELERY_BEAT, 'Celery Beat Lock'),
+        (RESOURCE_MANAGER, 'Resource Manager Lock')
+    )
+
+    name = models.TextField(db_index=True, unique=True)
+    timestamp = models.DateTimeField(auto_now_add=True)
+    lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
+
+
+class Task(Model):
+    """
+    Represents a task
+
+    Fields:
+
+    :cvar group: The group this task belongs to
+    :type group: models.UUIDField
+
+    :cvar state: The state of the task
+    :type state: models.TextField
+
+    :cvar started_at: The time the task started executing
+    :type started_at: models.DateTimeField
+
+    :cvar finished_at: The time the task finished executing
+    :type finished_at: models.DateTimeField
+
+    :cvar error: Collection of errors that might have occurred while task was running
+    :type error: models.JSONField
+
+    :cvar result: Return value of the task
+    :type result: models.JSONField
+
+    Relations:
+
+    :cvar parent: Task that spawned this task (if any)
+    :type parent: models.ForeignKey
+
+    :cvar worker: The worker that this task is in
+    :type worker: models.ForeignKey
+    """
+
+    WAITING = 'waiting'
+    SKIPPED = 'skipped'
+    ACCEPTED = 'accepted'
+    RUNNING = 'running'
+    SUSPENDED = 'suspended'
+    COMPLETED = 'completed'
+    ERRORED = 'errored'
+    CANCELED = 'canceled'
+    STATES = (
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (ACCEPTED, 'Accepted'),
+        (RUNNING, 'Running'),
+        (SUSPENDED, 'Suspended'),
+        (COMPLETED, 'Completed'),
+        (ERRORED, 'Errored'),
+        (CANCELED, 'Canceled')
+    )
+    group = models.UUIDField(null=True)
+    state = models.TextField(choices=STATES)
+
+    started_at = models.DateTimeField(null=True)
+    finished_at = models.DateTimeField(null=True)
+
+    error = JSONField()
+    result = JSONField()
+
+    parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks")
+    worker = models.ForeignKey("Worker", null=True)
+
+
+class TaskTag(Model):
+    """
+    Custom tags for a task
+
+    Fields:
+
+    :cvar name: The name of the tag
+    :type name: models.TextField
+
+    Relations:
+
+    :cvar task: The task this tag is associated with
+    :type task: models.ForeignKey
+    """
+    name = models.TextField()
+
+    task = models.ForeignKey("Task", related_name="tags", related_query_name="tag")
+
+
+class ScheduledCalls(Model):
+    """
+    Scheduled Call Request
+
+    Fields:
+
+    :cvar task: The task that should be run on a schedule
+    :type task: models.TextField
+
+    :cvar enabled: Indicates if schedule should be actively run by the scheduler
+    :type enabled: models.BooleanField
+
+    :cvar resource: Indicates a unique resource that should be used to find this schedule
+    :type resource: models.TextField
+
+    :cvar iso_schedule: ISO8601 string representing the schedule
+    :type iso_schedule: models.TextField
+
+    :cvar schedule: Pickled instance of celery.schedules.schedule that should be run.
+    :type schedule: models.TextField
+
+    :cvar first_run: The first time this schedule was ran
+    :type first_run: models.DateTimeField
+
+    :cvar last_run: Last time this schedule was ran
+    :type last_run: models.DateTimeField
+
+    :cvar total_run_count: Number of times this schedle has ran
+    :type total_run_count: models.IntegerField
+
+    :cvar last_updated: The last time this schedule was saved to the database
+    :type last_updated: models.DateTimeField
+
+    :cvar args: Arguments that should be passed to the apply_async function
+    :type args: models.JSONField
+
+    :cvar kwargs: Keyword arguments that should be passed to the apply_async function
+    :type kwargs: models.JSONField
+    """
+    task = models.TextField()
+    enabled = models.BooleanField(default=True)
+    resource = models.TextField(null=True)
+
+    iso_schedule = models.TextField()
+    schedule = models.TextField(null=True)
+
+    first_run = models.DateTimeField()
+    last_run = models.DateTimeField(null=True)
+    total_run_count = models.IntegerField()
+
+    last_updated = models.DateTimeField()
+
+    args = JSONField()
+    kwargs = JSONField()


platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    task_group = models.UUIDField()
+    state = models.TextField(choices=STATES)
+
+    start_time = models.DateTimeField()
+    finish_time = models.DateTimeField()
+
+    _error = models.TextField(null=True, db_column='error')
+    _result = models.TextField(null=True, db_column='result')
+
+    parent = models.ForeignKey("Task", null=True)
```
pcreech: I converted spwaned_tasks: [] into a parent foreign key.  The other option is to create a table with task ids.  This was based on the assumption that I came to that the new task.id uuid is the same as the uuid's describing tasks in spanwed_tasks array.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+        (CELERY_BEAT, 'Celery Beat Lock'),
+        (RESOURCE_MANAGER, 'Resource Manager Lock')
+    )
+
+    name = models.TextField(db_index=True, unique=True)
+    timestamp = models.DateTimeField(auto_now_add=True)
+    lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
+
+
+class Task(Model):
```
pcreech: I removed task_id from the model, with the assumption that the new uuid pk will act as the task_id.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    CELERY_BEAT = 'CeleryBeat'
+    RESOURCE_MANAGER = 'ResourceManager'
+    LOCK_STRINGS = (
+        (CELERY_BEAT, 'Celery Beat Lock'),
+        (RESOURCE_MANAGER, 'Resource Manager Lock')
+    )
+
+    name = models.TextField(db_index=True, unique=True)
+    timestamp = models.DateTimeField(auto_now_add=True)
+    lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
```
pcreech: I consolidated CeleryBeatLock and ResourceManagerLock into a single TaskLock, since the data structures were near identical, and utilized a TextField for the lock.  Instead of both using the string 'locked' they now use a context specific string to identify the lock.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    :cvar task: The task associated with this reservation
+    :type task: models.ForeignKey
+
+    :cvar worker: The worker associated with this reservation
+    :type worker: models.ForeignKey
+    """
+    resource = models.TextField()
+
+    task = models.ForeignKey(Task, unique=True)
+    worker = models.ForeignKey(Worker)
```
pcreech: Instead of utilizing strings for the task_id and worker_name, I converted these fields into foreign keys to tie them to the task and worker.  This is in line with converting to a relational model, but might need to be revisited if it causes headaches for the reserved resource code.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+
+    first_run = models.DateTimeField()
+
+    remaining_runs = models.IntegerField()
+    total_run_count = models.IntegerField()
+    last_run_at = models.DateTimeField()
+
+    last_updated = models.DateTimeField(auto_now=True)
+
+    principal = models.ForeignKey("User", null=True)
```
pcreech: This was a stored user object.  I converted this to be a foreign key on the currently non-existent "User" table.  There was question of if this is still needed/utilized.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    Relations:
+
+    :cvar task: The task this tag is associated with
+    :type task: models.ForeignKey
+    """
+    name = models.TextField()
+
+    task = models.ForeignKey(Task, related_name="tags", related_query_name="tag")
+
+class ScheduledCalls(Model):
```
pcreech: This model was never converted to mongoengine, and there were statements that the pymongo object had values that no longer were used.  This model should probably be scrutinized heavily to ensure data wasn't missed.  I did a best guess between what's in the pymongo object and what is stored in the database when a schedule is created.

platform/pulp/platform/models/task.py
mhrivnak: +1 I've been thinking the same thing for a while.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    :type task: models.ForeignKey
+
+    :cvar worker: The worker associated with this reservation
+    :type worker: models.ForeignKey
+    """
+    resource = models.TextField()
+
+    task = models.ForeignKey(Task, unique=True)
+    worker = models.ForeignKey(Worker)
+
```
mhrivnak: 2 lines for pep8

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    :cvar name: The name of the worker, in the format "worker_type@hostname"
+    :type name: models.TextField
+
+    :cvar last_heartbeat: A timestamp of this worker's last heartbeat
+    :type last_heartbeat: models.DateTimeField
+    """
+    name = models.TextField(db_index=True, unique=True)
+    last_heartbeat = models.DateTimeField()
+
```
mhrivnak: 2 lines for pep8

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    Relations:
+
+    :cvar task: The task this tag is associated with
+    :type task: models.ForeignKey
+    """
+    name = models.TextField()
+
+    task = models.ForeignKey(Task, related_name="tags", related_query_name="tag")
+
```
mhrivnak: 2 lines for pep8

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    :type _kwargs: models.TextField
+
+    Relations:
+    
+    :cvar principal: User who the teask should be run as
+    :type principal: models.ForeignKey
+
+    """
+    task = models.TextField()
+    enabled = models.BooleanField()
```
mhrivnak: I suggest a default of ``True`` since it will be the very common case.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    resource = models.TextField(null=True)
+
+    consecutive_failures = models.IntegerField()
```
mhrivnak: We should remove both of these fields and remove the feature of auto-disabling scheduled tasks. It adds a lot of complexity without much value.

I added a task just now to track that. Feedback welcome: https://pulp.plan.io/issues/2176

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,296 @@
...
+    :cvar remaining_runs: Number of runs left until this schedule gets disabled
+    :type remaining_runs: models.IntegerField
+
+    :cvar total_run_count: Number of times this schedle has ran
+    :type total_run_count: models.IntegerField
+
+    :cvar last_run_at: Last time this schedule was ran
+    :type last_run_at: models.DateTimeField
+
+    :cvar last_updated: The last time this schedule was saved to the database
```
mhrivnak: I think this is specifically the last time a user changed something about the schedule. This field helps the ``Scheduler`` know if it needs to re-load schedules from the database when considering whether any are due to be run. I do not think this field should be touched when something like the ``total_run_count`` is incremented.

platform/pulp/platform/models/task.py
mhrivnak: It wouldn't hurt to check code paths, but I'm pretty sure this isn't used. We don't track which user created any other objects in pulp. It's a reasonable idea to consider, but I think we should re-consider that from a global perspective later. I suggest dropping it from this model.

mhrivnak:
Don't forget to cross-link the PR with the redmine issue so people can easily find each from the other.

https://pulp.plan.io/issues/2087

mhrivnak:
@bmbouter should definitely give this a thorough review.

Once the model is agreed on, you'll of course still need to add the docs about how to map old data to new.

mhrivnak:
Speaking of mapping old data to new, I propose that we document that of these models, only the scheduled call will have data migrated from mongo to postgres. Most of the others should be empty anyway. Task status is something that gets cleaned up on a regular basis, so it's understood to not be indefinitely persistent. I think leaving that history behind should be fine, and will give us the flexibility to model it however is best in 3.0.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
+
+
+class Task(Model):
+    """
+    Represents a task
+
+    Fields:
+
+    :cvar task_type: The type of task this is
```
jortel: The task_ here is redundant.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    :type task_type: models.TextField
+
+    :cvar task_group: The group this task belongs to
```
jortel: The `task_` here is redundant.  Just `group` is sufficient.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+        (SKIPPED, 'Skipped'),
+        (ACCEPTED, 'Accepted'),
+        (RUNNING, 'Running'),
+        (SUSPENDED, 'Suspended'),
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
+        (CANCELED, 'Canceled')
+    )
+    task_type = models.TextField()
+    task_group = models.UUIDField()
```
jortel: Same `task_` comment as above.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    state = models.TextField(choices=STATES)
+
+    start_time = models.DateTimeField()
+    finish_time = models.DateTimeField()
+
+    _error = models.TextField(null=True, db_column='error')
+    _result = models.TextField(null=True, db_column='result')
```
jortel: I think we should avoid fields with leading `_`.  What is the purpose of this?

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+
+    parent = models.ForeignKey("Task", null=True)
+    worker = models.ForeignKey(Worker)
+
+    @property
+    def result(self):
```
jortel: I suggest instead of using `_error` field and a _property_ to deal with JSON, we use a custom field (if not already exists) that does this.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+        try:
+            return self._result_object
+        except AttributeError:
+            self._result_object = json.loads(self._result)
+            return self._result_object
+
+    @property
+    def error(self):
```
jortel: I suggest instead of using `_error` field and a _property_ to deal with JSON, we use a custom field (if not already exists) that does this.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+
+        try:
+            self._kwargs = json.dumps(self._result_object)
+        except AttributeError:
+            pass
+
+        super(Model, self).save(*args, **kwargs)
+
+
+class TaskTags(Model):
```
jortel: Tables are implicitly plural and each model is a instance (row).  `TaskTag` (singular) would be more appropriate.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    def error(self):
+        try:
+            return self._error_object
+        except AttributeError:
+            self._error_object = json.loads(self._result)
+            return self._error_object
+
+    def save(self, *args, **kwargs):
+        try:
+            self._args = json.dumps(self._error_object)
```
jortel: What is `self._args`?  Where does `self._error_object` come from?

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    failure_threshold = models.IntegerField()
+
+    iso_schedule = models.TextField()
+    schedule = models.TextField(null=True)
+
+    first_run = models.DateTimeField()
+
+    remaining_runs = models.IntegerField()
+    total_run_count = models.IntegerField()
+    last_run_at = models.DateTimeField()
```
jortel: Suggest dropping the `_at` and just be`last_run`.  The `_at` adds no value and better symmetry with `first_run`.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+
+    last_updated = models.DateTimeField()
+
+    principal = models.ForeignKey("User", null=True)
+
+    _args = models.TextField(db_column="args")
+    _kwargs = models.TextField(db_column="kwargs")
```
jortel: Same comment here about the `_` prefixed fields and properties.  Let make this simpler by using a custom field.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+    iso_schedule = models.TextField()
+    schedule = models.TextField(null=True)
+
+    first_run = models.DateTimeField()
+
+    remaining_runs = models.IntegerField()
+    total_run_count = models.IntegerField()
+    last_run_at = models.DateTimeField()
+
+    last_updated = models.DateTimeField()
```
jortel: Suggest:  `models.DateTimeField(auto_now=True, blank=True, null=True)`

platform/pulp/platform/models/__init__.py
```diff
@@ -4,3 +4,5 @@
 
 from .consumer import Consumer, ConsumerContent  # NOQA
 from .repository import Distributor  # NOQA
+
+from .task import (ReservedResource, Worker, Task, TaskTags, TaskLock, ScheduledCalls) #NOQA
```
jortel: Don't need the `()` since it all fits on 1 line.

platform/pulp/platform/models/task.py
jortel: Perhaps: https://docs.djangoproject.com/en/dev/ref/contrib/postgres/fields/#django.contrib.postgres.fields.JSONField

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,298 @@
...
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
+        (CANCELED, 'Canceled')
+    )
+    task_type = models.TextField()
+    task_group = models.UUIDField()
+    state = models.TextField(choices=STATES)
+
+    start_time = models.DateTimeField()
+    finish_time = models.DateTimeField()
```
jortel: I'm a little conflicted about the `_time` in these field names.  They're actually a `datetime` and we don't normally include the datatype in field names.  Perhaps: `started` and `finished`?  I don't feel strongly about this.

platform/pulp/platform/models/task.py
mhrivnak: I think we still want every user-facing model to have a "natural key" that is different from the pk created by django. @seandst I recall you being enthusiastic about that idea- can you comment on this case?

platform/pulp/platform/models/task.py
mhrivnak: I suggested the same. I think the note on the PR says he's still investigating options. Thanks for pointing out that ``JSONField``!

platform/pulp/platform/models/task.py
mhrivnak: Just to emphasize my note above, this field should only be updated when the user changes a property, like the schedule itself, or the arguments. It should not be update when something like the ``remaining_runs`` value gets updated.

You can see the usage here: https://github.com/pulp/pulp/blob/0128ae2c99d9d57c4f73bcde52b43e1c81891736/server/pulp/server/async/scheduler.py#L328

jortel: Makes sense.

platform/pulp/platform/models/task.py
pcreech: The purpose here is to use python's way of marking a variable 'private', i.e. don't use this.  In the naive implementation I did to get the PR up, the choice I made for abstracting the db implementation from the user is to utilize python properties to expose the functionality I wanted.  When using "task.result" the property below will intercept the variable call, and convert it for general consumption from the internal storage mechanism to the public interaction.

This is a common thing to do in .Net with properties, and it seemed to hold true for python as well.

That being said, so far my investigation is leaning me towards converting these into some type of PickleField type, whether we use django.picklefield or implement our own.

platform/pulp/platform/models/task.py
pcreech: I'm hesitant to utilize JSONField, as that locks us into very recent versions of postgres for little benefit (these fields don't get utilized in the database, they are just for persistent data storage (based on a conversation I had with @bmbouter ) 

With that, as mentioned in a previous comment, i'm looking into a PickleField implementation that will be more db-agnostic.

platform/pulp/platform/models/task.py
pcreech: As mentioned in earlier comments, I will be looking into a pickefield implementation

platform/pulp/platform/models/task.py
pcreech: Great catch!

platform/pulp/platform/models/task.py
pcreech: copy/paste errors.  should be self._error

self._error_object gets created in the property getter, therefore if it doesn't exist, it never got utilized and never modified

platform/pulp/platform/models/task.py
pcreech: Agreed, as mentioned in previous comments I'm looking into a picklefield implementation

platform/pulp/platform/models/task.py
pcreech: Was more in the interest here of 1-1 variable naming, facilitating hopefully minimal code refactoring to utilize this model.  Can easily change if others are in agreement.

platform/pulp/platform/models/task.py
pcreech: Was more in the interest here of 1-1 variable naming, facilitating hopefully minimal code refactoring to utilize this model. Can easily change if others are in agreement.

platform/pulp/platform/models/task.py
mhrivnak: I can see it both ways. Just intuitively, I can imagine someone's first impression being that ``task.started`` and ``task.finished`` could be booleans. Generally since python isn't typed, if it's not reasonably obvious what type something will be, I'm not opposed to making it clear in the variable or attribute name. But I could easily see it both ways here and would be happy with either.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
+        (CANCELED, 'Canceled')
+    )
+    type = models.TextField()
+    group = models.UUIDField()
+    state = models.TextField(choices=STATES)
+
+    started_at = models.DateTimeField()
+    finished_at = models.DateTimeField()
```
pcreech: Changed start_time and finish_time to started_ad and finished_at.

Per @jortel 's comments, I felt this was a sane modification as time felt off, but started and finished seemed like Booleans intuitively to me.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+
+    error = JSONField()
+    result = JSONField()
```
pcreech: Removed the property implementation, and replaced it with a custom JSONField implementation

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    first_run = models.DateTimeField()
+
+    remaining_runs = models.IntegerField()
+    total_run_count = models.IntegerField()
+    last_run = models.DateTimeField()
+
+    last_updated = models.DateTimeField()
+
+    args = JSONField()
+    kwargs = JSONField()
```
pcreech: Removed the property implementation, and replaced it with a custom JSONField implementation

pcreech:
Updated from latest feedback.

It was decided to create a JSONField implementation instead of utilizing a Picklefield due to potential issues with pickling.

I've added a skeleton, broken JSONFIeld implementation for the moment, so comment can be made on the models/task.py items.  I'm currently actively working on the JSONField implementation.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+class JSONField(models.TextField):
+    def __init__(self, *args, **kwargs):
```
mhrivnak: Not sure you'll need to override this, but obviously this class is a WIP. Looks good.

pcreech: Yeah, I just committed a skeleton version at the moment.  Definitely WIP

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,96 @@
...
+    def from_db_value(self, value, expression, connection, context):
+        """
+        Converts a value as returned by the database to a Python object
+
+        :param value: The value to convert to Python
+        :type value: Any
+
+        :param expression:
+        :param connection:
+        :param context:
```
pcreech: I'm not sure what to do to document these variables.  This method signature is the way it is to override the django method, but I don't take advantage of these variables.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_db_prep_value(self, value, connection, prepared=False):
+        """
+        Converts value to a backend-specific value
+
+        :param value: The value to be converted
+        :type value: Any
+
+        :param connection:
+
+        :param prepared:
```
pcreech: I'm not sure what to do to document these variables.  This method signature is the way it is to override the django method, but I don't take advantage of these variables.

pcreech:
I've updated the PR with the JSONField implemenation

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Relations:
+
+    :cvar task: The task associated with this reservation
+    :type task: models.ForeignKey
+
+    :cvar worker: The worker associated with this reservation
+    :type worker: models.ForeignKey
+    """
+    resource = models.TextField()
```
pcreech: This needs to be unique, looks like i forgot to add it

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+
+    task = models.ForeignKey("Task", unique=True)
```
pcreech: Changing this to a 1-1 field, per Django's suggestion:

`pulp_platform.ReservedResource.task: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
        HINT: ForeignKey(unique=True) is usually better served by a OneToOneField.`

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,96 @@
...
+Custom django fields provided by pulp platform
+"""
+import json
+
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils.translation import six, ugettext_lazy as _
+
+
+def _parse_json(json_string):
```
jortel: Instead of a separate function, perhaps you could put this code in `JSONField.to_python()` and call `self.to_python()` from `JSONField.from_db_value()`.

pcreech: Good point.  The development was based off of 

https://docs.djangoproject.com/en/1.8/howto/custom-model-fields/#converting-values-to-python-objects

Which has separate code in the functions to catch different scenarios.  I originally had followed something similar, while now both functions are identical.  Should be no problem to move it into to_python, and we can always move it back out if needed.

platform/pulp/platform/fields.py
jortel: Might be prudent to call super() first and then decode that value.

Something like:
```
value = super(JSONField, self).from_db_value(value, expression, connection, context)
if isinstance(value, six.string_types):
    return self.to_python(value)
else:
    return value
```

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    :type total_run_count: models.IntegerField
+
+    :cvar last_run: Last time this schedule was ran
+    :type last_run: models.DateTimeField
+
+    :cvar last_updated: The last time this schedule was saved to the database
+    :type last_updated: models.DateTimeField
+
+    :cvar args: Arguments that should be passed to the apply_async function
+    :type args: models.TextField
```
jortel: JSONField

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+
+    :cvar kwargs: Keyword arguments that should be passed to the apply_async function
+    :type kwargs: models.TextField
```
jortel: JSONField

platform/pulp/platform/fields.py
pcreech: Per [the docs](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.from_db_value)

> For performance reasons, from_db_value is not implemented as a no-op on fields which do not require it (all Django fields). Consequently you may not call super in your definition.

jortel:
This **LGTM** but want @bmbouter to add the tag when he's finished with his review.

I think the `JSONField` needs unit tests.

You should consider adding some tests for the models.  Mainly to provide working examples for how to interact with the json fields.

pcreech:
> I think the JSONField needs unit tests.
>
> You should consider adding some tests for the models. Mainly to provide working examples for how to interact with the json fields.

Working on this now, just wanted to make sure things settled down first

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
+"""
```
bmbouter: What do you think about having this in `pulp.platform.models.fields` perhaps? It's for the models layer so I figured it would be in there somewhere.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+import json
+
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils.translation import six, ugettext_lazy as _
+
+
+class JSONField(models.TextField):
+    """
+    A custom django field to serialize dicts into a text field
```
bmbouter: s/django/Django/
s/text field/text field and vice versa/

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+
+    """
+    def from_db_value(self, value, expression, connection, context):
+        """
+        Converts a value as returned by the database to a Python object
+
+        :param value: The value to convert to Python
+        :type value: Any
```
bmbouter: Any isn't a type but everything in Python is an object.   s/Any/object/  ?

platform/pulp/platform/fields.py
bmbouter: Consider these two resolutions to this docstring issue. One would be to adjust the function signature to be:

`from_db_value(self, value, *args, **kwargs):`

And then document it with:

```
:param args: unused positional arguments
:type args: list
:param kwargs: unused keyword arguments
:type kwargs: dict
```
This has the advantage of being resilient to most signature changes to the function as Django evolves. I've had this problem before with Django specifically. I think I prefer this option.

The other option is to link to the docs. If you do that I recommend pointing at latest and not just the 1.8 version since one day we will upgrade.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :param expression:
+        :param connection:
+        :param context:
+
+        :return: A python representation of value
+        :rtype: Any
```
bmbouter: s/Any/object/

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+        Converts a value as returned by the database to a Python object
+
+        :param value: The value to convert to Python
+        :type value: Any
+
+        :param expression:
+        :param connection:
+        :param context:
+
+        :return: A python representation of value
```
bmbouter: s/python/Python/

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+        :rtype: Any
+        """
+        if isinstance(value, six.string_types):
+            return self.to_python(value)
+
```
bmbouter: I generally don't put whitespace before the return statement. Pep 8 recommends using whitespace sparingly to separate logical blocks of code. It's a preference thing so this is not a required change but I wanted to mention it. It's possible someone would miss the return value since they already see one above.

platform/pulp/platform/fields.py
bmbouter: Similar recommendation here as above.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+        return value
+
+    def to_python(self, value):
+        """
+        Converts the value into the correct Python object
+
+        :param value: The value to convert to Python
+        :type value: Any
+
+        :return: A python representation of value
```
bmbouter: s/python/Python/
I'll stop commenting on these, but similar changes throughout

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+            return self.to_python(value)
+
+        return value
+
+    def to_python(self, value):
+        """
+        Converts the value into the correct Python object
+
+        :param value: The value to convert to Python
+        :type value: Any
```
bmbouter: s/Any/object/
I'll stop commenting on these, but similar changes throughout.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :return: A python representation of value
+        :rtype: Any
```
bmbouter: This should come with a `:raises ValidationError:` line documenting the exception possibility.

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        try:
+            return json.loads(value)
+        except Exception:
```
bmbouter: This would be better to be more restrictive. dumps() raises `ValueError` in situations where it couldn't deserialize due to a validation problem. s/Exception/ValueError/

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+            raise ValidationError(_("Invalid JSON stored in database"))
```
bmbouter: Pulp does this pattern in a lot of places and it is difficult to debug. I think anytime we catch an exception and mutate it we should log the exception at error level. Assuming you have a logger as `_logger` you could add the line which will log the entire traceback and the bad data as well.

`_logger.exception(_('The field "value" cannot be JSON deserialized. "value" has the contents: %s') % value)`

platform/pulp/platform/fields.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :param prepared:
+
+        :return: json string representing the object
+        :rtype: str
+        """
+        if value is None:
+            return None
+
+        return json.dumps(value)
```
bmbouter: Is the function we are overriding prepared to have a `ValueError` raised or should this be caught and transformed into another exception type similar to above?

Either way this method probably needs a `:raises:` statement in the docblock.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (ACCEPTED, 'Accepted'),
+        (RUNNING, 'Running'),
+        (SUSPENDED, 'Suspended'),
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
+        (CANCELED, 'Canceled')
+    )
+    type = models.TextField()
```
bmbouter: We can't use this name `type` because it's a reserve Python keyword.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    ERROR = 'error'
+    CANCELED = 'canceled'
+    STATES = (
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (ACCEPTED, 'Accepted'),
+        (RUNNING, 'Running'),
+        (SUSPENDED, 'Suspended'),
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
```
bmbouter: I'm hoping `Error` can be `Errored` so match the tense of all other states. I'm making this change on my progress reporting models too.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    FINISHED = 'finished'
+    ERROR = 'error'
+    CANCELED = 'canceled'
+    STATES = (
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (ACCEPTED, 'Accepted'),
+        (RUNNING, 'Running'),
+        (SUSPENDED, 'Suspended'),
+        (FINISHED, 'Finished'),
```
bmbouter: In the progress reporting models I'm changing `Finished` to `Completed` so that it conveys the idea that it finished without error. This is something I'm thinking about and not a required change.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    :type worker: models.ForeignKey
+    """
+
+    WAITING = 'waiting'
+    SKIPPED = 'skipped'
+    ACCEPTED = 'accepted'
+    RUNNING = 'running'
+    SUSPENDED = 'suspended'
+    FINISHED = 'finished'
+    ERROR = 'error'
```
bmbouter: I'm thinking about s/error/errored/ to match the tense of the others. I'm making this change in the progress reporting models too.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    :cvar worker: The worker that this task is in
+    :type worker: models.ForeignKey
+    """
+
+    WAITING = 'waiting'
+    SKIPPED = 'skipped'
+    ACCEPTED = 'accepted'
+    RUNNING = 'running'
+    SUSPENDED = 'suspended'
+    FINISHED = 'finished'
```
bmbouter: s/FINISHED/COMPLETED/  ?
s/finished/completed/  ?

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+class Task(Model):
+    """
+    Represents a task
+
+    Fields:
+
+    :cvar task_type: The type of task this is
+    :type task_type: models.TextField
+
+    :cvar task_group: The group this task belongs to
```
bmbouter: s/task_group/group/ to match the code.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    :type task_group: models.UUIDField
```
bmbouter: s/task_group/group/ to match the code.

platform/pulp/platform/models/task.py
bmbouter: Going back to `task_type` would also match the docstring. That is my recommendation.

platform/pulp/platform/models/task.py
seandst: In this case I think I'm not seeing anything on the model that would be useful in creating a natural key, since `task_type` combined with `task_group` is definitely not going to be unique, and there are enough potential problems (time collisions, mainly) using the `started_at` stamp that it's probably best to leave it out of the natural key. I don't think that leaves enough for constructing a proper natural key.

platform/pulp/platform/models/task.py
bmbouter: Actually this can just be removed. We don't use this anymore and it's been deprecated already.

http://docs.pulpproject.org/dev-guide/integration/rest-api/tasks.html#task-report

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    group = models.UUIDField()
+    state = models.TextField(choices=STATES)
+
+    started_at = models.DateTimeField()
+    finished_at = models.DateTimeField()
+
+    error = JSONField()
+    result = JSONField()
+
+    parent = models.ForeignKey("Task", null=True)
```
bmbouter: This makes sense for the data model, but I want to point something out. In terms of the purpose of this field to present this data to the user the parent has to name its children. Leaving it as-is is right for the data layer, but we need to have `spawned_tasks` or some similar name be returned when a Task is serialized through the API. We should capture that somehow or somewhere. I'm not sure what to do about that.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    worker = models.ForeignKey("Worker")
```
bmbouter: There is a time when a Task is in the DB, but Worker assignment is not yet known. This needs to allow null I think.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    :type args: models.JSONField
+
+    :cvar kwargs: Keyword arguments that should be passed to the apply_async function
+    :type kwargs: models.JSONField
+    """
+    task = models.TextField()
+    enabled = models.BooleanField(default=True)
+    resource = models.TextField(null=True)
+
+    consecutive_failures = models.IntegerField()
```
bmbouter: The feature that requires this is being removed from 3.0 so this can be removed from the data layer.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    failure_threshold = models.IntegerField()
```
bmbouter: The feature that requires this is being removed from 3.0 so this can be removed.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+
+    iso_schedule = models.TextField()
```
bmbouter: This probably isn't a data layer concern except that I was hoping for more validation of this field.

For 3.0 we should contemplate if we want to continue to use ISO8601. I haven't met a person yet who can correctly form their intended schedule on the first try after reading the ISO8601 docs. Something like what cron uses would be so much better.

What do you think? Maybe we should send this to pulp-dev mailing list for comment?

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,231 @@
...
+    schedule = models.TextField(null=True)
```
bmbouter: This says it's stored as a pickle field but we're going to store it as a `TextField`. This should work but I wanted to mention it. Maybe it will need some serialization/deserialization. We can also handle this when we hook code up to the models then since it probably will show a tangible version of the problem at that time.

bmbouter:
Here are two things that we should talk about how to handle:

1) In 3.0 we want to maintain the AMQP based signaling on each save of a Task. That is done in 2.y [here](https://github.com/bmbouter/pulp/blob/6868481c1a0eeab06d1f18330e7fd7dc533d8007/server/pulp/server/db/model/__init__.py#L567-L567). That would likely be a overridden save() method which calls super() and then sends the message. We could leave the send_message implementation blank for now and file an issue to wire that up later.

2) The reaper Mixin model which reaps Tasks along with a bunch of other model types. This is likely it's own separate problem so I would be satisfied if an issue was created to make sure this works in 3.0 with our new models.

bmbouter:
@pcreech I'm done w/ my review. I left a bunch of comments. Functionally this is a great start and will likely be easy to make work with the existing tasking system code. Want to fix some of the things and answer some of the questions and ping me with quesitons or to do another pass. I expect the next pass to be the last one. Thanks for this great work!

pcreech:
> 1) In 3.0 we want to maintain the AMQP based signaling on each save of a Task. That is done in 2.y here. That would likely be a overridden save() method which calls super() and then sends the message. We could leave the send_message implementation blank for now and file an issue to wire that up later.

In response to this statement specifically, we should avoid putting any code not related directly to managing the model in the model.

Django appears to have a robust system of signal handlers we can wire functions into to handle this situation.  Most importantly, in this case, is the post_save signal outlined [here](https://docs.djangoproject.com/en/1.8/ref/signals/#post-save)

When the person handling the signal wire-up does that task, they can take advantage of this to handle it.

platform/pulp/platform/models/task.py
pcreech: I'm a fan for pulp-dev'ing this convo.  ISO date formats have always seemed, to me, more of a way of 'serializing/deserializing' date data for machine consumption than allowing a human-friendly way of providing this information.

IMO, We could easily provide a more user-friendly scheduling system to work with for 3.0

platform/pulp/platform/models/task.py
pcreech: From the look of the docs, this relationship can be [followed backwards](https://docs.djangoproject.com/en/1.8/topics/db/queries/#following-relationships-backward) i.e. the parent will have a task_set value that you can use to get the children (and somewhere else in the code can be used to convert this into something useful for the api)  We should be able to rename it via related_name to something more intuitive

platform/pulp/platform/models/task.py
pcreech: Will remove, but as a point for future implementations, I shall provide @jortel 's quote on the Repository model PR about "type" variable names as opposed to using a redundant {classname}_type variable.

> Class and instance attributes are namespaced to the class (as I'm sure you know) and implicitly apply to the class or instance. The type() builtin function (not a reserved word) unfortunately exists in the global namespace. I'm opposed to redundantly qualifying it. Naming it something like repository_name does not add clarity and leads to awkward semantics like: repository.repository_type.
https://docs.python.org/2/library/functions.html#type

platform/pulp/platform/fields.py
pcreech: Is it common practice to document all possible exceptions that could be thrown from dependent code that you didn't write?  I'm more familiar with having to document things you specifically throw, but not the case of all possible exceptions from third parties.

platform/pulp/platform/fields.py
pcreech: After careful evalution, I'm thinking removing the exception handling here instead.  All this is doing is hiding information and re-raising, which (especially in this early stage) we don't want.  Especially here where we have invalid json stored in the db somehow.  My thought is to remove this, and let it bubble up to be handled by the normal mechanisms.  One example I saw only caught the ValueError to explicitly ignore it instead.

Thoughts?

platform/pulp/platform/fields.py
pcreech: I'm not to keen on it.  Looking at other code, it seems standard to have a separate fields.py and models.py  (While we use a package instead of a module, this is the reason we import models in the __init__.py iirc)

platform/pulp/platform/models/task.py
pcreech: I'll leave it as is for now, but we could easily turn in into a JSONField as well if needed.

pcreech:
Updated.  I tweaked some field defs as well for things like nullables or to keep defs in line with each other (i.e. strings instead of types for related fields)

platform/pulp/platform/fields.py
bmbouter: My comment wasn't well written. Yes fields.py and models.py should be separate. Where do Django app layouts recommend the fields.py live?

platform/pulp/platform/models/task.py
bmbouter: Oh yes I remember related_name. Is this something we would define on the class attribute L#141 or at the API representation level?

bmbouter: Ignore this comment, you already fixed it.

bmbouter: The `spawned_tasks` needs to go on `parent` not on `worker`.

platform/pulp/platform/fields.py
pcreech: The examples I've seen put them in the same folder as models.py

bmbouter: Cool I'm satisfied with this.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,218 @@
...
+    task = models.TextField()
+    enabled = models.BooleanField(default=True)
+    resource = models.TextField(null=True)
+
+    iso_schedule = models.TextField()
+    schedule = models.TextField(null=True)
+
+    first_run = models.DateTimeField()
+
+    remaining_runs = models.IntegerField(null=True)
```
bmbouter: This field is also able to be deleted and won't be used in 3.0 I believe. As far as I know we don't meaningfully rely on it even in 2.y. The only [example in the docs](http://docs.pulpproject.org/dev-guide/conventions/scheduled.html?highlight=remaining_runs) shows it as null.

We could delete it now and if we need it we can bring it back.

platform/pulp/platform/models/task.py
```diff
@@ -0,0 +1,218 @@
...
+    :type kwargs: models.JSONField
+    """
+    task = models.TextField()
+    enabled = models.BooleanField(default=True)
+    resource = models.TextField(null=True)
+
+    iso_schedule = models.TextField()
+    schedule = models.TextField(null=True)
+
+    first_run = models.DateTimeField()
```
bmbouter: Can this be grouped in with the `last_run` field since they go together? Also the docstrings could move similarly.

bmbouter:
I left a few more comments. You have a great way of not missing any of them so I'm going to go ahead and add the LGTM flag now. @pcreech great great work. Thank you so much! This is going to be awesome. :cat2: :fireworks: :grapes:

PR #2707, opened 2016-08-16
Merge master into 3.0

bmbouter:


seandst:
It looks like all the platform stuff gets deleted on 3.0-dev with this, which doesn't seem right

bmbouter:
@seandst it only deletes some of the platform stuff because other parts were added onto 3.0-dev. I'm trying to clean it up and will push a new version soon.

bmbouter:
I believe this is the merge that we want. The conflicts resolved with my merge commit I believe are right. There was only 1 additoinal cherry pick of @seandst 's commit 17467a7 to retain all authorship. There are no commits on 3.0-dev that are missing from here, you can see with:

https://github.com/bmbouter/pulp/compare/merge_master_into_3.0...pulp:3.0-dev

PR #2708, opened 2016-08-17
Adds docs for re-using files on disk when DB is lost

mhrivnak:
fixes #236
https://pulp.plan.io/issues/236

### DIFF
--- a/docs/user-guide/deferred-download.rst
+++ b/docs/user-guide/deferred-download.rst
@@ -326,6 +326,41 @@ downloading.
     access to both the Content Sources configuration directory and the ``file://`` URL.
 
 
+Re-Using Files On Disk When DB Is Lost
+--------------------------------------
+
+Given a situation where the database has been lost, but the files within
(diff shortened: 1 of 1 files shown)


docs/user-guide/deferred-download.rst
```diff
@@ -326,6 +326,41 @@ downloading.
...
+   their integrity, only downloading a file from the remote source if it is not
+   found on disk. From the command line::
+
+    pulp-admin repo download --repo-id=myrepo --verify-all
+
+#. Change the download policy of the repository to whichever policy you
+   want to use long-term.
+
+.. note::
+    There is a known issue that will prevent the units from having their
```
jeremycline: `known issue` could be a named link to the issue rather than having it be a link at the bottom of the note. I don't mind one way or the other, though.

mhrivnak: Good suggestion. I'll make the change.

PR #2701, opened 2016-08-15
Issue #2134 - Removing option from dist config which is not in it

BrnoPCmaniak:
closes #2134
https://pulp.plan.io/issues/2134

PR #2709, opened 2016-08-17
Fix crash from removing missing option from dist conf

BrnoPCmaniak:
Fixed crash that arise when we try to remove an option from
distributor confing which was not before set. This ends up
in KeyError because this option was not in config dict.
This was fixed and also added TestCase for it.

Summary:
* Fixed pulp.server.controllers.distributor.update to ignore
missing options from distributor config
* Added TestCase for this behaviour
TestUpdate.test_remove_none_values_non_existing()

closes #2134
https://pulp.plan.io/issues/2134

Reopen of https://github.com/pulp/pulp/pull/2701

### DIFF
--- a/server/pulp/server/controllers/distributor.py
+++ b/server/pulp/server/controllers/distributor.py
@@ -208,7 +208,7 @@ def update(repo_id, dist_id, config=None, delta=None):
 
     for k, v in config.iteritems():
         if v is None:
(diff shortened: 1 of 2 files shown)


ipanova:
@BrnoPCmaniak when you close a PR and re-open it against different branch don't forget to update the issue with new link to the PR

PR #2710, opened 2016-08-18
Add a new settings file for Django

jeremycline:
In order to allow for easy configuration, this patch loads settings from
`/etc/pulp/server.yaml`, merges them with default settings for Pulp, and
then applies them to settings.py (which is in turn merged with the
global default settings).

fixes #2109

Reviewer note: Tests are incoming, but I wanted to see what people thought about a) switching to YAML and b) allowing users to set _any_ Django-supported setting

### DIFF
--- a/platform/etc/pulp/server.yaml
+++ b/platform/etc/pulp/server.yaml
@@ -1,0 +1,44 @@
+# Pulp server configuration
+#
+# Values shown are the default values used, unless otherwise indicated.
+
+
+# `allowed_hosts`: A list of strings representing the host/domain names that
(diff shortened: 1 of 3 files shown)


jeremycline:
There are plenty of other options I've found https://code.djangoproject.com/wiki/SplitSettings. One thing we could do is just use Python for the configuration file and import from it, which is also appealing to me.

seandst:
This is almost exactly what I had in mind. It might be ridiculous, but we should probably have a task in redmine for converting server.conf in ini format to yaml. :\

PR #2712, opened 2016-08-20
Adds Progress Reporting models into pulp.platform.models.

bmbouter:
This creates three new models: ProgressBar, ProgressReport,
and ProgressSpinner. These can be used to create progress reports.

https://pulp.plan.io/issues/2092
closes #2092

### DIFF
--- a/platform/pulp/platform/models/progress.py
+++ b/platform/pulp/platform/models/progress.py
@@ -1,0 +1,233 @@
+"""
+Django models related to Progress Reporting
+"""
+from gettext import gettext as _
+import logging
+
+from django.db import models
+
+from pulp.platform.models import Model, Task
+from pulp.server.async.tasks import get_current_task_id
+
+
+__all__ = [
+    'ProgressBar',
+    'ProgressSpinner',
+]
+
+
+_logger = logging.getLogger(__name__)
+
+
+class ProgressReport(Model):
+    """
+    A base model for all progress reporting.
+
+    All progress reports have a message, state, and are related to a Task.
+
+    Fields:
+
+    :cvar message: A short message for the progress update, typically shown to the user. (required)
+    :type message: models.TextField
+
+    :cvar state: The state of the progress update. Defaults to `WAITING`. This field uses a limited
+                 set of choices of field states. See `STATES` for possible states.
+    :type state: models.TextField
+
+    :cvar total: The total count of items to be handled by the ProgressBar (required)
+    :type total: models.IntegerField
+
+    :cvar done: The count of items already processed. Defaults to 0.
+    :type done: models.IntegerField
+
+    Relations:
+
+    :cvar task: The task associated with this progress report. If left unset when save() is called
+                it will be set to the current task_id.
+    :type task: models.ForeignKey
+    """
+    WAITING = 'waiting'
+    SKIPPED = 'skipped'
+    RUNNING = 'running'
+    COMPLETED = 'completed'
+    ERRORED = 'errored'
+    CANCELED = 'canceled'
+    STATES = (
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (RUNNING, 'Running'),
+        (COMPLETED, 'Completed'),
+        (ERRORED, 'Errored'),
+        (CANCELED, 'Canceled')
+    )
+    message = models.TextField()
+    state = models.TextField(choices=STATES, default=WAITING)
+
+    total = models.IntegerField(null=True)
+    done = models.IntegerField(default=0)
+
+    task = models.ForeignKey("Task", on_delete=models.CASCADE)
+
+    def save(self, *args, **kwargs):
+        """
+        Auto-set the task_id if running inside a task
+
+        If the task_id is already set it will not be updated. If it is unset and this is running
+        inside of a task it will be auto-set prior to saving.
+
+        :param args: positional arguments to be passed on to the real save
+        :type args: list
+        :param kwargs: keyword arguments to be passed on to the real save
+        :type kwargs: dict
+        """
+        if self.task_id is None:
+            self.task_id = Task.objects.get(id=get_current_task_id())
+        super(ProgressReport, self).save(*args, **kwargs)
+
+
+    def __enter__(self):
+        """
+        Saves the progress report state as RUNNING
+        """
+        if self.state != self.RUNNING:
+            self.state = self.RUNNING
+            self.state.save()
+
+    def __exit__(self, type, value, traceback):
+        """
+        Update the progress report state to COMPLETED or ERRORED.
+
+        If an exception occurs the progress report state is saved as ERRORED and the exception is
+        not suppressed. If the context manager exited without exception the progress report state
+        is saved as COMPLETED.
+
+        See the context manager documentation for more info on __exit__ parameters
+        """
+        if type is None:
+            self.state = self.COMPLETED
+            self.save()
+        else:
+            self.state = self.ERRORED
+            self.save()
+
+
+class ProgressSpinner(ProgressReport):
+    """
+    Shows progress reporting when the count of items is not known.
+
+    Plugin writers should create these objects to show progress reporting of a single step or
+    aspect of work which has a name and a state. For example:
+
+        >>> ProgressSpinner(message='Publishing Metadata')  # default state is 'waiting'
+
+        >>> ProgressSpinner(message='Publishing Metadata', state='running')  # specify 'running'
+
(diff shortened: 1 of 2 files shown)


platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            if self.state != self.RUNNING:
+                self.state = self.RUNNING
+                self.state.save()
+            return func(*args, **kwargs)
+        except Exception:
+            self.state = self.ERROR
+            self.save()
+            raise
+        finally:
```
bmbouter: This should should be an `else` clause

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+        Fields:
+
+    :cvar max: The total count of items to be handled by the ProgressBar (required)
+    :type max: models.IntegerField
+
+    :cvar current: The count of items already processed. Defaults to 0.
+    :type current: models.IntegerField
+    """
+    max = models.IntegerField()
+    current = models.IntegerField(default=0)
```
bmbouter: This should not have a default probably. It should be a required parameter.

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+        (CANCELED, 'Canceled')
+    )
+    message = models.TextField()
+    state = models.TextField(choices=STATES, default=WAITING)
+    task = fields.GenericRelation(Task)
+
+    class Meta:
+        abstract = True
+
+    def call_and_observe(self, func, args, kwargs):
```
bmbouter: This should become a decorator. Thanks for the idea @jortel

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    :cvar max: The total count of items to be handled by the ProgressBar (required)
+    :type max: models.IntegerField
+
+    :cvar current: The count of items already processed. Defaults to 0.
+    :type current: models.IntegerField
+    """
+    max = models.IntegerField()
+    current = models.IntegerField(default=0)
+
+    def next(self):
```
bmbouter: s/next/increment/   Call it this because it doesn't implement the Python iterator interface since it never raises `StopIterator`

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    'ProgressBar',
+    'ProgressReportingBase',
+    'ProgressSpinner',
+]
+
+
+_logger = logging.getLogger(__name__)
+
+
+class ProgressReportingBase(Model):
```
bmbouter: s/ProgressReportBase/ProgressReport/

bmbouter: Verify that you cannot save the abstract model itself.

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Fields:
+
+    :cvar max: The total count of items to be handled by the ProgressBar (required)
+    :type max: models.IntegerField
+
+    :cvar current: The count of items already processed. Defaults to 0.
+    :type current: models.IntegerField
+    """
+    max = models.IntegerField()
```
bmbouter: s/max/total/

platform/pulp/platform/models/progress.py
bmbouter: s/current/completed/

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    RUNNING = 'running'
+    FINISHED = 'finished'
+    ERROR = 'error'
+    CANCELED = 'canceled'
+    STATES = (
+        (WAITING, 'Waiting'),
+        (SKIPPED, 'Skipped'),
+        (RUNNING, 'Running'),
+        (FINISHED, 'Finished'),
+        (ERROR, 'Error'),
```
bmbouter: This is in present tense, but it likely should be in past tense since it's a terminal state.
s/Error/Failed/

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    Relations:
+
+    :cvar task: The task associated with this progress report. If left unset when save() is called
+                it will be set to the current task_id.
+    :type task: models.ForeignKey
+    """
+    WAITING = 'waiting'
+    SKIPPED = 'skipped'
+    RUNNING = 'running'
+    FINISHED = 'finished'
```
bmbouter: s/FINISHED/COMPLETED/

platform/pulp/platform/models/progress.py
mhrivnak: This looks easy to use. Another helpful way to use this would be as a context manager. I think the use would be something like:

```
report = ProgressSpinner(...)
with report:
    do_stuff()
```

Then you could just put the code in the ``try`` block below in an ``__enter__`` method, and everything else in the ``__exit__`` method.

The ``__enter__`` method could optionally return something to track a loop, similar to what you've done below with  ``ProgressBar``. A from-the-hip rough-draft of use:

```
things = get_me_some_things()
report = ProgressBar(max=len(things))
with report as loop_watcher:
    for thing in loop_watcher(things):
        do_stuff(thing)
```

or some might prefer this:

```
things = get_me_some_things()
report = ProgressBar(max=len(things))
with report as loop_watcher:
    for thing in things:
        do_stuff(thing)
        loop_watcher.item_done()
```

mhrivnak:
I'll just observe that the task as written says "Once the model is done, make any required tasks to create an internal API for plugins to use it." You're diving right into the API part, which might make sense to do now if you're finding that to be easiest, but don't feel obligated.

platform/pulp/platform/models/progress.py
mhrivnak: s/Error/OMG/

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    Shows progress reporting when the count of items is not known.
+
+    Plugin writers should create these objects to show progress reporting of a single step or
+    aspect of work which has a name and a state. For example:
+
+    ProgressSpinner(message='Publishing Metadata') # default state is 'waiting'
+
+    ProgressSpinner(message='Publishing Metadata', state='running') # specify 'running' instead
+
+    # Update the state to FINISHED and save it
```
jeremycline: I like to format these sorts of comments like an interactive interpreter:
```
Update the state to ``FINISHED`` and save it:

    >>> metadata_progress = ProgressSpinner(message='Publishing Metadata', state='running')
    >>> metadata_progress.state = ProgressSpinner.FINISHED
    >>> metadata_progress.save()
```
I find it helps to clearly show Python code in a comment, but this is purely a personal preference.

platform/pulp/platform/models/progress.py
```diff
@@ -0,0 +1,195 @@
...
+    metadata_progress.state = 'finished'
+    metadata_progress.save()
+
+    The call_and_observe() method simplifies state management by providing automatic state
+    transitions for the RUNNING FINISHED and ERROR states.
+
+    ProgressSpinner objects are associated with a Task and auto-discover and populate the task id
+    when saved.
+    """
+    pass
```
jeremycline: Does it make sense for this to have a ``count``/``current`` field or similar that's allowed to be null? It's nice to see a number go up, even if you don't know how long it's going up for.

platform/pulp/platform/models/progress.py
bmbouter: I went for the context manager here. I may add it to ProgressBar also I'm not sure.

@jortel I went for the context manager instead of the decorator because it requires the user to create the spinner which was troublesome with the decorator approach. It allows for this:

```
report = ProgressSpinner(...)
with report:
    do_stuff()
```

Or this:

```
with ProgressSpinner(...):
    do_stuff()
```

platform/pulp/platform/models/progress.py
bmbouter: This is a great idea. I've adjusted all of my docstring examples to this. Thanks @jeremycline

platform/pulp/platform/models/progress.py
bmbouter: I decided to go with `done` since 'completed' is also the name of a state now.

jortel:
This `LGTM`. I'm not sold on `ProgressBar` and `ProgressSpinner` but can't think of anything better.

PR #2713, opened 2016-08-21
Fixes publish with predistributor logic

dkliban:
closes #2187
https://pulp.plan.io/issues/2187

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -993,7 +993,6 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
     :return: publish result containing information about the publish
     :rtype:  pulp.server.db.model.repository.RepoPublishResult
     """
-
     force_full = call_config.get('force_full', False)
     config_override = call_config.override_config
     last_published = conduit.last_publish()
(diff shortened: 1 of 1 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -1006,16 +1006,17 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
...
-    if predistributor_id and last_published:
+    if predistributor_id:
         predistributor = model.Distributor.objects.get_or_404(repo_id=repo_obj.repo_id,
                                                               distributor_id=predistributor_id)
         predistributor_last_published = predistributor["last_publish"]
-        if predistributor_last_published:
+        if predistributor_last_published and last_published:
             published_after_predistributor = last_published > predistributor_last_published
+        else:
+            published_after_predistributor = True
```
ipanova: i am still trying to get the general logic here, but i think this should be set to false

dkliban: I think that the else clause just needs to be removed.

ipanova: I tested with else clause set to False and task failed with an exception, i tested without else clause and task failed also.

ipanova: with set to True task has skipped result but, the logic is not correct for the case when predistributor_last_published is True and last_published is False. When i trigger rsync publish is has skipped result saying that pre_distributor has not been used,which is not true.

ipanova:
Tested 2 scenarios: 
1. create repo
2. add rsync dist
3. publish with rsync dist.
4. task should have skipped result

1. create repo
2. publish repo with predistributor
3 add rsync dist
4 publish with rsync dist
5. task should succeed and all publish steps should be performed

dkliban:
ok test

PR #2719, opened 2016-08-23
add content models.

jortel:
https://pulp.plan.io/issues/2098

Okay folks, a few updated semantics to discuss.

The *Unit* part of `ContentUnit` although firmly entrenched in the pulp glossary, is a little awkward.  The term *unit* is synonymous with *item*, *thing*, *object* (and others) which makes it a bit redundant.  Consider it in another context - like a class named `RepositoryUnit` or `TaskObject`.  I believe that a class name of: `Content` and table name of `content` seems more appropriate. 

*"Repositories contain content."*

Another new concept is `Artifacts`.  *"Content has associated artifacts"*.  Each artifact represents a file.  I originally named this `ContentFile` but this quickly lead to odd semantics such as:

```
file = ContentFile(content=content)
file.content        # This is a reference to Content, not the content of the file.
file.file           #  Hmm.. . so, a file contains a file .. odd.
```

The term `artifact` resolves this.

```
artifact = Artifact(content=content)
artifact.content  # yep the associated Content.
artifact.file     # yep, artifacts reference files.
```

See the `test_content.py` for more usage examples.

I tried to use the default `FileSystemStorage` with `FileField` but has some unsuitable behavior.
- some concurrency issues with *if exists* checks.
- Insists on storage paths being relative to `MEDIA_ROOT`
- Mangles storage paths to prevent file name collisions.

The `Storage` interface is straight forward so I just implemented what we needed.  Docstrings and tests to follow.

Also worth noting is the lack of `FileContent` and `SharedContent`.  The separation of Content and Artifacts makes the distinction of `FileContent` unnecessary.  As for `SharedContent`, We only have 1 use case for this (ostree) and I believe this should be handled by plugins.  Not in the platform.

An update to the `models.README.md` to follow.

### DIFF
--- a/platform/pulp/platform/models/content.py
+++ b/platform/pulp/platform/models/content.py
@@ -1,0 +1,134 @@
+"""
+Content related Django models.
+"""
+import hashlib
+
+from django.contrib.contenttypes import fields
+from django.db import models
+
+
+from pulp.platform.models import Model, MasterModel, Notes
+from pulp.platform.models.storage import StoragePath
+
+
+class Content(MasterModel):
+    """
+    A piece of managed content.
+
+    :cvar natural_key_fields: Tuple of natural fields.  Must be: models.Field.
+    :type natural_key_fields: tuple
+
+    Fields:
+
+    :cvar type: The content type.
+    :type type: models.TextField
+
+    Relations:
+
+    :cvar notes: Arbitrary information stored with the content.
+    :type notes: fields.GenericRelation
+    """
+    natural_key_fields = ()
+
+    type = models.TextField(blank=False, default=None)
+
+    notes = fields.GenericRelation(Notes)
+
+    def natural_key(self):
+        """
+        Get the model's natural key based on natural_key_fields.
+
+        :return: The natural key.
+        :rtype: tuple
+        """
+        return (getattr(self, f.name) for f in self.natural_key_fields)
+
+    def natural_key_digest(self):
+        """
+        Get the SHA-256 digest of the natural key.
+        The digest is updated with each field name followed by its value.
+        The field names are only necessary to preserve backward compatibility
+        with digests generated in pulp 2.
+
+        :return: The hex digest.
+        :rtype: str
+        """
+        h = hashlib.sha256()
+        for name in sorted(f.name for f in self.natural_key_fields):
+            h.update(name)
+            value = getattr(self, name)
+            if not isinstance(value, basestring):
+                h.update(str(value))
+            else:
+                h.update(value)
+        return h.hexdigest()
+
+
+class Artifact(Model):
+    """
+    A file associated with a piece of content.
+
+    Fields:
+
+    :cvar file: The stored file.
+    :type file: models.FileField
+
+    :cvar downloaded: The associated file has been successfully downloaded.
+    :type downloaded: BooleanField
+
+    :cvar published_path: The relative published path.
+    :type published_path: models.FileField
+
+    :cvar size: The size of the file in bytes.
+    :type size: models.IntegerField
+
+    :cvar md5: The MD5 checksum of the file.
+    :type md5: models.CharField
+
+    :cvar sha1: The SHA-1 checksum of the file.
+    :type sha1: models.CharField
+
+    :cvar sha224: The SHA-224 checksum of the file.
+    :type sha224: models.CharField
+
+    :cvar sha256: The SHA-256 checksum of the file.
+    :type sha256: models.CharField
+
+    :cvar sha384: The SHA-384 checksum of the file.
+    :type sha384: models.CharField
+
+    :cvar sha512: The SHA-512 checksum of the file.
+    :type sha512: models.CharField
+
+    Relations:
+
+    :cvar content: The associated content.
+    :type content: Content
+    """
+
+    # Note: The FileField does not support unique indexes and has
+    # other undesirable behavior and complexities.  Using a custom
+    # field should be investigated.
+
+    file = models.FileField(db_index=True, upload_to=StoragePath(), max_length=255)
+    downloaded = models.BooleanField(db_index=True, default=False)
+    published_path = models.TextField(db_index=True, blank=False, default=None)
+
+    size = models.IntegerField(blank=True, null=True)
+
+    md5 = models.CharField(max_length=32, blank=True, null=True)
+    sha1 = models.CharField(max_length=40, blank=True, null=True)
+    sha224 = models.CharField(max_length=56, blank=True, null=True)
+    sha256 = models.CharField(max_length=64, blank=True, null=True)
+    sha384 = models.CharField(max_length=96, blank=True, null=True)
+    sha512 = models.CharField(max_length=128, blank=True, null=True)
+
+    content = models.ForeignKey(Content, related_name='artifacts', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('content', 'published_path')
+
+    def delete(self, *args, **kwargs):
+        if self.downloaded:
+            self.file.delete()
+        super(Artifact, self).delete(*args, **kwargs)
--- a/platform/pulp/platform/models/storage.py
+++ b/platform/pulp/platform/models/storage.py
@@ -1,0 +1,255 @@
+import os
+import errno
+import shutil
+
+from datetime import datetime
+
+from django.conf import settings
+from django.core.files import File
+from django.core.files.storage import Storage
+from django.utils.deconstruct import deconstructible
+
+
+@deconstructible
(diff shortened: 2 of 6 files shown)


platform/pulp/platform/models/content.py
```diff
@@ -0,0 +1,94 @@
...
+    :cvar notes: Arbitrary information stored with the content.
+    :type notes: fields.GenericRelation
+    """
+    natural_key_fields = ()
+
+    type = models.TextField(blank=False, default=None)
+
+    notes = fields.GenericRelation(Notes)
+
+    def natural_key(self):
```
mhrivnak: Maybe this should be a property. I'm not a huge fan of properties in general, because it seems to really just save a pair of ```()``` while obscuring the fact that it's a derived value. That said, I thought I'd point out the option. I'd be happy either way.

platform/pulp/platform/models/content.py
```diff
@@ -0,0 +1,94 @@
...
+        return (getattr(self, f.name) for f in self.natural_key_fields)
+
+    def natural_key_digest(self):
+        h = hashlib.sha256()
+        for name in sorted(f.name for f in self.natural_key_fields):
+            h.update(name)
+            value = getattr(self, name)
+            if not isinstance(value, basestring):
+                h.update(str(value))
```
mhrivnak: I can already feel a non-ascii character bug in our future. Let's make sure there's a unit test for this where at least one value has a non-ascii character, and I suspect you'll have to change this a bit.

mhrivnak: On second thought, I suspect you had in mind things like numbers and dates here, so maybe it's ok. I guess it would be unexpected to have some non-string object as the value and have that object's string representation fail.

platform/pulp/platform/models/storage.py
```diff
@@ -0,0 +1,101 @@
...
+        return (directories, files)
+
+    def path(self, path):
+        return path
+
+    def size(self, path):
+        return os.path.getsize(path)
+
+    def url(self, path):
+        'file://{}'.format(path)
```
mhrivnak: Can you explain the reasoning for this?

platform/pulp/platform/models/content.py
jortel: Normally, I'd agree with you but the django serializer expects this to be a plain method.  See: https://docs.djangoproject.com/en/1.8/topics/serialization/#serialization-of-natural-keys

platform/pulp/platform/models/storage.py
jortel: According to the `Storage.url()` docstring - "Returns an absolute URL where the file's contents can be accessed directly by a Web browser."  This implementation probably needs a bit more work and likely needs to incorporate `settings.MEDIA_URL`.  I also don't know how this get's used by django.

mhrivnak:
This generally looks very good. I'm happy to see it get merged once others have weighed in.

platform/pulp/platform/models/content.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+    Relations:
+
+    :cvar content: The associated content.
+    :type content: Content
+
+    """
+
+    file = models.FileField(upload_to=StoragePath(), max_length=255)
```
jeremycline: In addition to the checksum, it'd be nice to have the file's size

platform/pulp/platform/models/content.py
jeremycline: I don't believe this is Python 3 compatible. Django returns Unicode codepoints and you cannot feed those to hashlib:
```
(pulp3) [vagrant@dev platform]$ python3 manage.py shell
Python 3.5.1 (default, Aug  9 2016, 15:35:51) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux
(InteractiveConsole)
>>> from pulp.platform.models import Repository
>>> repo = Repository(name='test')
>>> repo.save()
>>> repos = Repository.objects.filter()
>>> repos[0].name
'test'
>>> import hashlib
>>> h = hashlib.sha256()
>>> h.update(repos[0].name)
  File "/home/vagrant/.virtualenvs/pulp3/lib/python3.5/site-packages/Django-1.8.14-py3.5.egg/django/core/management/commands/shell.py", line 69, in handle
    self.run_shell(shell=options['interface'])
  File "/home/vagrant/.virtualenvs/pulp3/lib/python3.5/site-packages/Django-1.8.14-py3.5.egg/django/core/management/commands/shell.py", line 61, in run_shell
    raise ImportError
ImportError

During handling of the above exception, another exception occurred:

  File "<console>", line 1, in <module>
TypeError: Unicode-objects must be encoded before hashing
>>> 
```

If the value is not already raw bytes, it needs to be encoded.

Something like:
```
h = hashlib.sha256()
for name in sorted(f.name for f in self.natural_key_fields):
    h.update(name.encode('utf-8'))  # assuming name is also unicode
    value = getattr(self, name)
    # I honestly don't know the best way to be Python 2 and Python 3 compatible
    # Maybe this works? Or maybe this doesn't work at all and is stupid.
    if isinstance(value, bytes):
        h.update(value)
    else:
        h.update(value.encode('utf-8'))
    return h.hexdigest()
```

platform/pulp/platform/models/content.py
```diff
@@ -0,0 +1,94 @@
...
+            h.update(name)
+            value = getattr(self, name)
+            if not isinstance(value, basestring):
+                h.update(str(value))
+            else:
+                h.update(value)
+        return h.hexdigest()
+
+
+class Artifact(Model):
```
jeremycline: I know we said we didn't want to do drastic remodeling as part of this conversion, but I'm wondering if it would be good to use this model to hold (or maybe just reference) the various catalog models from Pulp 2.

The LazyCatalogEntry model for lazy sync is the ``file`` (we have that here), ``importer_id``, ``unit_id`` (since this relates to content we don't need that), ``unit_type_id``(``content`` contains the type so we don't need that either), ``url``, checksum + type (we have that here), ``revision`` (do we need that if we have transactions?), and ``data`` (currently unused).

So if we included the URL(s) the artifact came from and the importer(s) that pulled it in, we've duplicated the lazy catalog. If we added a weight to each of those URLs we have the alternate content sources (I think). Does that sound crazy?

jortel: We'd need to talk through this but the `Artifact` and `LazyCatalogEntry` represent very different things.  The `Artifact` is a file related to a content unit.  The LazyCatalogEntry describes how an artifact can be downloaded.  *One* artifact can be referenced by *many* catalog entries (one for each importer that can download it).

The modeling of the artifact does mean we need less information in the catalog entry.  Each entry can be reduced to something like:
```
class LazyCatalogEntry(Model):
    url = models.TextField(blank=False, default=None)

    artifact = models.ForeignKey('Artifact', related_name='catalog', on_delete=models.CASCADE)
    importer = models.ForeignKey('RepositoryImporter', on_delete=models.CASCADE)
```

The downloading code can find the artifact based on `Artifact.file.name` which is the file storage path.  Then iterate though `artifact.catalog` just to get importers and urls that are used to download it.  As you mentioned, the checksum etc can just be gotten from the artifact.

The `FileField` supports `db_index=True` but not `unique` which may require a little work.

platform/pulp/platform/models/content.py
jortel: Good idea.  Added.

platform/pulp/platform/models/content.py
jeremycline: Yeah, I think I talked myself into having a ``Catalog`` model that's stripped down and relates to an ``Artifact``.

platform/pulp/platform/models/content.py
```diff
@@ -0,0 +1,134 @@
...
+    :cvar content: The associated content.
+    :type content: Content
+    """
+
+    # Note: The FileField does not support unique indexes and has
+    # other undesirable behavior and complexities.  Using a custom
+    # field should be investigated.
+
+    file = models.FileField(db_index=True, upload_to=StoragePath(), max_length=255)
+    downloaded = models.BooleanField(db_index=True, default=False)
```
jeremycline: Does it make sense to have a field on this model, say ``cached``, to indicate it's been handled by the streamer? Currently the ``DeferredDownload`` collection is just the unit_id and unit_type_id. Then the periodic task that downloads the cached content just queries all artifacts with ``cached = True`` and gets to the catalog entries by way of the foreign key relationship between ``LazyCatalogEntry`` and ``Artifact``.

seandst:
I think this is a great starting point for getting our hands on the new Content models, and am prettty confident that the commented issues will get dealt with as we continue development on 3.0-dev. LGTM.

PR #2720, opened 2016-08-25
The predistributor's history is no longer considered if force_full is…

dkliban:
… passed in. The predistributor's history is also ignored any time posdistributor's configuration changes.

closes #2202
https://pulp.plan.io/issues/2202

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1022,14 +1022,19 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
             repo_id=repo_obj.repo_id,
             distributor_id=dist_id).update(set__last_override_config=config_override)
 
-    # Skip if a predistributor is configured and the predistributor has not published since the
-    # last publish. Skip if content has not changed since last publish and force_full has not been
-    # requested and no predistributor is defined.
+    # Check if a predistributor is configured and the predistributor has not published since the
+    # last publish.
     skip_for_predistributor = (predistributor_id and (published_after_predistributor or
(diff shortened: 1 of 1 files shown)


server/pulp/server/controllers/repository.py
```diff
@@ -1027,9 +1027,8 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
     # requested and no predistributor is defined.
     skip_for_predistributor = (predistributor_id and (published_after_predistributor or
                                                       not predistributor_last_published))
-    if skip_for_predistributor or \
-            (last_published and not force_full and not last_updated and not units_removed and
-             not dist_updated and same_override and not predistributor_id):
+    if not force_full and not dist_updated and same_override and (skip_for_predistributor or \
+            (last_published and not last_updated and not units_removed and not predistributor_id)):
```
ipanova: please consider to bring up the `last_published` in the front, by checking last_published first it will be no need to check force_full neither dist_updated in case repo was not published previously at all. 
```
if last_published and not force_full and not dist_updated and same_override and (skip_for_predistributor or \
        (not last_updated and not units_removed and not predistributor_id)):
```

ipanova:
Tested:
create repo
add iso dist and iso rsync dist
publish with iso rsync dist
observe no-op publish

publish with iso dist
publish with iso rsync dist
see operation publish of ^

publish again with iso rsync dist
see no-op publish

Checked force-full works on predistributor normally, checked force full works on rsync dist even if predistirbutor has not published after rsync publish.
Update of rsync dist config also causes operational publish even if predistirbutor has not published after rsync publish.

PR #2722, opened 2016-08-26
ansible: fix pulp_facts when selinux-policy not installed

rohanpm:
If selinux-policy package is not installed, rpmspec produces spurious
requirements for packages named "selinux-policy", "is", "not", "installed"
which breaks subsequent ansible tasks.  This is due to the macro in
pulp.spec:

  %define selinux_policyver %(rpm --qf "%%{version}-%%{release}" -q %%selinux-policy)

We know we'll need this package anyway, so just install it a bit
earlier to fix this.

### DIFF
--- a/playpen/ansible/roles/core/tasks/main.yml
+++ b/playpen/ansible/roles/core/tasks/main.yml
@@ -14,6 +14,7 @@
   dnf: name={{ item }} state=present
   with_items:
       - rpm-build
+      - selinux-policy
 
 - name: Gathering Pulp facts
   pulp_facts:


jeremycline:
This is fine with me, but we recently moved all our Ansible stuff to https://github.com/pulp/devel. If you could make a PR against that repository that would be great. It doesn't hurt to merge it here if you need it in this repository for 2.9, but it can't be merged forward to 2.10/master.

bmbouter:
@rohanpm +1 to moving this to pulp/devel When you do want to close this one?

As an aside, the spec files starting with pulp 3.0 will live in the repo https://github.com/pulp/packaging The line causing problems have been removed from from that repo altogether so this won't be an issue with pulp 3.0.

rohanpm:
OK, I was targeting it at the earliest applicable branch as standard practice, but I'm actually only interested in using the fix in 2.11.

So, closing this in favor of https://github.com/pulp/devel/pull/7

PR #2723, opened 2016-08-26
Implement PULP_MAX_TASKS_PER_CHILD

jkroepke:
https://pulp.plan.io/issues/2172


jeremycline:
I certainly don't have a problem with this. Anyone else want to weigh in? @bmbouter?

server/pulp/server/async/celery_instance.py
```diff
@@ -40,13 +40,21 @@
...
 celery.conf.update(CELERY_ACCEPT_CONTENT=['json'])
 
+"""
+Configures the celery option CELERYD_MAX_TASKS_PER_CHILD to avoid memory leaks.
+See: https://pulp.plan.io/issues/2172
+"""
+_PULP_MAX_TASKS_PER_CHILD = os.environ.get("PULP_MAX_TASKS_PER_CHILD")
+
+if _PULP_MAX_TASKS_PER_CHILD.isdigit() and _PULP_MAX_TASKS_PER_CHILD >= 1:
+    celery.conf.update(CELERYD_MAX_TASKS_PER_CHILD=_PULP_MAX_TASKS_PER_CHILD)
```
seandst: At this point, even though `_PULP_TASKS_PER_CHILD` is `isidigit`, it's still a string. So...this happens:

```python
>>> '1' > 1
True
```

So `_PULP_MAX_TASKS_PER_CHILD` is always `>=` 1.

This should maybe be converted from LBYL to EAFP, e.g.
```python
try:
    _PULP_MAX_TASKS_PER_CHILD = int(os.environ.get("PULP_MAX_TASKS_PER_CHILD", 1)  # if not 1 here, then whatever a sane default is
    celery.conf.update(CELERYD_MAX_TASKS_PER_CHILD=_PULP_MAX_TASKS_PER_CHILD
except (TypeError, ValueError):
    # do something reasonable...maybe explode?
    # it seems like silently failing if the value in the config makes no sense is a bad idea, so
    # probably raising a good exception explaining why the value for this envvar is wrong would be good?
    do_something()
```

jeremycline: Good catch, I didn't even notice this

jkroepke: Should I modify this PR?

jeremycline: Yup, you can either push additional commits, or you can amend this commit and force push it. Either way is fine with me. Also feel free to add yourself to the AUTHORS file!

bmbouter:
@jkroepke Thank you so much for this PR I had meant to comment on it earlier.

I want to suggest an alternate way to do this and encourage you to modify it. Relying on the environment variable could be problematic. Also we need to do the same for upstart environment and this only handles systemd. Here is my recommendation:

* Revert the edits to the celery_instance.py file.
* Adjust the default in PULP_MAX_TASKS_PER_CHILD to be 0. I was suggesting 2 for 3.0, but this would be for 2.y which needs to default to off (which I think of as 0).
* Add the PULP_MAX_TASKS_PER_CHILD configuration value with 0 default to Upstart configs, those are [here](https://github.com/pulp/pulp/blob/master/server/etc/default/upstart_pulp_workers).
* Have it use the `--maxrequestsperchild` argument to the worker. On systemd you can have it check the config value and add the argument if enabled [here](https://github.com/pulp/pulp/blob/0ec80d17/server/pulp/server/async/manage_workers.py#L25). For Upstart systems you can do the same thing [here](https://github.com/pulp/pulp/blob/d1735ec63109d19705b36e03c036521cd7126a19/server/etc/rc.d/init.d/pulp_workers).
* It needs a release note so that Pulp users will know about this
* consider filing an [issue against pulp-smash](https://github.com/PulpQE/pulp-smash/issues) so that someone can automate test cases related to this great addition.

Please ask questions and I will see them. You can also ping me in #pulp-dev on freenode.

jkroepke:
@bmbouter Thanks for your improvments. Hopefully I do it right.

Can someone review this?

Tasks TODO, if the code quallity is good enough:

- [ ] It needs a release note so that Pulp users will know about this
- [ ] consider filing an [issue against pulp-smash](https://github.com/PulpQE/pulp-smash/issues) so that someone can automate test cases related to this great addition.

AUTHORS
```diff
@@ -45,3 +45,4 @@ Florian Weimer (fw@deneb.enyo.de)
 Adam Young (ayoung@redhat.com)
 Lukas Zapletal (lzap+git@redhat.com)
 Bihan Zhang (bizhang@redhat.com)
+Jan-Otto Kröpke (mail@jkroepke.de)
```
bmbouter: Can you add a newline at the end of the file?

server/etc/default/systemd_pulp_workers
```diff
@@ -6,3 +6,7 @@
 
 # Configure Python's encoding for writing all logs, stdout and stderr
 PYTHONIOENCODING="UTF-8"
+
+# To avoid memory leaks, pulp can restart the worker after X tasks. Defaults to 0.
+# for more informations, see http://docs.celeryproject.org/en/latest/userguide/workers.html#max-tasks-per-child-setting
+PULP_MAX_TASKS_PER_CHILD=0
```
bmbouter: Can you add a newline at the end of the file?

server/etc/rc.d/init.d/pulp_workers
```diff
@@ -44,6 +44,7 @@ DEFAULT_LOG_LEVEL="INFO"
 DEFAULT_NODES="celery"
 DEFAULT_CELERYD="-m celery worker --detach"
 DEFAULT_PULP_CONCURRENCY=$(nproc --all)
+DEFAULT_PULP_MAX_TASKS_PER_CHILD=0
```
bmbouter: The PULP_MAX_TASKS_PER_CHILD already defines the default of 0 so this probably can be removed.and the logic below reworked to read from the PULP_MAX_TASKS_PER_CHILD itself.

server/etc/rc.d/init.d/pulp_workers
```diff
@@ -59,6 +60,10 @@ if [ ! -z "$CELERY_APP" ]; then
     CELERY_APP_ARG="--app=$CELERY_APP"
 fi
 
+# Sets --maxrequestsperchild argument for CELERY_BIN
+PULP_MAX_TASKS_PER_CHILD=${PULP_MAX_TASKS_PER_CHILD:-$DEFAULT_PULP_MAX_TASKS_PER_CHILD}
+CELERY_MAX_TASKS_PER_CHILD_ARG="--maxrequestsperchild=$PULP_MAX_TASKS_PER_CHILD"
```
bmbouter: This will include `maxrequestsperchild` on all workers that are started. That is OK if a value of 0 leaves the feature disabled. [The docs](http://celery.readthedocs.io/en/latest/configuration.html#std:setting-CELERYD_MAX_TASKS_PER_CHILD) don't indicate what a value of 0 would do. Can you confirm that a default of 0 will disable the feature?

server/pulp/server/async/manage_workers.py
```diff
@@ -18,11 +18,12 @@
 
 [Service]
 EnvironmentFile=%(environment_file)s
+Environment=PULP_MAX_TASKS_PER_CHILD=0
```
bmbouter: This is no longer needed since the value is specified in L#26.

server/pulp/server/async/manage_workers.py
```diff
@@ -18,11 +18,12 @@
...
 User=apache
 WorkingDirectory=/var/run/pulp/
 ExecStart=/usr/bin/celery worker -n reserved_resource_worker-%(num)s@%%%%h -A pulp.server.async.app\
           -c 1 --events --umask 18 --pidfile=/var/run/pulp/reserved_resource_worker-%(num)s.pid\
-          --heartbeat-interval=30
+          --heartbeat-interval=30 --maxrequestsperchild=$PULP_MAX_TASKS_PER_CHILD
```
bmbouter: Similar question here about this being set 100% of the time with a default value of 0.

bmbouter:
@jkroepke This PR is very close to what it needs to be for acceptance. I left some comments. Can you look at them and make adjustments or ask more questions. Thank you again for your PR submission this will be a great feature for Pulp 2.y that you are making!

bmbouter:
@jkroepke I wanted to check in on this PR. If there is anything I can help with, please let me know.

jkroepke:
@bmbouter I still have to check, if there is a difference between `--maxrequestsperchild=0` and not define this parameter.

jkroepke:
Added improvments from @bmbouter 

Branch rebased.

bmbouter:
@jkroepke Thank you so much for the update. The upstart parts look mostly correct, but the systemd part needs to be updated some. The updates are really small so I'm going to merge this and make a PR on top of your commit with the adjustments.

Thank you again for the contribution! This is a really great feature which if I merge my other fix quickly will be included in Pulp 2.11.

PR #2724, opened 2016-08-26
handles mkdirs race condition

werwty:
closes #2047
https://pulp.plan.io/issues/2047


server/pulp/plugins/util/publish_step.py
```diff
@@ -45,6 +45,20 @@ def _post_order(step):
...
+    A wrapper for os.makedirs that handles the race condition of two directories being
+    created at the same time. An OSError is thrown if a file exists with the same name.
+
+    :param args: path[, mode] that goes to os.makedirs
+    :return:
+    """
+    try:
+        os.makedirs(*args)
+    except OSError:
+        if not os.path.isdir(args[0]):
```
dkliban: The newer style of exception handling is preferred. Python provides errno module that has all the error codes. The codes can vary from platform to platform. So it's better to use the module when checking IO errors and such. 

`except OSError as e:
    if e.errno != errno.EEXIST:
        raise`

dkliban: As discussed on IRC, let's just leave it as is to handle the case where a file exists with the same name as directory requested.

server/pulp/plugins/util/publish_step.py
```diff
@@ -904,7 +918,7 @@ def process_main(self, item=None):
             # Create the parent directory of the published repository tree, if needed
             publish_dir_parent = os.path.dirname(publish_location)
             if not os.path.exists(publish_dir_parent):
-                os.makedirs(publish_dir_parent, 0750)
+                _safe_makedirs(publish_dir_parent, 0750)
```
dkliban: why is the change in the mode needed?

jortel:
We already have an implementation of this:  https://github.com/pulp/pulp/blob/master/server/pulp/plugins/util/misc.py#L39

server/pulp/plugins/util/publish_step.py
```diff
@@ -45,6 +45,20 @@ def _post_order(step):
         yield v
     yield step
 
+def _safe_makedirs(*args):
```
jortel: If you're going to do `(*args)` instead of matching the signature in `os.makedirs()`, it would be better to do `(*args, **kwargs)` in case the user calls as `_safe_makedirs(path, mode=0775)`.

jortel:
Suggest updating one of the unit tests to pass `mode` and assert it get's passed on.

werwty:
updated commit to add a unit test for passing `mode`

PR #2725, opened 2016-08-30
Update coverage_hook script

elyezer:
Patch the pulp_streamer service file and pulp_workers service template in order
to properly capture coverage data.

### DIFF
--- a/playpen/coverage/coverage_hook
+++ b/playpen/coverage/coverage_hook
@@ -6,10 +6,13 @@ In addition to calling %(prog)s with --help, each subcommand has its own options
 that can also be seen by passing --help after the subcommand, e.g. %(prog)s <command> --help
 """
 import argparse
-import hashlib
 import glob
+import hashlib
+import importlib
 import os
+import platform
 import shutil
+import subprocess
 import sys
 import warnings
 
@@ -56,6 +59,9 @@ report the saved coverage reports and write the output to the specified director
 report will fail if coverage has not yet been installed
 """
 _root_epilog = "This script is intended to be run with root privileges."
+
+_DISTRIBUTION = platform.dist()[0]
+_DISTRIBUTION_MAJOR_VERSION = platform.dist()[1].split('.')[0]
 
 
 def _files():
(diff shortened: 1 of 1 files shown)


elyezer:
I've tested this on the vagrant machine and it is working, I have scheduled a provisioning for RHEL6, RHEL7 in order to make sure this will work on all platforms. In the mean time you can provide any feedback about the changes.

playpen/coverage/coverage_hook
```diff
@@ -57,6 +60,9 @@ report will fail if coverage has not yet been installed
 """
 _root_epilog = "This script is intended to be run with root privileges."
 
+_DISTRIBUTION = platform.dist()[0]
+_DISTRIBUTION_MAJOR_VERSION = platform.dist()[1].split('.')[0]
```
seandst: We've had problems with this not working in f24, so I think it'd be good to check the `platform.dist` returns are sane and fail quickly here if not.

elyezer: Did some testing and it appears to work:

```console
[jenkins@host-172-16-46-235 ~]$ python
iPython 2.7.12 (default, Aug  9 2016, 15:48:18) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux2
>>> import platform
>>> platform.dist()
('fedora', '24', 'Twenty Four')
>>> _DISTRIBUTION_MAJOR_VERSION = platform.dist()[1].split('.')[0]
>>> print(_DISTRIBUTION_MAJOR_VERSION)
24
>>> _DISTRIBUTION = platform.dist()[0]
>>> print(_DISTRIBUTION)
fedora
```

playpen/coverage/coverage_hook
```diff
@@ -200,6 +206,75 @@ def _patch_entry_module(install):
...
+        if _DISTRIBUTION == 'redhat' and _DISTRIBUTION_MAJOR_VERSION == '6':
+            subprocess.check_call([
+                'sed', '-i',
+                '/^VERSION=/ a CELERYD_OPTS="-P solo"',
+                '/etc/rc.d/init.d/pulp_workers',
+            ])
+        else:
+            subprocess.check_call([
+                'sed', '-i',
+                r's/\(--heartbeat-interval=[0-9]\+\).*$/\1 -P solo/',
```
Ichimonji10: The `.*` will drop all characters after `--heartbeat-interval=n`, correct? Is this intended? What happens if another option is after `--heartbeat-interval`?

playpen/coverage/coverage_hook
```diff
@@ -200,6 +206,75 @@ def _patch_entry_module(install):
...
+    """
+    mod = None
+    msg = None
+    try:
+        mod = importlib.import_module('pulp.server.async.manage_workers')
+        mod = mod.__file__.rstrip('co')
+    except ImportError:
+        msg = 'Module {} import failed'.format('pulp.server.async.manage_workers')
+    if msg:
+        print >>sys.stderr, msg + ', coverage reporting may be incomplete.'
```
Ichimonji10: This logic could be simplified to:

```python
    try:
        mod = importlib.import_module('pulp.server.async.manage_workers')
    except ImportError:
        print >>sys.stderr, 'Module {} import failed, coverage reporting may be incomplete.'.format('pulp.server.async.manage_workers')
        return
    mod = mod.__file__.rstrip('co')
    with open(mod) as file:
        ...
```

This is more of a coding style suggestion than anything else. Do what you like.

playpen/coverage/coverage_hook
elyezer: This patch needs to be updated if Pulp gets updated since we are editing the source code directly.

playpen/coverage/coverage_hook
elyezer: I've just copied from another function but hadn't took time to improve it. Thank you very much for the suggestion.

playpen/coverage/coverage_hook
Ichimonji10: Since we're editing a unit file that's tightly integrated with Pulp itself, I definitely see why this line of code may need to be frequently updated. That said, I still have a big question mark floating over my head.

Here's some content from  `pulp_resource_manager.service` file:

```
ExecStart=/usr/bin/celery worker -A pulp.server.async.app -n resource_manager@%%h\
          -Q resource_manager -c 1 --events --umask 18 --pidfile=/var/run/pulp/resource_manager.pid\
          --heartbeat-interval=30
```

Let's imagine that the above was updated to:

```
ExecStart=/usr/bin/celery worker -A pulp.server.async.app -n resource_manager@%%h\
          -Q resource_manager -c 1 --umask 18 --pidfile=/var/run/pulp/resource_manager.pid\
          --heartbeat-interval=30 --events
```

Wouldn't the current regex be problematic, because it would drop the `--events` flag?

```
s/\(--heartbeat-interval=[0-9]\+\).*$/\1 -P solo/
```

Wouldn't something like this make the regex more resilient?

```
s/\(--heartbeat-interval=[0-9]\+\)/\1 -P solo /
```

bmbouter:
I recommend not putting this workaround in place. I posted some about why [here](https://www.redhat.com/archives/pulp-dev/2016-August/msg00020.html) along with a recommendation that this be filed as a bug in the upstream Celery community.

playpen/coverage/coverage_hook
seandst: ....weeeeeeeird

elyezer:
Instead of workaround on Pulp we decided to actually fix the coverage report to be able to capture data even when the prefork POOL_CLS is used.

PR #2726, opened 2016-08-31
Add a DownloadCatalog model to track Artifacts

jeremycline:
Lazy sync has a LazyCatalogEntry collection Pulp 2. This covers the
LazyCatalogEntry functionality by relating a URL to an ``Artifacts``
and an ``Importer`` that contains the necessary download configuration
to access the URL (basic auth, client certificates, CA certificates,
etc).

closes #2088

### DIFF
--- a/app/pulp/app/models/catalog.py
+++ b/app/pulp/app/models/catalog.py
@@ -1,0 +1,48 @@
+"""
+Django models for artifact catalogs.
+"""
+from django.core import validators
+from django.db import models
+from django.utils.translation import ugettext as _
+
+from pulp.app.models import Model, Artifact, Importer
+
+
+class DownloadCatalog(Model):
+    """
+    Each :class:`DownloadCatalog` maps an :class:`Artifact` to a URL where
+    it is stored remotely and to the :class:`Importer` which contains the
+    network configuration required to access that URL.
+
+    Fields:
+
+    :cvar url: The URL used to download the related artifact.
+    :type url: django.db.models.TextField
+
+    Relations:
+
+    :cvar artifact: The artifact that is expected to be present at ``url``.
+    :type artifact: pulp.platform.models.Artifact
+
+    :cvar importer: The importer that contains the configuration necessary
+                    to access ``url``.
+    :type importer: pulp.platform.models.Importer
+    """
+    # Although there is a Django field for URLs based on CharField, there is
+    # not technically any limit on URL lengths so it's simplier to allow any
+    # length here and let the upstream server deal with the problem of size.
+    url = models.TextField(blank=True, validators=[validators.URLValidator])
+
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE)
+    importer = models.ForeignKey(Importer, on_delete=models.CASCADE)
+
+    def __str__(self):
+        """
+        Human-readable representation of this model.
+
+        This is helpful for interactive prompts and is used in Django's admin
+        interface.
+        """
+
+        return _('{artifact} is retrievable at {url} by {importer}'.format(
+            artifact=self.artifact, url=self.url, importer=self.importer))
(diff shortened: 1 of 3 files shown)


platform/pulp/platform/models/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+                    to access ``url``.
+    :type importer: pulp.platform.models.RepositoryImporter
+
+    artifact (file information)
+    importer (download configuration)
+    """
+    url = models.TextField()
+
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE)
+    importer = models.ForeignKey(RepositoryImporter, on_delete=models.CASCADE)
```
jeremycline: I initially wanted to relate to ``Importer``, but it's abstract. One thing I talked about a bit with jortel is the future of alternate content sources, and the possibility of having an importer that isn't associated with a ``Repository``. I'm not sure how forward-thinking we want to be here, but would it be a good idea to make the ``Importer`` model not abstract and relate to a plain ``Importer`` here?

platform/pulp/platform/models/content.py
```diff
@@ -112,6 +115,7 @@ class Artifact(Model):
 
     file = models.FileField(db_index=True, upload_to=StoragePath(), max_length=255)
     downloaded = models.BooleanField(db_index=True, default=False)
+    requested = models.BooleanField(db_index=True, default=False)
```
jeremycline: This particular field is primarily useful for the ``deferred_download`` task since it would just query for all artifacts that have ``downloaded = False`` and ``requested = True``, but it could also be helpful if people want to know what content is or isn't used. It could also be a number and we could track the number of times a piece of content is requested, but that adds the overhead of a database write for each request.

platform/pulp/platform/models/catalog.py
jortel: I *think* making the import non-abstract seems reasonable given this relationship and the likelihood of plans for alt-content sources.

jeremycline: So it turns out I can't make the ``Importer`` concrete because that breaks the ``unique_together = (repository, name)`` requirement on ``RepositoryImporter`` because ``repository`` and ``name`` live in different tables.

We _could_ condense ``RepositoryImporter`` and ``Importer`` and allow the ``repository`` foreign key to be null I guess, but that still breaks the natural key and uniqueness of ``(repository, name)``. I'm not sure if we want to tackle this problem now or later, though.

platform/pulp/platform/models/repository.py
```diff
@@ -214,7 +214,6 @@ class Importer(Plugin):
     scratchpad = fields.GenericRelation(Scratchpad)
 
     class Meta:
-        abstract = True
         unique_together = ('name', 'repository')
```
jortel: Just noticed that the `unique_together` should not be on the `Importer.Meta`.  These models went through several revisions and this `unique_together` was moved to `RepositoryImporter` along with the FK but I missed removing it here.  Please remove it.

jeremycline: So the unfortunate deal with this is that if ``Importer`` is not abstract, a table gets created for and now ``name`` is in the ``Importer`` table and ``repository`` is in the ``RepositoryImporter`` table. Unfortunately it's not possible to use ``unique_together`` with multi-table inheritance.

We had a conversation earlier today about all this, and I'm not sure whether we want to defer dealing with the Importer/RepositoryImporter and Distributor/RepositoryDistributor thing until later, or if we want to address it now.

app/pulp/app/models/catalog.py
```diff
@@ -0,0 +1,48 @@
...
+from django.core import validators
+from django.db import models
+from django.utils.translation import ugettext as _
+
+from pulp.app.models import Model, Artifact, Importer
+
+
+class DownloadCatalog(Model):
+    """
+    A catalog that tracks where an :class:`Artifact` originated from
```
jortel: Suggest something like:

Each catalog entry maps a URL  to an `Artifact` and to the `Importer` that contributed the entry.  The importer provides information needed to download the URL.

Since multiple importers can contribute catalog entries for the same artifact, the *originated from* is misleading.

PR #2729, opened 2016-09-02
skip rsync if there are no files to sync but also only if delete is F…

ipanova:
…alse

closes #2221
https://pulp.plan.io/issues/2221

Signed-off-by: Ina Panova <ipanova@redhat.com>


ipanova:
Hand tested on ISO repo example.
1. create iso repo and sync&publish it
2. add rsync dist
3. publish with rsync dist with --delete=True
4. check all files are on remote host
5. remove 1 file from repo and publish it
6. publish with rsync dist with --delete=True
7. check 1 file was removed on remote host
8. remove all files from repo and publish it
9. publish with rsync dist with --delete=True
10. check that there are no files on the remote host

server/pulp/plugins/rsync/publish.py
```diff
@@ -226,8 +226,11 @@ def rsync(self):
                   generated by rsync command)
         :rtype: tuple
         """
-        if not self.file_list:
+        if not self.file_list and not self.delete:
             return (True, _("Nothing to sync"))
+        if not os.path.exists(self.src_directory):
```
dkliban: In what situation would the path not exist?

ipanova: this condition is needed in the situation when an empty repo is being synced

PR #1823, opened 2015-04-24
Update docs for task notification messages

beav:
The previous docs were not suitable for someone who wanted to integrate with
Pulp but was not already familiar with TaskStatus objects.

This patch adds some examples and further explanation on task notifications and
how to use them.

PR #2730, opened 2016-09-02
Verify size of a file at its final location

goosemania:
re #1823
https://pulp.plan.io/issues/1823


server/pulp/plugins/util/verification.py
```diff
@@ -14,6 +14,9 @@ class VerificationException(ValueError):
 
 def verify_size(file_object, expected_size):
     """
+    Depricated. Do not use it anymore. Use os.stat to find out the file size to compare
```
jortel: :+1:

PR #2731, opened 2016-09-02
Merge master to 3.0 dev

bmbouter:
There was one modify/delete conflict with this on the pulp.spec file. I resolved it on 3.0-dev by removing the pulp.spec file since we don't want spec files there anymore.


asmacdo:
Since this is just a merge forward I dont think it needs to go through the normal review process. Still, I glanced over it and all looks fine!

PR #2732, opened 2016-09-04
Fix syslog facility setting

goosemania:


jeremycline:
I thought our log handler would be kind to strings, but I guess not. Sorry I didn't test this more thoroughly. 

I was about to suggest something fancy like adding an ``__init__`` override to our log handler and checking to see if it was a Python path, but if people want to configure it they can just add the actual string values from https://docs.python.org/2/library/logging.handlers.html#logging.handlers.SysLogHandler so that would be silly.

goosemania:
Our handler is kind to the strings but just to the certain ones ;)
And I agree it is easy to configure facility using its string value, so probably it is not worth doing anything more complicated

PR #2734, opened 2016-09-05
Creates basic pulp.plugin

bmbouter:
This PR adds a base Cataloger and Profiler objects to
pulp.plugin. It as imports the following objects from
pulp.platform.models for usage by plugins:

* Distributor
* Importer
* Repository
* RepositoryContent

It removes several modules from the old pulp.plugins
package.

This does not have an interface for upload. After thinking about it I believe platform can do everything for upload assuming that a plugin writer has defined the Content type correctly. If this turns out not to be the case we can easily add the upload interface.

closes #2181
https://pulp.plan.io/issues/2181


jeremycline:
I think it'd be super neat if we used Spinx-y docstrings where mentions of classes, for example, are written as ``:class:`class.path.Here```. That way we can generate API docs for our public Python interfaces that look like [this](http://urllib3.readthedocs.io/en/latest/reference/index.html#module-urllib3.poolmanager) ([source](https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L83))! It's really handy to have links to the classes and their documentation right there, with syntax-highlighted examples, etc.

bmbouter:
I'm going to put some work into making the docstrings great since this is the new Plugin API, but having some more comments on the code currently here would be great.

plugin/pulp/plugin/__init__.py
```diff
@@ -0,0 +1,6 @@
+from pulp.platform import ProgressBar, ProgressSpinner, Repository, RepositoryContent
```
jortel: This needs to be `from pulp.platform.models import ....`

plugin/pulp/plugin/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+
+from .cataloger import Cataloger
+from .distributor import RepositoryDistributor
+from .importer import RepositoryImporter
+from .profiler import Profiler
```
jortel: Do we really want to use a relative import?

plugin/pulp/plugin/cataloger.py
```diff
@@ -0,0 +1,33 @@
...
+        Initialize the Cataloger with a config and url.
+
+        :param config: The content source configuration.
+        :type config: dict
+
+        :param url: The URL for the content source.
+        :type url: str
+        """
+        self.config = config
+        self.url = url
```
jortel: :+1:  for constructing the object with the configuration.  However, I think the `url` should be passed into `get_downloader()` instead because the configuration is part of the object's *state* and the `url` is not.  For example, the cataloger could used as:
```
cataloger = Cataloger(configuration)
for url in url_list:
   downloader = cataloger.get_downloader(url)
   ...
```

jortel:
So far, I like this approach the best.  It provides clear separation of concerns between the platform (core) model and the plugin integration (layer) model.  Plugin (contributed) objects are *used* differently by the platform so having them be an extension of the platform model instead of part of the platform model seems appropriate.  Also, this achieves the goal of simplifying the plugin writer's universe.

It reinforces this:
```
--------------------------
|         REST           |
--------------------------
|    Platform (core)     |
--------------------------
|   Plugin (integration) |
--------------------------
| p1 | p2 | p3 | p4 | p5 |
|    |    |    |    |    |

```

I assume you'll be adding additional methods on these classes for feature parity with pulp2 model after we have consensus on the approach?

bmbouter:
@jortel Thanks a lot for the review. Everything suggested makes sense and I will revise it today along with improving the docstrings.

The one thing I had not planned on doing was adding more methods. The upload in 2.y was on the importer but I think that can become an all-platform activity and the plugin writer would not have to implement anything to support it. Similarly for Copy. It's likely other methods do need to be added though. What do you think those methods are?

jortel:
Agreed about the `copy()`.  

After review of the pulp2 models only a few things come to mind.

- Do we still need anything like `ensure_all_units_downloaded()`?  I'm hoping not but don't know why it's needed in pulp2.  Could be added later if we discover it's needed.

- Do we still need the distributor to participate in clean up after it (distributor) has been deleted (like in pulp2)?

- And, lastly, something new for pulp3.  I did not model anything in the platform for *share* content storage.  After much consideration, it seemed more appropriate for the plugin to solely manage that.  Given this, the plugin would need to participate in orphan purging.  Perhaps the importer?  Something like: `content_deleted(content)`.  This could be added later.  Thoughts?

plugin/pulp/plugin/distributor.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class RepositoryDistributor(RepositoryDistributor):
+    """
+    The base RepositoryDistributor object
+    """
+
+    def publish(self):
+        """
+        Perform a publish
```
mhrivnak: It would be great to emphasize here that the plugin writer should definitely not try to implement their publish logic on this class. They should implement it somewhere else, and just call their workflow from here.

plugin/pulp/plugin/importer.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class RepositoryImporter(RepositoryImporter):
+    """
+    The base RepositoryImporter object
+    """
+
+    def sync(self):
+        """
+        Perform a sync
```
mhrivnak: It would be great to emphasize here that the plugin writer should definitely not try to implement their sync logic on this class. They should implement it somewhere else, and just call their workflow from here.

bowlofeggs: Shouldn't that just be the job of the code reviewer to enforce? It doesn't make sense to document in each function that programmers should make sure to follow good coding practices.

plugin/pulp/plugin/profiler.py
```diff
@@ -0,0 +1,32 @@
+class Profiler(object):
```
mhrivnak: It's worth keeping an eye on this once we actually implement one, to see if it really deserves to be an entire class. As-is, it reminded me of this classic pycon talk. Although I recognize that this is really just an interface at the moment, so that logic may not apply.

https://youtu.be/o9pEzgHorH0?t=129

I think it's fine to re-visit this once a profiler actually gets implemented it, but I just wanted to share my thoughts. Another option would be to make it a static method on a unit model, which would also solve the discovery problem nicely.

plugin/pulp/plugin/importer.py
mhrivnak: That's a fair point. The broader context is that I was initially surprised to see that we'd have a method on a model that creates and/or changes state in lots of other models. It feels more like controller code, and inappropriate as a model method. But having talked through it with @bmbouter , the reasoning for doing it this way is to simplify the plugin writer experience, and simplify the platform's discovery process.

If we ask plugin writers to make a subclass of a separate ``Importer`` class of some kind, that would add complexity for them, and we would need a way to discover that class. (we have a nice simple way to discover the models that is outlined elsewhere) This design is a compromise to make the connection between the platform and the plugin's workflow simple, easy, and discoverable.

All of that said, my suggestion here is in large part to explain some of the above in the doc block, so that plugin writers and other readers understand why we have a method on a model class that is going to change state with lots of other models. I think a small hint to the plugin writer is appropriate, lest they think we are trying to lead them down the path of putting their sync code here.

mhrivnak:
Looking very good! I think this will be mergeable when your next changes are done.

mhrivnak:
@jortel some responses:

> Do we still need the distributor to participate in clean up after it (distributor) has been deleted (like in pulp2)?

I think removing published data when the distributor gets deleted will continue to be important. There may be a more elegant way to trigger that on the model than how we did it in pulp 2, such as using the post_delete signal.

> And, lastly, something new for pulp3. I did not model anything in the platform for share content storage. After much consideration, it seemed more appropriate for the plugin to solely manage that. Given this, the plugin would need to participate in orphan purging. Perhaps the importer? Something like: content_deleted(content). This could be added later. Thoughts?

That sounds like it might be a good fit for a class or static method on the unit model, but I'm light on details of what it would need to do.

jeremycline:
> Do we still need the distributor to participate in clean up after it (distributor) has been deleted (like in pulp2)?

My preference would be to have the platform provide a set of methods to interact with the filesystem which records the ownership of files and purges them (accounting for shared directories) automatically. It doesn't seem like something the plugin writer should ever need to worry about.

We already provide a utility method to attempt to handle race conditions during directory creation, anyway.

mhrivnak:
Recapping discussion with @bmbouter 

Upload could (as one option) be handled by putting a class method on each unit model called something like ``from_files(cls, path)``. The platform would enable the user to upload files and associate them with a particular upload operation, then submit/finalize the upload. In that last step, the platform would call the class method on the model.

Copy and remove are both great candidates for a default implementation in the platform, with the option for the plugin writer to override and provide custom logic. Examples for why that's useful include dependency resolution in pulp_rpm, copying docker manifests and blobs along with the tags that reference them, etc.

I agree that all of that can be considered and implemented in the future and not on this PR.

mhrivnak:
> My preference would be to have the platform provide a set of methods to interact with the filesystem which records the ownership of files and purges them (accounting for shared directories) automatically. It doesn't seem like something the plugin writer should ever need to worry about.
> 
> We already provide a utility method to attempt to handle race conditions during directory creation, anyway.

Majorly agree. This lines up well with the "managed publications" idea I've described in the past. We'll get there!

plugin/pulp/plugin/__init__.py
bmbouter: I think it' ok to use them. These modules should always be relative to this `__init__.py` file. Even a few weeks ago I was arguing against not using them. Recently I re-evaluated my thoughts on them and now I think they are OK. Django uses them in its examples so they can't be that bad. The issue we had in pulp 2.y was that you couldn't run only one test. There was something about the way the tests were setup, run, or the test runner which would cause a line like this to fail:

`from ....base import PulpServerTests`

I think that was due to some other issue and wasn't related to it being a relative import necessarily.

I plan to leave them here for now and also leave them in in [pulp.platform.models](https://github.com/pulp/pulp/blob/6bb793295b65d041a7e10820d67e7c9eaa560865/platform/pulp/platform/models/__init__.py). If you want to change them I'm OK with that too.

plugin/pulp/plugin/cataloger.py
bmbouter: Good point, done.

plugin/pulp/plugin/__init__.py
bmbouter: good catch, done.

plugin/pulp/plugin/profiler.py
bmbouter: Agreed this may go away before 3.0 is released. I think a lot of what is being built currently is still subject to change as we get into the plugin conversion process.

plugin/pulp/plugin/importer.py
bmbouter: @mhrivnak The new docstrings are pretty verbose, but they don't specifically mention this. Would you give them a read through and leave some comments?

plugin/pulp/plugin/distributor.py
```diff
@@ -0,0 +1,74 @@
...
+
+        It is expected that plugins wanting to support publish will provide an implementation on the
+        subclassed Distributor.
+
+        The publish to perform uses the model attributes to encapsulate all of the information
```
mhrivnak: I find the first few words a little awkward to read. I think you could just cut them out and reduce this to "The model attributes encapsulate all of the information required."

plugin/pulp/plugin/distributor.py
```diff
@@ -0,0 +1,74 @@
...
+        required. This includes the platform :class: `pulp.platform.models.Publish` base
+        attributes and any custom attributes defined by the subclass.
+
+        The model attributes were loaded from the database and then had the user specified override
+        config applied on top. Additionally the distributor is read-only and prevents the saving of
+        changes to the Distributor instance
```
mhrivnak: missing a period at the end

plugin/pulp/plugin/importer.py
```diff
@@ -0,0 +1,71 @@
...
+    * Plugin specific sync functionality
+    * Add persistent data attributes for a plugin importer subclass
+    * Additional validation of plugin importer subclass configuration
+
+    The sync implementation is provided by :meth: `Importer.sync` which provides more details.
+    Failing to implement this method will prevent sync functionality for this plugin type.
+
+    This object is a Django model that inherits from :class: `pulp.platform.models.Importer` which
+    provides the platform persistent attributes for an importer object. Plugin authors can add
+    additional persistent importer data by subclassesing this object and adding Django fields. We
```
mhrivnak: s/subclassesing/subclassing/

plugin/pulp/plugin/distributor.py
```diff
@@ -0,0 +1,74 @@
...
+    * Add persistent data attributes for a plugin distributor subclass
+    * Additional validation of plugin distributor subclass configuration
+
+    The publish implementation is provided by :meth: `Distributor.publish` which provides more
+    details. Failing to implement this method will prevent publish functionality for this plugin
+    type.
+
+    This object is a Django model that inherits from :class: `pulp.platform.models.Distributor`
+    which provides the platform persistent attributes for a distributor object. Plugin authors can
+    add additional persistent distributor data by subclassesing this object and adding Django
```
mhrivnak: s/subclassesing/subclassing/

plugin/pulp/plugin/importer.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def sync(self):
+        """
+        Perform a sync
+
+        It is expected that plugins wanting to support sync will provide an implementation on the
+        subclassed Importer.
+
+        The sync to perform uses the model attributes to encapsulate all of the information
```
mhrivnak: Same suggestion as above on ``Distributor``

plugin/pulp/plugin/importer.py
```diff
@@ -0,0 +1,71 @@
...
+        following:
+
+            1. The plugin provides an implementation called WidgetImporter which subclasses Importer
+
+            2. The user makes a call to sync widget_importer (say id=10) with some override config
+
+            3. The platform loads the saved
+                >>> wi = WidgetImporter.objects.get(id=10)
+
+            4. The platform puts the WidgetImporter into read-only mode, preventing a save to the
```
mhrivnak: The last three words of this sentence are a little ambiguous and I think could be removed.

plugin/pulp/plugin/distributor.py
bmbouter: agreed

mhrivnak:
I made some suggestions, but otherwise it looks fine to me.

bmbouter:
Thanks @mhrivnak I updated with all those revisions. They were good. I'll wait for questions/ideas or an LGTM from a second reviewer. @pcreech @jortel @seandst

plugin/pulp/plugin/cataloger.py
```diff
@@ -0,0 +1,48 @@
+class Cataloger(object):
+    """
+    The base cataloger to index content sources.
+
+    This is meant to be subclassed by plugin authors to given the plugin control where content is
+    fetched from. For a given url the plugin should return a
+    :class `~nectar.downloaders.base.Downloader` object. Using this interface, content that is
+    stored locally or in multiple locations can be fetched more efficiently.
+
+    The platform will use one :class `~pulp.plugin.Cataloger` instance per content source so this
```
jortel: should be `:class:`, right?

jortel:
I'm satisfied.  The suggestion made by @mhrivnak is good (though, I strongly oppose using signals) that the derived plugin `Content` class can override `delete()` and do what is necessary.  No plugin API call is needed.

pcreech:
Based on everything here, this looks good to me as well

PR #2736, opened 2016-09-06
Remove the playpen directory

jeremycline:
We have a good place for developer scripts and Python code, `pulp/devel`. Let's stop dumping things in here.


bowlofeggs:
LGTM

asmacdo:
There is a link to the Mongoengine README in our docs:
https://github.com/pulp/pulp/blob/master/docs/user-guide/release-notes/2.8.x.rst#before-upgrade

asmacdo:
This [issue](https://pulp.plan.io/issues/1238) and associated [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1168604) required a cleanup script, `playpen/cleanup/remove_orphaned_dirs.py` which is removed by this PR. I agree that the place for this should be pulp/devel. I would support moving it there and putting a new link to that script in the issue and BZ. However I *only* know that because I wrote that particular script. Fixing that one case would be fine, but I wonder if there are other cases as well, creating broken links.

For now, I am a -1 on deleting the whole directory.

asmacdo:
Ok, I am concvinced :)

PR #2739, opened 2016-09-07
Changed copyright year to current year

werwty:
closes #2097
https://pulp.plan.io/issues/2097

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@
 
 import sys
 import os
+from datetime import date
 
 try:
     import sphinx_rtd_theme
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -48,7 +49,9 @@
 
 # General information about the project.
 project = u'Pulp Project'
-copyright = u'2012-2014, Pulp Team'
+
+# Set copyright to current year
+copyright = u'2012-' + str(date.today().year).decode("utf-8") + u', Pulp Team'
```
jeremycline: Since this isn't Python 3 compatible anyway, it might be better to be explicit and cast ``date.today().year`` to unicode directly ``unicode(date.today().year)``

PR #2740, opened 2016-09-07
add filter docs to 3.0 dev section

asmacdo:
Demonstrates how filtering will work, covers proof of concept for replacing searches

https://pulp.plan.io/issues/1873


asmacdo:
@seandst, @bmbouter since these are new files, mentionbot didn't mention, but I think that you two would be good reviewers!

seandst:
This commit will be included in (and merged by) another PR that I'm putting together.

PR #2744, opened 2016-09-08
The Artifact `published_path` field is now `relative_path`

jeremycline:
This updates the Artifact model's `published_path` field to be
`relative_path` which I think is clearer. This also updates the
documenation for the field.

### DIFF
--- a/platform/pulp/platform/models/content.py
+++ b/platform/pulp/platform/models/content.py
@@ -76,8 +76,13 @@ class Artifact(Model):
     :cvar downloaded: The associated file has been successfully downloaded.
     :type downloaded: BooleanField
 
-    :cvar published_path: The relative published path.
-    :type published_path: models.FileField
+    :cvar relative_path: The artifact's path relative to the associated
+                         :class:`Content`. This path is incorporated in
+                         the absolute storage path of the file and its
+                         published path relative to the root publishing
+                         directory. At a minimum the path will contain the
+                         file name but may also include sub-directories.
+    :type relative_path: models.TextField
 
     :cvar size: The size of the file in bytes.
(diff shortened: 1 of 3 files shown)


platform/pulp/platform/models/content.py
```diff
@@ -76,8 +76,11 @@ class Artifact(Model):
     :cvar downloaded: The associated file has been successfully downloaded.
     :type downloaded: BooleanField
 
-    :cvar published_path: The relative published path.
-    :type published_path: models.FileField
+    :cvar relative_path: The path to publish this artifact to relative to
+                         the :class:`Content` it is associated with; if the
+                         :class:`Content` has a single :class:`Artifact`
+                         this should simply be the filename.
```
jortel: I think this would be clearer.
"""
The artifact's path relative to the associated :class:`Content`.
This path is incorporated in the absolute storage path of the `file` and its
published path relative to the root publishing directory.  At a minimum the path
will contain the file *name* but may also include sub-directories.
"""

jeremycline: That works for me

jortel:
Did you test this?  You need to change to `relative_path` here: https://github.com/pulp/pulp/blob/3.0-dev/platform/pulp/platform/models/storage.py#L50 as well.

Running this test would have caught it:  https://github.com/pulp/pulp/blob/3.0-dev/platform/pulp/platform/tests/models/test_content.py

jeremycline:
I neglected to run the tests because I'm so used to looking at the PR and not worrying if I see a green check mark. I set up Travis to run tests against 3.0: https://github.com/pulp/pulp/pull/2746

I also fixed the problem here.

PR #2746, opened 2016-09-09
Add travis configuration for Pulp 3

jeremycline:
This includes some re-arrangement of code as well as the necessary updates to the .travis.yml file. This is because I don't want install pulp.server and its dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,16 @@
 language: python
 python:
     - "2.7"
+    - "3.4"
+    - "3.5"
+addons:
+    postgresql: "9.3"
 install:
-    - "pip install flake8"
+    - "pip install -r test_requirements.txt"
+    - "pushd platform/ && python setup.py develop && popd"
+before_script:
+    - psql -U postgres -c "CREATE USER pulp WITH SUPERUSER LOGIN;"
+    - psql -U postgres -c "CREATE DATABASE pulp OWNER pulp;"
 script:
-    - "git diff HEAD^ | flake8 --diff --config flake8.cfg"
+    - "flake8 --config flake8.cfg platform/"
+    - "cd platform/pulp/platform && ./db-reset.sh && python manage.py test"
--- a/platform/pulp/platform/logs.py
+++ b/platform/pulp/platform/logs.py
@@ -1,90 +1,18 @@
 """
 This module defines and configures Pulp's logging system.
 """
-import ConfigParser
+from logging import handlers
 import logging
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,7 +1,16 @@
 language: python
 python:
     - "2.7"
+    - "3.4"
+    - "3.5"
+addons:
+    postgresql: "9.3"
```
jeremycline: I couldn't remember what version we were targeting, but I assume we should work with any recent version.

platform/pulp/platform/logs.py
```diff
@@ -0,0 +1,165 @@
+"""
```
jeremycline: I left the other version of this file in server/. I'm interested to see whether others think we should try to keep the Pulp 2 setup working with a Pulp 3 checkout or not. If not (which is what I'm in favor of), I'll just move the server/ version here.

platform/pulp/platform/models/progress.py
```diff
@@ -19,6 +19,19 @@
 _logger = logging.getLogger(__name__)
 
 
+def get_current_task_id():
```
jeremycline: I copied this this function out of pulp.server.async.tasks, but I don't know that it should live here. @bmbouter?

jeremycline:
A few things to note here: someone needs to flip the repository setting switches to start running this, and we are already incompatible with Python 3 in platform/ so tests will fail until we fix that.

platform/pulp/platform/models/progress.py
bmbouter: I'm actually moving this function as part of the conversion of the tasking code to use the new models. What if we let my PR clean up this import and you continue to use it from `pulp.server.async.tasks` ?

.travis.yml
```diff
@@ -1,7 +1,16 @@
 language: python
 python:
     - "2.7"
+    - "3.4"
+    - "3.5"
```
bmbouter: :+1:

bmbouter:
I posted just one piece of feedback, but I'm adding the LGTM now because I don't need to re-review.

platform/pulp/platform/models/progress.py
jeremycline: Cool. I'll change it back, but I won't install the ``pulp.server`` package. The tests will fail until your PR is merged, but the Python 3.y tests fail in any case (I've got a branch ready for that as well).

PR #2748, opened 2016-09-09
Convert Pulp 3 to be Python 3 compatible.

jeremycline:


bmbouter:
@jeremycline I read through this and it looks good to me

seandst:
```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals
```
I think it might be easier if we just go all-in on py3. We've got SCL support for python 3.4 in el6 and el7, so while I think this is a great thing to do, it would be easier if we just committed to py3 entirely.

mhrivnak:
I am looking forward to pulp working with python 3. Before merging, can we make a redmine task and document exactly why this is needed now? I'm not doubting that it is, but it would be very helpful to have it documented and go through something approximating our normal planning process.

That would also be the right place to discuss scope questions, such as whether we should commit to only python3 or try to support both 2 and 3.

That would also be a good place to confirm whether all of our dependencies also work in python 3, and if not, how to handle that.

bowlofeggs:
LGTJCvD

JCvD

jeremycline:
Given https://pulp.plan.io/issues/2246 I've backed out the Python 2 compatibility code.

jeremycline:
https://travis-ci.org/jeremycline/pulp/builds/159371584

PR #2750, opened 2016-09-12
Moves the get_current_task_id function

bmbouter:
This creates the pulp.platform.tasks.task_system package
and moves the get_current_task_id function into it.


jeremycline:
The only other thing is to change the import statement here: https://github.com/pulp/pulp/blob/3.0-dev/platform/pulp/platform/models/progress.py#L10. Other than that, it looks good to me.

bmbouter:
@jeremycline I adjusted the important. Thanks for pointing that out. I pushed a new version and now I will merge it.

PR #2751, opened 2016-09-12
Document Pulp 3 only supports Python 3

jeremycline:


docs/dev-guide/3.0-development/index.rst
```diff
@@ -1,6 +1,15 @@
 .. warning::
     All documents within the 3.0-Development section should be considered temporary, and will be removed or relocated prior to the release of Pulp 3.0.
 
+Pulp 3 and Python 3
+===================
+
+Pulp 3 will only support Python 3. When introducing new dependencies to Pulp,
```
asmacdo: Would it be helpful to be more specific? Do we have a specific 3.y?

PR #2752, opened 2016-09-12
Converts tasking system to use Django models

bmbouter:
Moves all tasking code to pulp.platform.tasks and updates
existing usage to use the new location. Also the code
itself is converted to use the new Django models.

https://pulp.plan.io/issues/2154
closes #2154


platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+    via "startswith()" since we do not know which host the worker is running on.
+    """
+
+    if worker_name.startswith(SCHEDULER_WORKER_NAME) or \
+       worker_name.startswith(RESOURCE_MANAGER_QUEUE):
+        return False
+    return True
+
+
+def _get_unreserved_worker():
```
ipanova: I am not sure if you forgot or did not mean to, on purpose, but i don't see `get_worker_for_reservation()`

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+
+    :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with them.
+
+    :returns:          The Worker instance that has no reserved_resource
+                       entries associated with it.
+    :rtype:            pulp.platform.model.Worker
+    """
+    free_workers = [Worker.objects.annotate(Count('reservations')).filter(reservations__count=0)]
+    if len(free_workers) == 0:
+        raise ReservedResource.DoesNotExist()
```
ipanova: should not it be Worker.DoesNotExist? at least on L86 you're trying to catch it.

ipanova: and in the dock block you state the same

platform/pulp/platform/tasks/task_system.py
ipanova: oh now i see that it is not necessary taking into account #L77

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+
+    Any tasks associated with this worker are explicitly canceled.
+
+    :param name:            The name of the worker you wish to delete.
+    :type  name:            basestring
+    :param normal_shutdown: True if the worker shutdown normally, False otherwise.  Defaults to
+                            False.
+    :type normal_shutdown:  bool
+    """
+    if normal_shutdown is False:
```
ipanova: another way of writing this could be - if not normal_shutdown:

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+        msg = _('The task status %(task_id)s exited immediately for some reason. Marking as '
+                'errored. Check the logs for more details')
+        runtime_exception = RuntimeError(msg % {'task_id': task_id})
+
+        class MyEinfo(object):
+            traceback = None
+
+        new_task.on_failure(runtime_exception, task_id, (), {}, MyEinfo)
+
+    ReservedResource.objects.filter(task__id=task_id).delete()
```
ipanova: s/task__id/task

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+        self.error = error
+        self.spawned_tasks = []
+        if spawned_tasks:
+            for spawned_task in spawned_tasks:
+                if isinstance(spawned_task, dict):
+                    self.spawned_tasks.append({'task_id': spawned_task.get('task_id')})
+                elif isinstance(spawned_task, AsyncResult):
+                    self.spawned_tasks.append({'task_id': spawned_task.id})
+                elif isinstance(spawned_task, TaskStatus):
+                    self.spawned_tasks.append({'task_id': spawned_task.task_id})
```
ipanova: s/spawned_task.task_id/spawned_task.id

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+                              Alternately an AsyncResult, or the task_id of the task created.
+        :type spawned_tasks: list of TaskStatus, AsyncResult, or str objects
+        """
+        self.return_value = result
+        self.error = error
+        self.spawned_tasks = []
+        if spawned_tasks:
+            for spawned_task in spawned_tasks:
+                if isinstance(spawned_task, dict):
+                    self.spawned_tasks.append({'task_id': spawned_task.get('task_id')})
```
ipanova: i do not really know where this dict comes from but make sure that it has 'task_id' key and not just 'id'

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+    def from_task_status_dict(cls, task_status):
+        """
+        Create a TaskResult object from a celery async_result type
+
+        :param task_status: The dictionary representation of a TaskStatus
+        :type task_status: dict
+        :returns: a TaskResult containing the task in it's spawned_tasks lsit
+        :rtype: TaskResult
+        """
+        return cls(spawned_tasks=[{'task_id': task_status.task_id}])
```
ipanova: s/task_status.task_id/task_status.id

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+        # Form a resource_id for reservation by combining given resource type and id. This way,
+        # two different resources having the same id will not block each other.
+        resource_id = ":".join((resource_type, resource_id))
+        inner_task_id = str(uuid.uuid4())
+        task_name = self.name
+        tag_list = kwargs.get('tags', [])
+        group_id = kwargs.get('group_id', None)
+
+        # Create a new task status with the task id and tags.
+        task_status = TaskStatus(id=inner_task_id, state=TaskStatus.ERRORED, tags=tag_list,
```
ipanova: Can you explain why the state of the task was changed from waiting to errored?

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+    if task_status.state in constants.CALL_COMPLETE_STATES:
+        # If the task is already done, just stop
+        msg = _('Task [%(task_id)s] already in a completed state: %(state)s')
+        _logger.info(msg % {'task_id': task_id, 'state': task_status.state})
+        return
+
+    if task_status.worker.name == 'agent':
+        tag_dict = dict(
+            [
+                tags.parse_resource_tag(t) for t in task_status['tags'] if tags.is_resource_tag(t)
```
ipanova: what about task_status.tags?

platform/pulp/platform/tasks/task_system.py
```diff
@@ -1,4 +1,548 @@
...
+            ])
+        agent_manager = managers.consumer_agent_manager()
+        consumer_id = tag_dict.get(tags.RESOURCE_CONSUMER_TYPE)
+        agent_manager.cancel_request(consumer_id, task_id)
+    else:
+        celery_controller.revoke(task_id, terminate=True)
+
+    qs = TaskStatus.objects.get(id=task_id, state__neq=constants.CALL_COMPLETE_STATES)
```
ipanova: i suppose you wanted to use the mentioned above `qs`? i.e to cancel task

ipanova: i am not sure, but i think state__neq will not work, it should be state__nin since `CALL_COMPLETE_STATES` is a list

platform/pulp/platform/tasks/task_system.py
bmbouter: Yeah the code got so simple I moved it directly to L#77

platform/pulp/platform/tasks/task_system.py
bmbouter: I did mean `Worker.DoesNotExist` thanks for point this out. It will be updated on my next push.

platform/pulp/platform/tasks/task_system.py
bmbouter: Good idea. Look for this on my next push.

platform/pulp/platform/tasks/task_system.py
bmbouter: I believe this is correct as is because task_id is a UUID. If task_id was an actual pulp.app.models.Task object then what you are suggesting would work.

platform/pulp/platform/tasks/task_system.py
bmbouter: The next revision will have a lot of this code deleted. The `TaskStatus` object is also going to be made available through the `pulp.plugin` API.

platform/pulp/platform/tasks/task_system.py
bmbouter: The next revision will have a lot of this code deleted. The `TaskStatus` object is also going to be made available through the `pulp.plugin` API.

platform/pulp/platform/tasks/task_system.py
bmbouter: The next revision will have a lot of this code deleted. The `TaskStatus` object is also going to be made available through the `pulp.plugin` API.

platform/pulp/platform/tasks/task_system.py
bmbouter: This was a mistake! I changed it back, thanks.

platform/pulp/platform/tasks/task_system.py
ipanova: so does it mean that if using double underscore id you are accessing id of the task object?( and you can do that because task field on ReservedResource model is one-to-one field?)

platform/pulp/platform/tasks/task_system.py
bmbouter: This if path has been deleted.

platform/pulp/platform/tasks/task_system.py
bmbouter: Good point. This code has been reworked so this won't be an issue on the next push.

bmbouter: Good catch, this has been fixed.

platform/pulp/platform/tasks/task_system.py
ipanova: makes sense, neat :metal:

server/pulp/server/async/scheduler.py
```diff
@@ -12,13 +12,13 @@
...
 from pulp.server.async import worker_watcher
-from pulp.server.async.celery_instance import celery as app
-from pulp.server.async.tasks import _delete_worker
 from pulp.server.db import connection as db_connection
 from pulp.server.db.connection import UnsafeRetry
 from pulp.server.db.model.dispatch import ScheduledCall, ScheduleEntry
 from pulp.server.db.model import Worker, CeleryBeatLock
 from pulp.server.managers.schedule import utils
+from pulp.tasking.celery_instance import celery as app
+from pulp.tasking.system import _delete_worker
```
jortel: Since this is used outside the module it seems more appropriate to remove the `_` prefix.

plugin/pulp/plugin/task_system.py
```diff
@@ -0,0 +1,34 @@
+from pulp.platform.models import Task
+from pulp.tasking.system import get_current_task_id
+
+
+class TaskController(object):
```
jortel: Few things:
- Naming the module `plugin/pulp/plugin/tasking.py` is more consistent with `pulp.tasking`.  Also, the `_system` seems redundant.
- Having a controller outside of the django `app` seems odd. 
- Having a *controller* that manipulates a single model object seems inappropriate.
- The `_and_save_` part of the method name seems odd.  Why update the task and not save it?
- Why not add `Task.append_non_fatal_error()`? and include the `Task` model in the *plugin* package?

common/pulp/common/constants.py
```diff
@@ -31,15 +31,15 @@
 CALL_RUNNING_STATE = 'running'
 CALL_SUSPENDED_STATE = 'suspended'
 CALL_COMPLETED_STATE = 'completed'
-CALL_ERRORED_STATE = 'errored'
+CALL_FAILED_STATE = 'failed'
```
jortel: :+1:  `failed`

nodes/test/nodes_tests/base.py
```diff
@@ -13,12 +13,12 @@
 from pulp.client.extensions.core import PulpCli, ClientContext, PulpPrompt
 from pulp.client.extensions.exceptions import ExceptionHandler
 from pulp.common.config import Config
-from pulp.server.async import celery_instance
 from pulp.server.config import config as pulp_conf
 from pulp.server.db import connection
 from pulp.server.logs import start_logging, stop_logging
 from pulp.server.managers import factory as managers
 from pulp.server.managers.auth.cert.cert_generator import SerialNumber
+from pulp.tasking import celery_instance
```
jortel: :+1: `pulp.tasking`

platform/pulp/tasking/system.py
```diff
@@ -0,0 +1,474 @@
...
+
+        class MyEinfo(object):
+            traceback = None
+
+        new_task.on_failure(runtime_exception, task_id, (), {}, MyEinfo)
+
+    ReservedResource.objects.get(task__id=task_id).delete()
+
+
+class UserFacingPulpTask(PulpTask):
```
jortel: I know, I'm being pedantic :/
Since this `isA` `PulpTask` perhaps `UserFacingTask` would be sufficient.

platform/pulp/tasking/system.py
```diff
@@ -0,0 +1,474 @@
...
+    The :meth: `apply_async_with_reservation` dispatch interface will send the task through the
+    resource_manager queue and process and will be assigned a specific worker. A series of
+    reservations cause specific types of tasks to not be run concurrently. See the
+    :meth: `apply_async_with_reservation` documentation for more details.
+    """
+
+    # this tells celery to not automatically log tracebacks for these exceptions
+    throws = (PulpCodedException,)
+
+    def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwargs):
```
jortel: @seandst and I discussed deriving the *resource_type* and *resource_id* using the model object's class name and natural_key().  Basically getting rid of all the resource tag code splattered everywhere as well.

May not be appropriate here but maybe passing the model (the actual resource) would be a good thing here?

platform/pulp/tasking/system.py
```diff
@@ -0,0 +1,474 @@
...
+        :param resource_type: A string that identifies type of a resource
+        :type resource_type:  basestring
+
+        :param resource_id:   A string that identifies some named resource, guaranteeing that only
+                              one task reserving this same string can happen at a time.
+        :type  resource_id:   basestring
+
+        :param tags:          A list of tags (strings) to place onto the task, used for searching
+                              for tasks by tag
+        :type  tags:          list
```
jortel: I don't see `tags` in the function signature.

server/pulp/server/agent/direct/services.py
```diff
@@ -229,7 +229,7 @@ def failed(self, reply):
             finished = dateutils.format_iso8601_datetime(now)
 
         TaskStatus.objects(task_id=task_id).update_one(set__finish_time=finished,
-                                                       set__state=constants.CALL_ERRORED_STATE,
+                                                       set__state=constants.CALL_FAILED_STATE,
                                                        set__traceback=traceback)
 
         if action == 'bind':
```
jortel: I really need to strip out the agent stuff soon so we don't feel the need to make these changes.

server/pulp/server/async/scheduler.py
bmbouter: I agree. This is fixed in my next push.

plugin/pulp/plugin/task_system.py
bmbouter: I renamed it to `plugin/pulp/plugin/tasking.py`. Good idea. Fixed in my next push.

This object helps me avoid giving access the `Task` model to plugin writers. I am hoping we don't have to include `Task` in the plugin API. The tasking system relies on that data for correctness so I'm also hoping to not give that access to that model for this reason too.

This object can collect any plugin-appropriate Task related calls can live. For now the only one I think we have is to record non-fatal exceptions as a task runs.

I'm removing the `_and_save_` part and I updated the docstring too. The other name was too verbose.

platform/pulp/tasking/system.py
bmbouter: I agree, my next version will use `UserFacingTask`.

platform/pulp/tasking/system.py
```diff
@@ -0,0 +1,474 @@
...
+        :type group_id:     uuid.UUID
+
+        :return:            An AsyncResult instance as returned by Celery's apply_async
+        :rtype:             celery.result.AsyncResult
+        """
+        routing_key = kwargs.get('routing_key',
+                                 defaults.NAMESPACES['CELERY']['DEFAULT_ROUTING_KEY'].default)
+        tag_list = kwargs.pop('tags', [])
+        group_id = kwargs.pop('group_id', None)
+        async_result = super(UserFacingPulpTask, self).apply_async(*args, **kwargs)
```
jeremycline: It'd be super cool if there was a check here for the Django setting `profile = True` or something and then start a cProfile that is stored in MEDIA_ROOT/profiles/<uuid> or something. That way users can easily generate profile information.

I know that's not really the point of this PR, but it'd be cool.

platform/pulp/tasking/system.py
bmbouter: I added some docstrings indicating that `tags` and `group_id` are both pulled out of `kwargs`. You'll see the changes on my next push. At some point in the future I would like to improve the argument interface on this function but it is more than I want to do now.

server/pulp/server/agent/direct/services.py
bmbouter: +1 to stripping out the agent stuff.

platform/pulp/tasking/system.py
bmbouter: We should consider doing something like this but I don't plan to do it with this PR. Is there some kind of issue tracking that idea?

BrnoPCmaniak:
Maybe just one small suggestion in Django it is more common and I would say even convention to use the ``pk`` instead of ``id`` when accessing the model. ``pk`` points to whatever the primary key on the model is (By default points to ``id``). It also makes the code more consistent for cases when the primary key is not ``id``. And moreover ``id`` is python keyword. 
[Djnago docs](https://docs.djangoproject.com/en/1.9/topics/db/queries/#the-pk-lookup-shortcut)

I would recommend to use it, but it's just my opinion.

bmbouter:
Closing to open a new PR so that a full review can be made without seeing all of the comments.

bmbouter:
New PR is here  https://github.com/pulp/pulp/pull/2771

PR #2753, opened 2016-09-12
Consolidation of repository models.

jortel:
Flattened importer and distributor models.
Renamed `Distributor` to `Publisher`.
Renamed `Plugin` to`ContentAdaptor`.  Never seemed quite right and took a stab at something better.

The natural key for an importer and publisher is `name` which requires it to be unique.  This puts *some* extra burden on the API user but I don't think it's unreasonable.

Updated the tests until we decide what direction we're going with on testing.

### DIFF
--- a/docs/dev-guide/3.0-development/data-modeling.rst
+++ b/docs/dev-guide/3.0-development/data-modeling.rst
@@ -4,13 +4,13 @@ Data Modeling
 Introduction
 ^^^^^^^^^^^^
 
-The pulp 3 data modeling effort is not just a translation or porting effort but instead
-an effort to build pulp 3 using pulp 2.  Remodeling will include changes to leverage the
-relational capabilities of postgres but also to improve upon the pulp 2 model.
+The Pulp 3 data modeling effort is not just a translation or porting effort but instead
+an effort to build Pulp 3 using Pulp 2.  Remodeling will include changes to leverage the
+relational capabilities of postgres but also to improve upon the Pulp 2 model.
 
-When creating each pulp 3 model object, consider the following:
+When creating each Pulp 3 model object, consider the following:
 
-- Understand what the pulp 2 model (collection) stores and how the information
+- Understand what the Pulp 2 model (collection) stores and how the information
   is used by pulp.
 
 - Name the model/table appropriately.  A table is implicitly plural.  Naming a
@@ -26,14 +26,14 @@ When creating each pulp 3 model object, consider the following:
   effort, a proper class hierarchy is almost always possible and will be better in the end.
 
 - Question the existence, type, and naming of all fields.  Field names beginning with `_` 
-  should not exist in pulp 3.  The primary key field name is `id` and already defined in the
+  should not exist in Pulp 3.  The primary key field name is `id` and already defined in the
   `Model` base class.  All `display_name` fields need to be renamed to `name`.  Also, avoid
   redundant scope by prefixing field names (or any attribute) with the name of the class.
   For example: `Task.task_type`.
 
 - Question all methods (including @property). We **only** want those methods that encapsulate
-  significant complexity and are widely used.  Most of the methods on pulp 2 models will likely
-  not be necessary or wanted in the pulp 3 models.  Let's keep the model interface as clean
+  significant complexity and are widely used.  Most of the methods on Pulp 2 models will likely
+  not be necessary or wanted in the Pulp 3 models.  Let's keep the model interface as clean
   as possible.
 
 - Most string fields will be `TextField`.  When the field is required and indexed,
@@ -44,7 +44,7 @@ When creating each pulp 3 model object, consider the following:
 - All fields containing *ISO-8601* strings must be converted to `DateTimeField`.
 
 - Think about *natural* keys.  For example, each repository has a unique name (instead of repo_id
-  like in pulp 2) that is known to users and is not the primary key (`id` is).  The `name` is
+  like in Pulp 2) that is known to users and is not the primary key (`id` is).  The `name` is
   the *natural* key.  Don't forget the index. See: https://en.wikipedia.org/wiki/Natural_key
 
 - Define a `natural_key()` method in each model.  This both documents the *natural* key and
@@ -57,7 +57,7 @@ Development Process
 
 - The *refactor* tasks (in Redmine) group related collections.
 
-- The pulp 3 models are grouped into modules within the `models` package.
+- The Pulp 3 models are grouped into modules within the `models` package.
 
 - Foreach model class defined, A *refactor* task needs to be created in Redmine for
   migrating the data.  See existing examples.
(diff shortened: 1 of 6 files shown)


docs/dev-guide/3.0-development/data-modeling.rst
```diff
@@ -4,13 +4,13 @@ Data Modeling
...
+The Pulp 3 data modeling effort is not just a translation or porting effort but instead
+an effort to build Pulp 3 using Pulp 2.  Remodeling will include changes to leverage the
+relational capabilities of postgres but also to improve upon the Pulp 2 model.
 
-When creating each pulp 3 model object, consider the following:
+When creating each Pulp 3 model object, consider the following:
 
-- Understand what the pulp 2 model (collection) stores and how the information
+- Understand what the Pulp 2 model (collection) stores and how the information
   is used by pulp.
```
pcreech: s/used by pulp/used by Pulp/

???

pcreech:
This looks good to me, with the minor consistency comment on pulp casing.  
(Since you changed all the others, I just didn't want that one missed)

docs/dev-guide/3.0-development/data-modeling.rst
jortel: Not sure it matters.  I actually made more changes here WRT case than I intended doing global replaces.  If folks feel strongly I'll do what's needed here.

platform/pulp/platform/models/repository.py
```diff
@@ -105,35 +105,44 @@ def natural_key(self):
...
     """
-    name = models.TextField(db_index=True)
+    name = models.TextField(db_index=True, unique=True)
     type = models.TextField(blank=False, default=None)
     last_updated = models.DateTimeField(auto_now=True, blank=True, null=True)
 
     class Meta:
         abstract = True
 
+    def natural_key(self):
```
jeremycline: Does this apply to children that inherit? If so, won't a unique `name` be required for each instance of importer? If not, should it be on this abstract class?

platform/pulp/platform/models/repository.py
```diff
@@ -213,44 +225,20 @@ class Importer(Plugin):
...
 
-class Distributor(Plugin):
+class Publisher(ContentAdaptor):
     """
-    An abstract content distributor.
+    A content publisher.
 
     Fields:
 
+    :cvar auto_publish: Indicates that the adaptor may publish automatically
```
jeremycline: When I read "may" I feel uncertain about what this field really means, since it's a Boolean field. I know this is very nitpicky, but maybe "will" is better here?

Do or do not. There is no "may"

platform/pulp/platform/models/repository.py
jortel: Yes.  Every importer and distributor must have a unique name.

platform/pulp/platform/models/repository.py
jortel: IIRC, I chose `may` because I wanted to leave the door open for a publisher to make the final decision to actually publish.  However, each publisher with auto-publish=True `will` be requested to publish.  I can make this clearer.  Or, I also fine with changing this to `will` here.

platform/pulp/platform/models/repository.py
jeremycline: Will that be somewhat awkward? I'm failing to imagine example values for this field that will be useful _and_ unique, but I'm also not really sure what people have in mind for this.

Even if this is left as-is, I think the natural key on the Publisher should include the repository somehow, since there's no longer an intermediate table and the Publisher requires a repository relationship.

docs/dev-guide/3.0-development/data-modeling.rst
```diff
@@ -95,22 +95,24 @@ on a consumer.
...
-The repository models include importers and distributors.  The main difference being the
-consolidation of the importer and distributor and it's configuration.  In the model, a
-`Plugin` is the base for plugin contributed models that can be associated to a repository.
-On importer and distributor base models, the *standard* configuration settings that were
-separate documents in pulp 2 are attributes of the importer and distributor itself.  These models
-follow the *master-detail* pattern.  Plugins needing additional configuration, need to extend the
+First, `Distributor` has been renamed to `Publisher` because it seems more appropriate.
+
+The repository models include importers and publishers.  The main difference being the
+consolidation of the importer and publisher and it's configuration.  In the model, a
```
mhrivnak: s/it's/its/

platform/pulp/platform/models/consumer.py
```diff
@@ -25,14 +25,14 @@ class Consumer(Model):
...
-    :type distributors: models.ManyToManyField
+    :cvar publishers: Associated publishers.
+    :type publishers: models.ManyToManyField
     """
     name = models.TextField(db_index=True, unique=True)
     description = models.TextField(blank=True)
 
     notes = fields.GenericRelation(Notes)
-    distributors = models.ManyToManyField('RepositoryDistributor')
+    publishers = models.ManyToManyField('Publisher')
```
mhrivnak: Not sure if this has been discussed elsewhere, but django's auto-naming of the relationship on the other model is awkward. I think it would appear in this case as ``consumer_set``. It's common to use the ``related_name`` parameter for the ``ManyToManyField`` and ``ForeignKey`` to provide a more natural name, such as ``consumers``. I suggest we do that here and anywhere else we have similar relationships.

asmacdo:
I love the change of distributors to publishers! And thanks for fixing the 3.0 docs while you are at it. Maybe we should make a redmine task to `grep -rn distributor` since it would be unreasonable to change it everywhere now.

platform/pulp/platform/tests/models/test_consumer.py
```diff
@@ -1,10 +1,25 @@
...
 class TestConsumer(TestCase):
 
     def test_natural_key(self):
         consumer = Consumer(name='test')
         self.assertEqual(consumer.natural_key(), (consumer.name,))
+
+    def test_bind(self):
+        consumer = Consumer(name='test')
+        consumer.save()
+        repository = Repository(name='test')
```
mhrivnak: Shouldn't this save the repository next before associating a publisher?

platform/pulp/platform/models/consumer.py
jortel: Adding it.

platform/pulp/platform/models/consumer.py
```diff
@@ -25,14 +25,14 @@ class Consumer(Model):
...
-    :type distributors: models.ManyToManyField
+    :cvar publishers: Associated publishers.
+    :type publishers: models.ManyToManyField
     """
     name = models.TextField(db_index=True, unique=True)
     description = models.TextField(blank=True)
 
     notes = fields.GenericRelation(Notes)
-    distributors = models.ManyToManyField('RepositoryDistributor')
+    publishers = models.ManyToManyField('Publisher', related_name='publishers')
```
mhrivnak: The ``related_name`` should be ``consumers``

jortel:
After further consideration, it seemed that making the repository FK optional on `Importer` to support un-associated importers (like alt-content-sources) was premature.  I'm not convinced that an alt-content-source `isA` importer because an importer provides operations like `sync()`.  We can revisit these models as part of the alt-content-source redesign.

platform/pulp/platform/tests/models/test_consumer.py
jortel: You'd think but it didn't raise an exception.

platform/pulp/platform/models/consumer.py
jortel: Fixed.  Github just doesn't know it for some reason.

PR #2754, opened 2016-09-13
Update the developer documentation to use SSHFS

jeremycline:
SSHFS is far easier to set up and simplier to use compared to NFS.

Note: I don't know if we want these in master or 3.0-dev. I don't mind either way.


asmacdo:
Awesome. It might be worth an email to pulp-dev when you merge this since the corresponding [vagrant change](https://github.com/pulp/devel/pull/10) occurs only in the vagrantfile example file and will require manual intervention from each dev.

PR #2757, opened 2016-09-15
Implement a useful download API for Pulp

jeremycline:
This API is intended to replace the Nectar library. It uses
requests-futures, a library that wraps requests with Python 3.2's
`concurrent.futures`. Rather than returning responses, this library
returns futures. It also provides a simply callback to allow users
to provide a function that works with the response in a background
thread.

By default, it uses a ThreadPoolExecutor, but in Python 3.4+ it is
possible to use a ProcessPoolExecutor (with a few caveats).

closes #2168

### DIFF
--- a/plugin/pulp/plugin/download.py
+++ b/plugin/pulp/plugin/download.py
@@ -1,0 +1,166 @@
+"""
+This module contains set of download APIs.
+"""
+import os
+import hashlib
+
+import requests
+from requests_futures.sessions import FuturesSession
+from requests.packages.urllib3.util import retry
+
+# TODO make config settings
+DEFAULT_CONNECT_TIMEOUT = 6.1
+DEFAULT_READ_TIMEOUT = 30
+DEFAULT_TRIES = 5
+DEFAULT_CONCURRENCY = 5
+# TODO Picked a bit out of a hat - I expect it should try to line up with the FS.
+DEFAULT_CHUNKSIZE = 1024 * 1024
+
+
+def default_callback_generator(filename=None, size=None, checksums=None):
+    """
+    Creates a closure to write a response to disk, checksum it, and verify
+    its size.
+
+    :param filename: The absolute path the response should be saved to.
+    :param size: The size, in bytes, that the file is expected to be; check is
+                 skipped if this value is not provided.
+    :param checksums: A dictionary of checksums to validate the file with; the
+                      key should be the hash algorithm and the value the
+                      expected hex digest.
+    """
+    # TODO potentially add a way to have this generate the entire checksum set.
+    # Maybe if there's {'algorithm': None}?
+    checksums = checksums or {}
+
+    def callback(self, response):
+        """
+        A callback for the requests_futures library.
+
+        :param resp: The requests response object
+        """
+        hashes = {algorithm: hashlib.new(algorithm) for algorithm in checksums}
+        with open(filename, mode='wb') as fp:
+            chunks = response.iter_content(chunk_size=DEFAULT_CHUNKSIZE)
+            for chunk in chunks:
+                for hash in hashes.values():
+                    hash.update(chunk)
+                fp.write(chunk)
+
+        # Validate that what we downloaded is what the user expects
+        # TODO add more reasonable exceptions
+        if size and size != os.stat(filename).st_size:
+            raise ValueError()
+
+        if checksums != {algorithm: hash.hexdigest() for algorithm, hash in hashes.items()}:
+            raise ValueError()
+
+    return callback
+
+
+class FileFuturesSession(FuturesSession):
+    """
+    This class implements the requests-futures API, with one  small adjustment. It provides
+    a default callback that writes a file to disk and validates its content. Once it
+    successfully returns, the file is ready to be imported into Pulp.
+
+    requests-futures allows you to queue up a large number of requests which are run
+    asynchonously on a pool of worker threads or processes. For examle:
+
+        >>> session = FileFuturesSession(max_workers=10)  # Create a pool of 10 threads
+        >>> requests = [
+        ...     {url='http://example.com/file1'},
+        ...     {url='http://example.com/file2'},
+        ... ]
+        >>> responses = []
+        >>> for request in requests:
+        ...     responses.append(session.get(**request))  # session.get does not block
+        >>> for reponse in responses:
+        ...     request_response, filename = response.result()
+        ...     with open(filename) as fd:
+        ...         django_file = django.core.files.File(fd)
+        ...     artifact = pulp.apps.models.Artifact(file=django_file)
+        ...     artifact.save()
+        >>>
+
+    Here are a list of things I think would be good to consider including in
+    this module and possibly this class:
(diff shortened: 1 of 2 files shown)


jeremycline:
This is very much a work in progress, but I think early feedback on this approach would be very helpful.

plugin/pulp/plugin/download.py
```diff
@@ -0,0 +1,122 @@
...
+        super(FileFuturesSession, self).__init__(*args, **kwargs)
+
+    def request(self, *args, **kwargs):
+        """
+        This provides the requests API as well as the requests-futures API.
+
+        The only difference is a few additional keyword arguments are accepted.
+        """
+        callback_generator = kwargs.pop('callback_generator', default_callback_generator)
+        callback_generator_kwargs = kwargs.pop(
```
jeremycline: I'm not crazy about this API, but it was the first thing I came up with. Alternate suggestions would be very much appreciated.

plugin/pulp/plugin/download.py
```diff
@@ -0,0 +1,122 @@
...
+# TODO Picked a bit out of a hat - I expect it should try to line up with the FS.
+DEFAULT_CHUNKSIZE = 1024 * 1024
+
+
+def default_callback_generator(filename=None, size=None, checksums=None):
+    """
+    Creates a closure to write a response to disk, checksum it, and verify
+    its size.
+
+    :param filename: The absolute path the response should be saved to.
```
jortel: If it's a path, let's call it a `path`.

plugin/pulp/plugin/download.py
jortel: I looked at the `requests_futures` and found it kind of clunky does not seem to add much value, IMHO.
I liked that it (and you) used a combination of  `concurrent.futures` and `requests`. here is a stab at something *light-weight* on top of them.  I wanted to be sure to support a *generator* of things to download and a streams processing approach.  It also seemed like supporting a plugin work flow of concurrently downloading files and iterating the results as they complete (in whatever order) to add the artifact etc interesting.  Not sure if anyone else would agree.

[1] https://gist.github.com/jortel/c6bf56ebe86035252cd899e412e91eb2

plugin/pulp/plugin/download.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        A callback for the requests_futures library.
+
+        :param resp: The requests response object
+        """
+        hashes = {algorithm: hashlib.new(algorithm) for algorithm in checksums}
+        with open(filename, mode='wb') as fp:
+            chunks = response.iter_content(chunk_size=DEFAULT_CHUNKSIZE)
+            for chunk in chunks:
+                for hash in hashes.values():
```
seandst: If we're worried about stomping builtins, `import builtins; builtins.hash` so it's explicit where they're coming from. The python `hash` builtin has no use in this context, and the `hash` identifier here is explicit and clear.

plugin/pulp/plugin/download.py
jeremycline: Although I'm not entirely satisfied with how the callback functions are provided or built, I _do_ think it's valuable to implement the `requests` API. It's clean, powerful, and is hugely popular.

I'm not sure I agree that `requests_futures` is clunky or that it doesn't add value. Can you expand on those comments a bit?

jeremycline:
I've pushed an updated commit with some expanded comment which contains a few examples

plugin/pulp/plugin/download.py
```diff
@@ -0,0 +1,166 @@
...
+    This class implements the requests-futures API, with one  small adjustment. It provides
+    a default callback that writes a file to disk and validates its content. Once it
+    successfully returns, the file is ready to be imported into Pulp.
+
+    requests-futures allows you to queue up a large number of requests which are run
+    asynchonously on a pool of worker threads or processes. For examle:
+
+        >>> session = FileFuturesSession(max_workers=10)  # Create a pool of 10 threads
+        >>> requests = [
+        ...     {url='http://example.com/file1'},
```
jeremycline: Note that as the implementation currently stands, this won't work. You'd need to do something like `((url,), {})` and then 

```
for args, kwargs in requests:
    session.get(*args, **kwargs)
```

plugin/pulp/plugin/download.py
jortel: I agree we should leverage the `requests` lib as much as possible and perhaps to the extent of plugin writers using its API directly.  I dissagree that the requests API is clean but it does seem to be popular.  And, there is value in that.  But, like many python libs, is popular mostly because its the best thing available but not because it's necessarily well done.  For example, the `requests.request(method, url, **kwargs)` taking `**kwargs` and documenting what the arguments are is pretty hacky and requires users to look them up in the docs.  I have other criticisms of the requests API but I don't think it useful to enumerate them here.

The `requests_futures` is a thin integration of `requests` and `futures` which sounds good but it's done in an odd way.  **First**, the `FuturesSession`  **IsA** `requests.Session` but it takes a `Session` constructor argument (I have to assume because it's poorly documented) and contains a `session` attribute.  That is the *wonky* part.  **Second**, the `FuturesSession.request()` sanitizes an undocumented keyword parameter named *background_callback* and simply calls `Executor.submit()`.  Then, returns a `Future` which is a deviation from the *requests* API.  Also, overriding a method and returning a different type is bad.  The `Executor.submit()` queues the request in a `Queue` that is configured without a *maxsize*.  So, each call to `request()` results in something stored in memory.  Not good for scalability.  The caller then has to store the returned future(s) in a list (more memory).  Then,  iterate them calling `Future.done()`.  Or, pass in a callback function.

jortel:
I doubt we can use the `ProcessPoolExecutor` because it will likely use `multiprocessing` which cannot be used with *daemon=True` within celery workers.

plugin/pulp/plugin/download.py
jeremycline: I have a suspicion there is a good reason for the oddity of having a `session` object and also being a `session` which has to do with the "Important" section at the bottom of the README that mentions you must provide a session for Python 3.4 and below, but I didn't investigate further.

>  Then, returns a Future which is a deviation from the requests API. Also, overriding a method and returning a different type is bad.

It is a documented deviation and one that is required to provide a requests-like API in an asynchronous way. Personally, I think this is fine.

As for the problem with the `Queue`, I have no doubt that upstream would be thrilled to receive a contribution that solves the problem. It's general and could be of great use to everyone, so I don't see a good reason to implement it in Pulp-specific code.

jortel: It's not worth the trouble.  As a plugin writer, I'd rather just use `requests` and `futures` directly.

jeremycline:
Based on my interpretation of https://github.com/pulp/pulp/pull/2757#discussion_r79454260 it sounds like the current proposal is to not provide an API at all and let the plugin authors each roll their own.

I'm not the best person to argue for or against that since I won't be using the API so I will leave it up to the current stakeholders to decide that and either continue this PR or close it.

PR #2759, opened 2016-09-16
Remove pulp_node

BrnoPCmaniak:
* Remove all code
* Remove from docs

closes #2193
https://pulp.plan.io/issues/2193

I removed everything in folder nodes and anything what had pointed to it. But there might be somewhere some dead code which pulp_node used.


bmbouter:
I looked through it and it looks right. @BrnoPCmaniak This PR makes me so happy. :+1: :fireworks: :earth_africa: 

It would be good to start some release notes for 3.0 and put in there that nodes as a feature was removed. Is that possible? I'll put the LGTM flag so that you can merge it when you decide how to handle that.

docs/user-guide/release-notes/3.x.x.rst
```diff
@@ -0,0 +1,12 @@
...
+=======================
+
+Pulp 3
+======
+
+Changelog
+---------
+
+* Nodes are no longer supported. And were completly removed please use just sync from one pulp
+  server to another instead of this.
```
BrnoPCmaniak: @bmbouter Something like this?

docs/user-guide/release-notes/index.rst
```diff
@@ -6,6 +6,7 @@ Contents:
 .. toctree::
    :maxdepth: 2
 
+   3.x.x
```
seandst: Should be 3.0.x, not 3.x.x.

docs/user-guide/release-notes/3.x.x.rst
bmbouter: Thanks for adding these. I couldn't help but elaborate on the language here. You can include it if you want.

```
 Nodes are no longer supported and were completely removed with Pulp 3.0. Instead of using nodes, publish content from one Pulp server and use that feed url to sync the content into another Pulp server. This is the recommended approach for a nodes use case.
```

docs/user-guide/release-notes/index.rst
BrnoPCmaniak: Fixed but not recognised by GitHub

jortel:
LGTM

PR #2768, opened 2016-09-23
Converts worker watcher to use django models and removes ScheduledCalls

dkliban:
This also renames the scheduler to pulp_celerybeat.


BrnoPCmaniak:
Just an suggestion on scheduler https://github.com/BrnoPCmaniak/pulp/commit/f8bd4cb12020c8adf0f616f77dcdc756c097a7b7?diff=split

dkliban:
@BrnoPCmaniak Thanks for the suggestion. I'll update my PR with your changes. 

I am also going to remove the ScheduledCall references because we will no longer have user defined scheduled tasks stored in Pulp.

tasking/pulp/tasking/worker_watcher.py
```diff
@@ -72,14 +72,16 @@ def handle_worker_heartbeat(event):
     :type event: dict
     """
     event_info = _parse_and_log_event(event)
-    worker = Worker.objects(name=event_info['worker_name']).first()
 
-    if not worker:
+    try:
+        worker = Worker.objects.get(name=event_info['worker_name'])
+    except worker.DoesNotExist:
+        worker = Worker(name=event_info['worker_name'])
```
BrnoPCmaniak: I think this would be better using [`worker, created = Worker.objects.get_or_create(name=event_info['worker_name'])`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create)

tasking/pulp/tasking/scheduler.py
```diff
@@ -117,38 +114,39 @@ def check_celery_processes(self):
...
+            _logger.error(msg)
+
+            if worker.name.startswith(constants.SCHEDULER_WORKER_NAME):
+                worker.delete()
             else:
-                worker_count = worker_count + 1
+                _delete_worker(worker.name)
 
-        if resource_manager_count == 0:
+        if not (Worker.objects.filter(name__startswith=constants.RESOURCE_MANAGER_WORKER_NAME)
```
dkliban: I rearranged things and now just use the count variable that used to be declared just after this code.

tasking/pulp/tasking/scheduler.py
```diff
@@ -117,38 +114,39 @@ def check_celery_processes(self):
...
+                      .exists()):
             msg = _("There are 0 pulp_resource_manager processes running. Pulp will not operate "
                     "correctly without at least one pulp_resource_mananger process running.")
             _logger.error(msg)
 
-        if scheduler_count == 0:
+        if not Worker.objects.filter(name__startswith=constants.SCHEDULER_WORKER_NAME).exists():
```
dkliban: I did the same thing as above. Using the scheduler_count to do the check.

tasking/pulp/tasking/scheduler.py
```diff
@@ -173,15 +171,12 @@ class Scheduler(beat.Scheduler):
     second, is a WorkerTimeoutMonitor thread that watches for cases where all workers disappear
     at once.
     """
-    Entry = ScheduleEntry
+    Entry = beat.ScheduleEntry
```
dkliban: I had the same suspicion also. I am removing this line.

tasking/pulp/tasking/scheduler.py
```diff
@@ -260,110 +255,34 @@ def tick(self):
         old_timestamp = datetime.utcnow() - timedelta(seconds=constants.CELERYBEAT_LOCK_MAX_AGE)
 
         # Updating the current lock if lock is on this instance of celerybeat
-        result = CeleryBeatLock.objects(celerybeat_name=self.celerybeat_name).\
-            update(set__timestamp=datetime.utcnow())
-
-        # If current instance has lock and updated lock_timestamp, call super
-        if result == 1:
+        try:
+            celerybeat_lock = TaskLock.objects.get(name=celerybeat_name)
```
dkliban: this should be self.celerybeat_name which is defined in the __init__(). Changing it to that.

tasking/pulp/tasking/scheduler.py
```diff
@@ -260,110 +255,34 @@ def tick(self):
...
+            celerybeat_lock.save()
+            # If current instance has lock and updated lock_timestamp, call super
             _logger.debug(_('Lock updated by %(celerybeat_name)s')
-                          % {'celerybeat_name': self.celerybeat_name})
-            ret = self.call_tick(self, self.celerybeat_name)
-        else:
-            # check for old enough time_stamp and remove if such lock is present
-            CeleryBeatLock.objects(timestamp__lte=old_timestamp).delete()
+                          % {'celerybeat_name': celerybeat_name})
+            ret = self.call_tick(self, celerybeat_name)
```
dkliban: same

tasking/pulp/tasking/scheduler.py
```diff
@@ -260,110 +255,34 @@ def tick(self):
...
+        except TaskLock.DoesNotExist:
+            TaskLock.objects(timestamp__lte=old_timestamp).delete()
```
dkliban: good catch! adding the name filter here.

tasking/pulp/tasking/scheduler.py
```diff
@@ -7,20 +8,15 @@
...
-from pulp.server.db.connection import UnsafeRetry
-from pulp.server.db.model import CeleryBeatLock, Worker
-from pulp.server.db.model.dispatch import ScheduledCall, ScheduleEntry
-from pulp.server.managers.schedule import utils
-from pulp.tasking import delete_worker
+from pulp.tasking import worker_watcher
 from pulp.tasking.celery_instance import celery as app
-from pulp.tasking.constants import TASKING_CONSTANTS
+from pulp.tasking import delete_worker
+from pulp.app.models.task import TaskLock, Worker
```
bmbouter: This import is out of order

tasking/pulp/tasking/scheduler.py
```diff
@@ -260,110 +251,34 @@ def tick(self):
...
+            celerybeat_lock.save()
+            # If current instance has lock and updated lock_timestamp, call super
             _logger.debug(_('Lock updated by %(celerybeat_name)s')
                           % {'celerybeat_name': self.celerybeat_name})
             ret = self.call_tick(self, self.celerybeat_name)
-        else:
-            # check for old enough time_stamp and remove if such lock is present
-            CeleryBeatLock.objects(timestamp__lte=old_timestamp).delete()
+        except TaskLock.DoesNotExist:
+            TaskLock.objects(name=self.celerybeat_name, timestamp__lte=old_timestamp).delete()
```
bmbouter: This needs to be `TaskLock.objects.get` I think

tasking/pulp/tasking/scheduler.py
```diff
@@ -260,110 +251,34 @@ def tick(self):
...
             try:
-                lock_timestamp = datetime.utcnow()
-
+                lock_timestamp = timezone.now()
                 # Insert new lock entry
-                new_lock = CeleryBeatLock(celerybeat_name=self.celerybeat_name,
-                                          timestamp=lock_timestamp)
-                new_lock.save()
+                TaskLock.objects.create(name=self.celerybeat_name, timestamp=lock_timestamp,
```
bmbouter: The `timestamp` will be set to now() automatically so it's not needed and it can be removed from L#265 too.

tasking/pulp/tasking/scheduler.py
```diff
@@ -117,38 +113,38 @@ def check_celery_processes(self):
...
+        resource_manager_count = Worker.objects.filter(
+            name__startswith=constants.RESOURCE_MANAGER_WORKER_NAME).count()
+
+        if not resource_manager_count:
             msg = _("There are 0 pulp_resource_manager processes running. Pulp will not operate "
                     "correctly without at least one pulp_resource_mananger process running.")
             _logger.error(msg)
 
-        if scheduler_count == 0:
+        if not scheduler_count:
```
bmbouter: For readability I think this was clearer with the ` == 0`. It would be good to make them all consistent.

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
     # The interval in seconds for which the Celery Process monitor thread sleeps between
     # checking for missing Celery processes.
     CELERY_CHECK_INTERVAL=60,
+    # The interval in seconds during which the Celery Processes being monitored need to report a
+    # hearbeat to be considered active.
+    CELERY_TIMEOUT_SECONDS=200,
```
bmbouter: Was this set to 200 before? I remember the age-out time being 300 seconds. Generally I want to lower these timing for Pulp3 anyway. Should we do that now or in another PR? I'm OK either way.

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
+    # The interval in seconds with which a secondary resource manager tries to acquire the lock
+    # that is needed to be the primary resource manager.
+    CELERY_TICK_DEFAULT_WAIT_TIME=90,
```
bmbouter: Consider renaming this to be `RESOURCE_MANAGER_LOCK_RETRY_ACQUIRE_TIME` or `RESOURCE_MANAGER_LOCK_RETRY_TIME`

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
     # Resource manager worker name
-    RESOURCE_MANAGER_WORKER_NAME='resource_manager'
+    RESOURCE_MANAGER_WORKER_NAME='resource_manager',
+    # celerybeat constants
+    SCHEDULER_WORKER_NAME = 'pulp_celerybeat',
+    # The maximum amount of time in seconds celerybeat sleeps before checking for more scheduled
+    # work to dispatch.
+    CELERY_MAX_INTERVAL=90,
```
bmbouter: s/CELERY_MAX_INTERVAL/CELERYBEAT_MAX_SLEEP_INTERVAL/
OR
s/CELERY_MAX_INTERVAL/CELERYBEAT_MAX_SLEEP/

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
+    # The amount of time in seconds that a celerybeat lock is valid for. If the lock is older than
+    # the value specified here, another instance of pulp_celerybeat is allowed to acquire the lock.
```
bmbouter: s/another instance of pulp_celerybeat is allowed to acquire the lock/the process holding the lock is considered to be dead/

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
+    # hearbeat to be considered active.
+    CELERY_TIMEOUT_SECONDS=200,
+    # The interval in seconds with which a secondary resource manager tries to acquire the lock
+    # that is needed to be the primary resource manager.
+    CELERY_TICK_DEFAULT_WAIT_TIME=90,
     # Resource manager worker name
-    RESOURCE_MANAGER_WORKER_NAME='resource_manager'
+    RESOURCE_MANAGER_WORKER_NAME='resource_manager',
+    # celerybeat constants
+    SCHEDULER_WORKER_NAME = 'pulp_celerybeat',
```
dkliban: sounds good to me. I am updating my PR wit this change.

tasking/pulp/tasking/constants.py
dkliban: i am going with `CELERYBEAT_MAX_SLEEP_INTERVAL`

tasking/pulp/tasking/constants.py
dkliban: i like `HEARTBEAT_MAX_AGE`.

tasking/pulp/tasking/constants.py
dkliban: thanks! i am updating the PR with this change.

tasking/pulp/tasking/constants.py
dkliban: So I looked at the code, and the comment was wrong. This value is used by the celerybeat only. It's how long celerybeat sleeps between trying to acquire the lock. I am changing this to `CELERYBEAT_LOCK_RETRY_TIME` and updating the comment.

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
+    # The interval in seconds during which the Celery Processes being monitored need to report a
+    # hearbeat to be considered active.
+    HEARTBEAT_MAX_AGE=200,
+    # The interval in seconds with which a secondary celerybeat tries to acquire the lock
+    # that is needed to be the primary celerybeat.
+    CELERYBEAT_LOCK_RETRY_TIME=90,
     # Resource manager worker name
-    RESOURCE_MANAGER_WORKER_NAME='resource_manager'
+    RESOURCE_MANAGER_WORKER_NAME='resource_manager',
+    # celerybeat constants
```
bmbouter: s/celerybeat constants/The name of the celerybeat entry in the workers table/

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
...
+    SCHEDULER_WORKER_NAME = 'celerybeat',
```
bmbouter: s/SCHEDULER/CELERYBEAT/

tasking/pulp/tasking/constants.py
```diff
@@ -4,6 +4,20 @@
     # The interval in seconds for which the Celery Process monitor thread sleeps between
     # checking for missing Celery processes.
     CELERY_CHECK_INTERVAL=60,
+    # The interval in seconds during which the Celery Processes being monitored need to report a
+    # hearbeat to be considered active.
+    HEARTBEAT_MAX_AGE=200,
+    # The interval in seconds with which a secondary celerybeat tries to acquire the lock
+    # that is needed to be the primary celerybeat.
+    CELERYBEAT_LOCK_RETRY_TIME=90,
     # Resource manager worker name
```
bmbouter: s/Resource manager worker name/The name of the resource manager entry in the workers table/

PR #2771, opened 2016-09-27
Converts tasking system to use Django models

bmbouter:
Moves all tasking code to pulp.platform.tasks and updates
existing usage to use the new location. Also the code
itself is converted to use the new Django models.

https://pulp.plan.io/issues/2154
closes #2154


app/etc/pulp/server.yaml
```diff
@@ -42,3 +42,16 @@
...
+# `broker`: Broker configuration for Pulp. By default, Pulp uses the Qpid
+# broker without SSL. It is possible to change this by providing a different
+# broker string. The remaining options are used to configure SSL settings.
+#
+# broker:
+#   url: qpid://localhost/
+#   celery_require_ssl: false
+#   cacert: /etc/pki/pulp/qpid/ca.crt
+#   keyfile: /etc/pki/pulp/qpid/client.crt
+#   certfile: /etc/pki/pulp/qpid/client.crt
```
jortel: Nitpicking but sure would be nice if the 3 ssl properties were named more consistency.

Consider:
```
ssl_ca_certificate:
ssl_client_key:
ssl_client_certificate:
```

Might be trying to fix too much and be outside the scope of this PR.

app/pulp/app/models/progress.py
```diff
@@ -44,14 +44,14 @@ class ProgressReport(Model):
     SKIPPED = 'skipped'
     RUNNING = 'running'
     COMPLETED = 'completed'
-    ERRORED = 'errored'
+    FAILED = 'failed'
```
jortel: :+1:

app/pulp/app/models/task.py
```diff
@@ -27,7 +27,7 @@ class ReservedResource(Model):
     resource = models.TextField()
 
     task = models.OneToOneField("Task")
-    worker = models.ForeignKey("Worker")
+    worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
```
jortel: Good catch.

common/pulp/common/constants.py
```diff
@@ -0,0 +1,17 @@
...
+CALL_FINISHED_STATE = 'finished'
+CALL_ERROR_STATE = 'error'
+CALL_CANCELED_STATE = 'canceled'
+
+CALL_INCOMPLETE_STATES = (CALL_WAITING_STATE, CALL_ACCEPTED_STATE, CALL_RUNNING_STATE,
+                          CALL_SUSPENDED_STATE)
+CALL_COMPLETE_STATES = (CALL_SKIPPED_STATE, CALL_FINISHED_STATE, CALL_ERROR_STATE,
+                        CALL_CANCELED_STATE)
+CALL_STATES = (CALL_WAITING_STATE, CALL_SKIPPED_STATE, CALL_ACCEPTED_STATE, CALL_RUNNING_STATE,
+               CALL_SUSPENDED_STATE, CALL_FINISHED_STATE, CALL_ERROR_STATE, CALL_CANCELED_STATE)
```
jortel: Again, maybe taking on too much but thinking `TASK_` would be more accurate.

common/pulp/common/error_codes.py
```diff
@@ -0,0 +1,160 @@
...
+                  "Binding creation was attempted for the repository %(repo_id)s and "
+                  "distributor %(distributor_id)s"),
+                ['repo_id', 'distributor_id', 'group_id'])
+
+PLP0005 = Error("PLP0005",
+                _("Errors occurred deleting bindings for the repository group %(group_id)s.  "
+                  "Binding deletion was attempted for the repository %(repo_id)s and "
+                  "distributor %(distributor_id)s"),
+                ['repo_id', 'distributor_id', 'group_id'])
+PLP0006 = Error("PLP0006", _("Errors occurred while updating the distributor configuration for "
```
jortel: Probably outside the scope of this PR and could file a separate task but `s/distributor/publisher/g`.

server/pulp/server/controllers/content.py
```diff
@@ -9,9 +9,9 @@
 from pulp.plugins.conduits.mixins import (ContentSourcesConduitException, StatusMixin,
                                           PublishReportMixin)
 from pulp.plugins.util.publish_step import Step
-from pulp.server.async.tasks import Task
 from pulp.server.content.sources.container import ContentContainer
 from pulp.server.exceptions import PulpCodedTaskException
+from pulp.tasking import UserFacingTask
```
jortel: :+1:  for importing from the package.

app/etc/pulp/server.yaml
bmbouter: This is a good suggestion. I'm fixing it in my next push

app/pulp/app/models/progress.py
mhrivnak: We need to make sure this gets into a list of backwards-incompatible changes. Perhaps it's time to start at least a skeleton for 3.0 release notes with a list of changes. It doesn't have to be well-written; we just need to not forget about "little" changes like this that have a huge impact on people's integration with pulp.

app/pulp/app/models/task.py
mhrivnak: I saw @seandst use this model-level setting which is very likable: https://docs.djangoproject.com/en/1.8/ref/models/options/#django.db.models.Options.default_related_name

common/pulp/common/constants.py
mhrivnak: Agreed. Keeps it simple and consistent.

common/pulp/common/constants.py
```diff
@@ -0,0 +1,17 @@
...
+CALL_ACCEPTED_STATE = 'accepted'
+CALL_RUNNING_STATE = 'running'
+CALL_SUSPENDED_STATE = 'suspended'
+CALL_FINISHED_STATE = 'finished'
+CALL_ERROR_STATE = 'error'
+CALL_CANCELED_STATE = 'canceled'
+
+CALL_INCOMPLETE_STATES = (CALL_WAITING_STATE, CALL_ACCEPTED_STATE, CALL_RUNNING_STATE,
+                          CALL_SUSPENDED_STATE)
+CALL_COMPLETE_STATES = (CALL_SKIPPED_STATE, CALL_FINISHED_STATE, CALL_ERROR_STATE,
```
mhrivnak: While we're nitpicking, only one of these states represents one where the task's work was completed, which could be slightly misleading. I suggest renaming to ``TASK_FINAL_STATES`` and ``TASK_NONFINAL_STATES``.

common/pulp/common/error_codes.py
```diff
@@ -0,0 +1,160 @@
+from collections import namedtuple
+from gettext import gettext as _
+
+
+Error = namedtuple('Error', ['code', 'message', 'required_fields'])
+"""
+The error named tuple has 4 components:
```
mhrivnak: s/4/3/

common/pulp/common/error_codes.py
bmbouter: I changed my mind about bringing in all of these exceptions. My next version only brings in the 4 that are in use by the new code. Others can be brought in as the rest of `pulp.server` is converted.

plugin/pulp/plugin/tasking.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    An interface to the currently running task.
+
+    With this object you can record non-fatal exceptions as they occur.
+
+    This object is natively task aware. When called inside of a task it will append the non-fatal
+    exception to the currently running task. When called outside of a task a
+    :class: `pulp.app.models.Task.DoesNotExist` Exception is raised.
+
+    Fatal errors should not use this. Instead they should raise an Exception, preferable one that
```
mhrivnak: s/preferable/preferably/

server/pulp/server/async/worker_watcher.py
```diff
@@ -72,14 +71,14 @@ def handle_worker_heartbeat(event):
     :type event: dict
     """
     event_info = _parse_and_log_event(event)
-    worker = Worker.objects(name=event_info['worker_name']).first()
-
-    if not worker:
+    try:
```
mhrivnak: This looks like a good opportunity for ``get_or_create``:

https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create

server/pulp/server/controllers/consumer.py
```diff
@@ -46,31 +40,29 @@ def unbind(consumer_id, repo_id, distributor_id, options):
...
-
-    response = TaskResult(result=binding)
-
     bind_manager.delete(consumer_id, repo_id, distributor_id, True)
-
-    return response
+    return binding
 
 
 def force_unbind(consumer_id, repo_id, distributor_id, options):
```
mhrivnak: Maybe out of scope for this PR, but I don't think we need this behavior anymore. Normally a consumer would be notified of an unbinding, and the agent would perform necessary changes on the consumer. This function was intended for a case where a consumer was not available to participate in its own unbinding.

common/pulp/common/constants.py
bmbouter: I've updated these, and also switched them to use [SimpleNamespace](https://docs.python.org/3.3/library/types.html#types.SimpleNamespace).

bmbouter: Thanks @seandst for the SimpleNamespace suggestion.

common/pulp/common/error_codes.py
bmbouter: fixed

common/pulp/common/constants.py
bmbouter: done

plugin/pulp/plugin/tasking.py
bmbouter: thanks. fixed

server/pulp/server/async/worker_watcher.py
bmbouter: A great idea, I'm about to do this.

server/pulp/server/controllers/consumer.py
bmbouter: These codepath will get removed by virtue of the eventual deletion of `pulp.server`.

common/pulp/common/constants.py
```diff
@@ -0,0 +1,15 @@
...
+    WAITING='waiting',
+    SKIPPED='skipped',
+    RUNNING='running',
+    COMPLETED='completed',
+    FAILED='failed',
+    CANCELED='canceled'
+)
+
+TASK_FINAL_STATES = (TASK_STATES.SKIPPED, TASK_STATES.COMPLETED, TASK_STATES.FAILED,
+                        TASK_STATES.CANCELED)
```
mhrivnak: pep8 indentation

server/pulp/server/async/worker_watcher.py
```diff
@@ -72,14 +71,12 @@ def handle_worker_heartbeat(event):
...
-    if not worker:
+    existing_worker, created = Worker.objects.get_or_create(name=event_info['worker_name'])
+    if created:
         msg = _("New worker '%(worker_name)s' discovered") % event_info
         _logger.info(msg)
-
-    Worker.objects(name=event_info['worker_name']).\
-        update_one(set__last_heartbeat=event_info['local_received'], upsert=True)
+    else:
+        existing_worker.save()
```
mhrivnak: Is this done to increment a timestamp as a side effect? I think it's worth an in-line comment saying so. Otherwise this appears unnecessary.

common/pulp/common/constants.py
bmbouter: fixed

server/pulp/server/async/worker_watcher.py
bmbouter: This is for exactly that reason. I added a comment.

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+
+    This function also provides randomization for worker selection.
+
+    :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with them.
+
+    :returns:          The Worker instance that has no reserved_resource
+                       entries associated with it.
+    :rtype:            pulp.app.model.Worker
+    """
+    free_workers = [Worker.objects.annotate(Count('reservations')).filter(reservations__count=0)]
```
mhrivnak: I haven't used ``annotate`` before, but if I'm understanding correctly, ``filter(...)`` should return a query set, right? And you want to evaluate that into a list? If so, I think the ``[...]`` notation will not do what you want, and will instead give you a list whose only member is a query set. I think you want ``list(...)`` here.

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+    if len(free_workers) == 0:
+        raise Worker.DoesNotExist()
+    random.shuffle(free_workers)
```
mhrivnak: This would be a slight improvement:
``return random.choice(free_workers)``

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+                            False.
+    :type normal_shutdown:  bool
+    """
+    if not normal_shutdown:
+        msg = _('The worker named %(name)s is missing. Canceling the tasks in its queue.')
+        msg = msg % {'name': name}
+        _logger.error(msg)
+
+    # Cancel all of the tasks that were assigned to this worker's queue
+    for task_status in Worker.objects.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
```
mhrivnak: Doesn't there need to be a ``get(name=name)`` in here somewhere to scope it to one worker?

For that matter, it probably makes sense to get the worker object first, then use that object to both find tasks and later call its ``delete()`` method.

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+    :type  task_id: basestring
+    """
+    try:
+        TaskStatus.objects.get(pk=task_id, state=TASK_STATES.RUNNING)
+    except TaskStatus.DoesNotExist:
+        pass
+    else:
+        new_task = PulpTask()
+        msg = _('The task status %(task_id)s exited immediately for some reason. Marking as '
+                'errored. Check the logs for more details')
```
mhrivnak: s/errored/failed/

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+        This updates the :attr: `started_at` and sets the :attr: `state` to :attr: `RUNNING`.
+
+        If the :class: `pulp.app.models.Task` status attribute :attr: `state` is
+        :attr: `CANCELED` the task is skipped.
+
+        Skip the status updating if the task is called synchronously.
+        """
+        if not self.request.called_directly:
+            task_status = TaskStatus.objects.get(pk=self.request.id)
+            if task_status.state != TaskStatus.WAITING:
```
mhrivnak: I'd consider moving the rest of this block into a method on the TaskStatus model. That would allow it to manage its own state change and any consistency rules among attributes. As-is, this code expresses and enforces consistency rules (1. state should be WAITING when moving to RUNNING, 2. ``started_at`` value must be set when moving to RUNNING) far from the actual data structure.

Instead, you'd just call something like:

``task_status.set_running()``

tasking/pulp/tasking/base.py
bmbouter: Oh yeah good catch. Will be fixed on next push.

tasking/pulp/tasking/base.py
bmbouter: I had meant to do this differently base on what @seandst told me last week about `order_by('?').first()`. The next push uses that.

tasking/pulp/tasking/base.py
bmbouter: good catch, fixed.

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+
+        :param args:    Original arguments for the executed task.
+        :type args:     list
+
+        :param kwargs:  Original keyword arguments for the executed task.
+        :type kwargs:   dict
+        """
+        _logger.debug("Task successful : [%s]" % task_id)
+        if not self.request.called_directly:
+            task_status = TaskStatus.objects.get(pk=task_id)
```
mhrivnak: Same as above. This would be appropriate logic to have on the ``TaskStatus`` model itself. Perhaps you'd call:

``task_status.set_completed(retval)``

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+            _logger.error(_('Task failed : [%(task_id)s] : %(msg)s') %
+                         {'task_id': task_id, 'msg': str(exc)})
+            _logger.error(traceback.format_exc())
+        else:
+            _logger.error(_('Task failed : [%s]') % task_id)
+            # celery will log the traceback
+
+        if not self.request.called_directly:
+            task_status = TaskStatus.objects.get(pk=task_id)
+            task_status.state = TaskStatus.FAILED
```
mhrivnak: Same here.

tasking/pulp/tasking/base.py
bmbouter: Good catch. I've made both of these improvements.

tasking/pulp/tasking/base.py
```diff
@@ -0,0 +1,444 @@
...
+
+def get_current_task_id():
+    """"
+    Get the current task id from celery. If this is called outside of a running
+    celery task it will return None
+
+    :return: The ID of the currently running celery task or None if not in a task
+    :rtype: str
+    """
+    if current_task and current_task.request and current_task.request.id:
```
mhrivnak: A nice opportunity for my favorite contextmanager! https://docs.python.org/3/library/contextlib.html#contextlib.suppress

```
with supress(AttributeError):
    return current_task.request.id
```

tasking/pulp/tasking/base.py
bmbouter: I just did this, look for it in the next push.

tasking/pulp/tasking/base.py
bmbouter: I implemented this.

tasking/pulp/tasking/base.py
bmbouter: I implemented this.

mhrivnak:
Love the changes. LGTM++

PR #2772, opened 2016-09-28
Convert pulp_resource_manager to use a django lock model

BrnoPCmaniak:
closes #2156
https://pulp.plan.io/issues/2156

This PR have to be merged into code created by https://github.com/pulp/pulp/pull/2771.

### DIFF
--- a/app/pulp/app/models/task.py
+++ b/app/pulp/app/models/task.py
@@ -54,6 +54,15 @@ class Worker(Model):
     name = models.TextField(db_index=True, unique=True)
     last_heartbeat = models.DateTimeField(auto_now=True)
 
+    def save_heartbeat(self):
+        """Save a worker heartbeat
+
+        Update the last_heartbeat field to now and save it.
+
+        !!! Only the last_heartbeat field will be saved. No other changes will be saved.
+        """
+        self.save(update_fields=['last_heartbeat'])
+
 
 class TaskLock(Model):
     """
--- a/server/pulp/server/async/worker_watcher.py
+++ b/server/pulp/server/async/worker_watcher.py
@@ -76,8 +76,7 @@ def handle_worker_heartbeat(event):
         msg = _("New worker '%(worker_name)s' discovered") % event_info
         _logger.info(msg)
     else:
-        # Saving the worker updates the timestamp.
-        existing_worker.save()
+        existing_worker.save_heartbeat()
 
 
 def handle_worker_offline(event):
--- a/tasking/pulp/tasking/base.py
+++ b/tasking/pulp/tasking/base.py
@@ -1,20 +1,21 @@
-from contextlib import suppress
-from gettext import gettext as _
 import logging
 import time
 import traceback
 import uuid
+from contextlib import suppress
+from gettext import gettext as _
 
-from celery import task, Task as CeleryTask, current_task
+from celery import Task as CeleryTask, current_task, task
 from celery.app import control
 from celery.result import AsyncResult
 from django.db.models import Count
 
-from pulp.app.models import Worker, ReservedResource, Task as TaskStatus
+from pulp.app.models import ReservedResource, Task as TaskStatus, TaskLock, Worker
 from pulp.common import TASK_FINAL_STATES, TASK_INCOMPLETE_STATES, TASK_STATES
-from pulp.exceptions import PulpException, MissingResource, PulpCodedException
-from pulp.tasking.celery_instance import celery, RESOURCE_MANAGER_QUEUE, DEDICATED_QUEUE_EXCHANGE
-
+from pulp.exceptions import MissingResource, PulpCodedException
+from pulp.tasking.celery_instance import celery
+from pulp.tasking.celery_instance import DEDICATED_QUEUE_EXCHANGE, RESOURCE_MANAGER_QUEUE
+from pulp.tasking.constants import TASKING_CONSTANTS
 
 celery_controller = control.Control(app=celery)
 _logger = logging.getLogger(__name__)
@@ -146,6 +147,9 @@ def delete_worker(name, normal_shutdown=False):
         cancel(task_status.pk)
 
     worker.delete()
+
+    if name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME):
+        TaskLock.objects.filter(name=name).delete()
 
 
 @task(base=PulpTask)
@@ -345,7 +349,7 @@ class UserFacingTask(PulpTask):
         """
         if isinstance(exc, PulpCodedException):
             _logger.error(_('Task failed : [%(task_id)s] : %(msg)s') %
-                         {'task_id': task_id, 'msg': str(exc)})
+                          {'task_id': task_id, 'msg': str(exc)})
             _logger.error(traceback.format_exc())
         else:
             _logger.error(_('Task failed : [%s]') % task_id)
--- a/tasking/pulp/tasking/celery_app.py
+++ b/tasking/pulp/tasking/celery_app.py
@@ -6,14 +6,17 @@ Celery setup finishes.
 
 import contextlib
 import logging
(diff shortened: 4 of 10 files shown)


BrnoPCmaniak:
The #2771  has been merged in Pulp 3 so I merged my code into it.

app/pulp/app/models/task.py
```diff
@@ -54,6 +54,15 @@ class Worker(Model):
     name = models.TextField(db_index=True, unique=True)
     last_heartbeat = models.DateTimeField(auto_now=True)
 
+    def heartbeat(self):
```
bmbouter: :+1:

app/pulp/app/models/task.py
```diff
@@ -54,6 +54,15 @@ class Worker(Model):
...
+        """Do a heartbeat on worker.
```
bmbouter: s/Do a heartbeat on a worker/Save a worker heartbeat/

app/pulp/app/models/task.py
```diff
@@ -54,6 +54,15 @@ class Worker(Model):
...
+
+        Update last_heartbeat field on Instance to now and perform save.
```
bmbouter: s/last_heartbeat/the last_heartbeat/
s/on Instance//
s/perform save/save it/

app/pulp/app/models/task.py
bmbouter: Consider renaming this to be s/heartbeat/save_heartbeat/

server/pulp/server/async/worker_watcher.py
```diff
@@ -76,8 +76,7 @@ def handle_worker_heartbeat(event):
         msg = _("New worker '%(worker_name)s' discovered") % event_info
         _logger.info(msg)
     else:
-        # Saving the worker updates the timestamp.
-        existing_worker.save()
+        existing_worker.heartbeat()
```
bmbouter: yes!

tasking/pulp/tasking/base.py
```diff
@@ -1,20 +1,22 @@
...
-from pulp.common import TASK_FINAL_STATES, TASK_INCOMPLETE_STATES, TASK_STATES
-from pulp.exceptions import PulpException, MissingResource, PulpCodedException
-from pulp.tasking.celery_instance import celery, RESOURCE_MANAGER_QUEUE, DEDICATED_QUEUE_EXCHANGE
-
+from pulp.app.models import Task as TaskStatus
+from pulp.app.models import ReservedResource, TaskLock, Worker
+from pulp.common import (
+    RESOURCE_MANAGER_WORKER_NAME, TASK_FINAL_STATES, TASK_INCOMPLETE_STATES, TASK_STATES)
+from pulp.exceptions import MissingResource, PulpCodedException
+from pulp.tasking.celery_instance import DEDICATED_QUEUE_EXCHANGE, RESOURCE_MANAGER_QUEUE, celery
```
bmbouter: maybe move `celery` to its own import line? It's un-alphabetized as it is now.

tasking/pulp/tasking/celery_app.py
```diff
@@ -69,6 +72,51 @@ def initialize_worker(sender, instance, **kwargs):
     # Delete any potential old state
     delete_worker(sender, normal_shutdown=True)
 
+    if sender.startswith(constants.RESOURCE_MANAGER_WORKER_NAME):
+        get_resource_manager_lock(sender)
+
+
+def get_resource_manager_lock(name):
+    """
+    Tries to acquire the resource manager lock. If the lock cannot be acquired immediately, it
```
bmbouter: It would be good to add a single-line docstring above this multi-line docstring. https://www.python.org/dev/peps/pep-0257/#id17

For example it could be:  `Block until the the resource manager lock is acquired.`

tasking/pulp/tasking/celery_app.py
```diff
@@ -69,6 +72,51 @@ def initialize_worker(sender, instance, **kwargs):
...
+    will wait until the currently active instance becomes unavailable, at which point the worker
+    cleanup routine will clear the lock for us to acquire. A worker record will be created so that
+    the waiting resource manager will appear in the Status API. We override the SIGTERM signal
+    handler so that that the worker record will be immediately cleaned up if the process is killed
+    while in this states.
+
+    :param name:   The hostname of the worker
+    :type  name:   basestring
+    """
+    assert name.startswith(constants.RESOURCE_MANAGER_WORKER_NAME)
```
bmbouter: I like putting assertions like this in code, but I haven't seen much of it in the pulp codebase before. :+1:

tasking/pulp/tasking/celery_app.py
```diff
@@ -69,6 +72,51 @@ def initialize_worker(sender, instance, **kwargs):
...
+        # Whether this is the first lock availability check for this instance
+        _first_check = True
+
+        while True:
+            worker.heartbeat()
+
+            try:
+                lock.save()
+
+                msg = _("Resource manager '%s' has acquired the resource manager lock" % name)
```
bmbouter: Can these L#106 - L#108 be moved to an `else:` clause since we aren't interested in the `except` clause applying to it.

common/pulp/common/constants.py
```diff
@@ -13,3 +13,9 @@
 TASK_FINAL_STATES = (TASK_STATES.SKIPPED, TASK_STATES.COMPLETED, TASK_STATES.FAILED,
                      TASK_STATES.CANCELED)
 TASK_INCOMPLETE_STATES = (TASK_STATES.WAITING, TASK_STATES.RUNNING)
+
+# The interval in seconds for which the Celery Process monitor thread sleeps between
+# checking for missing Celery processes.
+CELERY_CHECK_INTERVAL = 60
```
BrnoPCmaniak: I was even thinking if it shouldn't be included in settings instead of hard-coded constant.

app/pulp/app/models/task.py
BrnoPCmaniak: Fixed.

tasking/pulp/tasking/celery_app.py
BrnoPCmaniak: Fixed

BrnoPCmaniak:
I'm thinking about letting the tests how they are now edited, hoping it will ease the job for the one who will be writing them later.

tasking/pulp/tasking/base.py
bmbouter: I recommend making it two lines:
This is really a matter of opinion. I'm OK with whatever you decided to do here. My recommendation is:

```
from pulp.tasking.celery_instance import celery
from pulp.tasking.celery_instance import DEDICATED_QUEUE_EXCHANGE, RESOURCE_MANAGER_QUEUE
```

common/pulp/common/constants.py
bmbouter: I think leaving them hard-coded would be OK for now. If we make them settings then it's possible users could get undesirable behaviors if they aren't set correctly. I'm not ready to write all those docs so I propose leaving it hard-coded for now.

bmbouter: I do think they should be moved out of pulp/common/constants though. That area is more formally reserved in pulp3 for code that is common to the server and client. These constants will never be accessed on the client side so they can be moved into somewhere in pulp.tasking.

bmbouter:
@BrnoPCmaniak Leaving the tests fixed up sounds fine. It was more of an FYI that (1) the tests in pulp/server/... are not being run. Also (2) that you can fix or not fix these in the future. It's up to you.

tasking/pulp/tasking/constants.py
```diff
@@ -0,0 +1,9 @@
+from types import SimpleNamespace
+
+TASKING_CONSTANTS = SimpleNamespace(
```
bmbouter: :+1:

PR #2775, opened 2016-09-29
Update progress examples to use Django's create syntax

asmacdo:


bmbouter:
@asmacdo Why wouldn't think work?

app/pulp/app/models/progress.py
```diff
@@ -181,15 +181,6 @@ class ProgressBar(ProgressReport):
...
         >>>     for file in progress_bar.iter(files_iterator):
         >>>         handle(file)
-
-    You can also use this short form:
-
-        >>> with ProgressBar(message='Publishing files', total=len(files_iterator)):
-        >>>     for file in progress_bar.iter(files_iterator):
-        >>>         handle(file)
-
-    ProgressBar objects are associated with a Task and auto-discover and populate the task id when
```
bmbouter: This should stay so we can tell users that the task_id will be auto-set

app/pulp/app/models/progress.py
```diff
@@ -181,15 +181,6 @@ class ProgressBar(ProgressReport):
         >>> with progress_bar:
         >>>     for file in progress_bar.iter(files_iterator):
         >>>         handle(file)
-
-    You can also use this short form:
-
-        >>> with ProgressBar(message='Publishing files', total=len(files_iterator)):
```
bmbouter: Maybe this should be?

`with ProgressBar.objects.create(message='Publishing files', total=len(files_iterator)):`

asmacdo: The problem is that progress_bar is undefined.

bmbouter: Oh you are right. Instead of deleting it, what about updating it to be this:

`with ProgressBar.objects.create(message='Publishing files', total=len(files_iterator)) as progress_bar:`

bmbouter: Also L#127 and L#135 would both probably benefit from switching to `create()` also.

seandst:
I think the current interface is superior. Unless I'm missing something, making sure the model instance is saved in the `__enter__` method (and documenting the content manager appropriately to reflect that htis will happen) would let us keep the cleaner interface.

bmbouter:
@seandst that is an excellent idea. We should do that instead.

PR #2776, opened 2016-09-30
Implement worker and task directory creation and cleanup

BrnoPCmaniak:
closes #2281
https://pulp.plan.io/issues/2281

This PR have to be merged into code created by https://github.com/pulp/pulp/pull/2771


BrnoPCmaniak:
The #2771 has been merged in Pulp 3 so I merged my code into it.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
+import errno
+import os
+import shutil
+import stat
+
+from celery import task
+
```
bmbouter: `django.conf` and `celery` can be included in the same block. In terms of grouping they are both third-party libraries.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+from django.conf import settings as pulp_settings
+
+
+def _working_dir_root(worker_name):
+    """
+    Returns the path to the working directory of a worker
+    :param worker_name:     Name of worker for which path is requested
```
bmbouter: I usually have a line of whitespace between the description and the arguments.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    :type  name:            basestring
+    :return working_directory setting from server config
```
bmbouter: Similar for the rest of the docstrings below.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    :return Path to the working directory for a worker
+    :rtype  str
+    """
+    working_dir_root = _working_dir_root(worker_name)
+    os.mkdir(working_dir_root)
+
+
+def delete_worker_working_directory(worker_name):
+    """
+    Deletes a working directory inside the working_directory as specified in /etc/pulp/server.yaml
```
bmbouter: We should probably replace all usage of `/etc/pulp/server.yaml` with `the settings file`. The location could change in the future.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    default path for working_directory is /var/cache/pulp
```
bmbouter: We should probably not document the defaults here. It could change. Let the settings file be the canonical place where defaults are specified.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    :param worker_name:     Name of worker that uses the working directory being deleted
+    :type  name:            basestring
+    """
+    working_dir_root = _working_dir_root(worker_name)
+    if os.path.exists(working_dir_root):
```
bmbouter: Can this be adjusted to not check and then delete. It doubles the filesystem operations and [it's better to ask forgiveness](http://docs.quantifiedcode.com/python-code-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html).

bmbouter: I also know you didn't write this. It was in the old codebase.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _working_directory_path():
+    """
+    Gets the path for a task's working directory inside a workers working directory
+
+    @return: full path on disk to the working directory for current task
+    @rtype:  str
+    """
+    current_task = task.current
```
bmbouter: This could likely be done with much less code. Also to get the task ID I recommend calling [get_current_task_id()](https://github.com/pulp/pulp/blob/6e81d8d723144c586933192754b8d99fe8af3414/tasking/pulp/tasking/base.py#L403-L412). Can this be simplified some?

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+def delete_working_directory():
+    """
+    Deletes working directory for a particular task.
+
+    @return: None
+    @rtype:  None
+    """
+    working_dir_root = _working_directory_path()
+
+    if working_dir_root and os.path.exists(working_dir_root):
```
bmbouter: Similar change requested here. Try to delete it and handle exceptions accordingly.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    is created only once per task. The path to the existing working directory is returned for all
+    subsequent calls for that task.
+
+    @return: full path on disk to the working directory for current task
+    @rtype:  str
+    """
+    working_dir_root = _working_directory_path()
+
+    if working_dir_root:
+        if os.path.exists(working_dir_root):
```
bmbouter: It would be better to attempt to create it and if it already exists handle the exception.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+            return working_dir_root
+        os.mkdir(working_dir_root)
+        return working_dir_root
+    else:
+        # If path is None, this method is called outside of an asynchronous task
+        raise RuntimeError("Working Directory requested outside of asynchronous task.")
+
+
+def _rmtree_fix_permissions(directory_path):
```
bmbouter: I'm not sure why this is a thing we have. Was there a bug that caused us to write this code?

bmbouter: Based on the comment this is to recover from user error. I propose we remove this from pulp3.

mhrivnak: I think we should keep this behavior when removing working directories, but the comment could be improved to give better context.

I believe the original problem this addressed was that when extracting the contents of a puppet module, the permissions on files could be unexpected. We've seen a surprising number of such modules. Cleanup of the working directory was not able to remove some files without first changing their permissions.

That particular plugin now makes efforts to fix those permissions so they are reasonable, but it's a best-effort. The plugin does not guarantee that all files it leaves behind will have reasonable permissions.

Any other plugin that engages in extraction of files from an archive could face the same issue. Having the platform be defensive in this way is reasonable I think.

bmbouter: @mhrivnak that makes sense; sounds fine
@BrnoPCmaniak let's keep this functionality then.

tasking/pulp/tasking/storage.py
BrnoPCmaniak: I simplified a lot this function, but the `get_current_task_id()` is in base and that causes a circular import. So the only thing I can do is to move the function or le it be.

BrnoPCmaniak:
@bmbouter  It's done ready for next round. But i'm thinking of renaming the functions to be in some consistent manner.

tasking/pulp/tasking/storage.py
BrnoPCmaniak: Fixed

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+from celery import task
+
+from django.conf import settings as pulp_settings
+
+
+def _working_dir_root(worker_name):
+    """
+    Returns the path to the working directory of a worker
+    :param worker_name:     Name of worker for which path is requested
+    :type  name:            basestring
```
BrnoPCmaniak: Fixed

tasking/pulp/tasking/storage.py
BrnoPCmaniak: Fixed

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,123 @@
...
+    working_dir_root = _working_dir_root(worker_name)
+    if os.path.exists(working_dir_root):
+        _rmtree_fix_permissions(working_dir_root)
+
+
+def _working_directory_path():
+    """
+    Gets the path for a task's working directory inside a workers working directory
+
+    @return: full path on disk to the working directory for current task
```
BrnoPCmaniak: I forgot this at `get_working_directory` but on all others fixed.

tasking/pulp/tasking/storage.py
BrnoPCmaniak: Fixed

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    Create a working directory inside the working_directory as specified in the setting file.
+
+    :param worker_name:     Name of worker that uses the working directory created
+    :type  worker_name:     basestring
+
+    :returns: Path to the working directory for a worker
+    :rtype:   str
+    """
+    working_dir_root = _working_dir_root(worker_name)
+    os.mkdir(working_dir_root)
```
bmbouter: Should this suppress the exceptions if the directory already exists?

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+
+def _working_directory_path():
+    """Get path of task working directory.
+
+    Get the path for a task working directory inside a workers working directory.
+
+    :returns: full path on disk to the working directory for current task
+    :rtype:   basestring
+    """
+    current_task = task.current
```
bmbouter: Could this become?

```
    # above put  `from pulp.tasking import get_current_task_id`
     task_id = get_current_task_id()
     worker_name = current_task.request.hostname
     if task_id:
         return os.path.join(_working_dir_root(worker_name), task_id)
```

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    worker_name = current_task.request.hostname
+    if current_task and current_task.request and task_id and worker_name:
+        return os.path.join(_working_dir_root(worker_name), task_id)
+
+
+def delete_working_directory():
+    """Delete working directory for a particular task."""
+    working_dir_root = _working_directory_path()
+
+    if working_dir_root:
```
bmbouter: I believe it would be better to not check and instead try to delete and suppress exceptions.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    @return: full path on disk to the working directory for current task
+    @rtype:  str
+    """
+    working_dir_root = _working_directory_path()
+
+    if working_dir_root:
+        try:
+            os.mkdir(working_dir_root)
+        except OSError as error:
+            if error.errno is errno.EEXIST:
```
bmbouter: :+1: good use of `errno`

tasking/pulp/tasking/storage.py
bmbouter: I see that `_rmtree` already does the exception suppression so you could just move the _rmtree implemntation into this function.

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    """Create a working directory with task id as name.
+
+    This directory resides inside a particular worker's working directory.
+    The 'working_directory' setting in ``server`` section of settings defines the local path
+    to the root of working directories.
+
+    The directoryis created only once per task. The path to the existing working directory
+    is returned for all subsequent calls for that task.
+
+    @return: full path on disk to the working directory for current task
```
bmbouter: Use `:return:` instead

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    @rtype:  str
```
bmbouter: use `:rtype:` instead

tasking/pulp/tasking/storage.py
BrnoPCmaniak: @bmbouter Should I use `:return:` or `:returns:`?

tasking/pulp/tasking/storage.py
BrnoPCmaniak: @bmbouter I can't I'm using it also in `delete_worker_working_directory`, but I removed the check

tasking/pulp/tasking/storage.py
bmbouter: That is a fine reason to leave it here. Maybe make it cleaner by using `suppress`? This implementation looks similar to what `get_current_task_id()` before it was refactored.

tasking/pulp/tasking/storage.py
bmbouter: Not suppressing it sounds fine. It sounds like that is what the code that calls this function wants too.

tasking/pulp/tasking/storage.py
bmbouter: in imports

tasking/pulp/tasking/storage.py
bmbouter: ok sounds good

tasking/pulp/tasking/storage.py
bmbouter: `:returns:` good catch!

tasking/pulp/tasking/storage.py
```diff
@@ -0,0 +1,139 @@
...
+            return working_dir_root
+    else:
+        # If path is None, this method is called outside of an asynchronous task
+        raise RuntimeError("Working Directory requested outside of asynchronous task.")
+
+
+def _rmtree(path):
+    """Delete if exists an entire directory tree in path.
+
+    Uses _rmtree_fix_permissions but catches out 'No such file or directory' Exception.
```
bmbouter: s/catches out/suppresses/

bmbouter:
Thanks @BrnoPCmaniak  great job

PR #2778, opened 2016-10-03
Updates saving behaviors of `ProgressReport`

bmbouter:
The context manager now provides saving in all cases.
Also the docstrings have been updated to make this
behavior clear to the user.

Also the short form of `ProgressBar` was fixed
to create the 'pb' variable within the 'with'
section.

https://pulp.plan.io/issues/2092
re #2092

### DIFF
--- a/app/pulp/app/models/progress.py
+++ b/app/pulp/app/models/progress.py
@@ -82,9 +82,8 @@ class ProgressReport(Model):
         """
         Saves the progress report state as RUNNING
         """
-        if self.state != self.RUNNING:
-            self.state = self.RUNNING
-            self.state.save()
+        self.state = self.RUNNING
+        self.save()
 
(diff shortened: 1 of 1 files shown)


app/pulp/app/models/progress.py
```diff
@@ -82,9 +82,8 @@ def __enter__(self):
         """
         Saves the progress report state as RUNNING
         """
-        if self.state != self.RUNNING:
-            self.state = self.RUNNING
-            self.state.save()
+        self.state = self.RUNNING
+        self.state.save()
```
seandst: Should this just be `self.save()`? `self.state` isn't a relation, so there's nothing to save there.

bmbouter: yup...

bmbouter: new version is pushed

PR #2780, opened 2016-10-07
Updates the nodes install and quickstart

bmbouter:
The nodes install and quickstart was leaving out
an important step on the child node to configure
the server.conf on the child node.

https://pulp.plan.io/issues/1392
closes #1392

### DIFF
--- a/docs/user-guide/nodes.rst
+++ b/docs/user-guide/nodes.rst
@@ -113,9 +113,9 @@ To install *Nodes* child support, follow the instructions below.
 
  $ sudo yum install pulp-nodes-child pulp-nodes-consumer-extensions pulp-agent
 
-2. The communication between the child and parent nodes is secured using OAuth. The child node
-   must have OAuth enabled and configured. Please see :ref:`OAuth <oauth-config>` for instructions
-   on enabling and configuring OAuth.
+2. The agent on the child node uses the local pulp REST API which is secured using OAuth.
+   The child node must have OAuth enabled and configured. Please see :ref:`OAuth <oauth-config>`
+   for instructions on enabling and configuring OAuth.
 
 ::
 
@@ -399,15 +399,24 @@ On the Pulp server to be used as the parent node:
 ::
 
   $ sudo yum install pulp-nodes-parent pulp-nodes-admin-extensions
-  $ sudo service httpd restart
 
-2. Enable the ``pulp-goodness`` repository.
+2. Configure and enable :ref:`OAuth <oauth-config>`.
+
+3. Restart Apache.  For upstart::
+
+     $ sudo service httpd restart
+
+   For systemd::
+
+     $sudo systemctl restart httpd
+
+4. Enable the ``pulp-goodness`` repository.
 
 ::
 
  $ pulp-admin node repo enable --repo-id pulp-goodness
 
-3. Publish the ``pulp-goodness`` repository.
+5. Publish the ``pulp-goodness`` repository.
 
 ::
 
@@ -425,7 +434,10 @@ On the Pulp server to be used as the child node:
 
   $ sudo yum install pulp-nodes-child pulp-nodes-consumer-extensions pulp-agent
 
-2. Edit ``/etc/pulp/nodes.conf`` and set the parent OAuth *key* and *secret* to match values found in
+2. Configure and enable :ref:`OAuth <oauth-config>`. This should use different credentials from
+   the parent for security.
+
+3. Edit ``/etc/pulp/nodes.conf`` and set the parent OAuth *key* and *secret* to match values found in
    ``/etc/pulp/server.conf`` on the parent node.
 
 ::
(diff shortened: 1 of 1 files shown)


docs/user-guide/nodes.rst
```diff
@@ -113,44 +113,29 @@ To install *Nodes* child support, follow the instructions below.
...
  [oauth]
- enabled: true
- oauth_key: Xohkaethaama5eki
- oauth_secret: eePa7Bi3gohdir1pai2icohvaidai0io
-
-.. warning::
-
-   Do not use the key or secret given in the above example. It is important that you use unique and
-   secret values for these configuration items.
+ user_id:  <EDIT: admin user on parent node>
```
jortel: This is the admin `user_id` on the child.  The `[oauth]` section in nodes.conf is for the *local* child REST API.  Not the parent.  Those settings are in `[parent_oauth]`.

docs/user-guide/nodes.rst
```diff
@@ -113,44 +113,29 @@ To install *Nodes* child support, follow the instructions below.
...
 
-3. Edit ``/etc/pulp/nodes.conf`` on the child node and set the parent OAuth *key* and *secret* to
-   match values found in ``/etc/pulp/server.conf`` on the parent node. The *user_id* must be updated
-   as needed to match a user with administration privileges on the parent node.
```
jortel: The example here is good but by removing the description which includes editing `/etc/pulp/nodes.conf' results in incorrectly instructing the user to edit `/etc/pulp/server.conf` for the `[parent_node]` settings.

docs/user-guide/nodes.rst
```diff
@@ -425,31 +418,47 @@ On the Pulp server to be used as the child node:
 
   $ sudo yum install pulp-nodes-child pulp-nodes-consumer-extensions pulp-agent
 
-2. Edit ``/etc/pulp/nodes.conf`` and set the parent OAuth *key* and *secret* to match values found in
+2. Edit ``/etc/pulp/server.conf`` on the child node to configure OAuth to match the values found in
```
jortel: This change is incorrect.  The matching `[parent_oauth]` settings are in `/etc/pulp/nodes.conf` not `server.conf`.

jortel:
Looking at the original documentation, it's seems clearly stated that OAuth must be enabled and configured on the child node.  I appreciate the attempt to clarify this further here but many of the changes are incorrect.

bmbouter:
@jortel I understand nodes less than I thought. Maybe the best thing to do is for me to close this PR? What do you think I should do?

jortel:
@bmbouter I would suggest reverting most the changes except the example you added showing how to edit `[parent_oauth]`.  Then, perhaps edit step 2 to read like:

> 2. The agent on the child node uses the local pulp REST API which is secured using OAuth. The child node must have OAuth enabled and configured. Please see OAuth for instructions on enabling and configuring OAuth.

bmbouter:
I reverted the previous code, and I brought back the suggestions from [this comment](https://github.com/pulp/pulp/pull/2780#issuecomment-252664185). @jortel would you be able to give me some more feedback?

The current commit I've made doesn't seem like enough because it updates the guide but not the quickstart. The bug report and related pulp-list e-mails suggest that the quickstart needs updating. Are those reports wrong or are there updates to the quickstart that I need to make also?

jortel:
Looking at https://docs.pulpproject.org/ I don't see any sections named *quickstart* to be updated.  Can you help me find it?

bmbouter:
@jortel I believe they are referring to the  [section at the bottom of that page](http://docs.pulpproject.org/user-guide/nodes.html#quick-start). What do you think about that content?

jortel:
Oh, right.  I see it.  

Suggest you edit the `Quick Start` and remove all install setup instructions.  Begin the section saying that it's assumed that the parent and child nodes have been setup as described above.  Then, the  `Quick Start` only contains example commands for what to do with it.

bmbouter:
@jortel I updated it per our discussion. Would you be able to give it another look? Is there anything else I could do to improve or correct what is here? Thanks!

docs/user-guide/nodes.rst
```diff
@@ -396,18 +396,20 @@ On the Pulp server to be used as the parent node:
 
 1. Install the pulp-nodes-parent package group.
 
+2. Configure and enable :ref:`OAuth <oauth-config>`.
+
```
bmbouter: @jortel Thanks for pointing this out. I pushed a new version. Would you be able to take another look?

docs/user-guide/nodes.rst
```diff
@@ -401,13 +401,15 @@ On the Pulp server to be used as the parent node:
   $ sudo yum install pulp-nodes-parent pulp-nodes-admin-extensions
   $ sudo service httpd restart
```
jortel: Sorry, just noticed the example includes restarting http. The http restart needs to happen after the oauth configuration change. The example needs to be split. For consistency with On The Child instructions, you probably should cut-n-paste from there.

PR #2781, opened 2016-10-07
Prompt user of running workers/tasks on migration

werwty:
Add close() method to celery scheduler for clean exit.
Catch ServerSelectionTimeoutError when mongo is down.

closes #2186
https://pulp.plan.io/issues/2186

### DIFF
--- a/docs/user-guide/release-notes/2.11.x.rst
+++ b/docs/user-guide/release-notes/2.11.x.rst
@@ -10,3 +10,5 @@ New Features
 
 * For RPM content, a full sync will be forced if the sync configuration has been changed or content
   has been removed since the last sync.
+
+* When pulp-manage-db is run, prompt the user to continue if there are Pulp services still running.
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -192,6 +192,9 @@ class Scheduler(beat.Scheduler):
         self._schedule = None
         self._loaded_from_db_count = 0
 
+        # Setting the celerybeat name
+        self.celerybeat_name = constants.SCHEDULER_WORKER_NAME + "@" + platform.node()
+
         # Force the use of the Pulp celery_instance when this custom Scheduler is used.
         kwargs['app'] = app
 
@@ -241,8 +244,6 @@ class Scheduler(beat.Scheduler):
         :return:    number of seconds before the next tick should run
         :rtype:     float
         """
-        # Setting the celerybeat name
-        celerybeat_name = constants.SCHEDULER_WORKER_NAME + "@" + platform.node()
 
         # this is not an event that gets sent anywhere. We process it
         # immediately.
@@ -250,7 +251,7 @@ class Scheduler(beat.Scheduler):
             'timestamp': time.time(),
             'local_received': time.time(),
             'type': 'scheduler-event',
-            'hostname': celerybeat_name
+            'hostname': self.celerybeat_name
         }
 
         worker_watcher.handle_worker_heartbeat(scheduler_event)
@@ -258,14 +259,14 @@ class Scheduler(beat.Scheduler):
         old_timestamp = datetime.utcnow() - timedelta(seconds=constants.CELERYBEAT_LOCK_MAX_AGE)
 
         # Updating the current lock if lock is on this instance of celerybeat
-        result = CeleryBeatLock.objects(celerybeat_name=celerybeat_name).\
+        result = CeleryBeatLock.objects(celerybeat_name=self.celerybeat_name).\
             update(set__timestamp=datetime.utcnow())
 
         # If current instance has lock and updated lock_timestamp, call super
         if result == 1:
             _logger.debug(_('Lock updated by %(celerybeat_name)s')
-                          % {'celerybeat_name': celerybeat_name})
-            ret = self.call_tick(self, celerybeat_name)
+                          % {'celerybeat_name': self.celerybeat_name})
+            ret = self.call_tick(self, self.celerybeat_name)
         else:
             # check for old enough time_stamp and remove if such lock is present
             CeleryBeatLock.objects(timestamp__lte=old_timestamp).delete()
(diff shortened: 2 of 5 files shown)


bmbouter:
@werwty FYI there is another PR which moves code you are editing. We need to somehow coordinate these two efforts.

server/pulp/server/async/scheduler.py
```diff
@@ -368,3 +369,6 @@ def add(self, **kwargs):
         entries to the database, and they will be picked up automatically.
         """
         raise NotImplementedError
+
+    def close(self):
```
bmbouter: What calls this?

server/pulp/server/db/manage.py
```diff
@@ -191,6 +193,14 @@ def main():
         options = parse_args()
         _start_logging()
         connection.initialize(max_timeout=1)
+
+        # Prompt the user if there are active workers or running tasks
+        if status.get_workers() or filter(lambda x: x.state in constants.CALL_INCOMPLETE_STATES,
```
bmbouter: I think we should disinclude the check for tasks in the incomplete states. The only unsafe aspect really is if workers are running. Sometimes bugs put task states incorrectly in an incomplete state so I don't want those bugs to have an impact on this functionality.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +193,14 @@ def main():
...
+                                          model.TaskStatus.objects(group_id=None)):
+
+            if not _user_input_continue('There are still running workers/tasks,'
```
bmbouter: Also what if we don't give the user an option to continue at all. It seems unsafe to continue. If we do continue to prompt we need to reword this to clearly state that it is likely unsafe to continue.

server/pulp/server/db/manage.py
bmbouter: Also what about checking the timestamps? This will only work if all processes are gracefully shut down. If we check the timestamps also we could pick up cases for non-graceful shutdown. What do you think?

server/pulp/server/db/manage.py
bmbouter: These user facing strings should be i8ln

server/pulp/server/async/scheduler.py
werwty: scheduler close is called by the celery beat service object when it does a sync:
https://github.com/celery/celery/blob/master/celery/beat.py#L544

By default scheduler.close just calls scheduler.sync, which does not seem to do much. Overriding it to clean up our db seems like the cleanest solution.
https://github.com/celery/celery/blob/master/celery/beat.py#L331

server/pulp/server/db/manage.py
```diff
@@ -200,6 +210,9 @@ def main():
         return os.EX_DATAERR
     except models.MigrationRemovedError:
         return os.EX_SOFTWARE
+    except ServerSelectionTimeoutError:
```
werwty: If the mongodb service is down the exception is raised when _CONNECTION.server_info() is called at:
https://github.com/pulp/pulp/blob/master/server/pulp/server/db/connection.py#L119

currently we're not catching this exception and show a stack trace to the user

server/pulp/server/db/manage.py
werwty: Good point, I'll add a worker timestamp checking and remove the task checking

server/pulp/server/db/manage.py
werwty: I would rather give the user an option to continue, since personally, I prefer software to trust me.
Instead I will reword this to induce more fear.

server/pulp/server/async/scheduler.py
bmbouter: We call `super()` here. It may not do much but since we are overriding it and we didn't before we should preserve to codepath.

server/pulp/server/db/manage.py
bmbouter: Niiiice.  :+1:

server/pulp/server/db/manage.py
```diff
@@ -191,6 +195,15 @@ def main():
         options = parse_args()
         _start_logging()
         connection.initialize(max_timeout=1)
+
+        # Prompt the user if there are workers that have not timed out
+        if filter(lambda worker: (UTCDateTimeField().to_python(datetime.now()) -
+                                  worker['last_heartbeat']).total_seconds() <
+                  constants.CELERY_TIMEOUT_SECONDS, status.get_workers()):
+            if not _user_input_continue('There are still running workers, continuing can '
+                                        'cause major data issues. Are you sure you wish '
```
bmbouter: I like the language update. I was thinking about how to introduce an `corrupt` as an even scarier word. Consider s/can cause major data issues/could corrupt your Pulp installation/  I'm OK with it as is, so this is not a required change.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +195,15 @@ def main():
         options = parse_args()
         _start_logging()
         connection.initialize(max_timeout=1)
+
+        # Prompt the user if there are workers that have not timed out
+        if filter(lambda worker: (UTCDateTimeField().to_python(datetime.now()) -
+                                  worker['last_heartbeat']).total_seconds() <
+                  constants.CELERY_TIMEOUT_SECONDS, status.get_workers()):
+            if not _user_input_continue('There are still running workers, continuing can '
```
bmbouter: Can this string be i8ln

server/pulp/server/db/manage.py
```diff
@@ -274,3 +290,13 @@ def _start_logging():
     console_handler = logging.StreamHandler()
     console_handler.setLevel(logging.INFO)
     _logger.root.addHandler(console_handler)
+
+
+def _user_input_continue(question):
+    reply = str(raw_input(_(question + ' (y/n): '))).lower().strip()
```
bmbouter: Can this default to no? I think the prompt would become `(y/N`)` What do you think about this?

bmbouter:
A release note needs to be added for this feature

docs/user-guide/release-notes/2.11.x.rst
```diff
@@ -10,3 +10,5 @@ New Features
 
 * For RPM content, a full sync will be forced if the sync configuration has been changed or content
   has been removed since the last sync.
+
+* When pulp-manage-db is called, prompt the user to continue if there are running workers.
```
bmbouter: s/called/run/
s/running workers/Pulp services still running/

PR #2782, opened 2016-10-10
Finish Repository serializer

asmacdo:
Add remaining fields and move fiter_fields back to ViewSet.

closes #2298

### DIFF
--- a/app/pulp/app/serializers/repository.py
+++ b/app/pulp/app/serializers/repository.py
@@ -1,7 +1,7 @@
 from rest_framework import serializers
 
 from pulp.app import models
-from pulp.app.serializers import GenericKeyValueRelatedField, ModelSerializer
+from pulp.app.serializers import NotesKeyValueRelatedField, ModelSerializer
 
 
 class RepositorySerializer(ModelSerializer):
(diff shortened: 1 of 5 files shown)


app/pulp/app/serializers/repository.py
```diff
@@ -15,11 +15,26 @@ class RepositorySerializer(ModelSerializer):
         help_text='A unique name for this repository.'
     )
 
+    description = serializers.CharField(
+        help_text='An optional description.',
+        required=False
+    )
+
+    last_content_added = serializers.DateTimeField(
+        help_text='When content was last added.',
```
seandst: Could this help text be more verbose, like "Timestamp of the most recent addition of content to this repository." ?

Same for `last_content_removed`.

seandst: For some background, my thinking here is that the API is user-facing, so it might be worth being a little bit more verbose than what we've got in the Model docstrings due to the different audience (API user vs. Python developer).

asmacdo: Agreed. Fixed in most recet commit.

PR #2785, opened 2016-10-13
Repository syncs show all units updated even when there are no changes.

ipanova:
closes #2328
https://pulp.plan.io/issues/2328


ipanova:
@mhrivnak  this changed was required to cover this situation:
1. create repo with immediate policy and sync it
2 . drop DB
3. create repo with whether on_demand policy but then trigger 'download_repo' or just with backgroung.
Without this change the downloaded flag after step 3 would still be set to False.

PR #2786, opened 2016-10-13
Adds pulpproject.org updates as build steps

bmbouter:
There are updates to the http://pulpproject.org/docs/
page that need to happen with different build types.
Those instructions are now added present.


bmbouter:
This does not well capture the update that would need to happen when a new X.Y nightly is created on Jenkins. I'm not sure where to write those docs.

PR #2790, opened 2016-10-17
Fix SSLError resulting in http 503 when using proxy and SSL.

jortel:
https://pulp.plan.io/issues/2344

The `ProxyManager` IsA `PoolManager` and the code needs to update the SSL settings on either pool manager when retrieving a connection.

### DIFF
--- a/streamer/pulp/streamer/adapters.py
+++ b/streamer/pulp/streamer/adapters.py
@@ -123,19 +123,24 @@ class PulpHTTPAdapter(HTTPAdapter):
 
         return self.proxy_manager[proxy]
 
-    def _update_poolmanager_ssl_kw(self, verify, cert):
+    @staticmethod
+    def _update_poolmanager_ssl_kw(pool_manager, verify, cert):
         """Update the :class:`PoolManager <urllib3.poolmanager.PoolManager>`
         connection_pool_kw with the necessary SSL configuration. This method
         should not be called from user code, and is only exposed for use when
         subclassing the :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
 
+        :param pool_manager: A pool manager to be updated.
+        :type  pool_manager: urllib3.poolmanager.PoolManager
         :param verify: Whether we should actually verify the certificate;
                        optionally a path to a CA certificate bundle or
                        directory of CA certificates.
+        :type  verify: str
         :param cert: The path to the client certificate and key, if any.
                      This can either be the path to the certificate and
                      key concatenated in a single file, or as a tuple of
                      (cert_file, key_file).
+        :type  cert: tuple
         """
         if verify:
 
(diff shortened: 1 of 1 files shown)


streamer/pulp/streamer/adapters.py
```diff
@@ -212,22 +216,31 @@ def get_connection(self, url, proxies=None, verify=None, cert=None):
...
+                pool_manager = self.proxy_manager_for(proxy)
             else:
-                # Only scheme should be lower case
-                parsed = urlparse(url)
-                url = parsed.geturl()
-                conn = self.poolmanager.connection_from_url(url)
+                pool_manager = self.poolmanager
+
+            if scheme == 'https':
+                self._update_poolmanager_ssl_kw(pool_manager, verify, cert)
```
jortel: This is the main fix.  To apply the SSL settings to either the main `PoolManager` or the `ProxyManager`.

jeremycline:
I did my best in the module comments to make this clear, but this is a vendored module. You definitely don't want to do this and you definitely don't want to re-write this for Pulp 3. The idea of re-implementing pooling by holding on to sessions was raised and dismissed during the initial fix. I believe there is an issue somewhere that discusses why that is not a good idea.

mhrivnak:
@jeremycline thanks for the feedback. Just to confirm: I assume when you say "You definitely don't want to do this", you are referring to the proposed future changes, and not the current bug fix in this PR. If that's not the case, and you do see a flaw in this bug fix, please add some details. Thanks!

streamer/pulp/streamer/adapters.py
```diff
@@ -212,22 +216,31 @@ def get_connection(self, url, proxies=None, verify=None, cert=None):
...
 
             if proxy:
                 proxy = prepend_scheme_if_needed(proxy, 'http')
-                proxy_manager = self.proxy_manager_for(proxy)
-                conn = proxy_manager.connection_from_url(url)
+                pool_manager = self.proxy_manager_for(proxy)
             else:
-                # Only scheme should be lower case
-                parsed = urlparse(url)
-                url = parsed.geturl()
```
mhrivnak: These two lines are interesting. I assume they're here just to normalize the URL, but it's unclear what value that provides. Maybe that helps with the pooling algorithm? In any case, since this came from upstream ``requests``, I think we should keep these two lines.

streamer/pulp/streamer/adapters.py
```diff
@@ -123,19 +123,23 @@ def proxy_manager_for(self, proxy, **proxy_kwargs):
...
         connection_pool_kw with the necessary SSL configuration. This method
         should not be called from user code, and is only exposed for use when
         subclassing the :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
 
+        :param pool_manager: A pool manager to be updated.
+        :type  pool_manager: urllib3.poolmanager.PoolManager
         :param verify: Whether we should actually verify the certificate;
                        optionally a path to a CA certificate bundle or
                        directory of CA certificates.
+        :type  verify: str
```
jortel: I think you're right so I changed to a `str` but the name and code below are confusing:

```
        if verify:

            cert_loc = None

            # Allow self-specified cert location.
            if verify is not True:
                cert_loc = verify
```

streamer/pulp/streamer/adapters.py
mhrivnak: Definitely.

jeremycline:
@mhrivnak to be very clear, when I said "you don't want to do this" I mean this PR. This is a backported requests 3.0 HTTPAdapter, so the main problem with this proposed change is that it's not actually fixing the problem. Since this is in upstream requests, it *needs* to get fixed in upstream requests unless you plan to fork requests.

As for the patch itself, I have not reviewed so I cannot comment on it. I recommend that you fix this in the proper place (upstream) and backport the accepted patch.

mhrivnak:
@jeremycline Ah ok, thanks for that clarification. We will definitely make sure this gets fixed upstream. Right now we have a lot of users dead in the water, so we are going to get a fix to them ASAP. We will also get the fix submitted upstream, and adjust here as necessary if that upstream fix ends up being different.

jortel:
@jeremycline After reading the module docstring this all makes more sense now.  I should have started there but in most pulp modules, the module docstring is useless.  I will submit the same patch upstream.  Though may rename the lock to be clearer.  Disregard my suggestion to rewrite this in pulp3 since this is patching `requests` and nothing to do with `pulp`.  I still disagree with modifying the `PoolManager.connection_pool_kw()` each time in `get_connection()` instead of perhaps passing `verify` and `cert` to `PoolManager.connection_from_url()`.  But the upstream community has excepted so it's "water under the bridge".  Given how this all works in `requests`, I'm no longer interested in revisiting anything to do with session pooling in pulp streamer.

PR #2792, opened 2016-10-19
Add Task Serializer

werwty:
re #2300

### DIFF
--- a/app/pulp/app/serializers/task.py
+++ b/app/pulp/app/serializers/task.py
@@ -1,0 +1,72 @@
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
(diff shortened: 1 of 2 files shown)


app/pulp/app/serializers/task.py
```diff
@@ -0,0 +1,75 @@
...
+
+    parent = serializers.HyperlinkedRelatedField(
+        help_text="The parent task that spawned this task.",
+        read_only=True,
+        view_name='task-detail'
+    )
+
+    tags = TaskTagSerializer(
+        many=True,
+        read_only=True
```
seandst: I suspect we'll want this to be read/write at some point, but don't think that's worth blocking the PR.

seandst: Hahahah this is for a task nevermind sorry. :D

PR #2794, opened 2016-10-19
Add consumer serializer

pcreech:
Add consumer serializer

closes #[2306](https://pulp.plan.io/issues/2306)

### DIFF
--- a/app/pulp/app/serializers/consumer.py
+++ b/app/pulp/app/serializers/consumer.py
@@ -1,0 +1,32 @@
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import NotesKeyValueRelatedField, ModelSerializer
+
+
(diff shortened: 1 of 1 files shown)


app/pulp/app/serializers/consumer.py
```diff
@@ -0,0 +1,27 @@
...
+        required=False
+    )
+
+    notes = NotesKeyValueRelatedField()
+
+    publishers = serializers.HyperlinkedRelatedField(
+        many=True,
+        read_only=True,
+        view_name='publishers-detail'
+    )
```
seandst: If we're going to be defining this a lot, it might make sense to create a PublisherRelatedField in serializers.fields, but I'm happy for that to be an optimization that happens later.

pcreech: So far the only relation to publishers that exists is on the consumer model

PR #2301, opened 2016-01-07
Adding python-semantic_version to the external deps list.

ipanova:
https://pulp.plan.io/issues/1408

PR #2798, opened 2016-10-21
Add catalog serializer

BrnoPCmaniak:
closes #2301
https://pulp.plan.io/issues/2301

### DIFF
--- a/app/pulp/app/serializers/catalog.py
+++ b/app/pulp/app/serializers/catalog.py
@@ -1,0 +1,29 @@
+from gettext import gettext as _
+
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer, DetailRelatedField
+
+
+class DownloadCatalogSerializer(ModelSerializer):
+    url = serializers.CharField(
+        help_text=_("The URL used to download the related artifact."),
+        allow_blank=True, read_only=True,
+    )
+
+    artifact = serializers.HyperlinkedRelatedField(
+        help_text=_("The artifact that is expected to be present at url"),
+        queryset=models.Artifact.objects.all(),
+        view_name="artifact-details"
+    )
+
+    importer = DetailRelatedField(
+        help_text=_("The importer that contains the configuration necessary to access url."),
+        queryset=models.Importer.objects.all(),
+        view_name="importer-details"
+    )
+
+    class Meta:
+        model = models.DownloadCatalog
+        fields = ModelSerializer.Meta.fields + ("artifact", "importer", "url",)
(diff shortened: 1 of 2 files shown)


app/pulp/app/serializers/catalog.py
```diff
@@ -0,0 +1,27 @@
...
+        help_text="The URL used to download the related artifact.",
+        allow_blank=True, required=False, validators=[validators.URLValidator],
+    )
+
+    artifact = serializers.PrimaryKeyRelatedField(
+        help_text="The artifact that is expected to be present at url",
+        queryset=models.Artifact.objects.all()
+    )
+
+    importer = serializers.PrimaryKeyRelatedField(
```
seandst: Never use PrimaryKeyReleatedField. This should (probably) be a DetailRelatedField referencing the Importer model.

app/pulp/app/serializers/catalog.py
```diff
@@ -0,0 +1,27 @@
...
+from pulp.app.serializers import ModelSerializer
+
+
+class DownloadCatalogSerializer(ModelSerializer):
+    url = serializers.CharField(
+        help_text="The URL used to download the related artifact.",
+        allow_blank=True, required=False, validators=[validators.URLValidator],
+    )
+
+    artifact = serializers.PrimaryKeyRelatedField(
```
seandst: Never use PrimaryKeyRelatedField. This should be a HyperlinkedRelatedField referencing the Artifact model.

app/pulp/app/serializers/catalog.py
```diff
@@ -0,0 +1,27 @@
...
+from django.core import validators
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
+class DownloadCatalogSerializer(ModelSerializer):
+    url = serializers.CharField(
+        help_text="The URL used to download the related artifact.",
+        allow_blank=True, required=False, validators=[validators.URLValidator],
```
seandst: Is this something we want API users to be able to create? I would have expected no validators and `read_only=True`, but I don't remember how these get created so this might be an unhelpful note.

app/pulp/app/serializers/catalog.py
BrnoPCmaniak: Fixed

app/pulp/app/serializers/catalog.py
BrnoPCmaniak: Fixed

app/pulp/app/serializers/catalog.py
BrnoPCmaniak: I will investigate this one. AFAIK in pulp 2 there was no API for this.

mhrivnak: Agreed; we don't need an API to create these.

app/pulp/app/serializers/catalog.py
```diff
@@ -0,0 +1,29 @@
...
+
+from django.core import validators
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer, DetailRelatedField
+
+
+class DownloadCatalogSerializer(ModelSerializer):
+    url = serializers.CharField(
+        help_text="The URL used to download the related artifact.",
```
ipanova: please internationalize strings here and in other occurrences too.

app/pulp/app/serializers/catalog.py
seandst: So, to make it clear what change is requested here: This field should be read-only, and does not need validation.

BrnoPCmaniak:
@seandst All fixed.

PR #2800, opened 2016-10-24
Allow nested routing for objects related to repos

asmacdo:
This work allows objects that are related to repositories to have urls
that are nested within their related repository. For example, an
`rpm` importer named `zoo-importer` related to the `zoo` repository
will live in `repositories/zoo/importers/rpm/zoo-importer/`

### DIFF
--- a/app/pulp/app/serializers/repository.py
+++ b/app/pulp/app/serializers/repository.py
@@ -1,7 +1,8 @@
 from rest_framework import serializers
 
 from pulp.app import models
-from pulp.app.serializers import NotesKeyValueRelatedField, ModelSerializer
+from pulp.app.serializers import (ModelSerializer, NotesKeyValueRelatedField,
+                                  MasterModelSerializer)
 
 
 class RepositorySerializer(ModelSerializer):
@@ -37,3 +38,19 @@ class RepositorySerializer(ModelSerializer):
         fields = ModelSerializer.Meta.fields + ('name', 'description', 'notes',
                                                 'last_content_added', 'last_content_removed')
 
+
+class ImporterSerializer(MasterModelSerializer):
+    """
+    Every importer defined by a plugin should have an Importer serializer that inherits from this
+    class. Please import from `pulp.app.serializers` rather than from this module directly.
+
+    Every subclass must override the `_href` field with a `RepositoryNestedIdentityField` that
+    defines the view_name.
+    """
+    name = serializers.CharField(
+        help_text='A name for this importer, unique within the associated repository.'
+    )
+
+    class Meta:
+        abstract = True
+        fields = MasterModelSerializer.Meta.fields + ('name',)
--- a/app/pulp/app/viewsets/base.py
+++ b/app/pulp/app/viewsets/base.py
@@ -11,6 +11,7 @@ class NamedModelViewSet(viewsets.ModelViewSet):
     "Normal" Django Models and Master/Detail models are supported by the ``register_with`` method.
     """
     endpoint_name = None
+    nested_parent = None
 
     @classmethod
     def is_master_viewset(cls):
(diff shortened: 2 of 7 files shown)


seandst:
I think we'll never merge anything that's got those "Example*" models in there. You might be able to move some things around and incidentally get https://pulp.plan.io/issues/2369 completed.

app/pulp/app/models/repository.py
```diff
@@ -232,6 +232,18 @@ class Meta(ContentAdaptor.Meta):
         default_related_name = 'importers'
 
 
+class ExampleDetailImporter(Importer):
+    # I didn't see this variable being used?
```
seandst: Which variable?

You should be setting `TYPE`, not `type`.

app/pulp/app/serializers/repository.py
```diff
@@ -37,3 +38,34 @@ class Meta:
...
+        abstract = True
+        fields = MasterModelSerializer.Meta.fields + ('name',)
+
+
+class ExampleDetailImporterSerializer(ImporterSerializer):
+    """
+    Example of a Detail serializer that also utilizes nested routes.
+    """
+    # Content adapters need to use two kwargs because the name is only unique within a
+    # repository and alone, `Importer.name` is not enough to uniquely identify it.
```
seandst: This comment doesn't make sense to me in this context. It talks about needing to use two kwargs, but only one kwarg is used on the next line. Probably reword this to be more like "Serializers for ContentAdapter models have a special identity that is only valid in the context of the repositories that they're related to, which RepositoryNestedIdentityField knows how to handle", or something.

seandst: Also, and again, there's no way any of these "Example*" classes get merge into the platform app code.

app/pulp/app/viewsets/base.py
```diff
@@ -87,11 +106,9 @@ def register_with(cls, router):
                        'correctly subclass the Master ViewSet, and do both have endpoint_name '
                        'set to different values?').format(cls.__name__)
                 warnings.warn(msg, RuntimeWarning)
```
seandst: You actually got bit by this warning while working on this. Should it be a RuntimeError instead? I'm thinking probably it should.

app/pulp/app/viewsets/repository.py
```diff
@@ -41,3 +44,34 @@ def content(self, request, name):
...
+    @decorators.detail_route()
+    def importers(self, request, name):
+        # Creates a repositories/<repo>/importers/ endpoint.
+        # This will link to all importers that are associated within the repository.
+        repo = self.get_object()
+        importers = Importer.objects.filter(repository__name=repo.name)
+        paginator = UUIDPagination()
+        page = paginator.paginate_queryset(importers, request)
+        serializer = ImporterSerializer(page, many=True, context={'request': request})
+        return paginator.get_paginated_response(serializer.data)
```
seandst: This is so similar to the content detail route that it would be useful to have a sort of factory function to generalize the combination of a queryset, paginator class, and serializer class into a paginated response.

app/pulp/app/viewsets/repository.py
```diff
@@ -41,3 +44,34 @@ def content(self, request, name):
...
+    # Indicates that importer urls should be routed through their associated repository.
+    nested_parent = RepositoryViewSet
+    # Name the parameter to generate the link. This is important because `name` in this url refers
+    # to `Importer.name`, not `Repository.name`
+    nested_parent_lookup_name = 'repo_name'
+    endpoint_name = 'importers'
+    serializer_class = ImporterSerializer
+
+
+class ExampleDetailImporterViewSet(ImporterViewSet):
```
seandst: 'nother one that'll never get merged to the platform app.

asmacdo:
Yeah, the example model/serializer/viewset aren't intended to be merged either, but I felt that they were necessary to show how the other changes will be used.

Good idea about 2369, though I assumed that you probably had a more complete working example than what I have here. If not, lets chat about how that might work.

seandst:
I mainly focused on the loading code, which proved to be a minefield frought with peril. The branch I linked should hopefully be pretty easy to merge what what you've created here by moving the example stuff over into that framework.

I didn't comment my branch much at all, just committed it as soon as it didn't explode and moved on.

app/pulp/app/viewsets/base.py
asmacdo: Yeah that makes sense. I don't see a benefit to letting Pulp be successful with missing endpoints. We want to make sure a problem like that fails obviously.

asmacdo:
@seandst, the non-production code has been removed, ready for review.

seandst:
LGTM, but travis is indicating a problem with the viewset registration mechanism.

asmacdo:
@seandst, travis fix was a 1 line change, I had to add a queryset to the Importer ViewSet.

PR #2801, opened 2016-10-24
Removes extraneous permission from pulp-celery SELinux policy

dkliban:
re: #2277
https://pulp.plan.io/issues/2277


bmbouter:
I don't think that `re:` syntax will work in the commit message. Can you update it with sometime from here:  https://pulp.plan.io/issues/103

PR #2802, opened 2016-10-25
Test app

asmacdo:
Add a test app. This can also function as an example plugin that is absolutely minimal.

https://pulp.plan.io/issues/2369

~~@goosemania, I just got this working, but I haven't fully checked everything yet. This should be good to get you started with Publishers, but be aware, there might be some git-foo that happens here tomorrow to clean this up.~~

### DIFF
--- a/app/pulp/app/tests/testapp/apps.py
+++ b/app/pulp/app/tests/testapp/apps.py
@@ -1,0 +1,11 @@
+"""
+This is where we connect our plugin to Pulp. Having `PulpPluginAppConfig` as a parent class
+identifies this app as a Pulp plugin. We specify the location our app so it may be
+installed.
+"""
+from pulp.app.apps import PulpPluginAppConfig
+
+
+class TestAppConfig(PulpPluginAppConfig):
+    name = 'pulp.app.tests.testapp'
+    label = 'testapp'
--- a/app/pulp/app/tests/testapp/models.py
+++ b/app/pulp/app/tests/testapp/models.py
@@ -1,0 +1,36 @@
+"""
+A plugin is required to implement at least these three models. They should inherrit from these base
+classes.
+"""
+from django.db import models
+
+from pulp.app.models import Content, Importer, Publisher
+
+
+class TestContent(Content):
+    TYPE = 'test'
+
+    name = models.TextField()
+
+    class Meta:
+        pass
+
+
+class TestImporter(Importer):
+    TYPE = 'test'
+
+    test_field = models.TextField()
+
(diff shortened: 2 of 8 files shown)


asmacdo:
git-foo complete.  I wanted to preserve @seandst 's commit, so this will be 2 commits. The nested routing commit will be merged with https://github.com/pulp/pulp/pull/2800.

app/pulp/app/serializers/fields.py
```diff
@@ -10,3 +11,36 @@ class RepositoryRelatedField(serializers.HyperlinkedRelatedField):
     view_name = 'repositories-detail'
     lookup_field = 'name'
     queryset = models.Repository.objects.all()
+
+
+class RepositoryNestedIdentityField(serializers.HyperlinkedIdentityField):
+    """
+    Used to represent an href identity field includes the associated repository name in the url.
```
goosemania: Maybe s/includes/which includes/ ?

app/pulp/app/tests/testapp/models.py
```diff
@@ -0,0 +1,36 @@
...
+from pulp.app.models import Content, Importer, Publisher
+
+
+class TestContent(Content):
+    TYPE = 'test'
+
+    name = models.TextField()
+
+    class Meta:
+        pass
```
goosemania: Isn't it a requirement for plugin writers to specify `unique_together` for the content?

app/pulp/app/serializers/fields.py
asmacdo: This was a change introduced in #2800 so when that merged, this commit is no longer part of this PR, I'll clean that up in another PR.

app/pulp/app/tests/testapp/apps.py
```diff
@@ -0,0 +1,10 @@
+"""
+This is where we connect our plugin to Pulp. Having `PulpPluginAppConfig` as a parent class
+identifies this app as a Pulp plugin. We specify the location our app so it may be
+installed.
+"""
+from pulp.app.apps import PulpPluginAppConfig
+
+
+class TestAppConfig(PulpPluginAppConfig):
+    name = 'pulp.app.tests.testapp'
```
seandst: This would normally includ the app's "label", which in this case is just "testapp". For a normal plugin, where the app would be pulp_rpm.app, for example, this would just be "pulp_rpm". tl;dr add a `label = 'testapp'`. I'm not sure if I called this out in the docs, but it's explained in PulpAppConfig's comments.

PR #2803, opened 2016-10-25
Fixes to the process recycling feature

bmbouter:
- updates the configs for upstart and systemd
- updates the systemd unit manager
- updates the upstart script
- adds a release note
- adds docs on the feature
- adds a troubleshooting note

I hand tested these against systemd and upstart so I
expect them to work.

https://pulp.plan.io/issues/2172
closes 2172


bmbouter:
The 1 test failure left is not related. I'm merging.

PR #2804, opened 2016-10-25
datetime.total_time() not supported by python 2.6

werwty:


server/pulp/server/db/manage.py
```diff
@@ -198,8 +198,8 @@ def main():
 
         # Prompt the user if there are workers that have not timed out
         if filter(lambda worker: (UTCDateTimeField().to_python(datetime.now()) -
-                                  worker['last_heartbeat']).total_seconds() <
-                  constants.CELERY_TIMEOUT_SECONDS, status.get_workers()):
+                                  worker['last_heartbeat']) <
+                  timedelta(seconds=constants.CELERY_TIMEOUT_SECONDS), status.get_workers()):
```
seandst: It's not clear to me that the left side of the `<` operand is a datetime object except by seeing that we were calling `.total_seconds()` on it before. I like this change as a fix to make things happy on el6 again, but would love to see this code get a readability pass at some point.

server/pulp/server/db/manage.py
```diff
@@ -294,6 +294,4 @@ def _start_logging():
 
 def _user_input_continue(question):
     reply = str(raw_input(_(question + ' (y/N): '))).lower().strip()
-    if reply[0] == 'y':
-        return True
-    return False
+    return reply[0] == 'y'
```
seandst: Nice.

PR #2806, opened 2016-10-26
Add remaining fields to Importer Serializer

asmacdo:
Part of https://pulp.plan.io/issues/2298, but does not close the issue.

I chose to do just this one because I want someone else ( @goosemania ) to write the Publisher Serializer to ~make sure that using the nested routing is clear.`~

### DIFF
--- a/app/pulp/app/serializers/repository.py
+++ b/app/pulp/app/serializers/repository.py
@@ -72,15 +72,94 @@ class ImporterSerializer(MasterModelSerializer):
     Every importer defined by a plugin should have an Importer serializer that inherits from this
     class. Please import from `pulp.app.serializers` rather than from this module directly.
     """
+    # _href is normally provided by the base class, but Importers's
+    # "name" lookup field means _href must be explicitly declared.
+    _href = serializers.HyperlinkedIdentityField(
+        view_name='importers-detail',
+        lookup_field='name',
+    )
+
     name = serializers.CharField(
         help_text=_('A name for this importer, unique within the associated repository.')
+    )
+    last_updated = serializers.DateTimeField(
+        help_text='Timestamp of the most recent update of this configuration.',
+        read_only=True
+    )
+
+    feed_url = serializers.CharField(
+        help_text='The URL of an external content source.',
+        required=False,
+    )
+
+    validate = serializers.BooleanField(
+        help_text='Whether to validate imported content.',
+        required=False,
+    )
+
+    ssl_ca_certificate = serializers.CharField(
+        help_text='A PEM encoded CA certificate used to validate the server '
+                  'certificate presented by the external source.',
+        write_only=True,
+        required=False,
+    )
+    ssl_client_certificate = serializers.CharField(
+        help_text='A PEM encoded client certificate used for authentication.',
+        write_only=True,
+        required=False,
+    )
+    ssl_client_key = serializers.CharField(
+        help_text='A PEM encoded private key used for authentication.',
+        write_only=True,
+        required=False,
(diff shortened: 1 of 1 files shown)


app/pulp/app/serializers/repository.py
```diff
@@ -46,11 +46,72 @@ class ImporterSerializer(MasterModelSerializer):
 
     Every subclass must override the `_href` field with a `RepositoryNestedIdentityField` that
     defines the view_name.
+
+    `repository` is defined on the `Importer` model, but is excluded here because the user must
+    already know about a `Repository` and its relationship to the Importer because the REST
+    endpoint includes `Repository.name`.
```
goosemania: The part
> user must already know about a `Repository`...

is not very clear for me.
What about something like "excluded because related repository is specified in the URL (or API endpoint )"? And it  implies that user just can't even make a request without knowing it, and at the same time this info is available to the viewsets to be used.
I do not feel strong about it though, maybe it is just me :)

seandst: I'm not sure that this belongs in the docstring at all, really. I think a comment where the repository field would normally go (probably right down around _href) is enough to let devs know that it was omitted on purpose, but it's not a necessary part of the ImporterSerializer interface as declared here so much as a non-obvious implementation detail. Similar to @goosemania, though, maybe it's just me.

Something like `# The repository related field is excluded from this serializer because all of its API views exist in the context of its related repository` might be fine.

app/pulp/app/serializers/repository.py
```diff
@@ -46,11 +46,72 @@ class ImporterSerializer(MasterModelSerializer):
...
+    basic_auth_password = serializers.CharField(
+        help_text='The password to be used in HTTP basic authentication when syncing.'
+    )
+    max_download_bandwidth = serializers.IntegerField(
+        help_text='The max amount of bandwidth used per download (Bps).'
+    )
+    max_concurrent_downloads = serializers.IntegerField(
+        help_text='The number of concurrent downloads permitted.'
+    )
+    download_policy = serializers.MultipleChoiceField(
```
goosemania: There is [no default value in the model](https://github.com/pulp/pulp/blob/3.0-dev/app/pulp/app/models/repository.py#L226) for that field.

app/pulp/app/serializers/repository.py
```diff
@@ -63,14 +63,88 @@ class ImporterSerializer(MasterModelSerializer):
...
+        help_text='Timestamp of the most recent update of this configuration.',
+        read_only=True
+    )
+
+    feed_url = serializers.CharField(
+        help_text='The URL of an external content source.',
+        required=False,
+    )
+
+    validate = serializers.BooleanField(help_text='Whether to validate imported content.')
```
goosemania: Consider adding required=False. This field has default value in the model but that does not implies required=False, I've tested it. It is only the case when default value is specified in serializer.

app/pulp/app/serializers/repository.py
```diff
@@ -63,14 +63,88 @@ class ImporterSerializer(MasterModelSerializer):
...
+        help_text='A PEM encoded client certificate used for authentication.',
+        write_only=True,
+        required=False,
+    )
+    ssl_client_key = serializers.CharField(
+        help_text='A PEM encoded private key used for authentication.',
+        write_only=True,
+        required=False,
+    )
+    ssl_validation = serializers.BooleanField(
```
goosemania: Consider adding required=False. This field has default value in the model but that does not implies required=False.

PR #2807, opened 2016-10-27
Replaces PulpCodedException with PulpException

dkliban:
This commit also standardizes the format of the result field for failed Tasks.

closes #2271
https://pulp.plan.io/issues/2271


exceptions/pulp/exceptions/__init__.py
```diff
@@ -1,2 +1,2 @@
-from .base import PulpCodedException, PulpException
+from .base import PulpException
 from .http import MissingResource
```
bmbouter: Can a newline be added below here.

exceptions/pulp/exceptions/base.py
```diff
@@ -11,19 +11,20 @@ class PulpException(Exception):
...
-
-        # child exceptions are those that are wrapped within this exception, validation errors
-        # for example would have one overall validation error and then a separate sub error
-        # for each validation that failed
-
-        self.child_exceptions = []
-
-    def add_child_exception(self, exception):
-        self.child_exceptions.append(exception)
+    def __init__(self, error_code=None, **kwargs):
```
bmbouter: I imagined all `PulpException`'s are required to have codes. If someone subclasses this they should be required to specify a code. What do you think?

exceptions/pulp/exceptions/base.py
```diff
@@ -11,19 +11,20 @@ class PulpException(Exception):
...
+        super(PulpException, self).__init__()
```
bmbouter: I don't think we need to call super() here.

app/pulp/app/models/task.py
```diff
@@ -204,8 +204,9 @@ def set_failed(self, exc, einfo):
         self.finished_at = timezone.now()
         if isinstance(exc, PulpException):
             self.result = exc.to_dict()
+            self.result['traceback'] = einfo.traceback
```
bmbouter: What do you think about moving the to_dict() code here?

exceptions/pulp/exceptions/http.py
```diff
@@ -1,35 +1,25 @@
...
+import http.client
 
-from .base import PulpExecutionException
+from .base import PulpException
 
 
-class MissingResource(PulpExecutionException):
+class MissingResource(PulpException):
     """"
     Base class for exceptions raised due to requesting a resource that does not
```
bmbouter: Similar one line request here.

exceptions/pulp/exceptions/http.py
```diff
@@ -1,35 +1,25 @@
...
-        super(MissingResource, self).__init__(resources)
-        self.error_code = error_codes.PLP0009
+        super(MissingResource, self).__init__("PLP0009")
         self.resources = resources
-        self.error_data = {'resources': resources}
 
     def __str__(self):
         resources_str = ', '.join('%s=%s' % (k, v) for k, v in self.resources.items())
-        msg = self.error_code.message % {'resources': resources_str}
+        msg = _("The following resources are missing: %s") % (resources_str)
```
bmbouter: The last parentheses are not needed. Good i8ln. :+1:

exceptions/pulp/exceptions/http.py
```diff
@@ -1,35 +1,25 @@
...
     Base class for exceptions raised due to requesting a resource that does not
     exist.
     """
     http_status_code = http.client.NOT_FOUND
 
-    def __init__(self, *args, **resources):
+    def __init__(self, **resources):
         """
-        @param args: backward compatibility for for positional resource_id argument
         @param resources: keyword arguments of resource_type=resource_id
```
bmbouter: s/@/:/
Also can a `:type resources:` be added

bmbouter:
Can this also be looked at to see if its correct?

https://github.com/pulp/pulp/blob/aba8ba20f32c54c20a08d48c734bbc85c8bec8d4/plugin/pulp/plugin/tasking.py#L33-L39

app/pulp/app/models/task.py
```diff
@@ -202,10 +202,8 @@ def set_failed(self, exc, einfo):
         """
         self.state = Task.FAILED
         self.finished_at = timezone.now()
-        if isinstance(exc, PulpException):
-            self.result = exc.to_dict()
-        else:
-            self.result = {'traceback': einfo.traceback}
+        result = exception_to_dict(exc, einfo.traceback)
```
bmbouter: I think this could assign to `self.result` and the next line could be removed.

app/pulp/app/models/task.py
```diff
@@ -146,7 +146,7 @@ class Task(Model):
     finished_at = models.DateTimeField(null=True)
 
     non_fatal_errors = JSONField()
-    result = JSONField()
+    result = JSONField(default="[]")
```
bmbouter: Should this be `default=[]`? Testing this would be good.

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
-        self.child_exceptions.append(exception)
-
-    def to_dict(self):
+    def __init__(self, error_code):
         """
-        The to_dict method is used to provide a standardized dictionary
-        of the exception information for usage storing to the database
-        or converting to json to send back via an API call
+        :param error_code: unique error code
+        :type error_code: str
```
bmbouter: s/str/basestring/

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
-
-
-class PulpCodedException(PulpException):
-    """
-    Base class for exceptions that put the error_code and data as init arguments
-    """
-    def __init__(self, error_code=error_codes.PLP0001, **kwargs):
-        super(PulpCodedException, self).__init__()
+        if not isinstance(error_code, basestring):
+            raise RuntimeError("Error code must be an instance of basestring.")
```
bmbouter: Can this be i8ln?

bmbouter: FYI, this is close to something Python3 provides.  https://docs.python.org/3/library/typing.html

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
     def __str__(self):
-        msg = self.error_code.message % self.error_data
-        return msg.encode('utf-8')
+        """
+        Returns the string representation of the exception.
+
+        Each concrete class that inherits from :class: `pulp.server.exception.PulpException` is
+        expected to implement it's own __str__() method.
+        """
+        pass
```
bmbouter: This should probably be `NotImplementedError` so that subclasses are required to implement it.

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
+def exception_to_dict(exc, traceback):
     """
-    Base class of exceptions raised during the execution of Pulp.
+    Return a dictionary representation of an Exception.
 
-    This class should be used as a graceful server-side error while running
-    an operation. It is acceptable to instantiate and use this class directly.
+    :param exc: Exception that is being serialized
+    :type exc: Exception
+    :param traceback:
```
bmbouter: needs to be completed.

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
+    :type traceback:
```
bmbouter: needs to be completed.

tasking/pulp/tasking/base.py
```diff
@@ -350,7 +350,7 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
         :param einfo:   celery's ExceptionInfo instance, containing serialized traceback.
         :type einfo:    ???
         """
-        if isinstance(exc, PulpCodedException):
+        if isinstance(exc, PulpException):
```
bmbouter: Do we want this if statment at all? I suspect celery will log the exception correctly in all situations right?

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
 
-    Subclasses to this exception can be used to further describe any problems
-    encountered by the server.
+    :return: dictionary representing the Exception
+    :rtype: dict
     """
-    # NOTE intermediate exception class, no overrides will be provided
-    pass
+    result = {'code': None, 'description': str(exc), 'traceback': traceback}
```
bmbouter: Does `traceback` need to be converted to a string format here? The traceback module provides methods to convert a traceback to a string repr version. I'm not which one to use. Do we need to do something in this area?

plugin/pulp/plugin/tasking.py
```diff
@@ -25,15 +26,13 @@ def append_non_fatal_error(cls, error):
...
         :raises :class: `pulp.app.models.Task.DoesNotExist`: If not currently running inside a
                                                              task.
         """
         task_id = get_current_task_id()
         task_obj = Task.objects.get(id=task_id)
-        if not isinstance(task_obj.non_fatal_errors, list):
-            task_obj.non_fatal_errors = [error]
-        else:
-            task_obj.non_fata_errors.append(error)
+        serialized_error = exception_to_dict(error, None)
```
dkliban: Since we are not planning to store tracebacks for non fatal errors, that is the way I will implement it. thanks for the suggestion.

exceptions/pulp/exceptions/base.py
dkliban: It now raises the exception.

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,44 @@
...
-                raise PulpCodedException(error_codes.PLP0008, code=self.error_code.code,
-                                         field=key)
 
     def __str__(self):
-        msg = self.error_code.message % self.error_data
-        return msg.encode('utf-8')
+        """
+        Returns the string representation of the exception.
+
+        Each concrete class that inherits from :class: `pulp.server.exception.PulpException` is
```
dkliban: i am adding a sentence about this

exceptions/pulp/exceptions/base.py
dkliban: i changed this back to str because there is no basestring in python 3

app/pulp/app/models/task.py
dkliban: done

app/pulp/app/models/task.py
dkliban: I am switching this to an empty list object, but I have not tested it yet.

exceptions/pulp/exceptions/base.py
dkliban: Done!

exceptions/pulp/exceptions/base.py
dkliban: Done!

tasking/pulp/tasking/base.py
dkliban: I agree that celery should be logging the exception before we even get into this method.

exceptions/pulp/exceptions/base.py
dkliban: The traceback is already a string at this point. We get the value of it from this[0] attribute.
[0] http://docs.celeryproject.org/en/latest/internals/reference/celery.datastructures.html#celery.datastructures.ExceptionInfo.traceback

bmbouter:
@dkliban This has flake 8 errors that need to be fixed. Also all of these commits should be squashed.

exceptions/pulp/exceptions/base.py
bmbouter: This makes sense thanks.

exceptions/pulp/exceptions/base.py
```diff
@@ -1,93 +1,47 @@
...
+        expected to implement it's own __str__() method. The return value is used by Pulp when
+        recording the exception in the database.
+        """
+        raise NotImplementedError
```
bmbouter: I think you want this to be an instance right?  `raise NotImplementedError()` or `raise NotImplementedError('subclasses of PulpException must implement a __str__() method')`

plugin/pulp/plugin/__init__.py
```diff
@@ -1,4 +1,5 @@
 from pulp.app.models import ProgressBar, ProgressSpinner, Repository, RepositoryContent
+from pulp.exceptions import PulpException, MissingResource
```
bmbouter: I don't think the plugin api needs MissingResource. I had imagined that all url parameter validation for the view layer will happen in the platform views. Also we need to have an issue for how to handle missing resource, but that is a separate matter.

exceptions/pulp/exceptions/base.py
dkliban: It does not have to be an instance actually, but supplying a message is a good idea. I am adding the message.

plugin/pulp/plugin/__init__.py
dkliban: I did not add the MissingResource import here. I just added the PulpException and a newline.

PR #2809, opened 2016-11-01
Set celery logger to pulp logger

BrnoPCmaniak:
closes #2275
https://pulp.plan.io/issues/2275

Have to be yet tested in some real life scenario.

### DIFF
--- a/app/etc/pulp/server.yaml
+++ b/app/etc/pulp/server.yaml
@@ -37,10 +37,8 @@
 #
 # logging:
 #   loggers:
-#     django:
-#       level: INFO
-#     pulp.app:
-#       level: INFO
+#     '':
+#       'handlers': ["stream"]
 
 # Broker configuration
 #
--- a/app/pulp/app/settings.py
+++ b/app/pulp/app/settings.py
@@ -170,13 +170,9 @@ _DEFAULT_PULP_SETTINGS = {
             },
         },
         'loggers': {
-            'pulp.app': {
-                'handlers': ['stream'],
-                'level': 'INFO',
-            },
-            'django': {
-                'handlers': ['stream'],
-                'level': 'INFO',
+            '': {
+                # The root logger
+                'handlers': ["stream"],
             },
         }
     },


BrnoPCmaniak:
Tested on celery django example and working.
[pulp_logger_test.tar.gz](https://github.com/pulp/pulp/files/572452/pulp_logger_test.tar.gz)

app/pulp/app/settings.py
```diff
@@ -165,18 +165,22 @@
             'simple': {'format': 'pulp: %(name)s:%(levelname)s: %(message)s'},
         },
         'handlers': {
-            'stream': {
+            'pulp': {
```
bmbouter: Can this be named the 'stream' handler since it is actually `logging.StreamHandler`?

app/pulp/app/settings.py
```diff
@@ -165,18 +165,22 @@
...
             },
         },
         'loggers': {
-            'pulp.app': {
-                'handlers': ['stream'],
+            '': {
+                # The root logger
+                'handlers': ["pulp"],
+            },
+            'pulp': {
```
bmbouter: Does this still need to be `pulp.app` still?

app/pulp/app/settings.py
```diff
@@ -165,18 +165,22 @@
...
-            'stream': {
+            'pulp': {
                 'class': 'logging.StreamHandler',
                 'formatter': 'simple',
             },
         },
         'loggers': {
-            'pulp.app': {
-                'handlers': ['stream'],
+            '': {
```
bmbouter: What is the effect of having this be the root logger. Is that the default logger used anytime someone calls: `_logger = logging.getLogger(__name__)`?

app/pulp/app/settings.py
bmbouter: Should we just do away with the `pulp` logger altogether? It only adds the `level=INFO` option which I think is not what we want anyway. Having a single root logger with a default `level` would probably be good.

app/pulp/app/settings.py
```diff
@@ -165,18 +165,22 @@
...
+                # The root logger
+                'handlers': ["pulp"],
+            },
+            'pulp': {
+                'handlers': ["pulp"],
                 'level': 'INFO',
             },
             'django': {
-                'handlers': ['stream'],
+                'handlers': ['pulp'],
```
bmbouter: Similar question here, do we even want a `django` logger that restricts the `level` or should we just use a root logger?

bmbouter:
I can see how setting the default logger is the right thing to do, but I don't see how this could work.

The loggers are called during django's `setup()` call. Do the celery processes call that Django `setup()` call?

Also celery installs its own logging handler so the order of logging installation matters. Assuming the answer to the above question is yes, are we sure that the `setup()` call occurs after the celery logger is installed? For example [on the 2.y line](https://github.com/pulp/pulp/blob/6837b902f9babdab26c3dae3ab49f975b591ae63/server/pulp/server/logs.py#L34) we needed to wait until celery signals us to install our own logger over the one they just installed.

bmbouter:
@seandst I'm a little late to the logging game, but I know you've been working with this. Do these comments make any sense to you?

Also can you confirm that we are not using our custom logger anymore? I thought I remember us deciding that but a confirmation would be great.

app/pulp/app/settings.py
BrnoPCmaniak: Well the idea behind renaming it was that it uses `pulp: %(name)s:%(levelname)s: %(message)s` formatter and anybody can from the config just change the logging class and use whatever they want without need to create one on their own. But it was just an Idea.

BrnoPCmaniak:
@bmbouter I was unable to find any information about where or how the Celery calls the setup. But I remember when doing this I found that there were some changes about logging. But I don't recall.
But I assume from the fact that to interact with Django ORM setup have to have happened it will be called somehow somewhere on every process. 

First version of this PR was with the custom logger included, but I was told by @seandst or at least I think it was him, that we no longer support any platform on which system logger is used and therefore I should remove it.

BrnoPCmaniak:
Tested on @goosemania 's changes and find out it doesn't works. But works perfectly fine in my django-celery barebone setup... So I'll look up what's the difference.

bmbouter:
To test tasks I've been running `workon pulp` and then `./manage.py shell_plus`. Once I'm in the shell I paste in this snippet, originally provided by @mhrivnak 

```
#!/usr/bin/env python3

from pulp.app import models
from pulp.app.tasks.importer import delete as delete_importer
from pulp.app.tasks.publisher import delete as delete_publisher


models.Repository.objects.all().delete()
models.Task.objects.all().delete()


N = 1


for n in range(N):
    repo = models.Repository(name='foo-%s' % n)
    repo.save()
    importer = models.Importer(name='bar', repository=repo)
    importer.save()
    publisher = models.Publisher(name='bar', repository=repo)
    publisher.save()


for n in range(N):
    repo_name = 'foo-%s' % n 
    #delete_importer.apply_async_with_reservation(kwargs={'repo_name': repo_name, 'importer_name': 'bar'}, resource_id=repo_name, resource_type='repo')
    delete_publisher.apply_async(kwargs={'repo_name': repo_name, 'publisher_name': 'bar'})

```

bmbouter:
I was testing some logging using this diff, but I'm going to stop for now. We need to start the celery processes using systemd unit files which is what I think is blocking the logging effort currently. See [this related conversation](https://github.com/pulp/devel/pull/21#issuecomment-264502594).

```
[bmbouter@localhost pulp]$ git diff
diff --git a/app/pulp/app/tasks/publisher.py b/app/pulp/app/tasks/publisher.py
index 30c21b0..b5c2172 100644
--- a/app/pulp/app/tasks/publisher.py
+++ b/app/pulp/app/tasks/publisher.py
@@ -4,6 +4,10 @@ from pulp.app import models
 from pulp.tasking.tasks import UserFacingTask
 
 
+from logging import getLogger
+_logger = getLogger(__name__)
+
+
 def delete(repo_name, publisher_name):
     """
@@ -14,4 +18,5 @@ def delete(repo_name, publisher_name):
     :param publisher_name:  the name of a publisher
     :type  publisher_name:  str
     """
+    _logger.error('there is a log error')
     models.Publisher.objects.filter(name=publisher_name, repository__name=repo_name).delete()
```

BrnoPCmaniak:
@bmbouter I just have found where the issue probably is. The task have to be invoked by the webserver to appear in logs. If you will run it from the shell the logs should be logged to the shell.

BrnoPCmaniak:
I have add commit ca327c4 Test suit. Now when you start webserver and acess url `/log/?message=ItWorks` 

This will appear in the logs.
```
pulp: pulp.app.tasks.publisher:ERROR: ItWorks
[10/Mar/2017 19:01:07] "GET /log/?message=ItWorks HTTP/1.1" 200 7
```

app/pulp/app/tasks/publisher.py
```diff
@@ -1,8 +1,12 @@
 from celery import shared_task
 
+from logging import getLogger
+
 from pulp.app import models
 from pulp.tasking.tasks import UserFacingTask
 
+_logger = getLogger(__name__)
```
bmbouter: Can this be moved into a page in the docs called logging.rst? I would put it here https://github.com/pulp/pulp/tree/3.0-dev/docs/contributing and it would provide a basic logging example.

app/pulp/app/settings.py
```diff
@@ -170,13 +170,9 @@
...
-            'pulp.app': {
-                'handlers': ['stream'],
-                'level': 'INFO',
-            },
-            'django': {
-                'handlers': ['stream'],
-                'level': 'INFO',
+            '': {
+                # The root logger
+                'handlers': ["stream"],
```
bmbouter: These changes need to be also updated in the server.yaml file itself.  https://github.com/pulp/pulp/blob/3.0-dev/app/etc/pulp/server.yaml#L38

BrnoPCmaniak:
@bmbouter Fixed.

app/etc/pulp/server.yaml
```diff
@@ -37,10 +37,7 @@
 #
 # logging:
 #   loggers:
-#     django:
-#       level: INFO
-#     pulp.app:
-#       level: INFO
+#     '':
```
bmbouter: Doesn't this also need the `'handlers': ["stream"],` part?
Also how are you validating this yaml?

BrnoPCmaniak: @bmbouter Validated by www.yamllint.com/ codebeautify.org/yaml-validator beautifytools.com/yaml-validator.php and all says it's okay. About the ` 'handlers': ["stream"],` It was missing from it before so I also leave it out, but probably it should be there, I will add it.

BrnoPCmaniak:
@bmbouter Fixed

app/etc/pulp/server.yaml
bmbouter: When I uncomment the current section and push it through yamllint it says it's valid , but it also mutated it some:

```
--- 
logging: 
  loggers: 
    ? ""
    : 
      handlers: 
        - stream
```

BrnoPCmaniak: @bmbouter I noticed, but no idea why it does that. But yaml parses it fine.

```
Python 2.7.13 (default, Jan 12 2017, 17:59:37) 
[GCC 6.3.1 20161221 (Red Hat 6.3.1-1)] on linux2
>>> from yaml import load, dump
>>> load("""
... logging:
...     loggers:
...         "":
...             'handlers': ["stream"]
... """)
{'logging': {'loggers': {'': {'handlers': ['stream']}}}}
>>> 
```

bmbouter:
@BrnoPCmaniak I see what you mean now w/ the parser example.

Thank you so much for this PR. It's not many lines but it represents a lot of good testing and follow through from you. 👍 🌵 🍰 🌜 🌻

PR #2814, opened 2016-11-02
Renames Task to TaskController

dkliban:


jortel:
I know this has already been merged but I think this is starting down the wrong path.  This is a _procedural_ programming approach and I'd much prefer keeping pulp3 _object-oriented_.  Appending a fatal exception to a running task is a natural operation on the object representing the task itself.

For example:

```
class Task(object):
    """ A facade for the current running task. """

    def __init__(self):
        self.id = get_current_task_id()

    def append_non_fatal_error(self, error):
        task = Task.objects.get(id=self.id)
        task.non_fatal_errors.append(error)  # This should be default=[] in the model.
        task.save()
```

PR #2815, opened 2016-11-02
Adds OperationPostponedResponse object

dkliban:
This patch also removes the middleware that used to handle OperationPostponed exceptions.

closes #2273
https://pulp.plan.io/issues/2273


plugin/pulp/plugin/response.py
```diff
@@ -0,0 +1,24 @@
+from gettext import gettext as _
+
+from rest_framework.response import Response
+
```
bmbouter: PEP8 wants 2 spaces before a class definition

plugin/pulp/plugin/response.py
```diff
@@ -0,0 +1,24 @@
...
+class OperationPostponedResponse(Response):
+    """
+    An HTTP response class for returning 202 and a list of spawned tasks.
+    """
+
+    def __init__(self, task_results, task_group_id=None):
+        """
+        :param task_results: List of AsyncResult objects that are used to generate the response.
```
bmbouter: This should get types on the next line. Also consider using the :class: stuff.

server/pulp/server/webservices/middleware/postponed.py
```diff
@@ -1,57 +0,0 @@
-import logging
-
-from celery.result import AsyncResult
```
bmbouter: :+1: for removing all of this. Thanks!

bmbouter:
Does this need more, or can the WIP come off?

seandst:
I suspect this is something we'll want to return in normal API views, and not only in plugins. I think it makes sense for this to be exposed in `pulp.plugin`, but maybe it should be defined in `pulp.app.response`?

bmbouter:
I had totally missed the path this code lives at. I think it should live outside of the plugin API. Maybe somewhere where the views are kept?

I also don't think we plan to allow plugin writers to implement views with 3.0 so we probably shouldn't include it with the plugin API at this point. What do you think?

seandst:
Plugin writers have to implement view*sets*, at least, for detail models defined in that plugin, but I think you're correct when it comes to tasks (in particular, views that would return this response) only needing to exist in platform.

app/pulp/app/response.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self, task_results, task_group_id=None):
+        """
+        :param task_results: List of AsyncResult objects that are used to generate the response.
+        :type task_group_id: List of AsyncResult
+        :param task_group_id: Task group id, if any, for all the task results
+        :type task_gropu_id: str
+        """
+        if not isinstance(task_results, list):
+            raise RuntimeError(_('task_results must be an instance of list.'))
```
mhrivnak: ``TypeError`` would be an appropriate exception to use, and has the advantage of being more specific.

dkliban: Thanks! You are totally right. I am changing it here and in my other PR.

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
+from gettext import gettext as _
+
+from rest_framework.response import Response
+
+
+class OperationPostponedResponse(Response):
+    """
+    An HTTP response class for returning 202 and a list of spawned tasks.
+
+    This reponse object should be used by views that dispatch asynchronous tasks. Most common use
```
bmbouter: s/Most/The most/

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
...
+    case is for sync and publish operations. When the task_group_id is specified, the task_results
+    list should only contain 10 items. Task groups can consist of many tasks, so returning their
+    ids should occur in a paginated fashion.
+    """
+
+    def __init__(self, task_results, task_group_id=None):
+        """
+        :param task_results: List of AsyncResult objects that are used to generate the response.
```
bmbouter: Can this docstring get the :class: additions with the full path to the celery AsyncResult object?

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
...
+        :type task_group_id: List of AsyncResult
+        :param task_group_id: Task group id, if any, for all the task results
+        :type task_group_id: str
+        """
+        if not isinstance(task_results, list):
+            raise TypeError(_('task_results must be an instance of list.'))
```
bmbouter: Does it make sense to also check the type of the items in the list to be 'AsyncResult' similarly?

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
...
+from rest_framework.response import Response
+
+
+class OperationPostponedResponse(Response):
+    """
+    An HTTP response class for returning 202 and a list of spawned tasks.
+
+    This reponse object should be used by views that dispatch asynchronous tasks. Most common use
+    case is for sync and publish operations. When the task_group_id is specified, the task_results
+    list should only contain 10 items. Task groups can consist of many tasks, so returning their
```
bmbouter: How will this 10 item pagination be done? Is this something DRF provides?

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
...
+
+from rest_framework.response import Response
+
+
+class OperationPostponedResponse(Response):
+    """
+    An HTTP response class for returning 202 and a list of spawned tasks.
+
+    This reponse object should be used by views that dispatch asynchronous tasks. Most common use
+    case is for sync and publish operations. When the task_group_id is specified, the task_results
```
bmbouter: Having this one object provide both types when one is optional in most cases may be better as two separate objects. It could be called GroupOperationPostponedResponse and be a subclass or peer object of OperationPostponedResponse. What do you think?

app/pulp/app/response.py
dkliban: I was just assuming we can get some kind of pagination from django or DRF. However, as I said in previous comment, I am removing the group_id from this object.

app/pulp/app/response.py
```diff
@@ -0,0 +1,32 @@
...
+        :param task_group_id: Task group id, if any, for all the task results
+        :type task_group_id: str
+        """
+        if not isinstance(task_results, list):
+            raise TypeError(_('task_results must be an instance of list.'))
+
+        data = {'task_group_id': task_group_id}
+        tasks = []
+        for result in task_results:
+            task = {"_href": result.task_id, "task_id": result.task_id}
```
mhrivnak: Not having worked with the new REST API code, I might just be uninformed, but I'm surprised at the value for ``_href``. Is there something else going on under the hood to convert the task_id into a path? Or is a plain task_id what we actually want?

app/pulp/app/response.py
bmbouter: Let's not do this check. To check it every time at runtime to catch a developer error at developer time probably isn't worth it. The 1 check you have makes sense for the reason you identified.

app/pulp/app/response.py
bmbouter: Yes @mhrivnak is right. The body of this probably needs to be formed automatically from the URI's of the corresponding tasks themselves in order to stay DRY. This will involve getting the actual `pulp.app.models.Task` object for each AsyncResult, but I that is OK because staying DRY is important and the number of them will be one or two for all non-group tasks.

What do others think about this?

app/pulp/app/response.py
mhrivnak: Maybe it makes sense for an endpoint that creates a task group to just return a serialized task group? Perhaps a special response type isn't required at all, except to set the code to 202.

app/pulp/app/response.py
dkliban: This is just place holder code for the URL. The viewsets for Tasks have not been completed yet, so I am not able to utilize the machinery available for generating the URLs.

app/pulp/app/response.py
mhrivnak: If that's the only reason, then you may not need this check at all. The ``for`` loop will raise a ``TypeError`` if the object isn't iterable.

... unless of course an AsyncResult is iterable. Crazier things have happened. :)

app/pulp/app/response.py
bmbouter: We can comment on the task group use case on the redmine story. @dkliban can you link us to it when that is in posted?

app/pulp/app/response.py
bmbouter: I think @mhrivnak is right. Maybe check if AsyncResult supports iteration, and if it doesn't remove this part.

PR #2816, opened 2016-11-02
Add content filter class

goosemania:


app/pulp/app/viewsets/content.py
```diff
@@ -1,9 +1,21 @@
+from django_filters.rest_framework import filterset
+
 from pulp.app.models import Content
 from pulp.app.serializers import ContentSerializer
 from pulp.app.viewsets import NamedModelViewSet
 
 
+class ContentFilter(filterset.FilterSet):
+    # Plugin content filters should inherit from this class and add any plugin-specific filters
+    # if needed.
```
seandst: This should probably be in a docstring, and contain more guidance. I assume that a plugin writer would need to do a little bit more than inherit from `ContentFilter`, including extending `ContentFilter.Meta.fields` similar to how we do it in the serializers, and defining the correct model in Meta.

PR #2817, opened 2016-11-03
Adds a custom User model to pulp.app

dkliban:
closes #2243
https://pulp.plan.io/issues/2243

### DIFF
--- a/app/pulp/app/apps.py
+++ b/app/pulp/app/apps.py
@@ -72,7 +72,7 @@ class PulpAppConfig(PulpPluginAppConfig):
     # The app's importable name
     name = 'pulp.app'
 
-    # The app label to be used when creating tables, registering models,
-    # referencing this app with manage.py, etc. This cannot contain a dot,
-    # so for brevity's sake it's just "pulp", rather than e.g. "pulp_app".
-    label = 'pulp'
+    # The app label to be used when creating tables, registering models, referencing this app
+    # with manage.py, etc. This cannot contain a dot and must not conflict with the name of a
+    # package containing a Django app.
+    label = 'pulp_app'
--- a/app/pulp/app/models/auth.py
+++ b/app/pulp/app/models/auth.py
@@ -1,0 +1,79 @@
+"""
+Django models related to the Authentication system
+
+The PulpUser and PulpUserManager classes are based on Django documentation for creating custom User
+objects. More information can be found here:
+https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model
+"""
+from gettext import gettext as _
+
+from django.core import validators
+from django.contrib.auth.models import BaseUserManager, AbstractBaseUser, PermissionsMixin
+from django.db import models
+
+
+class PulpUserManager(BaseUserManager):
+    """
+    A custom manager for PulpUser objects.
+    """
+    use_in_migrations = True
+
+    def _create_user(self, username, password, is_superuser):
+        """
+        Creates and saves a PulpUser with the given username and password.
+        """
+        if not username:
+            raise ValueError('The given username must be set')
+        user = self.model(username=username, is_superuser=is_superuser)
+        user.set_password(password)
+        user.save(using=self._db)
+        return user
+
+    def create_user(self, username, password):
+        return self._create_user(username, password, False)
+
+    def create_superuser(self, username, password):
+        return self._create_user(username, password, True)
+
+
+class PulpUser(AbstractBaseUser, PermissionsMixin):
+    """
+    A custom Django User class for Pulp.
+    """
+    username = models.CharField(
+        _('username'),
+        max_length=150,
+        unique=True,
+        validators=[
+            validators.RegexValidator(
+                r'^[\w.@+-]+$',
+                _('Enter a valid username. This value may contain only letters, numbers '
+                  'and @/./+/-/_ characters.'),
+                'invalid'),
+        ],
+        error_messages={
+            'unique': _("A user with that username already exists.")
+        }
+    )
+
+    is_admin = models.BooleanField(default=False)
+
+    objects = PulpUserManager()
+
+    USERNAME_FIELD = 'username'
+
+    def get_full_name(self):
+        """
+        A longer formal identifier for the user.
+        :return: Username
+        :rtype: str
+        """
+        return self.username
+
+    def get_short_name(self):
+        """
+        Short, informal identifier for the user.
+        :return: Username
+        :rtype: str
+        """
+        return self.get_full_name()
--- a/app/pulp/app/db-reset.sh
+++ b/app/pulp/app/db-reset.sh
@@ -15,6 +15,6 @@ sleep 1
 fi
 
 python manage.py reset_db --noinput
-python manage.py makemigrations pulp --noinput
+python manage.py makemigrations pulp_app --noinput
 python manage.py migrate --noinput
 popd
--- a/app/pulp/app/models/__init__.py
+++ b/app/pulp/app/models/__init__.py
@@ -1,4 +1,5 @@
 # https://docs.djangoproject.com/en/1.10/topics/db/models/#organizing-models-in-a-package
+from .auth import PulpUser # NOQA
 from .base import Model, MasterModel  # NOQA
 from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyValueRelation,  # NOQA
                       GenericKeyValueModel, Config, Notes, Scratchpad)  # NOQA
--- a/app/pulp/app/settings.py
+++ b/app/pulp/app/settings.py
@@ -103,6 +103,8 @@ REST_FRAMEWORK = {
     'PAGE_SIZE': 100,
 }
 
+AUTH_USER_MODEL = 'pulp_app.PulpUser'
+
 # Password validation
 # https://docs.djangoproject.com/en/1.9/ref/settings/#auth-password-validators
 


app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,63 @@
+"""
+Django models related to the Authentication system
+"""
+from django.core import validators
+from django.contrib.auth.models import BaseUserManager, AbstractBaseUser, PermissionsMixin
+from django.db import models
+from django.utils import timezone
+
+
+class PulpUser(AbstractBaseUser, PermissionsMixin):
```
seandst: Can this be `User`? It usage in `settings.py` demonstrates the redundancy of including 'Pulp' in its name.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,63 @@
+"""
+Django models related to the Authentication system
```
seandst: There are things in here that don't seem like they're doing things in obvious ways, at least to me. It'd be helpful to have a link to https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model in here somewhere to make it clear that that's what's going on in here, and that future developers should probably be familiar with that page before changing things around in here.

app/pulp/app/models/auth.py
seandst: We probably want to use `pulp.models.Model` as an additional abstract base here to pick up the uuid pk.

seandst:
I accidentally made my review be "Request Changes" even though this is still WIP. I'll try to fix that, but in the meantime ignore my comments if you want since it is still WIP. :)

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _create_user(self, username, password, is_superuser):
+        """
+        Creates and saves a PulpUser with the given username and password.
+        """
+        now = timezone.now()
+        if not username:
+            raise ValueError('The given username must be set')
+        user = self.model(username=username, is_active=True, is_superuser=is_superuser,
+                          date_joined=now)
```
goosemania: It seems like you do not have `date_joined` in your model. But you would have it if you mimic the AbstractUser implementation I think.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+        return self._create_user(username, password, True)
+
+
+class PulpUser(AbstractBaseUser, PermissionsMixin):
+    """
+    A custom Django User class for Pulp.
+    """
+    username = models.CharField(_('username'), max_length=150, unique=True,
+        help_text=_('Required. 150 characters or fewer. Letters, digits and '
+                    '@/./+/-/_ only.'),
```
goosemania: In other models we did not specified the `help_text` in models but we did it later in serializers. For consistency (and maybe for the API docs which may use `help_text` from serializers if I am not mistaken) I suggest to move it to serializer. What do you think?

goosemania: And +1 for 150 characters in username,  like in Django 1.10

dkliban: I wasn't sure where we put the help text. I will move it  to the serializer.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+            'unique': _("A user with that username already exists."),
+        })
+
+    is_admin = models.BooleanField(default=False)
+
+    objects = PulpUserManager()
+
+    USERNAME_FIELD = 'username'
+
+    def get_full_name(self):
```
bmbouter: This should probably get a docstring.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+        return self.username
+
+    def get_short_name(self):
```
bmbouter: This should probably get a docstring.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+        now = timezone.now()
+        if not username:
+            raise ValueError('The given username must be set')
+        user = self.model(username=username, is_active=True, is_superuser=is_superuser,
+                          date_joined=now)
+        user.set_password(password)
+        user.save(using=self._db)
+        return user
+
+    def create_user(self, username, password=None):
```
bmbouter: I thought for the MVP all users were superusers?

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Creates and saves a PulpUser with the given username and password.
+        """
+        now = timezone.now()
+        if not username:
+            raise ValueError('The given username must be set')
+        user = self.model(username=username, is_active=True, is_superuser=is_superuser,
+                          date_joined=now)
+        user.set_password(password)
+        user.save(using=self._db)
```
bmbouter: Can't this just be `save()` or does it need the `using`?

app/pulp/app/models/auth.py
dkliban: good catch. I'll remove it.

app/pulp/app/models/auth.py
dkliban: I copied this from django's implementation of UserManager. I will experiment both ways and see what the difference is.

dkliban:
@goosemania I had the same thought about using the AbstractUser model that django provides. However, it seems like there is just too many extra fields that we will not be using. If we did use AbstractUser we would need to inherit from it  and then set REQUIRED_FIELDS=[] (since email is required by AbstractUser).

What do others think? https://github.com/django/django/blob/stable/1.8.x/django/contrib/auth/models.py#L378

goosemania:
@dkliban, the extra fields are first/last names, e-mail and date_joined, right? The most suspicious one for me is e-mail, it is required but can be blank and we probably won't use it. Other fields maybe ok, especially if you have several users and you will be able to specifiy their names. 
I do not know, it seems to me that it is easier to mimic Django implementation but maybe it is a wrong approach.

app/pulp/app/models/auth.py
dkliban: @bmbouter I played around with the code and it looks like self._db is actually None. I am not sure that there is value is specifying it.  However, I am also hesitant to remove because Django's implementation of UserManager object uses it.

app/pulp/app/apps.py
```diff
@@ -75,4 +75,4 @@ class PulpAppConfig(PulpPluginAppConfig):
     # The app label to be used when creating tables, registering models,
     # referencing this app with manage.py, etc. This cannot contain a dot,
     # so for brevity's sake it's just "pulp", rather than e.g. "pulp_app".
-    label = 'pulp'
+    label = 'pulp_app'
```
seandst: You've made a liar out of the preceding comment. It should now read something along the lines of "This cannot contain a dot and must not conflict with the name of a package containing a Django app, so it should be "pulp_app" rather than "pulp".

dkliban: Good catch. I'll fix the comment.

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        return self.username
+
+    def get_short_name(self):
+        """
+        Short, informal identifier for the user.
+        :return: Username
+        :rtype: str
+        """
+        return self.username
```
seandst: It might seem silly, but I think this could benefit from some DRYness. The Django docs say "A short, informal identifier for the user. A common interpretation would be the first name of the user, but it can be any string that identifies the user in an informal way. It may also return the same value as django.contrib.auth.models.User.get_full_name()," so this should probably do exactly that and `return self.get_full_name()`.

seandst:
Based on the Django docs, I think what's here is probably best. I don't think we gain anything by switching to the `AbstractUser` base. Django's docs recommend extending `AbstractBaseUser`[0] unless we're "entirely happy with Django’s User model"[1], which I don't think we are. Explicitly defining the things we do need, including a custom manager, and implementing Django's supported and well-documented interface for a User model feels correct to me.

[0]: https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model
[1]: https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#extending-django-s-default-user

app/pulp/app/models/auth.py
```diff
@@ -0,0 +1,74 @@
...
+    username = models.CharField(_('username'), max_length=150, unique=True,
+                                validators=[validators.RegexValidator(r'^[\w.@+-]+$',
+                                                                      _('Enter a valid username. '
+                                                                        'This value may contain '
+                                                                        'only letters, numbers '
+                                                                        'and @/./+/-/_ '
+                                                                        'characters.'), 'invalid'),
+                                            ],
+                                error_messages={'unique': _("A user with that username already "
+                                                            "exists.")})
```
seandst: flake8 is happy with it, so it definitely meets my style standards. That said, it's pretty hard to look at. What do you think about

```python
    username = models.CharField(
        _('username'),
        max_length=150,
        unique=True,
        validators=[
            validators.RegexValidator(
                r'^[\w.@+-]+$',
                _('Enter a valid username. This value may contain only letters, '
                  'numbers and @/./+/-/_ characters.'),
                'invalid'),
        ],
        error_messages={
            'unique': _("A user with that username already exists.")
        }
    )
```

This is definitely a personal preference thing, so ignore it if you even remotely disagree.

app/pulp/app/models/auth.py
dkliban: I am all about staying DRY. I'm making this change also.

app/pulp/app/models/auth.py
dkliban: I like it. Look for it in the next revision.

PR #2818, opened 2016-11-03
Replace task controller with facade object that represents the current running task.

jortel:
The `TaskController` is basically a *procedural* approach and proposing an *object-oriented* alternative.  Appending non-fatal errors is a natural operation on the currently running task.  Proposing a *facade* object that represents the currently running task that supports these operations.

I also noticed the models.Task.non_fatal_errors is expected to be a json document with its root being a list.  But, the field did not have a default=[].  So I added it.  I think this makes sense?

I feel like the docstring could use a little more but not really sure what that needs to be yet.  After the entire plugin API is complete, we should make another pass over them to ensure it's crystal clear to the plugin writer how the API is to be used.

### DIFF
--- a/app/pulp/app/models/task.py
+++ b/app/pulp/app/models/task.py
@@ -122,8 +122,8 @@ class Task(Model):
     started_at = models.DateTimeField(null=True)
     finished_at = models.DateTimeField(null=True)
 
-    non_fatal_errors = JSONField()
-    result = JSONField(default=[])
+    non_fatal_errors = JSONField(default=list)
+    result = JSONField(default=list)
 
     parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks")
     worker = models.ForeignKey("Worker", null=True, related_name="tasks")
--- a/plugin/pulp/plugin/tasking.py
+++ b/plugin/pulp/plugin/tasking.py
@@ -1,38 +1,37 @@
-from pulp.app.models import Task
+from pulp.app import models
 from pulp.exceptions import exception_to_dict
 from pulp.tasking import get_current_task_id
 
 
-class TaskController(object):
+class Task(object):
     """
-    An interface to the currently running task.
+    The task which is currently executing.
 
-    With this object you can record non-fatal exceptions as they occur.
+    Attributes:
+        id (str): The task identifier.
 
-    This object is natively task aware. When called inside of a task it will append the non-fatal
-    exception to the currently running task. When called outside of a task a
-    :class: `pulp.app.models.Task.DoesNotExist` Exception is raised.
-
-    Fatal errors should not use this. Instead they should raise an Exception, preferably one that
-    inherits from :class: `pulp.server.exception.PulpException`.
     """
 
-    @classmethod
-    def append_non_fatal_error(cls, error):
+    def __init__(self):
+        self.id = get_current_task_id()
+
+    def append_non_fatal_error(self, error):
         """
         Append and save a non-fatal error for the currently executing task.
+        Fatal errors should not use this. Instead they should raise an Exception,
+        preferably one that inherits from :class: `pulp.server.exception.PulpException`.
 
         This is saved in a structured way to the :attr: `~pulp.app.models.Task.non_fatal_errors`
         attribute on the :class: `~pulp.app.models.Task` model.
 
-        :param error: The non fatal error to be appended.
-        :type error: Exception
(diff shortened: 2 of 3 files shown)


plugin/pulp/plugin/tasking.py
```diff
@@ -1,26 +1,24 @@
...
-    @classmethod
-    def append_non_fatal_error(cls, error):
+    def __init__(self):
+        self.id = get_current_task_id()
+
+    def append_non_fatal_error(self, error):
         """
         Append and save a non-fatal error for the currently executing task.
+        Fatal errors should not use this. Instead they should raise an Exception,
+        preferably one that inherits from :class: `pulp.server.exception.PulpException`.
```
dkliban: s/pulp.server.exception.PulpException/pulp.exceptions.PulpException/

plugin/pulp/plugin/tasking.py
```diff
@@ -1,26 +1,24 @@
...
+from pulp.app import models
 from pulp.exceptions import exception_to_dict
 from pulp.tasking import get_current_task_id
 
 
-class TaskController(object):
+class Task(object):
     """
-    An interface to the currently running task.
+    The task which is currently executing.
```
dkliban: This probably needs a sentence about this being a facade. Something similar to what you say in your PR description.

app/pulp/app/models/task.py
```diff
@@ -145,7 +145,7 @@ class Task(Model):
     started_at = models.DateTimeField(null=True)
     finished_at = models.DateTimeField(null=True)
 
-    non_fatal_errors = JSONField()
+    non_fatal_errors = JSONField(default=[])
```
mhrivnak: I think this will result in a classic python bug. Exactly one list instance will be created and used as the default forever, and it'll be handed out to anyone who needs it. If something adds stuff to it, that stuff will still be there the next time the default gets used.

You can specify a callable as the default, which will be called each time a default value is required. So you can probably say ``default=list``.

bmbouter:
@mhrivnak I watched the youtube snippet and I agree with that perspective, but consider that we will almost certainly add additional methods to this over time. Granted, those could all be utility functions since the data being stored is so trivial, but for organization purposes I think have a class will make our future code more organized. Also note that we can't refactor this easily later because it's an interface governed by semver.

What do you think about all ^?

app/pulp/app/models/task.py
bmbouter: Actually Django guards against this by [requiring the value of default to be immutable](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.Field.default). So `JSONField(default=[])` won't work.

mhrivnak:
I see what you are saying, but I don't think a class should be used just to organize functions. Modules do that well.

jortel:
The Task is a real world object within pulp and the only reason it only has 1 method (currently) is because we have deliberately constrained it's interface.  This does not disqualify it from being an object.  The perspective promoted by the video views classes as a useful python construct.  I strongly disagree with this perspective.  Object oriented programming is larger than language constructs.  It's a programming paradigm where entire problem domain is modeled (as objects).  And, that functionality is achieved through collaboration between those objects.  The pulp code base is either OO or it's not.  I can go on, and, on, but I wont.

mhrivnak:
I see. I want to at least respond to that.

An OO design should not mean we have to put every function on a class. There can be balance; we should make objects whenever we have behaviors to pair with a data structure. I'm a big fan of that. But every data structure should not be forced into a class, and every collection of behaviors should not be forced into a class.

A class that exists just to hold functions, rather than being designed around a data structure, only serves as a namespace. Modules do the same thing more simply. One of these options is technically a class, but they otherwise behave exactly the same. I don't see that we gain anything by using a class.

``tasking.Task().do_something(*args)``

vs.

``tasking.do_something(*args)``

or even

``tasking.task.do_something(*args)``

There's no paradigm shift from one to the other. They are just namespaces, they behave the same, they can be passed around the same, and it just happens that one of the namespaces needs to be instantiated.

That said, we have bigger fish to fry right now, so I'm not concerned about this particular case. I think maybe we're getting hung up because this is being thought of as a facade for a legitimate class, and I suppose I'm suggesting that a facade might not be that useful. But maybe with more behavior filled in, that will start to look more natural.

In any case, let's continue the discussion somewhere, because I want the ability to write a function that's not on a class.

PR #2819, opened 2016-11-03
Add task and worker viewsets

werwty:
closes #2312

### DIFF
--- a/app/pulp/app/response.py
+++ b/app/pulp/app/response.py
@@ -1,21 +1,22 @@
 from rest_framework.response import Response
+from rest_framework.reverse import reverse
 
 
 class OperationPostponedResponse(Response):
     """
     An HTTP response class for returning 202 and a list of spawned tasks.
 
-    This reponse object should be used by views that dispatch asynchronous tasks. The most common
+    This response object should be used by views that dispatch asynchronous tasks. The most common
     use case is for sync and publish operations. When JSON is requested, the response will look
     like the following:
 
         [
             {
-                "_href": "/this/needs/to/be/updated/adlfk-bala-23k5l7-lslser",
+                "_href": "/api/v3/tasks/adlfk-bala-23k5l7-lslser",
                 "task_id": "adlfk-bala-23k5l7-lslser"
             },
             {
-                "_href": "/this/needs/to/be/updated/fr63x-dlsd-4566g-dv64m",
+                "_href": "/api/v3/tasks/fr63x-dlsd-4566g-dv64m",
                 "task_id": "fr63x-dlsd-4566g-dv64m"
             }
         ]
@@ -26,9 +27,9 @@ class OperationPostponedResponse(Response):
         Args:
             task_results (list): List of AsyncResult objects used to generate the response.
         """
-        raise NotImplementedError
         tasks = []
         for result in task_results:
-            task = {"_href": result.task_id, "task_id": result.task_id}
+            task = {"_href": reverse('tasks-detail', args=[result.task_id]),
+                    "task_id": result.task_id}
             tasks.append(task)
         super(OperationPostponedResponse, self).__init__(data=tasks, status=202)
--- a/app/pulp/app/viewsets/task.py
+++ b/app/pulp/app/viewsets/task.py
@@ -1,0 +1,41 @@
+from django_filters.rest_framework import filterset
+
+from pulp.app.models import Task, Worker
+from pulp.app.serializers import TaskSerializer, WorkerSerializer
+from pulp.app.viewsets import NamedModelViewSet
+from pulp.app.viewsets.custom_filters import CharInFilter
+from pulp.tasking.base import cancel as cancel_task
+
+from rest_framework.decorators import detail_route
+from rest_framework.response import Response
+from rest_framework import status
+
+
+class TaskFilter(filterset.FilterSet):
+    tags = CharInFilter(name='tag__name', lookup_expr='in')
+
+    class Meta:
+        model = Task
+        fields = ['state', 'worker__name', 'tags']
+
+
+class TaskViewSet(NamedModelViewSet):
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+    endpoint_name = 'tasks'
+    filter_class = TaskFilter
+    http_method_names = ['get', 'options']
+
+    @detail_route(methods=['post'])
+    def cancel(self, request, pk=None):
+        task = self.get_object()
+        cancel_task(task.pk)
+        return Response(status=status.HTTP_204_OK)
+
+
+class WorkerViewSet(NamedModelViewSet):
+    queryset = Worker.objects.all()
+    serializer_class = WorkerSerializer
+    endpoint_name = 'workers'
+    lookup_field = 'name'
+    http_method_names = ['get', 'options']
--- a/tasking/pulp/tasking/base.py
+++ b/tasking/pulp/tasking/base.py
@@ -377,7 +377,6 @@ def cancel(task_id):
     :type  task_id: basestring
 
     :raises MissingResource: if a task with given task_id does not exist
-    :raises PulpCodedException: if given task is already in a complete state
     """
     try:
         task_status = TaskStatus.objects.get(pk=task_id)
--- a/app/pulp/app/viewsets/__init__.py
+++ b/app/pulp/app/viewsets/__init__.py
@@ -1,4 +1,5 @@
 from pulp.app.viewsets.base import NamedModelViewSet  # NOQA
 from pulp.app.viewsets.content import ContentViewSet  # NOQA
 from pulp.app.viewsets.repository import ImporterViewSet, RepositoryViewSet  # NOQA
+from pulp.app.viewsets.task import TaskViewSet, WorkerViewSet  # NOQA
 from pulp.app.viewsets.user import UserViewSet  # NOQA


app/pulp/app/response.py
```diff
@@ -1,21 +1,22 @@
...
 
-    This reponse object should be used by views that dispatch asynchronous tasks. The most common
+    This response object should be used by views that dispatch asynchronous tasks. The most common
     use case is for sync and publish operations. When JSON is requested, the response will look
     like the following:
 
         [
             {
-                "_href": "/this/needs/to/be/updated/adlfk-bala-23k5l7-lslser",
+                "_href": "/tasks/adlfk-bala-23k5l7-lslser",
```
dkliban: i think this should be /api/v3/tasks/adlfk-bala-23k5l7-lslser

and the same for the other one

werwty: :+1: fixing now

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,30 @@
...
+        model = Task
+        fields = ['state', 'worker__name', 'tags']
+
+
+class TaskViewSet(NamedModelViewSet):
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+    endpoint_name = 'tasks'
+    filter_class = TaskFilter
+    http_method_names = ['get', 'delete', 'options']
```
mhrivnak: We used the DELETE method in pulp 2 to cancel a task, but I'm hoping that won't be the case in the v3 API, since it goes against the normally expected behavior of a REST endpoint. And I don't think we actually want clients to delete task records. So given that, I suspect we can remove ``delete`` from this list; does that match your understanding?

seandst: +1 to not using DELETE to mean cancel. If we still want delete behavior, we'll probably want to create a 'cancel' detail route that works with POST.

http://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

werwty: That makes a lot of sense. I'm adding [WIP] to this PR and implementing  a /cancel and a /resume detail route to update the status of a task. Are there any other detail routes we should consider?

mhrivnak: Just ``cancel`` should work. We don't actually have a ``resume`` behavior to offer.

@seandst ++ for pointing out the right way to do this.

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class TaskViewSet(NamedModelViewSet):
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+    endpoint_name = 'tasks'
+    filter_class = TaskFilter
+    http_method_names = ['get', 'options']
+
+    @detail_route(methods=['get'])
```
mhrivnak: s/get/post/

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,46 @@
...
+    serializer_class = TaskSerializer
+    endpoint_name = 'tasks'
+    filter_class = TaskFilter
+    http_method_names = ['get', 'options']
+
+    @detail_route(methods=['post'])
+    def cancel(self, request, pk=None):
+        task = self.get_object()
+        if task.state == Task.RUNNING or task.state == Task.WAITING:
+            # TODO call a controller method to cancel task
```
mhrivnak: Thinking about this a little more, I wonder if this would be better to have on the model itself, even if it's a temporary home while we think about moving it to a controller.

Most of the logic here is about managing the state of the task itself, and whether a particular change is ok. That clearly is the domain of the model.

The only part of cancellation that I think might not be a good fit for the model is a call to celery to actively carry out cancellation. Calling that from the model wouldn't be the worst thing, and would probably be better than calling it from here (and everywhere else that wants to cancel a task).

What do you think?

werwty: You're right, it would make more sense to have on the model given that having viewset know about the task states seems off. 
As far as having the model notify celery, maybe Django signals would be a good fit there?
https://docs.djangoproject.com/en/dev/topics/signals/

app/pulp/app/models/task.py
```diff
@@ -205,6 +206,22 @@ def set_failed(self, exc, einfo):
         self.result = exception_to_dict(exc, einfo.traceback)
         self.save()
 
+    def cancel(self):
+        """
+        Set the task to a canceled state if it is in an incomplete state.
+
+        This updates the :attr: `finished_at` attribute, sets the :attr: `state` to
+        :attr: `CANCELED`
+
```
mhrivnak: Documenting the exception is a good idea. Under old (but still current AFAIK) way, that looks like

```
:raises ValidationError: this is why
```

app/pulp/app/models/task.py
```diff
@@ -205,6 +206,22 @@ def set_failed(self, exc, einfo):
...
+        """
+        if self.state not in TASK_INCOMPLETE_STATES:
+            self.state = Task.CANCELED
+            self.finished_at = timezone.now()
+            self.save()
```
mhrivnak: This is approximately where a call should be made to celery to actually cancel the task. It would be a good idea to track that somehow either via a TODO here or by making a task in redmine. It might also be an easy 1-line addition to just knock out, but I'll leave that up to you.

app/pulp/app/models/task.py
```diff
@@ -205,6 +206,22 @@ def set_failed(self, exc, einfo):
...
 
+    def cancel(self):
+        """
+        Set the task to a canceled state if it is in an incomplete state.
+
+        This updates the :attr: `finished_at` attribute, sets the :attr: `state` to
+        :attr: `CANCELED`
+
+        """
+        if self.state not in TASK_INCOMPLETE_STATES:
```
mhrivnak: Isn't this the opposite of what we want? I assume incomplete states are "running" and "waiting". If the current state is one of those, then we can change it to canceled and save.

werwty: Oh man the lack of coffee is getting to me. Thanks for catching this.

mhrivnak:
This mostly looks good to me, but someone who has actually made one of these should approve.

One question: are we including doc blocks on functions and classes in the REST API? I can see where it might get redundant, but may still be worth it. Perhaps there's already been discussion and a conclusion about this? I just wanted to ask since there are no doc blocks in this PR.

tasking/pulp/tasking/base.py
```diff
@@ -388,7 +388,7 @@ def cancel(task_id):
         # If the task is already done, just stop
         msg = _('Task [%(task_id)s] already in a completed state: %(state)s')
         _logger.info(msg % {'task_id': task_id, 'state': task_status.state})
-        return
+        raise PulpException(msg % {'task_id': task_id, 'state': task_status.state})
```
mhrivnak: This may warrant some discussion. If a task has already been canceled, and a user requests again for it to be canceled, is there any case where they will care that it was already canceled? Is that an error, or an exceptional circumstance?

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,45 @@
...
+    http_method_names = ['get', 'options']
+
+    @detail_route(methods=['post'])
+    def cancel(self, request, pk=None):
+        task = self.get_object()
+        try:
+            cancel_task(task.pk)
+            return Response({'status': u'Task canceled'})
+        except (PulpException, MissingResource) as e:
+            return Response({'status': e.message}, status=status.HTTP_400_BAD_REQUEST)
```
mhrivnak: ``MissingResource`` should result in a 404, right?

mhrivnak: As noted below, I'm not sure 400 is the best response when the task has previously been canceled. Maybe it is, but it's worth discussion.

tasking/pulp/tasking/base.py
werwty: I do like telling the use exactly what's going on. But I do see your point about maybe just not worrying about this and telling the user we're making a best effort instead

werwty:
@mhrivnak There's no doc strings on this PR since it's doing the task/worker viewset in a very standardized way, and I think any info in a doc string would be redundant.

app/pulp/app/serializers/task.py
```diff
@@ -27,7 +27,6 @@ class TaskSerializer(ModelSerializer):
     state = serializers.CharField(
         help_text="The current state of the task. The possible values include:"
                   " 'waiting', 'skipped', 'running', 'completed', 'failed' and 'canceled'.",
-        read_only=True
     )
```
seandst: I just realized that if state isn't read_only, we probably want it to be a ChoiceField: http://www.django-rest-framework.org/api-guide/fields/#choicefield

I apologize for thinking of this at this point; if you think it would be better to open another issue, a seperate PR, or something, I'd be fine with that.

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,41 @@
...
+    serializer_class = TaskSerializer
+    endpoint_name = 'tasks'
+    filter_class = TaskFilter
+    http_method_names = ['get', 'options']
+
+    @detail_route(methods=['post'])
+    def cancel(self, request, pk=None):
+        task = self.get_object()
+        cancel_task(task.pk)
+        return Response(status=status.HTTP_200_OK)
```
mhrivnak: Am I right that the body of the response will be empty? If so, I think 204 is probably the best response code to be explicit that no representation of any entity should be expected.

werwty: :+1: the response is indeed empty.

app/pulp/app/viewsets/task.py
```diff
@@ -0,0 +1,41 @@
...
+from pulp.app.viewsets.custom_filters import CharInFilter
+from pulp.tasking.base import cancel as cancel_task
+
+from rest_framework.decorators import detail_route
+from rest_framework.response import Response
+from rest_framework import status
+
+
+class TaskFilter(filterset.FilterSet):
+    tags = CharInFilter(name='tag__name', lookup_expr='in')
```
asmacdo: Yeah!! Thanks for noticing the custom filter!

PR #2820, opened 2016-11-07
Adds serializer for the PulpUser model

dkliban:
closes #2356
https://pulp.plan.io/issues/2356

### DIFF
--- a/app/pulp/app/serializers/user.py
+++ b/app/pulp/app/serializers/user.py
@@ -1,0 +1,19 @@
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
+class UserSerializer(ModelSerializer):
+    username = serializers.CharField(
+        help_text="Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only."
+    )
+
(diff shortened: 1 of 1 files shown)


app/pulp/app/serializers/user.py
```diff
@@ -0,0 +1,29 @@
...
+
+    is_superuser = serializers.BooleanField(
+        help_text="Designates that this user has all permissions without explicitly assigning "
+                  "them."
+    )
+
+    groups = serializers.ManyRelatedField(
+        help_text="The groups this user belongs to. A user will get all permissions granted to "
+                  "each of their groups."
+    )
```
seandst: We should not be using ManyRelatedField:
https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/relations.py#L441-L452

`groups` and `user_permissions` should (probably) either be HyperlinkedRelatedFields, be nested, or not exist.

dkliban: We don't have any viewsets for groups and permissions. So I am not sure that I can add HyperlinkedRelatedFields for either of them. I am starting to think that I should remove these from the serializer from now.

PR #2823, opened 2016-11-07
Adds ViewSet for User model

dkliban:
Adds ability to:
  GET, POST to /api/v3/users/
  PUT, PATCH, DELETE to /api/v3/users/<username>/

closes #2357
https://pulp.plan.io/issues/2357

### DIFF
--- a/app/pulp/app/serializers/user.py
+++ b/app/pulp/app/serializers/user.py
@@ -1,19 +1,46 @@
-from rest_framework import serializers
+from gettext import gettext as _
 
-from pulp.app import models
+from django.contrib.auth.hashers import make_password
+from django.utils import six
+from rest_framework import serializers
+from rest_framework.validators import UniqueValidator
+
+from pulp.app.models import PulpUser
 from pulp.app.serializers import ModelSerializer
 
 
+class PasswordSerializer(serializers.CharField):
+    """
+    Serializer for the password field of User object.
+    """
+    def to_internal_value(self, data):
+        # We're lenient with allowing basic numerics to be coerced into strings,
+        # but other types should fail. Eg. unclear if booleans should represent as `true` or `True`,
+        # and composites such as lists are likely user error.
+        if isinstance(data, bool) or \
+                not isinstance(data, six.string_types + six.integer_types + (float,)):
+            self.fail('invalid')
+        value = six.text_type(make_password(data))
+        return value
+
+
 class UserSerializer(ModelSerializer):
+    _href = serializers.HyperlinkedIdentityField(view_name='users-detail',
+                                                 lookup_field='username')
+
     username = serializers.CharField(
-        help_text="Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only."
+        help_text=_("Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only."),
+        validators=[UniqueValidator(queryset=PulpUser.objects.all())],
     )
 
     is_superuser = serializers.BooleanField(
-        help_text="Designates that this user has all permissions without explicitly assigning "
-                  "them."
+        help_text=_("Designates that this user has all permissions without explicitly assigning "
+                    "them."),
+        required=False
     )
 
+    password = PasswordSerializer(write_only=True, help_text=_("Password"))
+
     class Meta:
-        model = models.PulpUser
-        fields = ModelSerializer.Meta.fields + ('username', 'is_superuser')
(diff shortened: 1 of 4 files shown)


app/pulp/app/viewsets/user.py
```diff
@@ -0,0 +1,23 @@
...
+    serializer_class = UserSerializer
+
+    @decorators.detail_route(methods=['post'])
+    def set_password(self, request, pk=None):
+        user = self.get_object()
+        serializer = UserSerializer(data=request.data)
+        if serializer.is_valid():
+            user.set_password(serializer.data['password'])
+            user.save()
+            return Response({'status': 'password set'})
```
asmacdo: This `Response` also needs `status=status.HTTP_200_OK`

app/pulp/app/viewsets/user.py
```diff
@@ -0,0 +1,23 @@
...
+
+class UserViewSet(NamedModelViewSet):
+    endpoint_name = 'users'
+    queryset = PulpUser.objects.all()
+    serializer_class = UserSerializer
+
+    @decorators.detail_route(methods=['post'])
+    def set_password(self, request, pk=None):
+        user = self.get_object()
+        serializer = UserSerializer(data=request.data)
```
asmacdo: The request data should contain only {'password': 'new_password'} right? I am curious if this would fail if any of the fields on the serializer other than password are changed to `required=True`.

app/pulp/app/viewsets/user.py
```diff
@@ -0,0 +1,23 @@
...
+        if serializer.is_valid():
+            user.set_password(serializer.data['password'])
```
asmacdo: I think this will return a 500 if the user hits the right endpoint but fails to include a password.

app/pulp/app/serializers/user.py
```diff
@@ -14,6 +14,9 @@ class UserSerializer(ModelSerializer):
                   "them."
     )
 
+    password = serializers.CharField(write_only=True, help_text="Password", min_length=5,
```
asmacdo: I think the min and max lengths are new, how did you choose these numbers?

Enforcing this might be fine, but this would need to be a release note. We don't really need to worry about backwards compatibility here because we won't be able to migrate the old passwords anyway.

app/pulp/app/viewsets/user.py
```diff
@@ -0,0 +1,23 @@
...
+class UserViewSet(NamedModelViewSet):
+    endpoint_name = 'users'
+    queryset = PulpUser.objects.all()
+    serializer_class = UserSerializer
+
+    @decorators.detail_route(methods=['post'])
+    def set_password(self, request, pk=None):
+        user = self.get_object()
+        serializer = UserSerializer(data=request.data)
+        if serializer.is_valid():
```
asmacdo: I am *pretty sure* that this will run this validation https://github.com/pulp/pulp/blob/3.0-dev/app/pulp/app/models/auth.py#L48, but I would need to double check.

Another way we could do this is to create a `validate_password` method on the serializer. If named that way, DRF will know about it and do the right thing.

app/pulp/app/serializers/user.py
```diff
@@ -1,19 +1,60 @@
...
+        :param value: Username being validated
+        :type value: str
+        :return: value
+        :rtype: str
+        :raises: :class:`rest_framework.serializers.ValidationError`
+        """
+        try:
+            models.PulpUser.objects.get(username=value)
+            raise serializers.ValidationError(_("Username is not unique."))
+        except ObjectDoesNotExist:
```
seandst: Models get their own version of this exception; use `PulpUser.DoesNotExist` here instead.

app/pulp/app/serializers/user.py
```diff
@@ -1,19 +1,60 @@
...
+        """
+        Check that the username does not exist in the database.
+        :param value: Username being validated
+        :type value: str
+        :return: value
+        :rtype: str
+        :raises: :class:`rest_framework.serializers.ValidationError`
+        """
+        try:
+            models.PulpUser.objects.get(username=value)
```
seandst: Existence checks should use `.exists`, e.g. `models.PulpUser.objects.filter(username=value).exists()`

If the username Model field is not indexed, we'll want it to be indexed.

https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.exists

PR #2826, opened 2016-11-09
Add serializer for the progress models.

ipanova:
closes #2302
https://pulp.plan.io/issues/2302

### DIFF
--- a/app/pulp/app/serializers/progress.py
+++ b/app/pulp/app/serializers/progress.py
@@ -1,0 +1,44 @@
+from gettext import gettext as _
+
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
+class ProgressReportSerializer(ModelSerializer):
+
+    message = serializers.CharField(
+        help_text=_("The message shown to the user for the progress report."),
+        read_only=True
+    )
+
+    state = serializers.CharField(
+        help_text=_("The current state of the progress report. The possible values are:"
+                    " 'waiting', 'skipped', 'running', 'completed', 'failed' and 'canceled'."
+                    " The default is 'waiting'."),
+        read_only=True
+    )
+
+    total = serializers.IntegerField(
+        help_text=_("The total count of items to be handled by the ProgressBar."),
+        read_only=True
+    )
+
+    done = serializers.IntegerField(
+        help_text=_("The count of items already processed. Defaults to 0."),
+        read_only=True
+    )
+
+    task = serializers.HyperlinkedRelatedField(
+        help_text=_("The task associated with this progress report."),
+        read_only=True,
+        view_name='tasks-detail'
+    )
+
+    class Meta:
+        model = models.ProgressReport
+        # this serializer is meant to be nested inside Task serializer,
+        # so it will not have its own endpoint, that's why
+        # we need to explicitly define fields to exclude '_href' field.
+        fields = ('message', 'state', 'total', 'done', 'task')
--- a/app/pulp/app/models/__init__.py
+++ b/app/pulp/app/models/__init__.py
@@ -13,4 +13,4 @@ from .catalog import DownloadCatalog  # NOQA
 from .task import ReservedResource, Worker, Task, TaskTag, TaskLock  # NOQA
 
 # Moved here to avoid a circular import with Task
-from .progress import ProgressBar, ProgressSpinner  # NOQA
+from .progress import ProgressBar, ProgressReport, ProgressSpinner  # NOQA
--- a/app/pulp/app/serializers/__init__.py
+++ b/app/pulp/app/serializers/__init__.py
@@ -9,6 +9,7 @@ from pulp.app.serializers.generic import (ConfigKeyValueRelatedField,  # NOQA
     NotesKeyValueRelatedField, ScratchpadKeyValueRelatedField)  # NOQA
 from pulp.app.serializers.consumer import ConsumerSerializer  # NOQA
 from pulp.app.serializers.content import ContentSerializer, ArtifactSerializer  # NOQA
+from pulp.app.serializers.progress import ProgressReportSerializer  # NOQA
 from pulp.app.serializers.repository import (ImporterSerializer, PublisherSerializer,  # NOQA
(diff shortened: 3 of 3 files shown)


ipanova:
i have here couple of questions/doubts:
- are we going to use progress in the api? if not then all this is not needed, i guess
- w/r to the viewsets i made both models point to one endpoint since they are proxied, not sure if it was a good idea or not, but we can make separate endpoints for progress bar and spinner.

ipanova:
another note, this will not work until PR with Task Viewset will not be merged

seandst:
> are we going to use progress in the api? if not then all this is not needed, i guess

Without this, I don't know how a client (such as the CLI) will be able to report progress on ongoing tasks. I think these endpoints are necessary.

> w/r to the viewsets i made both models point to one endpoint since they are proxied, not sure if it was a good idea or not, but we can make separate endpoints for progress bar and spinner.

This will not work, and in fact should be explicitly disallowed, if possible. I'll file a bug for this (or maybe just fix it).

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    state = serializers.CharField(
+        help_text="The current state of the progress update. The possible values include:"
+                  " 'waiting', 'skipped', 'running', 'completed', 'failed' and 'canceled'."
+                  " Default is waiting.",
+        read_only=True
+    )
+
+    total = serializers.IntegerField(
```
bmbouter: This field will always be updated on the server side. Given that, this should get a `read_only=True` also right?

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+        help_text="The total count of items to be handled by the ProgressBar."
+    )
+
+    done = serializers.IntegerField(
+        help_text="The count of items already processed. Defaults to 0."
```
bmbouter: This field will always be updated on the server side. Given that, this should get a `read_only=True` also right?

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+
+class ProgressReportSerializer(ModelSerializer):
+
+    _href = serializers.HyperlinkedIdentityField(
+    view_name='progress-detail',
+    lookup_field='id',
+    )
+
+    message = serializers.CharField(
+        help_text="The message shown to the user for the progress update.",
```
bmbouter: s/update/report/

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+        read_only=True
+    )
+
+    state = serializers.CharField(
+        help_text="The current state of the progress update. The possible values include:"
```
bmbouter: s/update/report/

bmbouter: s/include/are/

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+                  " 'waiting', 'skipped', 'running', 'completed', 'failed' and 'canceled'."
+                  " Default is waiting.",
```
bmbouter: s/Default is waiting/The default is 'waiting'/

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
...
+        read_only=True,
+        view_name='tasks-detail'
+    )
+
+    class Meta:
+        model = models.ProgressReport
+        fields = ModelSerializer.Meta.fields + ('message', 'state', 'total', 'done', 'task')
+
+
+class ProgressSpinnerSerializer(ProgressReportSerializer):
```
bmbouter: I don't think we need this serializer. We only need to serialize the ProgressReport model since the other two model types, `ProgressBar`, and `ProgressSpinner` are proxy models of `ProgressReport`.

bmbouter:
@ipanova @seandst I'm hoping we can not have a dedicated endpoint for ProgressReport or any of it's descendant models. Instead I hope we will include all associated ProgressReport entries in the task view. The task view should use the ProgressReportSerializer to serialize all ProgressReport objects that have a forieng key to the task being looked up in the view.

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,57 @@
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
+class ProgressReportSerializer(ModelSerializer):
+
+    _href = serializers.HyperlinkedIdentityField(
```
asmacdo: `_href` should be provided by the ModelSerializer, and will just work magically. The reason the other serializers have to define this is that their `lookup_field` is something other than `id`.

seandst:
> Thanks @ipanova for doing this. I don't think we need a viewset at all, but instead we need the task viewset to use the serializer this PR provides. We should reach an agreement with @seandst, whoever is making the task viewset PR, and others too.

This makes good sense to me. As an API user, it seems easier to get a task's progress from the task itself, and not have to bounce over to another API view.

ipanova:
@seandst @bmbouter I think all comments were addressed, please take another look

ipanova:
@seandst @bmbouter i think all comments have been addressed, please take another look.

app/pulp/app/serializers/progress.py
```diff
@@ -0,0 +1,42 @@
+from rest_framework import serializers
+
+from pulp.app import models
+from pulp.app.serializers import ModelSerializer
+
+
+class ProgressReportSerializer(ModelSerializer):
+
+    message = serializers.CharField(
+        help_text="The message shown to the user for the progress report.",
```
bmbouter: Do these need to be i8ln here or does that occur elsewhere?

seandst: That's a good question, but I don't know the answer to it. DRF's docs scarcely mention anything to do with internationalization or localization, so I expect this is the place to do it unless we want to get really fancy and do a lot of work to support localizing responses per-request.

bmbouter:
@ipanova it does need to be rebased prior to merging

PR #2829, opened 2016-11-10
Delete old settings file

werwty:
closes #2245
https://pulp.plan.io/issues/2245


werwty:
Those places already switched over to the new settings file/aren't referencing any settings file...
https://github.com/pulp/pulp/blob/3.0-dev/tasking/pulp/tasking/celery_instance.py#L8
https://github.com/pulp/pulp/blob/3.0-dev/app/pulp/app/wsgi.py#L14

The only mentions left of pulp.server.webservices.settings is in the /server directory

PR #2831, opened 2016-11-11
Handles exception when restorecon is called with SELinux disabled.

dkliban:
closes #2387
https://pulp.plan.io/issues/2387

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -902,7 +902,11 @@ class AtomicDirectoryPublishStep(PluginStep):
 
         try:
             os.rename(self.source_dir, timestamp_master_dir)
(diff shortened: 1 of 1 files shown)


server/pulp/plugins/util/publish_step.py
```diff
@@ -902,7 +902,11 @@ def process_main(self, item=None):
 
         try:
             os.rename(self.source_dir, timestamp_master_dir)
-            selinux.restorecon(timestamp_master_dir.encode('utf-8'), recursive=True)
+            try:
+                selinux.restorecon(timestamp_master_dir.encode('utf-8'), recursive=True)
+            except OSError as e:
+                if e.errno != errno.EPERM:
+                    raise
```
seandst: I like this better than the more crypic selinux_getenforcemode check in the previous commit.

dkliban:
ok test

PR #2834, opened 2016-11-15
Create 3.0 docs framework

asmacdo:
Creates the basic structure of the documentation that will go out with 3.0.0. 

This moves all of the old docs so they can be moved back into active docs and retain git history.

re #2348
https://pulp.plan.io/issues/2348

This PR contains a second commit that removes some extensions so sphinx can build. This commit is intended for review only and will not be merged.


bmbouter:
The rest_api.py file is being moved but it should be deleted instead.

bmbouter:
Also we need to not be moving `docs/_extensions/napoleon_django/` -> `old_docs/_extensions/napoleon_django/`

asmacdo:
@bmbouter, the travis issue was fixed in the last day on by a different commit.

docs/conf.py
```diff
@@ -19,7 +19,7 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #sys.path.insert(0, os.path.abspath('.'))
```
bmbouter: Can this line be removed since it is already commented

docs/conf.py
```diff
@@ -19,7 +19,7 @@
...
-sys.path.insert(0, os.path.abspath('./_extensions'))
+sys.path.insert(0, os.path.abspath('./extensions_src'))
```
bmbouter: Can this be s/extensions_src/extensions/ once again? I think just calling them extensions is good enough since we use them as-is and don't compile them into another asset unlike the diagrams for example.

PR #2835, opened 2016-11-15
Fix failing flake8

asmacdo:


app/pulp/app/models/base.py
```diff
@@ -142,4 +142,6 @@ def master_model(options):
     except IndexError:
         # Also None if this model is itself the master.
         return None
+
+
```
seandst: This is already fixed on 3.0-dev, which should be fixed with a rebase.

PR #2839, opened 2016-11-16
Changes how pulp-selinux RPM decides when to run restorecon statements

dkliban:
RHEL 7.3 was experiencing an bug that was preventing the pulp-selinux RPM from using semodule -l to
figure out the installed version of pulp-selinux policies during upgrades.

The other change is in relabel.sh. This change fixed version comparison to account for the fact that
2.10.0 is greater than 2.2.0.

closes #2434
https://pulp.plan.io/issues/2424

### DIFF
--- a/pulp.spec
+++ b/pulp.spec
@@ -967,8 +967,8 @@ SELinux policy for Pulp's components
 %pre selinux
 # Record old version so we can limit which restorecon statement are executed later
 test -e %{_localstatedir}/lib/rpm-state/%{name} || mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
-oldversion=$(semodule -l | grep pulp-server)
-echo ${oldversion:12} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
+oldversion=$(rpm -qa pulp-selinux)
+echo ${oldversion:13} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
 
 exit 0
 %post selinux
--- a/server/selinux/server/relabel.sh
+++ b/server/selinux/server/relabel.sh
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
 
+function version_less_than () {
+# Determines if the version passed in as the first argument is less than the version in the second
+# argument.
+    [[ $(echo -e $1'\n'$2|sort -V|head -n 1) != $2 ]]
+}
+
 # If upgrading from before 2.4.0
-if [[ $1 < '2.4.0' ]]
+if version_less_than $1 '2.4.0'
 then
     /sbin/restorecon -i -R /etc/httpd/conf.d/pulp.conf
     /sbin/restorecon -i -R /etc/pulp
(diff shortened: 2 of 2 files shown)


server/selinux/server/relabel.sh
```diff
@@ -1,7 +1,18 @@
...
+function version_less_than {
+# Determines if the version passed in as the first argument is less than the version in the second
+# argument.
+    if [[ $(echo -e $1'\n'$2|sort -V|head -n 1) != $2 ]]
+    then
+        return 0
+    else
+        return 1
+    fi
+}
```
seandst: bash does the implicit return thing, sorta like ruby, so this can just be
```bash
function version_less_than {
# Determines if the version passed in as the first argument is less than the version in the second
# argument.
    [[ $(echo -e $1'\n'$2|sort -V|head -n 1) != $2 ]]
}
```

pulp.spec
```diff
@@ -967,8 +967,8 @@ SELinux policy for Pulp's components
 %pre selinux
 # Record old version so we can limit which restorecon statement are executed later
 test -e %{_localstatedir}/lib/rpm-state/%{name} || mkdir -p %{_localstatedir}/lib/rpm-state/%{name}
-oldversion=$(semodule -l | grep pulp-server)
-echo ${oldversion:12} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
+oldversion=$(rpm -qa pulp-selinux)
+echo ${oldversion:13} > %{_localstatedir}/lib/rpm-state/%{name}/old-version
```
bmbouter: What is the purpose of this line?

dkliban: Line 971 takes a string that looks like 'pulp-selinux-2.10.0-0.1' and writes just '2.10.0-0.1' to a file called old-verion. This is done before the uninstalling the pulp-selinux package. The content of that file is then used in the post install step as the argument for the relabel.sh script.

bmbouter: Thank makes sense, thanks

server/selinux/server/relabel.sh
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
 
+function version_less_than {
```
bmbouter: For readability can this become:

`function version_less_than () {`

bmbouter:
@dkliban  What testing was done with this?

server/selinux/server/relabel.sh
dkliban: sure

server/selinux/server/relabel.sh
```diff
@@ -1,7 +1,13 @@
...
+# Determines if the version passed in as the first argument is less than the version in the second
+# argument.
+    [[ $(echo -e $1'\n'$2|sort -V|head -n 1) != $2 ]]
```
bmbouter: I did several comparisons locally, and I think this statement is exactly what we need.

dkliban:
I tested all the lines of code individually.  I created a copy of the relabel.sh file, but replaced all the restorecon statements with echo statements.  I also checked that the statements that use  rpm -qa pulp-selinux produce the correct version string. 

I did not build an RPM and try to install it.

PR #2841, opened 2016-11-16
Update troubleshooting guide [2.10-dev]

elyezer:
Inform the reader that the log level will be adjusted only when Pulp processes
are restarted.


seandst:
closes #2382

PR #2844, opened 2016-11-16
Reorder titles for hierarchy

asmacdo:


docs/contributing/3.0-development/index.rst
```diff
@@ -1,8 +1,13 @@
...
     All documents within the 3.0-Development section should be considered temporary, and will be removed or relocated prior to the release of Pulp 3.0.
 
+3.0 Development
+===============
+
+The goal of this section is to create a place for living documents related to the development of Pulp 3.0. Some parts may grow and replace current documentation (plugin API) and others may be temporary guides to making changes (translating from Mongo to Postgres).
+
 Pulp 3 and Python 3
-===================
+*******************
```
seandst: I prefer ------ for second-level headings, but as long as sphinx doesn't hate this, neither do I. :)

asmacdo: I recently discovered the frustrating and time consuming way that sphinx chooses the heading level based on the order that the title symbols are used in a document. The symbol itself doesn't matter to sphinx.

That said, I like `-----` better too, `*****` just came to mind first. I'm going to change before merge.

PR #2850, opened 2016-11-17
Various minor fixes to get workers/celerybeat/resource_manager run

goosemania:
One can also dispatch a task as a bonus

### DIFF
--- a/tasking/pulp/tasking/base.py
+++ b/tasking/pulp/tasking/base.py
@@ -8,6 +8,7 @@ from gettext import gettext as _
 from celery import Task as CeleryTask, current_task, task
 from celery.app import control
 from celery.result import AsyncResult
+from django.db import transaction
 from django.db.models import Count
 
 from pulp.app.models import ReservedResource, Task as TaskStatus, TaskLock, Worker
@@ -141,15 +142,20 @@ def delete_worker(name, normal_shutdown=False):
         msg = msg % {'name': name}
         _logger.error(msg)
 
-    worker = Worker.objects.get(name=name)
+    try:
+        worker = Worker.objects.get(name=name)
+    except Worker.DoesNotExist:
+        pass
+    else:
+        # Cancel all of the tasks that were assigned to this worker's queue
+        for task_status in worker.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
+            cancel(task_status.pk)
 
-    # Cancel all of the tasks that were assigned to this worker's queue
-    for task_status in worker.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
-        cancel(task_status.pk)
+        worker.delete()
 
-    worker.delete()
-
-    if name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME):
+    is_resource_manager = name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME)
+    is_celerybeat = name.startswith(TASKING_CONSTANTS.CELERYBEAT_WORKER_NAME)
+    if is_celerybeat or is_resource_manager:
         TaskLock.objects.filter(name=name).delete()
 
 
@@ -180,7 +186,7 @@ def _release_resource(task_id):
 
         new_task.on_failure(runtime_exception, task_id, (), {}, MyEinfo)
 
-    ReservedResource.objects.get(task__pk=task_id).delete()
+    ReservedResource.objects.filter(task__pk=task_id).delete()
 
 
 class UserFacingTask(PulpTask):
@@ -247,11 +253,14 @@ class UserFacingTask(PulpTask):
         group_id = kwargs.get('group_id', None)
 
         # Set the parent attribute if being dispatched inside of a Task
-        parent_arg = self._get_parent_arg
+        parent_arg = self._get_parent_arg()
 
         # Create a new task status with the task id and tags.
-        TaskStatus.objects.create(pk=inner_task_id, state=TaskStatus.WAITING, tags=tag_list,
-                                  group=group_id, **parent_arg)
+        with transaction.atomic():
+            task_status = TaskStatus.objects.create(pk=async_result.id, state=TaskStatus.WAITING,
+                                                    group=group_id, **parent_arg)
+            for tag in tag_list:
+                task_status.tags.create(name=tag)
 
         # Call the outer task which is a promise to call the real task when it can.
         _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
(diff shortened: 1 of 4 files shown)


goosemania:
This will work when django app label issue will be resolved

tasking/pulp/tasking/base.py
```diff
@@ -180,7 +184,7 @@ class MyEinfo(object):
 
         new_task.on_failure(runtime_exception, task_id, (), {}, MyEinfo)
 
-    ReservedResource.objects.get(task__pk=task_id).delete()
+    ReservedResource.objects.filter(task__pk=task_id).delete()
```
bmbouter: good catch

tasking/pulp/tasking/base.py
```diff
@@ -247,11 +251,14 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         group_id = kwargs.get('group_id', None)
 
         # Set the parent attribute if being dispatched inside of a Task
-        parent_arg = self._get_parent_arg
+        parent_arg = self._get_parent_arg()
```
bmbouter: nice

tasking/pulp/tasking/scheduler.py
```diff
@@ -258,7 +258,7 @@ def tick(self):
                           % {'celerybeat_name': self.celerybeat_name})
             ret = self.call_tick(self, self.celerybeat_name)
         except TaskLock.DoesNotExist:
-            TaskLock.objects.get(name=self.celerybeat_name, timestamp__lte=old_timestamp).delete()
+            TaskLock.objects.filter(name=self.celerybeat_name, timestamp__lte=old_timestamp).delete()
```
bmbouter: I think this line isn't needed at all. `TaskLock` is unique by `name` and because of the query at L#253 causing a `TaskLock.DoesNotExist` exception, `TaskLock.objects.filter(name=self.celerybeat_name).count()` should equal 0.

goosemania: Yeah, it looks like  it is not needed here.
But... Is it possible that it was an attempt to delete any old celerybeat lock (from other machine/node as well) which for some reason was not cleaned up properly before? 
In such case the lock should be specified and not the name:
`TaskLock.objects.filter(lock=TaskLock.CELERY_BEAT, timestamp__lte=old_timestamp).delete()`

@bmbouter , @dkliban what do you think?

goosemania:
@bmbouter , I think you meant #2851. I do not think they should be combined. The changes introduced in this PR are real fixes an should be merged, but PR #2851 is there just to unlock the task conversion work with the workaround which is questionable. We are still looking for the good way to make app label work. 
PR #2851 includes this commit just to make itself easier to use, I will rebase #2831 once this PR is merged.

tasking/pulp/tasking/scheduler.py
bmbouter: This doesn't look like the right location for this line, but I agree we need to handle lock cleanup. I'm looking into it to make a recommendation. I'll post back on this thread.

bmbouter: The resource manager lock during graceful shutdown gets [cleaned up here](https://github.com/goosemania/pulp/blob/b8abcc929cc3cc4d0a0c2a531d89f2d17480d24f/tasking/pulp/tasking/base.py#L156-L157). Note that the celerybeat code also calls that from [here](https://github.com/goosemania/pulp/blob/b8abcc929cc3cc4d0a0c2a531d89f2d17480d24f/tasking/pulp/tasking/scheduler.py#L286). We should add the celerybeat lock release next to the resource manager lock release in the same place.

Also, as an aside, I think the resource manager lock removal needs to be moved outside of [the if statement](https://github.com/goosemania/pulp/blob/b8abcc929cc3cc4d0a0c2a531d89f2d17480d24f/tasking/pulp/tasking/base.py#L156) that current guards it.

I think our non-graceful failover case is broken so I'm going to look into some recommendations there.

One question I have is what is the purpose of [this field](https://github.com/goosemania/pulp/blob/b8abcc929cc3cc4d0a0c2a531d89f2d17480d24f/app/pulp/app/models/task.py#L82)? I think it can be removed.

goosemania: My understanding is that `lock` field indicates that it is a celerybeat lock and it is the important one. The [`name` field](https://github.com/goosemania/pulp/blob/b8abcc929cc3cc4d0a0c2a531d89f2d17480d24f/app/pulp/app/models/task.py#L80) is probably not necessary but it contains the name of the node so it may be useful.

bmbouter: @goosemania The lock delete statement when it is moved to `delete_worker` should probably still delete by worker name since that is the graceful deletion case.

tasking/pulp/tasking/base.py
```diff
@@ -141,16 +142,19 @@ def delete_worker(name, normal_shutdown=False):
...
+        # Cancel all of the tasks that were assigned to this worker's queue
+        for task_status in worker.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
+            cancel(task_status.pk)
 
-    worker.delete()
+        worker.delete()
 
-    if name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME):
-        TaskLock.objects.filter(name=name).delete()
+        if name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME):
```
bmbouter: L#156 and L#157 need to be unindended one level so that they run every time `delete_worker` is called.

tasking/pulp/tasking/scheduler.py
dkliban: @goosemania I thikn you are right that the cleanup on line 261 is supposed to remove any locks that were created by other machines. So the 'lock' field should be filtered on instead of the 'name'.

tasking/pulp/tasking/base.py
```diff
@@ -141,16 +142,19 @@ def delete_worker(name, normal_shutdown=False):
         msg = msg % {'name': name}
         _logger.error(msg)
 
-    worker = Worker.objects.get(name=name)
-
-    # Cancel all of the tasks that were assigned to this worker's queue
-    for task_status in worker.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
-        cancel(task_status.pk)
+    try:
+        worker = Worker.objects.get(name=name)
```
bmbouter: What situation occurs where `delete_worker` is called yet the worker does not exist?

tasking/pulp/tasking/base.py
bmbouter: We also need to add a removal by name when the name startswith the celerybeat name

tasking/pulp/tasking/base.py
bmbouter: I think having the try/except additions here could be reverted.

bmbouter: Oh I see! Let's keep the `try/except/else`. but still move the lock cleanup outside of the else statement.

bmbouter: Thank you @goosemania

goosemania:
@bmbouter , @dkliban,  thanks for all the comments and discussion.
Please, re-review.
Here is [one change](https://github.com/pulp/pulp/pull/2850/files#diff-ff8367d6867b3f6557a36f4a482f9d25R156) and [the other](https://github.com/pulp/pulp/pull/2850/files#diff-271ca2b338d132bcca08e20e061e1fe6R261).

tasking/pulp/tasking/base.py
```diff
@@ -141,16 +142,18 @@ def delete_worker(name, normal_shutdown=False):
...
+        # Cancel all of the tasks that were assigned to this worker's queue
+        for task_status in worker.tasks.filter(state__in=TASK_INCOMPLETE_STATES):
+            cancel(task_status.pk)
 
-    worker.delete()
+        worker.delete()
 
-    if name.startswith(TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME):
-        TaskLock.objects.filter(name=name).delete()
+    TaskLock.objects.filter(name=name).delete()
```
goosemania: Sure, I will update it. It is just want I recalled from our discussion but I could easily mixed it up.
And I agree it is more explicit and guarded way to do it.

PR #2851, opened 2016-11-17
[Do not merge] Tasking sandbox

goosemania:
Workaround to avoid app label issue and start playing with tasks.

You would need to switch to RabbitMQ.
`sudo dnf install rabbitmq-server`
To use the default port for broker, stop/disable Qpid and start/enable RabbitMQ:
`sudo systemctl disable qpidd --now`
`sudo systemctl status rabbitmq-server`
Then change broker_url either in settings.py (as in PR), or in server.yaml config.
Start pulp_workers/celerybeat/resource_manager in a usual way via `systemctl` (find modified startup scripts in this PR).
Run `python manage.py shell` and try something out and continue the journey:
```
from pulp.tasking.registry import delete_publisher
delete_publisher.apply_async(...)
```
or
`delete_publisher.apply_async_with_reservation(...)`

Current status:
 - workers/resource_manager/celerybeat run and [almost] work
 - tasks can be dispatched but they are not handled well all the way

This is where I am so far. This is just to share where I am now. Feel free to join anytime ;)

### DIFF
--- a/tasking/pulp/celery_app.py
+++ b/tasking/pulp/celery_app.py
@@ -11,6 +11,11 @@ import sys
 import time
 from gettext import gettext as _
 
+import os
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
+import django
+django.setup()
+
 from celery.signals import celeryd_after_setup
 from django.db.utils import IntegrityError
 
@@ -19,10 +24,10 @@ from pulp.tasking import delete_worker, storage
 from pulp.tasking.constants import TASKING_CONSTANTS
 
 # This import is here so that Celery will find our application instance
-from pulp.tasking.celery_instance import celery  # noqa
+from pulp.celery_instance import celery  # noqa
 
 # This import is here so Celery will discover all tasks
-import pulp.tasking.registry  # noqa
+import pulp.tasking.registry # noqa
 
 
 _logger = logging.getLogger(__name__)
--- a/tasking/pulp/celery_instance.py
+++ b/tasking/pulp/celery_instance.py
@@ -3,13 +3,12 @@ import ssl
 
 from celery import Celery
 
-from pulp.app import settings
-
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
+
+from django.conf import settings
 
 broker_url = settings.BROKER['url']
 celery = Celery('tasks', broker=broker_url)
-
 
 DEDICATED_QUEUE_EXCHANGE = 'C.dq'
 RESOURCE_MANAGER_QUEUE = 'resource_manager'
@@ -18,7 +17,7 @@ CELERYBEAT_SCHEDULE = {
 
 
 celery.conf.update(CELERYBEAT_SCHEDULE=CELERYBEAT_SCHEDULE)
-celery.conf.update(CELERYBEAT_SCHEDULER='pulp.server.async.scheduler.Scheduler')
+celery.conf.update(CELERYBEAT_SCHEDULER='pulp.tasking.scheduler.Scheduler')
 celery.conf.update(CELERY_WORKER_DIRECT=True)
 celery.conf.update(CELERY_TASK_SERIALIZER='json')
 celery.conf.update(CELERY_ACCEPT_CONTENT=['json'])
--- a/tasking/pulp/tasking/registry.py
+++ b/tasking/pulp/tasking/registry.py
@@ -1,3 +1,4 @@
 """
 Any Celery task imported here will be known to Pulp
 """
+from pulp.tasking.tasks.repository import delete_publisher
--- a/tasking/pulp/tasking/scheduler.py
+++ b/tasking/pulp/tasking/scheduler.py
@@ -11,7 +11,7 @@ from django.db import IntegrityError
 
 from pulp.app.models.task import TaskLock, Worker
 from pulp.tasking import worker_watcher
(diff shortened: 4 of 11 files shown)


mhrivnak:
This works great. I pulled your branch, and then I just had to make these two fixes to the tasking system:

```
diff --git a/app/pulp/app/models/task.py b/app/pulp/app/models/task.py
index 9ad8239..9bf5462 100644
--- a/app/pulp/app/models/task.py
+++ b/app/pulp/app/models/task.py
@@ -151,7 +151,8 @@ class Task(Model):
             result (dict): The result to save on the :class:`~pulp.app.models.Task`
         """
-        self.result = result
+        if result is not None:
+            self.result = result
 
         # Only set the state to finished if it's not already in a complete state. This is
         # important for when the task has been canceled, so we don't move the task from canceled
diff --git a/tasking/pulp/tasking/base.py b/tasking/pulp/tasking/base.py
index 98b847b..198f789 100644
--- a/tasking/pulp/tasking/base.py
+++ b/tasking/pulp/tasking/base.py
@@ -336,7 +336,7 @@ class UserFacingTask(PulpTask):
         _logger.debug("Task successful : [%s]" % task_id)
         if not self.request.called_directly:
             task_status = TaskStatus.objects.get(pk=task_id)
-            task_status.set_success(retval)
+            task_status.set_completed(retval)
 
         storage.delete_working_directory()
```

Then I was able to run your task in a worker by doing this:

```
#!/usr/bin/env python3

from pulp.app import models
from pulp.tasking.registry import delete_publisher


models.Repository.objects.all().delete()
repo = models.Repository(name='foo')
repo.save()

publisher = models.Publisher(name='bar', repository=repo)
publisher.save()

task = delete_publisher.apply_async(kwargs={'repo_name': 'foo', 'publisher_name': 'bar'})
```

mhrivnak:
These two PRs fix the issues mentioned above:

https://github.com/pulp/pulp/pull/2855
https://github.com/pulp/pulp/pull/2856

tasking/pulp/tasking/scheduler.py
```diff
@@ -258,7 +258,7 @@ def tick(self):
                           % {'celerybeat_name': self.celerybeat_name})
             ret = self.call_tick(self, self.celerybeat_name)
         except TaskLock.DoesNotExist:
-            TaskLock.objects.get(name=self.celerybeat_name, timestamp__lte=old_timestamp).delete()
+            TaskLock.objects.filter(name=self.celerybeat_name, timestamp__lte=old_timestamp).delete()
```
bmbouter: I believe this line can be entirely eliminated. Since the TaskLock does not exist deleting it should not be necessary.

tasking/pulp/tasking/registry.py
```diff
@@ -1,3 +1,4 @@
 """
 Any Celery task imported here will be known to Pulp
 """
+from pulp.tasking.tasks.repository import delete_publisher
```
bmbouter: Can a newline be added at the end of this file please?

tasking/pulp/celery_app.py
```diff
@@ -11,6 +11,11 @@
 import time
 from gettext import gettext as _
 
+import os
```
bmbouter: This could be grouped with the other standard imports above

tasking/pulp/celery_app.py
```diff
@@ -11,6 +11,11 @@
...
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
+import django
```
bmbouter: It would be good to add a whitespace between this import and the `os.environ.setdefault`

bmbouter:
This branch has conflicts which need resolving.

tasking/pulp/celery_app.py
```diff
@@ -19,10 +24,10 @@
 from pulp.tasking.constants import TASKING_CONSTANTS
 
 # This import is here so that Celery will find our application instance
-from pulp.tasking.celery_instance import celery  # noqa
+from pulp.celery_instance import celery  # noqa
 
 # This import is here so Celery will discover all tasks
-import pulp.tasking.registry  # noqa
+import pulp.tasking.registry # noqa
```
bmbouter: I think 2 spaces are expected between code and an inline comment. This should likely be reverted.

tasking/pulp/celery_instance.py
```diff
@@ -3,22 +3,21 @@
 
 from celery import Celery
 
-from pulp.app import settings
-
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
 
+from django.conf import settings
```
bmbouter: I believe this is already fixed on 3.0-dev

https://github.com/pulp/pulp/blob/1a5e29cd797ae9300060b36ec234f9f364cfe9e8/tasking/pulp/tasking/celery_instance.py#L6

tasking/pulp/tasking/tasks/repository.py
```diff
@@ -0,0 +1,8 @@
+from celery import task
+
+from pulp.app.models import Publisher
+from pulp.tasking import UserFacingTask
+
```
bmbouter: Another line of whitespace is needed here for PEP8

goosemania:
@bmbouter, thanks for the review, I will fix all the style issues with the next push.

At the same time it is probably too early to merge these changes, I am looking for another solution for app label issue. But this PR can be used to test tasks till then.

PR #2853, opened 2016-11-18
Add RepositoryContent Serializer and Viewset

werwty:
closes #2311
https://pulp.plan.io/issues/2311


app/pulp/app/serializers/fields.py
```diff
@@ -11,6 +11,14 @@ class ContentRelatedField(DetailRelatedField):
     """
     queryset = models.Content.objects.all()
 
+    def use_pk_only_optimization(self):
```
asmacdo: I moved this up into the `DetailRelatedField` so it will work generally and this block can be removed.

PR #2855, opened 2016-11-18
no longer overwriting default value if task.result is None

mhrivnak:
Because this field cannot be null, this was causing an error when the result
was None.

### DIFF
--- a/app/pulp/app/models/task.py
+++ b/app/pulp/app/models/task.py
@@ -151,7 +151,8 @@ class Task(Model):
             result (dict): The result to save on the :class:`~pulp.app.models.Task`
         """
         self.finished_at = timezone.now()
-        self.result = result
+        if result is not None:
+            self.result = result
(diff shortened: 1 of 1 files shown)


BrnoPCmaniak:
What about even making the `result` optional argument?

mhrivnak:
I like that idea better also. Was there any particular reason it was given a default value of empty list to begin with? @dkliban it looks like you added that recently.

https://github.com/pulp/pulp/commit/9a37b213c2527fdbb0f2dec47f2e3d436e48be54#diff-ae04c17153454c7da50598aecabab77bR149

dkliban:
I added the default becuase there was some code that checked whether or not it was a list before appending errors to it or creating a list and then appending errors to it.

mhrivnak:
Closing in favor of https://github.com/pulp/pulp/pull/2857

PR #2859, opened 2016-11-22
Small fixes for tasking

bmbouter:
- adds a missing __init__.py
- switches default to RabbitMQ temporarily until the
  Qpid transport is Python 3.4 compatible

### DIFF
--- a/app/pulp/app/settings.py
+++ b/app/pulp/app/settings.py
@@ -184,7 +184,7 @@ _DEFAULT_PULP_SETTINGS = {
         'working_directory': '/var/cache/pulp',
     },
     'broker': {
-        'url': 'qpid://localhost/',
+        'url': 'amqp://guest@localhost//',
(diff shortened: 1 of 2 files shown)


app/pulp/app/settings.py
```diff
@@ -184,7 +184,7 @@
         'working_directory': '/var/cache/pulp',
     },
     'broker': {
-        'url': 'qpid://localhost/',
+        'url': 'amqp://guest@localhost//',
```
dkliban: I think the last slash is extra.

bmbouter: It's actually correct it signifies the default virtual host inside of rabbitMQ. Strange I know.

http://docs.celeryproject.org/en/3.1/getting-started/first-steps-with-celery.html#application

bmbouter:
Thanks @goosemania and @dkliban

PR #2860, opened 2016-11-23
Add a suffix attribute to the ProgressReport model.

ipanova:
closes #2319
https://pulp.plan.io/issues/2319

### DIFF
--- a/app/pulp/app/models/progress.py
+++ b/app/pulp/app/models/progress.py
@@ -27,6 +27,8 @@ class ProgressReport(Model):
         total: (models.IntegerField) The total count of items to be handled by the ProgressBar
             (required)
         done (models.IntegerField): The count of items already processed. Defaults to 0.
+        suffix (models.TextField): Customizable suffix rendered with the progress report
+            See `docs https://github.com/verigak/progress>`_. for more info.
 
     Relations:
 
@@ -54,6 +56,7 @@ class ProgressReport(Model):
     done = models.IntegerField(default=0)
 
     task = models.ForeignKey("Task", on_delete=models.CASCADE)
+    suffix = models.TextField(default='')
 
     def save(self, *args, **kwargs):
         """
--- a/app/pulp/app/serializers/progress.py
+++ b/app/pulp/app/serializers/progress.py
@@ -30,6 +30,11 @@ class ProgressReportSerializer(ModelSerializer):
         read_only=True
     )
 
+    suffix = serializers.CharField(
+        help_text=_("The suffix to be shown with the progress report."),
+        read_only=True
+    )
+
     task = serializers.HyperlinkedRelatedField(
         help_text=_("The task associated with this progress report."),
         read_only=True,
@@ -41,4 +46,4 @@ class ProgressReportSerializer(ModelSerializer):
         # this serializer is meant to be nested inside Task serializer,
         # so it will not have its own endpoint, that's why
         # we need to explicitly define fields to exclude '_href' field.
-        fields = ('message', 'state', 'total', 'done', 'task')
+        fields = ('message', 'state', 'total', 'done', 'suffix', 'task')


ipanova:
@bmbouter i did the changes the way it was described on the task, but i disagree with them, maybe i lack some piece of knowledge.
the suffix attribute is used just for ProgressBar, so theoretically ProgressReport should not contain it because the subclassed ProgressSpinner would inherit it, and the spinner does not have any kind of suffix,
instead it has phases.

If adding the suffix field just to the ProgressBar model, it would stop to be a proxy model.
I checked progress library code a bit, there is also possibility to change the filling of the progress bar, maybe we would like to add also a 'fill' attribute?

bmbouter:
@ipanova I see what you are saying, but we need to keep the Proxy model pattern. If we move away from that then `ProgressBar` and `ProgressSpinner` will be in different tables in postgresql. Having them be proxy models allows us to have different interfaces on the classes but still store them on 1 table.

Also note that combining `ProgressBar` and `ProgressSpinner` is not something we should do either otherwise they will end up with similarly inappropriate issues in terms of the interface.

Something to think about though is that just because the progress library doesn't support suffix on Spinner today doesn't mean it always won't. That could be a feature request that could be filed against their project.

One concern I have about this suffix idea altogether is that it will be difficult for my plans to aggregate progress bars and spinners, but I think we can figure something out in that area.

bmbouter:
@ipanova Oh also I don't think we should offer fill at all. I think we want a consistent visual presentation for both `ProgressBar` and `ProgressSpinner` and that should not be up to a plugin writer to choose.

ipanova:
@bmbouter i understand the benefits of proxy model pattern and i agree that we should keep it
Thanks for other feedback provided, that makes sense to me.

app/pulp/app/models/progress.py
```diff
@@ -27,6 +27,8 @@ class ProgressReport(Model):
         total: (models.IntegerField) The total count of items to be handled by the ProgressBar
             (required)
         done (models.IntegerField): The count of items already processed. Defaults to 0.
+        suffix (models.TextField): Customazible way of how progress report will be rendered
+            More info can be found here https://github.com/verigak/progress
```
bmbouter: Rewriting to active voice:    `See https://github.com/verigak/progress for more info.`

bmbouter: Also this docstring will be rendered by Sphinx. Can the url become an actual url with rst syntax so that the user can click on it in the docs.

app/pulp/app/models/progress.py
```diff
@@ -54,6 +56,7 @@ class ProgressReport(Model):
     done = models.IntegerField(default=0)
 
     task = models.ForeignKey("Task", on_delete=models.CASCADE)
+    suffix = models.TextField(default=None)
```
bmbouter: What about a `null=True` instead of `default=None`? As its written now the DB will store `None` in the db unless the plugin writer sets the `suffix`.

app/pulp/app/serializers/progress.py
```diff
@@ -30,6 +30,11 @@ class ProgressReportSerializer(ModelSerializer):
         read_only=True
     )
 
+    suffix = serializers.CharField(
+        help_text=_("Suffix that will render the progress report."),
```
bmbouter: The suffix doesn't do the rendering. Consider replacing with this:  `The suffix to be shown with the progress report`.

bmbouter: Good i8ln.  👍

app/pulp/app/models/progress.py
ipanova: i've read on django docs that it is not recommended to use null=true for string based fields https://docs.djangoproject.com/en/1.10/ref/models/fields/#null
maybe @seandst can provide more info here?

bmbouter: @ipanova oh that is right. In this case I think having it default to empty string would be the best. Since string are immutable a default of `default=''` would work fine.

bmbouter: @ipanova Thank you for re-raising this.

PR #2862, opened 2016-11-23
Workers can run and execute tasks

mhrivnak:
Thanks to ttereshc aka @goosemania who contributed some of this code.

There is a known issue that resource-reserving tasks still cannot run, but
there is a fix in the works that will follow this PR.

Fixes bugs in pulp.tasking
Rearranges pulp.tasking to put related code together
Adds two tasks (importer and publisher delete) in pulp.app.tasks
Fixes import order problems of django app vs. celery app

re #2440
re #2400
re #2408


app/pulp/app/tasks/__init__.py
```diff
@@ -0,0 +1,2 @@
+from pulp.app.tasks import repository  # noqa
```
goosemania: Is there any particular reason to import task modules here and no the tasks themselves? No to put all tasks in the same namespace?

tasking/pulp/tasking/celery_app.py
```diff
@@ -69,6 +66,10 @@ def initialize_worker(sender, instance, **kwargs):
     :param kwargs:   Other params (unused)
     :type  kwargs:   dict
     """
+    from pulp.tasking.services.worker_watcher import delete_worker
+    from pulp.tasking.services import storage
+    from pulp.tasking.constants import TASKING_CONSTANTS
```
goosemania: It is not necessary to import `constants` and `storage` inside callables. There should be no harm if they will be in their usual place because we import `pulp.tasking` anyway: `from pulp.tasking.celery_instance import celery`

tasking/pulp/tasking/celery_app.py
```diff
@@ -92,6 +93,8 @@ def get_resource_manager_lock(name):
     :param name:   The hostname of the worker
     :type  name:   basestring
     """
+    from pulp.app.models.task import TaskLock, Worker
+    from pulp.tasking.constants import TASKING_CONSTANTS
```
goosemania: It is not necessary to import `constants` inside callables.

tasking/pulp/tasking/services/scheduler.py
```diff
@@ -9,15 +10,20 @@
 from django.utils import timezone
 from django.db import IntegrityError
 
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
+
+# This import is here so that Celery will find our application instance. It's important that other
+# pulp and django code not get used until after the celery app is instantiated and does its "fixup"
+# of django.
```
goosemania: I do not think that order of the imports is important in scheduler.
To start celerybeat apart form scheduler you specify `--app=pulp.tasking.celery_app:celery` and it will be handled first. As for any other workers the first thing celery will do is import this app module and start worker. All the fixups will happen then. So at the time of scheduler import, everything will be ready and imports can go here in any order. If it was not the case you would not able to import models outside of callables here. That's my understanding.

app/pulp/app/tasks/repository.py
```diff
@@ -0,0 +1,22 @@
+from celery import shared_task
+
+from pulp.app import models
+from pulp.tasking.tasks import UserFacingTask
+
+
+@shared_task(base=UserFacingTask)
+def delete_publisher(repo_name, publisher_name):
+    models.Publisher.objects.filter(name=publisher_name, repository__name=repo_name).delete()
```
goosemania: Maybe it makes sense to remove the implementation of both of these tasks from pulp2 codebase in this 3.0-dev branch. 
And also we should not forget about cleanup by plugins which happened in these tasks in pulp 2.

app/pulp/app/tasks/__init__.py
mhrivnak: This could go multiple ways, and I'm not tied to this one. Here's how I settled on this one to try: If we import every single task, each of which is just a function, into the same namespace, then the function names will have to contain additional namespace to avoid collision. For example, ``delete_repository``, ``delete_importer``, ``delete_*``.

Would we do ``from repository import delete_repository``? That already looks slightly silly.

That would mean having a function at ``pulp.app.tasks.repository.delete_repository``. Why not just name it ``pulp.app.tasks.repository.delete`` and import the whole ``repository`` module?

That said, it's just semantics and naming, with minor changes. I'm not sure it matters a lot in any case.

Also, while typing this I've realized that the tasks in this PR should be moved to ``importer`` and ``publisher`` modules.

What do you think?

tasking/pulp/tasking/services/scheduler.py
```diff
@@ -9,15 +10,20 @@
 from django.utils import timezone
 from django.db import IntegrityError
 
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", 'pulp.app.settings')
```
mhrivnak: Removed.

tasking/pulp/tasking/services/scheduler.py
mhrivnak: That makes sense! I've changed the order to the usual convention and everything still works.

tasking/pulp/tasking/celery_app.py
mhrivnak: Moved.

app/pulp/app/tasks/repository.py
mhrivnak: Will do!

app/pulp/app/tasks/__init__.py
goosemania: That sounds really good :+1:

app/pulp/app/tasks/publisher.py
```diff
@@ -0,0 +1,17 @@
...
+from pulp.app import models
+from pulp.tasking.tasks import UserFacingTask
+
+
+@shared_task(base=UserFacingTask)
+def delete(repo_name, publisher_name):
+    """
+    Delete a publisher.
+
+    :param repo_name:       the name of a repository
```
bmbouter: This should be converted to use the new Google style docstring

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,29 @@ class ReservedResource(Model):
...
+        """
+        Find an unreserved Worker
+
+        Return the Worker instance that has no ReservedResource associated with it. If all workers have
+        ReservedResource relationships a Worker.DoesNotExist pulp.app.model.Worker exception is
+        raised.
+
+        This function also provides randomization for worker selection.
+
+        :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with them.
```
bmbouter: This should be converted to use the new Google style docstring

tasking/pulp/tasking/celery_app.py
```diff
@@ -14,15 +14,14 @@
 from celery.signals import celeryd_after_setup
 from django.db.utils import IntegrityError
 
-from pulp.app.models.task import TaskLock, Worker
-from pulp.tasking import delete_worker, storage
-from pulp.tasking.constants import TASKING_CONSTANTS
-
-# This import is here so that Celery will find our application instance
+# This import is here so that Celery will find our application instance. It's important that other
+# pulp and django code not get used until after the celery app is instantiated and does its "fixup"
```
bmbouter: s/pulp/Pulp/
s/django/Django/
s/celery/Celery/

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,42 @@ def handle_worker_offline(event):
...
+    Delete the Worker with name from the database, cancel any associated tasks and reservations
+
+    If the worker shutdown normally, no message is logged, otherwise an error level message is
+    logged. Default is to assume the worker did not shut down normally.
+
+    Any resource reservations associated with this worker are cleaned up by this function.
+
+    Any tasks associated with this worker are explicitly canceled.
+
+    :param name:            The name of the worker you wish to delete.
```
bmbouter: This should be converted to use the new Google style docstring

bmbouter: This should be converted to use the new Google style docstring

server/usr/lib/systemd/system/pulp_celerybeat.service
```diff
@@ -7,7 +7,7 @@ Wants=network-online.target
 EnvironmentFile=/etc/default/pulp_celerybeat
 User=apache
 WorkingDirectory=/var/run/pulp/
-ExecStart=/usr/bin/celery beat --app=pulp.tasking.celery_instance.celery --scheduler=pulp.tasking.scheduler.Scheduler
+ExecStart=/usr/bin/celery beat --app=pulp.tasking.celery_app:celery --scheduler=pulp.tasking.services.scheduler.Scheduler
```
bmbouter: 👍

server/usr/lib/systemd/system/pulp_resource_manager.service
```diff
@@ -7,7 +7,7 @@ Wants=network-online.target
 EnvironmentFile=/etc/default/pulp_resource_manager
 User=apache
 WorkingDirectory=/var/run/pulp/
-ExecStart=/usr/bin/celery worker -A pulp.tasking.celery_app -n resource_manager@%%h\
+ExecStart=/usr/bin/celery worker -A pulp.tasking.celery_app:celery -n resource_manager@%%h\
```
bmbouter: 👍

tasking/pulp/tasking/celery_app.py
```diff
@@ -14,15 +14,14 @@
...
+# of django.
 from pulp.tasking.celery_instance import celery  # noqa
+from pulp.tasking.constants import TASKING_CONSTANTS
+from pulp.tasking.services import storage
 
-# This import is here so Celery will discover all tasks
-import pulp.tasking.registry  # noqa
+celery.autodiscover_tasks()
```
bmbouter: This should get another line of whitespace above it separating it from imports

tasking/pulp/tasking/services/scheduler.py
```diff
@@ -10,14 +10,13 @@
...
-from pulp.tasking.celery_instance import celery as app
+from pulp.tasking.celery_app import celery as app
 from pulp.tasking.constants import TASKING_CONSTANTS as constants
-from pulp.tasking import delete_worker
+from pulp.tasking.services import worker_watcher
 
 
 # The import below is not used in this module, but it needs to be kept here. This module is the
-# first and only Pulp module to be imported by celerybeat, and by importing pulp.server.logs, it
+# first and only Pulp module to be imported by celerybeat, and by importing pulp.app.logs, it
```
bmbouter: I suspect this comment and the import on L#21 is probably out of date and can likely be removed along with the [`pulp.app.logs`](https://github.com/pulp/pulp/blob/3.0-dev/app/pulp/app/logs.py).

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
     worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
 
 
+class WorkerManager(models.Manager):
+    def get_unreserved_worker(self):
+        """
+        Find an unreserved Worker
```
bmbouter: Can the string 'Worker' be updated with the sphinx class path so it links to the actual `Worker` module?

bmbouter: This is the only reference I see to that module altogether:

```
[bmbouter@localhost pulp]$ git grep pulp.app.logs
tasking/pulp/tasking/services/scheduler.py:# first and only Pulp module to be imported by celerybeat, and by importing pulp.app.logs, it
tasking/pulp/tasking/services/scheduler.py:import pulp.app.logs  # noqa
```

bmbouter:
Can all of these commits be squashed?

bmbouter:
Do you want to remove the 'Work in Progress' label and also edit the title to remove [WIP]?

tasking/pulp/tasking/services/scheduler.py
bmbouter: This is the only reference I see to that module altogether:

```
[bmbouter@localhost pulp]$ git grep pulp.app.logs
tasking/pulp/tasking/services/scheduler.py:# first and only Pulp module to be imported by celerybeat, and by importing pulp.app.logs, it
tasking/pulp/tasking/services/scheduler.py:import pulp.app.logs  # noqa
```

app/pulp/app/models/task.py
bmbouter: Something like this for example:  https://github.com/bmbouter/pulp/blob/e806ab8a11a10fd1bbd9c7475f3187f0c8157d43/app/pulp/app/serializers/base.py#L86-L86

bmbouter: This shows some of the other options like `:attr:` for example

http://www.sphinx-doc.org/en/1.4.9/domains.html#directive-py:class

mhrivnak:
Yes and Yes.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
     worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
 
 
+class WorkerManager(models.Manager):
+    def get_unreserved_worker(self):
+        """
+        Find an unreserved :class:`pulp.app.models.Worker`
+
+        Return the Worker instance that has no :class:`pulp.app.models.ReservedResource`
```
bmbouter: Can Worker receive the linking also.

Note that you can use a ~ to write the full class path, but have Sphinx only show the classname itself.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
...
+        associated with it. If all workers have ReservedResource relationships a,
```
bmbouter: Can ReservedResource receive the docs linking also?

tasking/pulp/tasking/services/scheduler.py
bmbouter: I [commented](https://pulp.plan.io/issues/2275#note-11) about this on the logging PR so these comments can be ignored.

tasking/pulp/tasking/tasks.py
```diff
@@ -54,10 +52,10 @@ def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
 
     :param name:          The name of the task to be called
     :type name:           basestring
-    :param inner_task_id: The UUID to be set on the task being called. By providing
+    :param task_id:       The UUID to be set on the task being called. By providing
```
bmbouter: Oh you are fixing a consistency problem, but can the parameter itself be renamed from `task_id` -> `inner_task_id`. It isn't actually the task_id of the _queue reserve resource itself.

bmbouter:
Can the commit message be improved to include a summary of changes, a link to the issue, and a `re #` syntax so it will link to the issue. I think a `re` is right because my additional PR is likely also needed and will need to be associated also before it can go to MODIFIED.

mhrivnak:
Yes, I will squash, write a real log message, etc. before removing the WIP status.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
     worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
 
 
+class WorkerManager(models.Manager):
+    def get_unreserved_worker(self):
+        """
+        Find an unreserved :class:`~pulp.app.models.Worker`
+
+        Return the Worker instance that has no :class:`~pulp.app.models.ReservedResource`
+        associated with it. If all workers have ReservedResource relationships a,
```
bmbouter: I think the comma should be moved to be after the word 'relationships'.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
     worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
 
 
+class WorkerManager(models.Manager):
+    def get_unreserved_worker(self):
+        """
+        Find an unreserved :class:`~pulp.app.models.Worker`
```
bmbouter: Can the idea that randomness is used be introduced in this 1-line description?

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
...
+
+        Return the Worker instance that has no :class:`~pulp.app.models.ReservedResource`
+        associated with it. If all workers have ReservedResource relationships a,
+        :class:`pulp.app.models.Worker.DoesNotExist` exception is raised.
+
+        This function also provides randomization for worker selection.
```
bmbouter: s/function/method/

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
     worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
 
 
+class WorkerManager(models.Manager):
+    def get_unreserved_worker(self):
```
bmbouter: not a standard style request, but consider adding a line of whitespace between the class definition and the first method.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,30 @@ class ReservedResource(Model):
...
+        Return the Worker instance that has no :class:`~pulp.app.models.ReservedResource`
+        associated with it. If all workers have ReservedResource relationships a,
+        :class:`pulp.app.models.Worker.DoesNotExist` exception is raised.
+
+        This function also provides randomization for worker selection.
+
+        :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with
+                                     them.
+
+        :returns:          The Worker instance that has no reserved_resource
```
bmbouter: s/reserved_resource/ReservedResource/
Also consider replacing with `A random Worker instance that has no associated ReservedReservation entries`.

app/pulp/app/models/task.py
```diff
@@ -35,6 +35,31 @@ class ReservedResource(Model):
...
+        Return the Worker instance that has no :class:`~pulp.app.models.ReservedResource`
+        associated with it. If all workers have ReservedResource relationships, a
+        :class:`pulp.app.models.Worker.DoesNotExist` exception is raised.
+
+        This method also provides randomization for worker selection.
+
+        :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with
+                                     them.
+
+        :returns:          a randomly-selected Worker instance that has no ReservedResource
```
bmbouter: s/a/A/

tasking/pulp/tasking/celery_app.py
```diff
@@ -69,6 +69,8 @@ def initialize_worker(sender, instance, **kwargs):
     :param kwargs:   Other params (unused)
     :type  kwargs:   dict
     """
+    from pulp.tasking.services.worker_watcher import delete_worker
```
bmbouter: I really want to avoid this, but I'm not sure how. Any ideas?

tasking/pulp/tasking/celery_app.py
```diff
@@ -92,6 +94,7 @@ def get_resource_manager_lock(name):
     :param name:   The hostname of the worker
     :type  name:   basestring
     """
+    from pulp.app.models.task import TaskLock, Worker
```
bmbouter: Can we do this runtime lookup using the Django machinery. Can we try this instead?

```
from django.db.models import get_app
pulp_app =  get_app('pulp_app')
```

Then you could use `pulp_app.models.TaskLock` and `pulp_app.models.Worker`

I haven't tested this but I think it would work.

bmbouter: Another possible option would be to shorten an import at the module level to `from pulp import app as pulp_app` and refer to the modules in this method the same way as identified ^.

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
     msg = _("Worker '%(worker_name)s' shutdown") % event_info
     _logger.info(msg)
     delete_worker(event_info['worker_name'], normal_shutdown=True)
+
+
+def delete_worker(name, normal_shutdown=False):
+    """
+    Delete the :class:`~pulp.app.models.Worker` with name from the database, cancel any associated
```
bmbouter: I believe autodoc expects the short description to be 1 line. I relies on that because that is what the docstring pep requires. Can this be shortened to 1 line < 100 characters somehow?

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
...
+    tasks and reservations
+
+    If the worker shutdown normally, no message is logged, otherwise an error level message is
+    logged. Default is to assume the worker did not shut down normally.
```
bmbouter: The second sentence is a fragment.

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
...
     _logger.info(msg)
     delete_worker(event_info['worker_name'], normal_shutdown=True)
+
+
+def delete_worker(name, normal_shutdown=False):
+    """
+    Delete the :class:`~pulp.app.models.Worker` with name from the database, cancel any associated
+    tasks and reservations
+
+    If the worker shutdown normally, no message is logged, otherwise an error level message is
```
bmbouter: I think the comma after logged should either be a semicolon or split into two separate sentences.

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
...
+    logged. Default is to assume the worker did not shut down normally.
+
+    Any resource reservations associated with this worker are cleaned up by this function.
```
bmbouter: Does it make sense to s/resource reservations/ReservedResources/ ?

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
...
+
+    Any tasks associated with this worker are explicitly canceled.
```
bmbouter: Does it make sense to s/tasks/Task/ to refer to the model more explicitly?

tasking/pulp/tasking/services/worker_watcher.py
```diff
@@ -100,3 +102,43 @@ def handle_worker_offline(event):
...
+
+    :param name:            The name of the worker you wish to delete.
```
bmbouter: s/you wish//

tasking/pulp/tasking/tasks.py
```diff
@@ -1,23 +1,21 @@
...
 from celery.result import AsyncResult
 from django.db import transaction
-from django.db.models import Count
 
 from pulp.app.models import ReservedResource, Task as TaskStatus, TaskLock, Worker
 from pulp.common import TASK_FINAL_STATES, TASK_INCOMPLETE_STATES, TASK_STATES
 from pulp.exceptions import MissingResource, PulpException
-from pulp.tasking import storage
+from pulp.tasking import util
+from pulp.tasking.services import storage
```
bmbouter: This should be moved down so it comes after `pulp.tasking.celery_instance`

tasking/pulp/tasking/tasks.py
```diff
@@ -40,7 +38,7 @@ class PulpTask(CeleryTask):
 
 
 @task(base=PulpTask, acks_late=True)
-def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
+def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwargs):
```
bmbouter: thank you

tasking/pulp/tasking/tasks.py
```diff
@@ -75,7 +73,7 @@ def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
             break
 
         try:
-            worker = _get_unreserved_worker()
+            worker = Worker.objects.get_unreserved_worker()
```
bmbouter: 👍

tasking/pulp/tasking/util.py
```diff
@@ -0,0 +1,56 @@
...
+from pulp.exceptions import MissingResource, PulpException
+
+
+celery_controller = control.Control(app=current_app)
+_logger = logging.getLogger(__name__)
+
+
+def cancel(task_id):
+    """
+    Cancel the task that is represented by the given task_id. This method cancels only the task
```
bmbouter: Can a 1-line description be added above this. I think autodoc will do better with that.

bmbouter: I know you didn't write this (I probably did) but anyway

tasking/pulp/tasking/util.py
```diff
@@ -0,0 +1,56 @@
...
+    with given task_id, not the spawned tasks. This also updates task's state to 'canceled'.
```
bmbouter: s/not the/not/

tasking/pulp/tasking/util.py
```diff
@@ -0,0 +1,56 @@
...
+    task_status.save()
+
+    msg = _('Task canceled: %(task_id)s.')
+    msg = msg % {'task_id': task_id}
+    _logger.info(msg)
+
+
+def get_current_task_id():
+    """"
+    Get the current task id from celery. If this is called outside of a running
```
bmbouter: Can a 1-line description be added above here. I know you didn't write this originally.

tasking/pulp/tasking/celery_app.py
goosemania: It is so ^ because `delete_worker` imports models and also it imports `cancel` from `util` which also imports models.

tasking/pulp/tasking/util.py
```diff
@@ -0,0 +1,56 @@
...
+    celery task it will return None
```
bmbouter: s/celery/Celery/

bmbouter: This sentence needs a period.

tasking/pulp/tasking/util.py
```diff
@@ -0,0 +1,56 @@
...
+
+    :return: The ID of the currently running celery task or None if not in a task
```
bmbouter: s/ID/UUID/  ?

bmbouter: s/celery/Celery/
s/None/None,/

tasking/pulp/tasking/celery_app.py
bmbouter: Yeah I don't think we can avoid this easily at this time

mhrivnak: I suggest we go with this for now, and can try to improve it later if we find a better approach.

tasking/pulp/tasking/celery_app.py
mhrivnak: Interesting tidbit:

```
./manage.py:1: RemovedInDjango19Warning: django.db.models.get_app is deprecated.
```

goosemania: I guess it is expected to use `app.get_model(app_label, model)`:
```
from django.apps import apps
Task = apps.get_model('pulp_app', 'Task')
```

tasking/pulp/tasking/celery_app.py
goosemania: Is it worth creating a refactor task in redmine? (I can do it if needed) Or we are going to look for a better approach only in case of some problems so it will be resolved in some real issue?

tasking/pulp/tasking/celery_app.py
mhrivnak: I think we could do this:

```
from django.apps import apps  # yes you read that right

<other code here>

def get_resource_manager_lock(name):
    TaskLock = apps.get_app_config('pulp_app').models['tasklock']
```

This avoids doing an import, but I'm not sure it's an improvement. It requires knowing these two strings, which isn't very likable. The import seems simpler and more obvious to understand.

mhrivnak:
Thanks for all the feedback. I fixed a ton of doc block items, but I'd like to leave the rest for another time since I didn't write them (I just moved them). My feeling is that fixing so many things about these docs that I didn't write is becoming a distraction and pulling my focus away from the work I did do with this PR.

For future review, I suggest you consider saving such requests for a follow-up effort. A small number of changes while in the neighborhood can be convenient and productive, but it can quickly overwhelm the original goals of the PR.

tasking/pulp/tasking/celery_app.py
bmbouter: I don't think it's significant enough to have an issue tracking it

tasking/pulp/tasking/celery_app.py
bmbouter: I agree, this seems more fragile than just the 1 line import. I'm ok if there are no changes in this area.

bmbouter:
Thanks @mhrivnak this looks good. I can imagine how a lot of comments can be overwhelming, so I'll think more about that. The problem with follow-up efforts of this size is that they rarely get done. Anyway, a lot of the docblock improvements did get done, so I appreciate that. LGTM. 🎆 🎈

PR #2866, opened 2016-11-28
Remove sn.dat for serial number storage

werwty:
closes #1839
https://pulp.plan.io/issues/1839

### DIFF
--- a/docs/user-guide/release-notes/2.13.x.rst
+++ b/docs/user-guide/release-notes/2.13.x.rst
@@ -11,9 +11,13 @@ New Features
 * Log statements which are emitted from task processes now include a short task
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+
 * Users no longer need to have pulp_celerybeat running to know whether pulp_workers or
   pulp_resource_manager are online. All workers now write their own heartbeat records to the
   database, eliminating the dependence on pulp_celerybeat.
+
+* serial_number_path is removed from server.conf. Pulp now generates the serial number
+  randomly and does not use the sn.dat file. This is a part of the fix for :redmine:`1839`
 
 Bug Fixes
 ---------
--- a/pulp.spec
+++ b/pulp.spec
@@ -492,6 +492,9 @@ chown root:apache $KEY_PATH
 chown root:apache $KEY_PATH_PUB
 ln -fs $KEY_PATH_PUB %{_var}/lib/%{name}/static
 
+# Remove old serial number file
+rm -f /var/lib/pulp/sn.dat
+
 # CA certificate
 if [ $1 -eq 1 ]; # not an upgrade
 then
--- a/server/pulp/server/managers/auth/cert/cert_generator.py
+++ b/server/pulp/server/managers/auth/cert/cert_generator.py
@@ -1,6 +1,6 @@
-from threading import RLock
 import logging
 import subprocess
+import uuid
 
 from M2Crypto import X509, EVP, RSA, util
 
@@ -21,11 +21,11 @@ class CertGenerationManager(object):
         """
         Generates a x509 certificate for an admin user.
 
-        @param user: identification the certificate will be created for; may not be None
-        @type  user: pulp.server.db.model.User
+        :param user: identification the certificate will be created for; may not be None
+        :type  user: pulp.server.db.model.User
 
-        @return: tuple of PEM encoded private key and certificate
-        @rtype:  (str, str)
+        :return: tuple of PEM encoded private key and certificate
+        :rtype:  (str, str)
         """
         expiration = config.config.getint('security', 'user_cert_expiration')
         return self.make_cert(self.encode_admin_user(user), expiration)
@@ -35,15 +35,15 @@ class CertGenerationManager(object):
         Generate an x509 certificate with the Subject set to the cn passed into this method:
         Subject: CN=someconsumer.example.com
 
-        @param cn: ID to be embedded in the certificate
-        @type  cn: string
+        :param cn: ID to be embedded in the certificate
+        :type  cn: string
 
-        @param uid: The optional userid.  In pulp, this is the DB document _id
+        :param uid: The optional userid.  In pulp, this is the DB document _id
             for both users and consumers.
-        @type uid: str
+        :type uid: str
 
-        @return: tuple of PEM encoded private key and certificate
-        @rtype:  (str, str)
+        :return: tuple of PEM encoded private key and certificate
+        :rtype:  (str, str)
         """
         # Ensure we are dealing with a string and not unicode
         try:
@@ -72,7 +72,7 @@ class CertGenerationManager(object):
         ca_key = config.config.get('security', 'cakey')
 
         sn = SerialNumber()
-        serial = sn.next()
+        serial = sn.getSerialNumber()
 
         cmd = 'openssl x509 -req -sha1 -CA %s -CAkey %s -set_serial %s -days %d' % \
               (ca_cert, ca_key, serial, expiration)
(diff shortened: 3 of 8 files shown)


bmbouter:
@werwty This PR is opened against 2.7-dev, but it needs to be against the current bugfix z-stream which is 2.10-dev.

server/pulp/server/managers/auth/cert/cert_generator.py
```diff
@@ -161,42 +162,20 @@ def decode_admin_user(self, encoded_string):
...
-            fp.write('0')
-            fp.close()
-        finally:
-            self.__mutex.release()
+        # Generate a random 32 bit number
+        random_sn = str(getrandbits(32))
+        # Current utc time
+        curr_time = str(int(time.time()))
+
+        return curr_time + random_sn
```
seandst: I don't know of any size restrictions of serial length, and also don't see how concatenation of a random number with time.time is likely to be any better at avoiding a collision than the random number alone. I also don't see why there was a need to change the return type to string. Any reason not to `return int(uuid.uuid4())` here?

seandst: A comment explaining why it's okay for a method named "serial" to return something that is not a monotonically increasing integer would probably be helpful.

werwty: The size restriction of a sn is 20 octects. I was trying to keep the best of both worlds by using a timestamp + rand, so sn would be increasing. But the only requirement on a sn is uniqueness, not that it has to be increasing. I can see your point with the uuid.uuid4 being more elegant (and just as non-prone to collision, and only takes one import). I will update with comment and changes

werwty:
ok test

bmbouter:
Oh also, we should clean up the file that we created. The rpm installation has a way to detect if it's a fresh installation or an upgrade and from what version the previous one was. There could be a conditional statement that would remove it there. The removal of that file and the server.conf could become a separate issue if that is more than you want to do now. I +1 doing it now, but that is up to you.

werwty:
@bmbouter I don't think we need to worry about manually removing /var/lib/pulp/sn.dat 
It looks like that the rpm upgrade will take care of the old file removal for us: http://rpm5.org/docs/rpm-guide.html#id2980009 (I also confirmed this on #fedora-devel just to be sure)
As far as the server.conf goes I added a deprecated message. I assumed that since there's no functional changes this should be ok in a z-release. Please let me know if this is incorrect.

bmbouter:
I think it would autodelete if that file was managed by rpm, but I think it's created by the Python code at runtime. I confirmed this by grabbing one of the recent pulp-server rpms and running: `rpm -qpl pulp-server-2.11.1-1.fc24.noarch.rpm` which does not list `/var/lib/pulp/sn.dat`. Will that still automatically remove even if Pulp never put the file there? Maybe it will, and I just don't know.

In terms of y versus z stream for the file removal, generally the reason is that z-stream upgrades are expected to have no manual user intervention required. If a user has edited that file (the reporter has) they will have to manually merge their server.conf.new into their existing server.conf. That is the reasoning behind the y-stream request. We've never written this "no manual intervention required for z-stream upgrades" policy down so that is an area we could also improve.

Thank you for looking into these things. The code you've written looks really great.

werwty:
@bmbouter Ah you're right. sn.dat isn't managed by rpm. I will add the remove line. 

As far as the server.conf goes, you have a good point. I will open a PR to master to update it. Also what do you think about the deprecation message vs straight up removal?

bmbouter:
Since it was visible to the user via a conf file. +1 to a release note indicating that it's been removed as part of a bugfix. I don't think it's a deprecation since that means the thing is still currently functional but may not in the future.

seandst:
@bmbouter, so you're saying you wouldn't mind seeing this change in 2.12.1, or are you thinking it's more appropriate for 2.13.0?

bmbouter:
Since the server.conf modification was pulled out, this can go into 2.12-dev to be included in 2.12.1. However, I expected the `rm -f` to go with the server.conf modification. For simplicity doing all of the changes against master would just make this easier. I recommend doing that. That way we are sure it all gets done and the release notes go out with the changes.

bmbouter:
Also for cherry picking purposes, having a single commit would be easier.

seandst:
Yeah, I think that makes perfect sense. So we should get this down to one commit and retarget the PR to merge to master.

pulp.spec
```diff
@@ -492,6 +492,9 @@ chown root:apache $KEY_PATH
 chown root:apache $KEY_PATH_PUB
 ln -fs $KEY_PATH_PUB %{_var}/lib/%{name}/static
 
+# Remove old serial number file
+rm -f /var/lib/pulp/sn.dat
```
beav: will this fail if sn.dat doesn't already exist?

werwty: nope
`-f, --force ignore nonexistent files and arguments, never prompt`

docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -11,10 +11,15 @@ New Features
 * Log statements which are emitted from task processes now include a short task
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+
 * Users no longer need to have pulp_celerybeat running to know whether pulp_workers or
   pulp_resource_manager are online. All workers now write their own heartbeat records to the
   database, eliminating the dependence on pulp_celerybeat.
 
+* serial_number_path removed from server.conf. Pulp now generates the serial number
+  randomly and does not use the sn.dat file. This is a part of the `bugfix for #1839
```
bmbouter: I think using the macros we have would be an improvement. The second for [here](http://docs.pulpproject.org/dev-guide/contributing/documenting.html#editing-the-docs) is equivalent to what you've written.

server/pulp/server/managers/auth/cert/cert_generator.py
```diff
@@ -161,42 +161,19 @@ def decode_admin_user(self, encoded_string):
...
-    PATH = config.config.get('security', 'serial_number_path')
-    __mutex = RLock()
     __metaclass__ = Singleton
 
-    def next(self):
+    def getSerialNumber(self):
         """
-        Get the next serial#
-        @return: The next serial#
+        Returns a uuid to use for the next serial number. x509 standards indicate that serial
```
bmbouter: Since this is a multiline docstrings, can a 1-line summary line be added? PEP 257 talks about the multiline format [here](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

server/pulp/server/managers/auth/cert/cert_generator.py
```diff
@@ -161,42 +161,19 @@ def decode_admin_user(self, encoded_string):
...
+        number must be unique, not that it must be monotonically increasing. By using a uuid
+        the chance of collision is very low.
+
+        @return: A serial number
```
bmbouter: Can these be switched from s/`@`/`:`/ so that we use the newer form?

docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -11,10 +11,15 @@ New Features
 * Log statements which are emitted from task processes now include a short task
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+
 * Users no longer need to have pulp_celerybeat running to know whether pulp_workers or
   pulp_resource_manager are online. All workers now write their own heartbeat records to the
   database, eliminating the dependence on pulp_celerybeat.
 
+* serial_number_path removed from server.conf. Pulp now generates the serial number
```
bmbouter: s/removed/is removed/

werwty:
ok test

PR #2870, opened 2016-11-30
Remove RepositoryGroup model

asmacdo:


asmacdo:
@mhrivnak I wasn't sure what to say beyond the fact that we won't have repo groups in pulp 3. IIRC, the why was just that we didn't need it in the MVP. 

Either way, ready for re-review.

PR #2871, opened 2016-11-30
Adds docs about using ProgressBar with threads

dkliban:


dkliban:
Add test.py[0] to your pulp/app/pulp/app directory and start the django shell. Then run

import test



[0] https://gist.githubusercontent.com/dkliban/bc40c6cb28e5ab3cbedeb56bff6d55b0/raw/28ea46d12c297410f74449bc96cacbb0f00e6279/test.py

bmbouter:
I removed [WIP] from the title since we us the Label now for that

bmbouter:
@dkliban I did use the reproducer and I did reproduce the issue. I experimented with several other options and read some documentation. I've come to the conclusion that what is really important is that all threads use the same instance of ProgressBar and then we don't have to worry about thread race conditions inside of the instance methods.

I can write more on this tomorrow and we can discuss it some too with @mhrivnak who originally raised the thread safety concern. We likely will add some doc lines to the docstring, but I don' think a code change is necessary here.

bmbouter:
What about this?

```
When using threads to update a ProgressBar in parallel, it is recommended all threads share the same in-memory instance. Django does not synchronize in-memory model instances, so multiple instances of a specific ProgressBar will diverge as they are written to in separate threads.
```

app/pulp/app/models/progress.py
```diff
@@ -181,6 +181,10 @@ class ProgressBar(ProgressReport):
 
     ProgressBar objects are associated with a Task and auto-discover and populate the task id when
     saved.
+
+    When using threads to update a ProgressBar in parallel, it is recommended all threads share the
+    same in-memory instance. Django does not synchronize in-memory model instances, so multiple
+    instances of a specific ProgressBar will diverge as they are written to in separate threads.
```
bmbouter: s/in/from/

PR #2872, opened 2016-11-30
Fixes the Pulp tasking system

bmbouter:
This PR introduces three fixes:

- Updates the Celery dedicated queue name to be 'C.dq2'.
  This is a Celery backwards incompatible change so
  Pulp3 is only compatible with Celery 4.0.

- Fixes a bug in the worker selection which was dispatching
  work to non-dedicated workers.

- Resolves a traceback where a TaskStatus was being created
  twice during normal reserved task execution.

re #2440

### DIFF
--- a/app/pulp/app/models/task.py
+++ b/app/pulp/app/models/task.py
@@ -41,23 +41,25 @@ class WorkerManager(models.Manager):
         """
         Randomly selects an unreserved :class:`~pulp.app.models.Worker`
 
-        Return the Worker instance that has no :class:`~pulp.app.models.ReservedResource`
-        associated with it. If all workers have ReservedResource relationships, a
+        Return a random Worker instance that has no :class:`~pulp.app.models.ReservedResource`
+        associated with it. If all workers have at least one ReservedResource relationship, a
         :class:`pulp.app.models.Worker.DoesNotExist` exception is raised.
 
-        This method also provides randomization for worker selection.
+        This method provides randomization for Worker selection to distribute load across workers.
 
-        :raises Worker.DoesNotExist: If all workers have ReservedResource entries associated with
-                                     them.
+        Returns:
+            :class:`pulp.app.models.Worker`: A randomly-selected Worker instance that has zero
+                :class:`~pulp.app.models.ReservedResource` entries associated with it.
 
-        :returns:          A randomly-selected Worker instance that has no ReservedResource
-                           entries associated with it.
-        :rtype:            pulp.app.models.Worker
+        Raises:
+            Worker.DoesNotExist: If all Workers have at least one ReservedResource entry.
         """
-        free_workers_qs = self.annotate(models.Count('reservations')).filter(reservations__count=0)
-        if free_workers_qs.count() == 0:
+        workers_only_qs = self.filter(name__startswith='reserved')
+        workers_only_qs_with_counts = workers_only_qs.annotate(models.Count('reservations'))
+        try:
+            return workers_only_qs_with_counts.filter(reservations__count=0).order_by('?')[0]
+        except IndexError:
             raise self.model.DoesNotExist()
-        return free_workers_qs.order_by('?').first()
 
 
 class Worker(Model):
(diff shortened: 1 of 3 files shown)


app/pulp/app/models/task.py
```diff
@@ -54,10 +54,12 @@ def get_unreserved_worker(self):
                            entries associated with it.
         :rtype:            pulp.app.models.Worker
         """
-        free_workers_qs = self.annotate(models.Count('reservations')).filter(reservations__count=0)
+        workers_only_qs = Worker.objects.filter(name__startswith='reserved')
+        workers_only_qs_with_counts = workers_only_qs.annotate(models.Count('reservations'))
+        free_workers_qs = workers_only_qs_with_counts.filter(reservations__count=0)
         if free_workers_qs.count() == 0:
```
mhrivnak: Since you have me thinking about race conditions... :) is it possible there's one here?

Does ``free_workers_qs`` get fully evaluated here, with results loaded into RAM? Or (I suspect this is the case) does it only get the count from the DB, and does a separate query two lines below.

What if the reservation counts change in the mean time? Or a worker goes away entirely?

Perhaps this could instead follow the classic python try...except pattern, which would also eliminate the need for a ``count()`` call. Something like:

```
try:
  return workers_only_qs_with_counts.filter(reservations__count=0).order_by('?')[0]
except IndexError:
  raise self.model.DoesNotExist()
```

tasking/pulp/tasking/tasks.py
```diff
@@ -309,10 +315,14 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
         storage.delete_working_directory()
 
     def _get_parent_arg(self):
-        """Return a dictionary with the parent set if running inside of a Task"""
+        """Return a dictionary with the 'parent' set if running inside of a PulpUserFacingTask"""
```
mhrivnak: Should this say just ``UserFacingTask`` ?

bmbouter: Oh yes it should, this was the old name a few weeks ago. Good catch. updated with the next push.

bmbouter:
@mhrivnak I pushed a new version. In addition to the two requested changes, I updated the docstring for the Worker manager. Can you re-review?

app/pulp/app/models/task.py
```diff
@@ -41,23 +41,25 @@ def get_unreserved_worker(self):
...
 
-        :returns:          A randomly-selected Worker instance that has no ReservedResource
-                           entries associated with it.
-        :rtype:            pulp.app.models.Worker
+        Raises:
+            Worker.DoesNotExist: If all Workers have at least one ReservedResource entry.
         """
-        free_workers_qs = self.annotate(models.Count('reservations')).filter(reservations__count=0)
-        if free_workers_qs.count() == 0:
+        workers_only_qs = Worker.objects.filter(name__startswith='reserved')
```
mhrivnak: s/Worker.objects/self/

bmbouter: @mhrivnak good catch! this was a holdover from the pre-manager implementation. fixed with next push.

bmbouter:
@mhrivnak New changes are pushed. I also tested both `apply_async` and `apply_async_with_reservation` for `UserFacingTask` to verify everything is working. It all worked for me. Please re-review when you get a chance.

tasking/pulp/tasking/tasks.py
```diff
@@ -235,10 +235,16 @@ def apply_async(self, *args, **kwargs):
...
-            task_status = TaskStatus.objects.create(pk=async_result.id, state=TaskStatus.WAITING,
-                                                    group=group_id, **parent_arg)
-            for tag in tag_list:
-                task_status.tags.create(name=tag)
+            try:
+                task_status = TaskStatus.objects.create(pk=async_result.id,
+                                                        state=TaskStatus.WAITING,
+                                                        group=group_id, **parent_arg)
+            except IntegrityError:
+                # The TaskStatus was already created with the call to apply_async_with_reservation
```
goosemania: :+1:

bmbouter:
@mhrivnak @goosemania Thank you both for the reviews and testing!

PR #2874, opened 2016-12-05
Update pulp-manage-db to handle default answer, closes #2467

elyezer:
Default to no when any input is given for the continue question when running
pulp-manage-db.

I did some testing and the new condition works:

```python
>>> not len(a) == 0 and a[0] == 'y'
False
>>> a='n'; not len(a) == 0 and a[0] == 'y'
False
>>> a=''; not len(a) == 0 and a[0] == 'y'
False
>>> a='y'; not len(a) == 0 and a[0] == 'y'
True
>>> a='    '; not len(a) == 0 and a[0] == 'y'
False
```

Fixes https://pulp.plan.io/issues/2467

### DIFF
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -294,4 +294,7 @@ def _start_logging():
 
 def _user_input_continue(question):
     reply = str(raw_input(_(question + ' (y/N): '))).lower().strip()
-    return reply[0] == 'y'
+    try:
+        return reply[0] == 'y'
+    except IndexError:
+        return False


elyezer:
I wanted to write tests for the function but I was not sure where to add them. Let me know and I will be happy to add some unit tests in order to ensure all it is going to work as expected.

bmbouter:
@elyezer If this PR came with no tests that would be fine with me. Also can you paste a link to this PR on the issue and transition the issue to POST which is the correct state for issues which have a PR associated with them.

bmbouter:
@elyezer Also can you update the commit to use the `closes` syntax which will associate the commit with the issue in Redmine when merged. See the new commit messages section for details.

elyezer:
>@elyezer Also can you update the commit to use the closes syntax which will associate the commit with the issue in Redmine when merged. See the new commit messages section for details.

Definitely, I was expecting for a comment like this. Thanks for pointing me out to the proper documentation.

server/pulp/server/db/manage.py
```diff
@@ -294,4 +294,4 @@ def _start_logging():
 
 def _user_input_continue(question):
     reply = str(raw_input(_(question + ' (y/N): '))).lower().strip()
-    return reply[0] == 'y'
+    return not len(reply) == 0 and reply[0] == 'y'
```
bmbouter: I think this would be better as:

```
try:
    return reply[0] == 'y'
except IndexError:
    return False
```

elyezer: I thought using conditions would be much faster than try/catch but that is not the case:

```python
In [7]: def test1(s):
   ...:     try:
   ...:         return s[0] == 'y'
   ...:     except IndexError:
   ...:         return False
   ...:     

In [8]: test1('')
Out[8]: False

In [9]: test1('   ')
Out[9]: False

In [10]: def test2(s):
    ...:     return not len(s) == 0 and s[0] == 'y'
    ...: 

In [14]: %timeit "test1('')"
100000000 loops, best of 3: 7.81 ns per loop

In [15]: %timeit "test2('')"
100000000 loops, best of 3: 7.75 ns per loop
```

I will update the code.

elyezer:
@bmbouter updated the PR.

werwty:
Since we removed the feature in https://github.com/pulp/pulp/pull/2883 I am closing this PR.

PR #2875, opened 2016-12-05
Put only model imports into callables in pulp.tasking

goosemania:
re #2440

This is just a proposal for what we discussed some time ago, @bmbouter, @mhrivnak .
Just for style consistency the suggestion is to put only model imports inside callables. 
This `delete_worker` import is the last  one violating this approach.

If you disagree, feel free to close this PR.


bmbouter:
@goosemania I hope we can eliminate the runtime imports altogether. This PR adds more runtime imports than it removes so I propose we do not merge it. I appreciate you posting the idea though.

goosemania:
@bmbouter , thanks for the feedback! Closing this PR.

PR #2879, opened 2016-12-07
Convert delete_all_orphans task

goosemania:
closes #2394
https://pulp.plan.io/issues/2394


app/pulp/app/tasks/orphan.py
```diff
@@ -0,0 +1,14 @@
...
+from pulp.tasking.tasks import UserFacingTask
+
+
+@shared_task(base=UserFacingTask)
+def delete_all():
+    """
+    Delete all orphan content.
+    This task removes content from the filesystem as well.
+    """
+    orphans = models.Content.objects.exclude(id__in=models.RepositoryContent.objects.all())
```
goosemania: Here is query:
```
SELECT "pulp_app_content"."id", "pulp_app_content"."type" FROM "pulp_app_content"
    WHERE NOT ("pulp_app_content"."id" IN 
        (SELECT U0."content_id" FROM "pulp_app_repositorycontent" U0))
```
So I guess pagination is needed for the subquery. What paginator to use? We already have the DRF one but it requires `request`, iiuc. So should [Django paginator](https://docs.djangoproject.com/en/1.8/topics/pagination/#paginator-objects) be used in this case?

app/pulp/app/apps.py
```diff
@@ -40,6 +40,9 @@ def __init__(self, app_name, app_module):
         self.named_viewsets = None
 
     def ready(self):
+        # register signals
+        import pulp.app.signals  # noqa
```
bmbouter: Will this not work as an import at the top?

goosemania: [Django docs (see note)](https://docs.djangoproject.com/en/1.8/topics/signals/#connecting-receiver-functions) recommend to register signals in a  `ready` method of the app.
They also say that signals and their registration can live almost anywhere. I tried to follow Django recommendations though.

bmbouter:
When we used signals with mongoengine we had issues where a bulk delete would not trigger signals since it was a bulk action. In Django parlence that would be something like `<modelname>.objects.filter(...).delete()`. Does this also trigger a signal on each object contained in that queryset?

app/pulp/app/apps.py
goosemania: Sure , I will add them in the next push.

goosemania:
@bmbouter , that's a really good question, thanks! I was not aware of that.
[Docs says](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.delete) that it should work for `delete` signals at least. 

> The delete() method does a bulk delete and does not call any delete() methods on your models. It does, however, emit the pre_delete and post_delete signals for all deleted objects (including cascaded deletions).

It is working for cascaded deletions which happen in this celery task. That was actually a reason for using signals.

But in my quick test for bulk delete on the model directly fail. I am digging into it.

app/pulp/app/tasks/orphan.py
seandst: Does `orphans = Content.objects.filter(repositories=None)` work as you'd expect?

goosemania:
@bmbouter , bulk delete works well, it triggers pre_delete signals for each instance. It is just my quick test which was wrong.

app/pulp/app/tasks/orphan.py
goosemania: Thanks. @seandst , that works well!

goosemania:
I pushed all the suggested changes, PR is ready for re-review.

bmbouter:
@goosemania Thanks for the followup testing. That sounds really good. I'm not surprised though; it's Django. I don't plan to review this, but everything I've seen looks good.

server/pulp/server/managers/content/orphan.py
```diff
@@ -161,26 +161,6 @@ def get_orphan(self, content_type_id, content_unit_id):
                                               content_unit=content_unit_id)
```
goosemania: @jortel, all the changes are just deletion of the code which was converted from pulp2 to pulp3 in this PR.
Two lines are "fixed" here to satisfy flake8.

PR #2880, opened 2016-12-07
Deprecate Pulp nodes

daviddavis:
Update the documentation and release notes to indicate that nodes are
deprecated. Also, add warnings in the parent/child tasks and CLI commands.

fixes #1488
https://pulp.plan.io/issues/1488

### DIFF
--- a/docs/user-guide/nodes.rst
+++ b/docs/user-guide/nodes.rst
@@ -3,6 +3,14 @@
 Nodes
 =====
 
+.. warning::
+
+  As of Pulp 2.12, Pulp Nodes is deprecated and will be removed in Pulp 3.0.
+  Instead, users can set up a separate Pulp instance and sync content to this
+  instance using Pulp's built-in publish+sync functionality. For more
+  information, see `our blog post on the subject of deprecating Nodes
+  <http://pulpproject.org/2016/12/07/deprecating-nodes/>`_.
+
 Overview
 --------
 
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -9,3 +9,11 @@ New Features
 ------------
 
  * Task profiling can now be enabled. This will use cProfiling on an individual task and write the profile to a directory for a given task. While this can impact performance, this can enable users to get some insight into what a task is doing or use the output to give to a developer for debugging.
+
+Deprecation
+-----------
+
+* Pulp Nodes is now deprecated and will be removed in Pulp 3.0. Users can use a
+  Pulp instance and sync repositories to this instance from their main Pulp
+  instance instead. For more information, see `our blog post on the subject of
+  deprecating Nodes <http://pulpproject.org/2016/12/06/deprecating-nodes/>`_.
--- a/nodes/child/pulp_node/handlers/handler.py
+++ b/nodes/child/pulp_node/handlers/handler.py
@@ -1,11 +1,13 @@
 from logging import getLogger
+import warnings
 
 from pulp.agent.lib.handler import ContentHandler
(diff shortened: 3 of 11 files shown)


nodes/common/pulp_node/error.py
```diff
@@ -220,3 +228,7 @@ def update(self, **details):
         """
         for e in self:
             e.details.update(details)
+
+
+class NodeDeprecationWarning(Warning):
```
bmbouter: What is the behavior on 2.6? That is what the EL6 environments run.

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -0,0 +1,18 @@
+=======================
```
bmbouter: This new file needs to have a 1-line entry added in the index also

nodes/common/pulp_node/error.py
daviddavis: `DeprecationWarning` and `Warning` are loud on 2.6.

nodes/common/pulp_node/error.py
```diff
@@ -1,5 +1,13 @@
 from gettext import gettext as _
 
+CLI_DEPRECATION_WARNING = \
+    _('Warning: this node command is deprecated as the node functionality will be '
```
bmbouter: nice i8ln! 👍

nodes/common/pulp_node/error.py
bmbouter: Thanks!

docs/user-guide/nodes.rst
```diff
@@ -3,6 +3,14 @@
 Nodes
 =====
 
+.. warning::
+
+  As of Pulp 2.12, Pulp Nodes are deprecated and will be removed in Pulp 3.0.
+  Instead, users can set up a separate pulp instance and sync content to this
```
bmbouter: s/pulp/Pulp/

docs/user-guide/nodes.rst
```diff
@@ -3,6 +3,14 @@
...
+  instance using Pulp's built-in syncing functionality. For more information,
```
bmbouter: s/syncing/publish+sync/

nodes/common/pulp_node/error.py
```diff
@@ -1,5 +1,13 @@
 from gettext import gettext as _
 
+CLI_DEPRECATION_WARNING = \
```
daviddavis: Is this what I am looking for?

https://github.com/pulp/pulp/blob/master/pulp.spec#L616

bmbouter: Yes that is what you are looking for. So `pulp-nodes-common` looks like it's required from both `pulp-admin` and `pulp-consumer` which is good. This looks like it will be a non-issue, thank you for checking.

daviddavis:
@bmbouter this should be ready for re-review I think.

nodes/child/pulp_node/handlers/handler.py
```diff
@@ -10,6 +10,9 @@
 from pulp_node.handlers.reports import HandlerProgress, SummaryReport
 from pulp_node.handlers.strategies import find_strategy, Request
 
+import warnings
```
bmbouter: I should have pointed this out before (sorry). Can the imports be redone to follow the 3 level grouping of PEP8? Each group is also alphebatized for the Pulp style. warnings is part of standard lib so that is the top group and `pulp_node.error` can be combined with the import on L#8. Similar throughout PR.

bmbouter:
@daviddavis Thank you so much for putting this together. This was done very well. 🎆 😃 

Go ahead and merge it!

PR #2882, opened 2016-12-07
Fixes #1939: Enable cProfile of tasks

ehelms:
Adds the ability to turn on cProfiles of individual Pulp tasks
and dumping of the tasks into a directory. This can have an impact
on performance. The trade off is that users could investigate
individual tasks or provide these cProfiles to developers to help
detect anomalies or systematic issues.

### DIFF
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -1,0 +1,11 @@
+=======================
+Pulp 2.12 Release Notes
+=======================
+
+Pulp 2.12.0
+===========
+
+New Features
+------------
+
+ * Task profiling can now be enabled. This will use cProfiling on an individual task and write the profile to a directory for a given task. While this can impact performance, this can enable users to get some insight into what a task is doing or use the output to give to a developer for debugging.
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -394,3 +394,20 @@
 # https_retrieval: true
 # download_interval: 30
 # download_concurrency: 5
+
+# = Profiling =
+#
+# Settings for profiling Pulp tasks
+#
+# enable:
+#   This enables the cProfile Python module to profile individual tasks and
+#   store the output by task ID in a directory. Note that enabling this can
+#   impact performance.
+#
+# profile_directory:
+#   The directory that the cProfiles are stored in. This directory must be
+#   writeable and readable by Pulp.
+
+[profiling]
+# enable: false
+# directory: /var/lib/pulp/c_profiles
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -5,6 +5,7 @@ import signal
 import time
 import traceback
 import uuid
+import cProfile
 
 from bson.json_util import dumps as bson_dumps
 from bson.json_util import loads as bson_loads
@@ -20,6 +21,7 @@ from pulp.server.async.celery_instance import celery, RESOURCE_MANAGER_QUEUE, \
     DEDICATED_QUEUE_EXCHANGE
 from pulp.server.exceptions import PulpException, MissingResource, \
     PulpCodedException
+from pulp.server.config import config
 from pulp.server.db.model import Worker, ReservedResource, TaskStatus, ResourceManagerLock
 from pulp.server.exceptions import NoWorkers
 from pulp.server.managers.repo import _common as common_utils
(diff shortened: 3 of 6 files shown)


ehelms:
I'm not entirely sure how to add tests for this particular case, but if you feel they would be useful and point me in the right direction I am more than happy to add.

server/etc/pulp/server.conf
```diff
@@ -394,3 +394,20 @@
...
+#   This enables the cProfile Python module to profile individual tasks and
+#   store the output by task ID in a directory. Note that enabling this can
+#   impact performance.
+#
+# profile_directory:
+#   The directory that the cProfiles are stored in. This directory must be
+#   writeable and readable by Pulp.
+
+[profiling]
+# enable: true
```
bmbouter: This should default to off yes?

server/pulp/server/async/tasks.py
```diff
@@ -25,7 +25,10 @@
 from pulp.server.managers.repo import _common as common_utils
 from pulp.server.managers import factory as managers
 from pulp.server.managers.schedule import utils
+from pulp.server import config
 
+if config.get('profiling', 'enabled') is True:
```
bmbouter: Can we unguard this import statement and move it up with the other standard imports? We can import it 100 percent of the time I think.

bmbouter:
@ehelms Thank you for submitting this PR! Can a release note be added. I think you'll have to make this file, but [another PR already has one](https://github.com/pulp/pulp/pull/2880/files#diff-2bdedb44b0891fc2b1a9bf9ce03fcc65) so maybe just make something similar and the first one to merge wins.

bmbouter:
Also can the following be added to the commit so that Redmine associates it correctly:

```
https://pulp.plan.io/issues/1939
closes #1939
```

bmbouter:
Also Travis is saying there are flake8 errors which need to be fixed. I think I saw just one. Click details in the 'failing checks' area at the bottom of this PR to look at the travis report.

bmbouter:
How do you feel about updating the [debugging docs](http://docs.pulpproject.org/dev-guide/debugging.html#runtime-performance-analysis)? I imagine it would be a section at the top of the above link^. It would briefly explain the behavior of the feature. Beyond that, you can mostly just point them at the server.conf details instead of duplicating that text.

bmbouter:
ok test

bmbouter:
@ehelms regarding tests, I don't think you need to add any. I think ensuring all existing tests pass is important. I just kicked off a test runner. If you did really want to add tests you would put them [in this module here](https://github.com/pulp/pulp/blob/master/server/test/unit/server/async/test_tasks.py).

ehelms:
To be honest, I don't understand the flake8 output.

ehelms:
I have updated the following:

 * updated debugging docs
 * updated comments on in line code
 * added release note

bmbouter:
@ehelms I looked at the flake8 output and actually flake8 is confused. It's reporting a python flake8 issue on a .conf file. Please ignore the flake8 issue.

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -0,0 +1,11 @@
+=======================
```
bmbouter: 1 line needs to be added to the index file that indexes these release notes. Otherwise sphinx won't discover this file.

server/pulp/server/async/tasks.py
```diff
@@ -25,6 +26,7 @@
 from pulp.server.managers.repo import _common as common_utils
 from pulp.server.managers import factory as managers
 from pulp.server.managers.schedule import utils
+from pulp.server import config
```
bmbouter: Can this be moved above all `pulp.server` imports. That would put it just after the last `pulp.common` import.

ehelms:
Updated

server/pulp/server/config.py
```diff
@@ -136,6 +136,10 @@ def _load_config(self):
         'download_interval': '30',
         'download_concurrency': '5'
     },
+    'profiling': {
+        'enabled': False,
```
bmbouter: I think the tests are failing because this is `False` but it expects `'false'`. Crazy I know. Looking earlier in the file is the only reason I know. Please push a new version and I can kick off the tests again.

ehelms: Good find, updated!

bmbouter:
The good news is that the test failures now look sane. The bad news is there are 104 test failures. I think those 104 test failures are due to the mock objects which mock the config loading stuff at test time is not robust.

ehelms:
Woot! Tests are passing, stupid flak8

bmbouter:
Merging! Thanks @ehelms and @dkliban.

PR #2883, opened 2016-12-07
Revert "Prompt user of running workers/tasks on migration"

werwty:
re #2469 #2186
https://pulp.plan.io/issues/2469
https://pulp.plan.io/issues/2186

### DIFF
--- a/docs/user-guide/release-notes/2.11.x.rst
+++ b/docs/user-guide/release-notes/2.11.x.rst
@@ -24,9 +24,8 @@ New Features
 * For RPM content, a full sync will be forced if the sync configuration has been changed or content
   has been removed since the last sync.
 
-* When pulp-manage-db is run, prompt the user to continue if there are Pulp services still running.
-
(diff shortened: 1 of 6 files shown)


bmbouter:
The changes came in two commits, so can we negative commit each one separately?
Also the commits should have `re #` numbers so they associate with the issue for posterity (and downstream).

werwty:
@bmbouter The changes came in 4 commits: 70f8726b96decbcfa9ef1b799809beeb864d50a4, e0a950c517bde193992d9f862e091cad56b1c223, 75f0ac094ec3dd0bc0efd96670fe036ae7aa1982, 4f43a85dd568f4a0b50ae9b07bbec7138861e92b

There was also the issue of us wanting to keep some parts of the commits (celery cleanup and ServerSelectionTimeoutError catching) I think doing it in one commit is the better idea, since we can just revert this commit if we ever are in a state to add this feature back.

PR #2884, opened 2016-12-07
Store importer TLS keys and certificates on the filesystem.

jortel:
https://pulp.plan.io/issues/2455

Files stored in: `/var/lib/pulp/tls/`

The `StoragePath` needed to be more specific so renamed using `location` semantics as `ArtifactLocation` and `TLSLocation`.  Still need to update the serializer.

The `FileSystem._save()` *storage* also needed to change to use the django `File` appropriately as a stream.  This meant it could not use shutil.copy() (which was not providing much anyway).

Took a stab at serializer changes.  Since certificates are passed by-value we need a way to set the value of `FileField` with content instead of a *file-like* object opened to an actual file.

### DIFF
--- a/platform/pulpcore/app/models/__init__.py
+++ b/platform/pulpcore/app/models/__init__.py
@@ -1,4 +1,5 @@
 # https://docs.djangoproject.com/en/1.10/topics/db/models/#organizing-models-in-a-package
+
 from .auth import User  # NOQA
 from .base import Model, MasterModel  # NOQA
 from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyValueRelation,  # NOQA
@@ -7,9 +8,11 @@ from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyVa
 from .consumer import Consumer, ConsumerContent  # NOQA
 from .content import Content, Artifact  # NOQA
 from .repository import Repository, Importer, Publisher, RepositoryContent  # NOQA
-
-from .catalog import DownloadCatalog  # NOQA
+from .storage import FileContent  # NOQA
 from .task import ReservedResource, Worker, Task, TaskTag, TaskLock  # NOQA
+
+# Moved here: imports Artifact.
+from .catalog import DownloadCatalog  # NOQA
 
 # Moved here to avoid a circular import with Task
 from .progress import ProgressBar, ProgressReport, ProgressSpinner  # NOQA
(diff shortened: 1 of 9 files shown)


platform/pulp/app/models/content.py
```diff
@@ -100,7 +100,7 @@ class Artifact(Model):
     # other undesirable behavior and complexities.  Using a custom
     # field should be investigated.
```
asmacdo: Do you feel comfortable removing this comment now that this work is done?

platform/pulp/app/models/storage.py
```diff
@@ -1,6 +1,7 @@
 import os
 import errno
-import shutil
+import io
+import stat
```
asmacdo: alphabetical for PEP-8

platform/pulp/app/models/storage.py
```diff
@@ -142,8 +234,22 @@ def _save(path, content):
         Returns:
             str: Final storage page.
         """
+        # Create dir
```
asmacdo: I don't think this comment is necessary.

platform/pulp/app/models/storage.py
```diff
@@ -10,10 +11,98 @@
...
+class TLSLocation:
+    """
+    Determine storage location (path) for PEM encoded TLS keys and certificates
+    associated with any model.
+
+    Attributes:
+        name (str): The file name.
+    """
+
+    ROOT = 'tls'
```
asmacdo: s/ROOT/MEDIA_TYPE/

platform/pulp/app/models/storage.py
```diff
@@ -10,10 +11,98 @@
 from django.utils.deconstruct import deconstructible
 
 
+class FileContent(File):
```
asmacdo: Curious why this wouldnt work:
https://docs.djangoproject.com/en/1.11/_modules/django/core/files/base/#File

platform/pulp/app/models/storage.py
```diff
@@ -10,10 +11,98 @@
...
+    """
+
+    def __init__(self, content):
+        """
+        Args:
+            content (str): The file content.
+
+        Notes:
+            The name is set to '-' to ensure that model.FileField
+            will work properly.
```
asmacdo: More info would be helpful. Something like: For a regular Django `File` object, `name` is the path to the file-like object. Our `content` is already in memory, so `name` isn't necessary to open the file.

platform/pulp/app/models/storage.py
jortel: I don't mind ordering alphabetically but, fyi, alphabetizing imports is not part of pep8.
```
 Imports should be grouped in the following order:

    standard library imports
    related third party imports
    local application/library specific imports
```

PR #2446, opened 2016-02-22
Task 1671: Document the sudoers config to allow developers to vagrant…

shubham90:
… up without sudo password

https://pulp.plan.io/issues/1671

closes #1671

PR #2885, opened 2016-12-08
As a user, I can use REST API to remove an importer from a repository.

ipanova:
closes #2446
https://pulp.plan.io/issues/2446


app/pulp/app/viewsets/repository.py
```diff
@@ -72,6 +72,15 @@ class ImporterViewSet(NamedModelViewSet):
     serializer_class = ImporterSerializer
     endpoint_name = 'importers'
 
+    def destroy(self, request, name):
+        imp = self.get_object()
+        async_result = importer.delete.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_IMPORTER_TYPE, imp.name,
```
goosemania: The first argument for the `apply_async_with_reservation` is `resource_type` and the second - `resource_id`.
I believe for creating tasks we use only one resource type right now and it is repository and the id is its name. 
The rest of the "resource types" are specified only in tags.

This way we avoid running certain tasks on the same repository simultaneously:
```
importer.delete.apply_async_with_reservation(tags.RESOURCE_REPOSITORY_TYPE, imp.repository.name, ...)
```

app/pulp/app/viewsets/repository.py
```diff
@@ -72,6 +72,15 @@ class ImporterViewSet(NamedModelViewSet):
...
+            kwargs={'repo_name': imp.repository,
```
goosemania: If we'd like to specify repository name, isn't it supposed to be `imp.repository.name` and not just `imp.repository` ? It looks more explicit to me that way.

app/pulp/app/viewsets/repository.py
```diff
@@ -72,6 +72,15 @@ class ImporterViewSet(NamedModelViewSet):
...
+                    'importer_name': imp.name}
```
goosemania: Do we want to add tags here?  At least they were [specified in pulp 2](https://github.com/pulp/pulp/blob/3.0-dev/server/pulp/server/controllers/importer.py#L184) .

app/pulp/app/viewsets/repository.py
mhrivnak: @goosemania is correct.

app/pulp/app/tasks/tags.py
```diff
@@ -1 +1,2 @@
 RESOURCE_REPOSITORY_TYPE = 'repository'
```
goosemania: I suggest to move `tags.py` out of the `pulp.app.tasks` because it may be expected that only tasks will be imported from `pulp.app.tasks`. Let's discuss where to put it.

app/pulp/app/viewsets/repository.py
```diff
@@ -8,7 +8,7 @@
 from pulp.app.serializers import (ContentSerializer, ImporterSerializer, PublisherSerializer,
                                   RepositorySerializer, RepositoryGroupSerializer,
                                   RepositoryContentSerializer)
-from pulp.app.tasks import repository, tags
+from pulp.app.tasks import importer, repository, tags
```
goosemania: Another suggestion is to do `from pulp.app import tasks` and then `tasks.importer.delete.apply_async...` not to confuse task for importer and the instance of importer.

@mhrivnak , thanks for pointing to a simple solution here :)

app/pulp/app/viewsets/repository.py
ipanova: @goosemania seems like you are right, i got confused a bit with these two arguments

app/pulp/app/viewsets/repository.py
ipanova: i like it! thanks

app/pulp/app/viewsets/repository.py
ipanova: i think we do want tags eventually. shall we create a separate task for this? https://github.com/pulp/pulp/blob/master/common/pulp/common/tags.py#L41

PR #2887, opened 2016-12-08
Problem: Publish fails when SELinux is disabled

dkliban:
Solution: Explicitly check if SELinux is enabled before calling restorecon

closes #2476
https://pulp.plan.io/issues/2476


dkliban:
ok test

PR #2889, opened 2016-12-09
Add `--syncs` option to show repo syncs.

ulif:
With the new switch `pulp-admin repo list` supports a new option
``--syncs`` which triggers displaying a list of last sync operations run
on repositories.


daviddavis:
We already have a command for displaying tasks (`pulp-admin tasks list`). What do you think of building off that command instead of `repo list`?

bmbouter:
Similar request here as [in another comment](https://github.com/pulp/pulp/pull/2890#issuecomment-289563219). Can this be opened as a story in pulp.plan.io and use `pulp-admin tasks list --sync` instead? Also both of these PRs need to come with docs.

daviddavis:
Opened an RFE:

https://pulp.plan.io/issues/2740

Closing this PR due to inactivity.

PR #2890, opened 2016-12-09
Add --syncs option for 'pulp-admin node list'.

ulif:
Lists a table of sync tasks.

### DIFF
--- a/nodes/extensions/admin/pulp_node/extensions/admin/commands.py
+++ b/nodes/extensions/admin/pulp_node/extensions/admin/commands.py
@@ -1,11 +1,13 @@
+import datetime
 import os
 
 from gettext import gettext as _
 
 from pulp.bindings.exceptions import NotFoundException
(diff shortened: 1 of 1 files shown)


daviddavis:
Per https://github.com/pulp/pulp/pull/2889#issuecomment-266217325, I wonder if we should also have a separate `pulp-admin node tasks list` command?

bmbouter:
@ulif Can you open up a story request in pulp.plan.io about this feature? That way we can get the design vetted so we know if this code does what it needs to.

I think there are really two features here. One would be to list tasks for a given node `pulp-admin node tasks list` and the second is to only show sync tasks for the node tasks list using the `--sync` option such as `pulp-admin node tasks list --sync`.

daviddavis:
Closing this due to inactivity. I'm going to forgo creating a redmine issue since nodes are deprecated but feel free to reopen and create an issue.

PR #2891, opened 2016-12-09
Add publisher removal route to the REST API

goosemania:
closes #2447
https://pulp.plan.io/issues/2447

### DIFF
--- a/app/pulp/app/viewsets/repository.py
+++ b/app/pulp/app/viewsets/repository.py
@@ -137,6 +137,16 @@ class PublisherViewSet(NamedModelViewSet):
     queryset = Publisher.objects.all()
     filter_class = PublisherFilter
 
+    def destroy(self, request, pk):
+        publisher = self.get_object()
+        repo_name = publisher.repository.name
+        task_params = {'repo_name': repo_name,
+                       'publisher_name': publisher.name}
+        async_result = tasks.publisher.delete.apply_async_with_reservation(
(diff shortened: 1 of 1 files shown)


app/pulp/app/viewsets/repository.py
```diff
@@ -124,6 +124,16 @@ class PublisherViewSet(NamedModelViewSet):
     queryset = Publisher.objects.all()
     filter_class = PublisherFilter
 
+    def destroy(self, request, pk):
```
goosemania: I am not sure if it is worth a docstring here, since all the detailed routes will have the same set of arguments.
But this `pk` is a keyword argument and its name matters, it's a lookup field. In case of `RepositoryViewSet`, for example, it's not `pk` but `name`.
In [DRF docs](http://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) they define it like `pk=None` just to show that it is a keyword argument, `pk` can't be `None`, it will be always specified.

To be more clear what `pk` is, is anyone in favor of `pk=None`, docstring or both?

ipanova: i would not be in favour of pk=None not to create confusion, i'd go with just 'pk'

ipanova: i would suggest to put this note probably somewhere here https://github.com/pulp/pulp/blob/3.0-dev/docs/contributing/3.0-development/rest-api.rst#lookup_field-annoyances

PR #2892, opened 2016-12-10
Pins okaara dependency versions

bmbouter:
The okaara 1.0.36 release is incompatible with Python
2.7. This Pins the versions to 1.0.32 <= okaara <=
1.0.35 which are compatible with Pulp. This range limit
is made in the spec file and the setup.py requirements.

https://pulp.plan.io/issues/2464
closes #2464


bmbouter:
The flake8 errors are not correct. flake8 is trying to process the diff of a spec file.

PR #2894, opened 2016-12-10
Closes #2481: Adds kobo as setup.py dependency

bmbouter:
With 2.10.0 Kobo became a spec file dependency, which
handled installing it for users. For developers, the
Ansible playbook seems to get its Pulp dependencies from
the setup.py files. This adds the kobo dependency to
the pulp.server setup.py file which fixes `vagrant up`
for developers.

https://pulp.plan.io/issues/2481

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -33,6 +33,6 @@ setup(
     },
     install_requires=[
         'blinker', 'celery >=3.1.0, <3.2.0', 'httplib2', 'iniparse', 'isodate>=0.5.0',
-        'mongoengine>=0.10.0', 'oauth2>=1.5.211', 'pymongo>=3.0.0', 'setuptools',
+        'mongoengine>=0.10.0', 'oauth2>=1.5.211', 'kobo', 'pymongo>=3.0.0', 'setuptools',
         DJANGO_REQUIRES, SEMVER_REQUIRES, M2CRYPTO_REQUIRES],
 )


werwty:
I think that if the spec file for ansible isn't resolving automatically maybe we should add kobo in pulp/devel since it's resolving for users: https://github.com/pulp/devel/blob/master/ansible/roles/dev/tasks/main.yml#L43

Also I think pulp does check the spec file dependencies here (I'm not sure why it's not getting kobo though): https://github.com/pulp/devel/blob/master/ansible/library/pulp_facts.py#L40

ktdreyer:
Just a heads' up that no one really knows who owns the Kobo module on PyPI. I've been trying to adopt it, to no avail: https://sourceforge.net/p/pypi/support-requests/632/

The result is that the PyPI version is much older than one that you will find at https://github.com/release-engineering/kobo or in Fedora.

bmbouter:
@ktdreyer Thank you for the heads up. We will change this to install via rpm, which looks very consistent on rhel and fedoras. Thanks!

ktdreyer:
Another option is to install with Git and the tag, like:

```
setup(
    ...,
    dependency_links=[
        'git+https://github.com/release-engineering/kobo@kobo-0.5.2#egg=kobo-0.5.2'
    ],
```

... Whatever's easier!

bmbouter:
We had a bug in our Ansible deployment which was causing kobo to not be installed via rpm as it was expected to. I'm closing this PR. Thanks @ktdreyer for the help anyway. We use kobo!

PR #2896, opened 2016-12-12
Prevent exceptions raised in finally:

jortel:
https://pulp.plan.io/issues/2136

The issue proposes to catch and log the exception raised in the `try:` but this exception is most likely the root cause.  The code in `finally:` *should* just clean up resources.  Allowing an exception out of the `finally:` results in exceptions raised in the `try:` to be *swallowed*.

This seemed like a more appropriate solution and results in the root cause being reported as desired by the reporter.

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -162,7 +162,10 @@ class Step(object):
             for step in _post_order(self):
                 step.process()
         finally:
-            self.report_progress(force=True)
+            try:
+                self.report_progress(force=True)
+            except Exception:
+                _logger.exception(_('Progress reporting failed'))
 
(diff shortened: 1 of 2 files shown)


server/pulp/plugins/util/publish_step.py
```diff
@@ -162,7 +162,10 @@ def process_lifecycle(self):
             for step in _post_order(self):
                 step.process()
         finally:
-            self.report_progress(force=True)
+            try:
+                self.report_progress(force=True)
+            except Exception:
+                _logger.debug(_('Progress reporting failed'), exc_info=True)
```
mhrivnak: Why ``.debug(`` instead of ``.exception(`` like below?

jortel:
ok test

PR #2898, opened 2016-12-15
RSA key pair and SSL CA certificate generation removed from pulp.spec.

jortel:
https://pulp.plan.io/issues/2013

**Server:**
Moved the RSA key pair generation out of the _%post_ of the pulp.spec into a new script named: `pulp-gen-key-par` and removed call to `pulp-gen-ca-certificate`.  Updated the *installation* documentation and added steps to run these (2) scripts.  

I added another script named: `pulp-setup` that is intended to consolidate things.  I'm on the fence about this.  On on hand, it seems valuable to automate and encapsulate the post-install steps into a single script.  This makes it easier for users (run one thing) and hopefully supports future refactoring and/or additions without impacting the user (and documentation).  If we like this, we'll need to change the documentation to instruct the user to run `pulp-setup` instead.  On the other hand, the script currently only does (3) things ...

**Nodes:**
Removed calling `pulp-gen-nodes-certificate` from pulp.spec and updated the nodes _installation_ documentation.  I also noticed that the nodes *Quick Start*  was telling the user to restart `httpd` instead of all of the services.  Pretty sure both `httpd` and the workers need to be restarted to use the newly installed nodes importer and distributor.  Simpler to tell the user to restart all of the services.

### DIFF
--- a/docs/user-guide/installation/f23-.rst
+++ b/docs/user-guide/installation/f23-.rst
@@ -150,6 +150,11 @@ Server
      :ref:`Pulp Broker Settings Guide <pulp-broker-settings>`.
    * **server** if you want to change the server's URL components, hostname, or default credentials
 
+#. Generate RSA key pair and SSL CA certificate::
+
+   $ sudo pulp-gen-key-pair
+   $ sudo pulp-gen-ca-certificate
+
 #. Initialize Pulp's database. It is important that the broker is running before initializing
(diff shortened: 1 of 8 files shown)


jortel:
ok test

jortel:
ok test

jortel:
ok test

PR #2096, opened 2015-10-04
Fix error message on repo remove command

pcreech:
Adds repo validation for RepoUnassociate

closes [#1073](https://pulp.plan.io/issues/1073)

PR #2899, opened 2016-12-16
Check the search_tag before attempting to fast forward file

goosemania:
re #2096
https://pulp.plan.io/issues/2096


mhrivnak:
Thanks for adding that comment. I think it would be even better to specify in the ``__init__``, since that is where a developer has a decision to make about what value to provide.

PR #2901, opened 2016-12-17
Fix status API to not show stopped workers and scheduler

daviddavis:
This fix ensures that workers (including resource_manager) and the
scheduler properly clean up their database records when gracefully
shutdown.

fixes #2491
https://pulp.plan.io/issues/2491

Note: This diff is a bit easier to view if you turn off white space checking:

https://github.com/pulp/pulp/pull/2901/files?w=0

### DIFF
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -4,16 +4,13 @@ will see the "celery" module attribute and use it. This module also initializes 
 Celery setup finishes.
 """
 
-import contextlib
 import logging
-import sys
-import signal
 import time
 from datetime import datetime
 from gettext import gettext as _
 
 import mongoengine
-from celery.signals import celeryd_after_setup
+from celery.signals import celeryd_after_setup, worker_shutdown
 
 from pulp.common import constants
 from pulp.server import initialization
@@ -87,14 +84,32 @@ def initialize_worker(sender, instance, **kwargs):
         get_resource_manager_lock(sender)
 
 
+@worker_shutdown.connect
+def shutdown_worker(signal, sender):
+    """
+    Called when a worker is shutdown.
+
+    So far, this just cleans up the database by removing the worker's record in
+    the workers collection.
+
+    :param signal:   The signal being sent to the worker
+    :type  signal:   int
+    :param instance: The hostname of the worker
+    :type  instance: celery.apps.worker.Worker
+    """
+    tasks._delete_worker(sender.hostname, normal_shutdown=True)
+
+
 def get_resource_manager_lock(name):
     """
-    Tries to acquire the resource manager lock. If the lock cannot be acquired immediately, it
-    will wait until the currently active instance becomes unavailable, at which point the worker
-    cleanup routine will clear the lock for us to acquire. A worker record will be created so that
-    the waiting resource manager will appear in the Status API. We override the SIGTERM signal
-    handler so that that the worker record will be immediately cleaned up if the process is killed
-    while in this states.
+    Tries to acquire the resource manager lock.
+
+    If the lock cannot be acquired immediately, it will wait until the
+    currently active instance becomes unavailable, at which point the worker
+    cleanup routine will clear the lock for us to acquire. A worker record will
+    be created so that the waiting resource manager will appear in the Status
+    API. This worker record will be cleaned up through the regular worker
+    shutdown routine.
 
     :param name:   The hostname of the worker
     :type  name:   basestring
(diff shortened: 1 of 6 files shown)


daviddavis:
Note: This diff is a bit easier to view if you turn off white space checking:

https://github.com/pulp/pulp/pull/2901/files?w=0

server/pulp/server/async/scheduler.py
```diff
@@ -192,6 +192,9 @@ def __init__(self, *args, **kwargs):
         self._schedule = None
         self._loaded_from_db_count = 0
 
+        # Setting the celerybeat name
+        self.celerybeat_name = constants.SCHEDULER_WORKER_NAME + "@" + platform.node()
```
bmbouter: What about assigning this to a module constant instead of an instance variable? It's never modified so I think a constant would be better.

server/pulp/server/async/app.py
```diff
@@ -87,6 +84,13 @@ def initialize_worker(sender, instance, **kwargs):
         get_resource_manager_lock(sender)
 
 
+@worker_shutdown.connect
+def shutdown_worker(signal, sender):
+    msg = _("Worker '%s' shutdown" % sender.hostname)
```
bmbouter: This won't i8ln correctly because when gettext collects the strings it will already be replaced. I make this mistake often myself. Instead what you want is:

`_("Worker '%s' shutdown") % sender.hostname`

bmbouter:
I manually tested this and it all worked as expected! Also the ignoring whitespace github feature is cool; I've never seen that before. 👍

server/pulp/server/async/app.py
```diff
@@ -103,47 +107,26 @@ def get_resource_manager_lock(name):
...
+    _first_check = True
+
+    while True:
+        # Create / update the worker record so that Pulp knows we exist
+        Worker.objects(name=name).update_one(set__last_heartbeat=datetime.utcnow(),
+                                             upsert=True)
+        try:
+            lock.save()
+
+            msg = _("Resource manager '%s' has acquired the resource manager lock" % name)
```
bmbouter: You didn't write this, but this has the same i8ln problem.

server/pulp/server/async/app.py
```diff
@@ -103,47 +107,26 @@ def get_resource_manager_lock(name):
...
+            _logger.info(msg)
+            break
+        except mongoengine.NotUniqueError:
+            # Only log the message the first time
+            if _first_check:
+                msg = _("Resource manager '%s' attempted to acquire the the resource manager "
```
bmbouter: You didn't write this, but this has the same i8ln problem.
Also this needs to be converted to a dictionary based string replacement since this cannot be i8ln to all languages when the positional parameters are reordered.

server/pulp/server/async/scheduler.py
```diff
@@ -368,3 +368,7 @@ def add(self, **kwargs):
         entries to the database, and they will be picked up automatically.
         """
         raise NotImplementedError
+
+    def close(self):
```
bmbouter: Can 1-liner docstring be added that indicates celerybeat calls this when closing.

server/pulp/server/async/app.py
```diff
@@ -87,6 +84,13 @@ def initialize_worker(sender, instance, **kwargs):
         get_resource_manager_lock(sender)
 
 
+@worker_shutdown.connect
+def shutdown_worker(signal, sender):
+    msg = _("Worker '%s' shutdown" % sender.hostname)
+    _logger.info(msg)
```
bmbouter: Since `_delete_worker()` is also called by `Scheduler.close()` if this is moved to an `else` branch on [the `if` statement that checks `normal_shutdown`](https://github.com/pulp/pulp/blob/master/server/pulp/server/async/tasks.py#L246) then this will log for celerybeat also which I think would be good.

daviddavis:
@bmbouter updated based on your feedback. thanks for the review and let me know if I missed anything.

server/pulp/server/async/app.py
```diff
@@ -87,6 +84,20 @@ def initialize_worker(sender, instance, **kwargs):
         get_resource_manager_lock(sender)
 
 
+@worker_shutdown.connect
+def shutdown_worker(signal, sender):
+    """
+    Called when a worker is shutdown. So far, this just cleans up the database
```
bmbouter: The first sentence of a docstring should be a 1-line, then whitespace, then a multi-line paragraph which is optional. See 257 for convention details. In this case this second sentence can just be moved to its own paragraph. A similar change can be made for the other docstring.

bmbouter:
@daviddavis Can the base branch be changed to targe 2.11-dev instead of 2.10-dev now that 2.11 is GA.

server/pulp/server/async/app.py
bmbouter: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

server/pulp/server/async/app.py
```diff
@@ -87,14 +84,33 @@ def initialize_worker(sender, instance, **kwargs):
...
-    the waiting resource manager will appear in the Status API. We override the SIGTERM signal
-    handler so that that the worker record will be immediately cleaned up if the process is killed
-    while in this states.
+    Tries to acquire the resource manager lock.
+
+    If the lock cannot be acquired immediately, it will wait until the
+    currently active instance becomes unavailable, at which point the worker
+    cleanup routine will clear the lock for us to acquire. A worker record will
+    be created so that the waiting resource manager will appear in the Status
+    API. We override the SIGTERM signal handler so that that the worker record
```
dralley: Description should be fixed now that we're no longer overriding the signal handler :)

daviddavis:
@dralley thanks for the comment. I updated the description to remove the last line. Can you check my change? Thanks.

server/pulp/server/async/app.py
```diff
@@ -87,14 +84,31 @@ def initialize_worker(sender, instance, **kwargs):
...
-    the waiting resource manager will appear in the Status API. We override the SIGTERM signal
-    handler so that that the worker record will be immediately cleaned up if the process is killed
-    while in this states.
+    Tries to acquire the resource manager lock.
+
+    If the lock cannot be acquired immediately, it will wait until the
+    currently active instance becomes unavailable, at which point the worker
+    cleanup routine will clear the lock for us to acquire. A worker record will
+    be created so that the waiting resource manager will appear in the Status
+    API.
```
bmbouter: I think adding a line here that the worker record is expected to be cleaned up upon shutdown would be good. It can remain silent on where/how that happens.

daviddavis:
@dralley going to go ahead and merge this but if you're not happy with the description (or anything else), let me know and I'll fix it in a followup PR.

@bmbouter @dralley thanks for the reviews.

PR #2903, opened 2016-12-30
Add pulp-manage-db check for running workers

werwty:


server/pulp/server/db/manage.py
```diff
@@ -200,6 +234,9 @@ def main():
         return os.EX_DATAERR
     except models.MigrationRemovedError:
         return os.EX_SOFTWARE
+    except ServerSelectionTimeoutError:
+        _logger.info(_('Cannot connect to database, please validate that the database is up.'))
```
bmbouter: s/up/online and accessible/

I suggest ^ to capture that the db needs to be reachable via the network.

bmbouter: Also s/to /to the/

common/pulp/common/constants.py
```diff
@@ -59,13 +59,13 @@
 # Constant used to determine whether a CeleryBeatLock should be removed due to age
 CELERYBEAT_LOCK_MAX_AGE = 200
 # The amount of time in seconds before a Celery process is considered missing
-CELERY_TIMEOUT_SECONDS = 300
+CELERY_TIMEOUT_SECONDS = 60
```
bmbouter: This will change actual failover to be < 60 seconds for Pulp usage not involving `pulp-manage-db`. I like this change, but we should probably do it as a separate commit. It also needs to come with its own changes including docs updates and a dedicated release note. Is making this change with 2.12 something we want to do?

server/test/unit/server/async/test_scheduler.py
```diff
@@ -506,7 +506,7 @@ def test_debug_logging(self, mock__logger, mock_worker, mock_delete_worker):
         scheduler.CeleryProcessTimeoutMonitor().check_celery_processes()
         mock__logger.debug.assert_has_calls([
             mock.call('Checking if pulp_workers, pulp_celerybeat, or '
-                      'pulp_resource_manager processes are missing for more than 300 seconds'),
+                      'pulp_resource_manager processes are missing for more than 60 seconds'),
```
bmbouter: Can this use the variable value instead so that we can stay DRY. Yes I know I wrote this code originally. :-)

common/pulp/common/constants.py
mhrivnak: It looks like this change, along with the one below for ``CELERY_MAX_INTERVAL``, are already in a separate commit. It does also have a release note, although elaborating on its impact may be beneficial.

I think this is a good change for the 2.y line.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +202,29 @@ def main():
...
         connection.initialize(max_timeout=1)
+
+        if status.get_workers():
+            last_worker_timestamp = max([worker['last_heartbeat'] for worker
+                                         in status.get_workers()])
+            wait_time = timedelta(seconds=constants.CELERY_TIMEOUT_SECONDS) \
+                - (UTCDateTimeField().to_python(datetime.now()) - last_worker_timestamp)
+
+            if wait_time > timedelta(0):
+                print >> sys.stdout, _('\rThe following processes might still be running %s'
```
bmbouter: Can these print calls be converted to not use the `>>` operator. I can't find the documentation for that in the Python docs even. Changing it to `print 'stuff here'` would be consistent with the rest of the codebase.

server/pulp/server/db/manage.py
```diff
@@ -180,6 +185,12 @@ def ensure_database_indexes():
         model_class.ensure_indexes()
 
 
+def _get_running_worker_names():
```
bmbouter: Can this become part of a custom QuerySet that is attached to the Worker model? That would allow usage like `Worker.get_running_workers()` to be called from anywhere in the codebase.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +202,29 @@ def main():
...
+            last_worker_timestamp = max([worker['last_heartbeat'] for worker
+                                         in status.get_workers()])
+            wait_time = timedelta(seconds=constants.CELERY_TIMEOUT_SECONDS) \
+                - (UTCDateTimeField().to_python(datetime.now()) - last_worker_timestamp)
+
+            if wait_time > timedelta(0):
+                print >> sys.stdout, _('\rThe following processes might still be running %s'
+                                       % _get_running_worker_names())
+
+                for i in xrange(wait_time.seconds, 0, -1):
```
bmbouter: s/pulp/Pulp/

server/pulp/server/db/manage.py
```diff
@@ -191,6 +202,29 @@ def main():
...
+                    print >> sys.stdout, _('\rPlease wait %s seconds while pulp'
```
bmbouter: s/pulp/Pulp/

server/pulp/server/db/manage.py
bmbouter: What about making this print every 5 seconds instead? It would probably look prettier over a 60 second output.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +202,29 @@ def main():
...
+                                           ' confirms this.' % i),
+                    sys.stdout.flush()
+                    time.sleep(1)
+
+                if max([worker['last_heartbeat'] for worker in status.get_workers()]) \
+                        > last_worker_timestamp:
+                    print >> sys.stderr, _('\r\n%s are still running, please stop the running'
+                                           ' workers before retrying the pulp_manage_db command.'
```
bmbouter: s/pulp_manage_db/pulp-manage-db/

server/pulp/server/db/manage.py
bmbouter: Ohhh, I see. This rewrites the line in a fancy way.   Nice! Please disregard this comment.

bmbouter: For me this prints like:

```
The following processes might still be running [u'scheduler@dev', u'resource_manager@dev']
```

Can it instead print prettier with an indented, line separated list like:
```
The following processes might still be running:
    scheduler@dev
    resource_manager@dev
```

bmbouter: Similar request for the print output below.

bmbouter:
The flake8 errors can be ignored, and I think the test failure is not related, but after the next push please run `ok test` so that Jenkins can show it passing.

common/pulp/common/constants.py
bmbouter: I wrote commit, but I meant commit+redmine_ticket. We can implement this entire feature without changing this value. Also the scaling guide I believe talks about failover times which we would need to update. I'm hoping to establish for this PR that we only increase the heartbeat frequency and in a separate PR with a separate issue update the failover speed. What do you all think about that?

werwty: Currently the wait time for pulp-manage-db is calculated by CELERY_TIMEOUT_SECONDS - currtime - last_hearbeat if we leave this to be 300 then the user has a potentially very long time to wait. I'll file a new feature with the timeout improvement and update this PR with a separate commit since it's a big change and should get it's own line in the release note.

server/pulp/server/db/manage.py
werwty: haha nothing to do with the print style, I was trying to be consistent with the `print >> sys.stderr`. Will change to `print` instead

server/pulp/server/db/manage.py
werwty: I think it might make more sense to put this in `status.get_running_workers()`
@bmbouter what do you think?

bmbouter: Since the function provides filtering at the collection level, putting them on the QuerySet associated with the Worker is most consistent with the mongoengine (and future Django) design. In fact, the `status.get_workers()` method should not even exist, but since each PR should handle one thing leaving that there for now I think is OK.

common/pulp/common/constants.py
```diff
@@ -66,6 +66,9 @@
 # The maximum number of seconds that will ever elapse before the scheduler looks for
 # new or changed schedules
 CELERY_MAX_INTERVAL = 90
+# The amount of time the migration script will wait to confirm that no processes are running
+# This is the old CELERY_TICK_DEFAULT_WAIT_TIME + delta, to account for migration scenarios
```
bmbouter: Can this be more explicit by version those who are unfamiliar. I think a longer comment is fine. Consider something like:

`Pulp 2.12 lowered the heartbeat time from 90 seconds to 5 seconds. 92 seconds is used to allow the process check feature to work correctly even in cases where a user forgot to restart pulp_celerybeat while upgrading from 2.11 or earlier. The additional 2 seconds allows the database to quiesce so all processes have read consistency.`

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
         options = parse_args()
         _start_logging()
         connection.initialize(max_timeout=1)
+
+        if status.get_workers():
+            last_worker_time = max([worker['last_heartbeat'] for worker
+                                    in status.get_workers()])
+            wait_time = timedelta(seconds=constants.MIGRATION_WAIT_TIME) -\
+                (UTCDateTimeField().to_python(datetime.now()) - last_worker_time)
```
bmbouter: Can this expression be assigned to a variable to lower the complexity of this L203-L204. 1-liners like this are hard to understand and benefit from bring broken up. Eric Helms taught me that when statements are small, and variable names are really clear, readability goes way up and comments are almost never needed.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
...
+                print _('The following processes might still be running:')
+                for worker in status.get_workers():
+                    print _('\t%s' % worker['name'])
+
+                for i in range(wait_time.seconds, 0, -1):
+                    print _('\rPlease wait %s seconds while Pulp confirms this.' % i),
+                    sys.stdout.flush()
+                    time.sleep(1)
+
+                if status.get_workers() and \
```
bmbouter: Similar here, can this if statement 1-liner be decomposed into some intermediate variables with meaningful names.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
...
+                        max([worker['last_heartbeat'] for worker in
```
bmbouter: Is there a way to make this `max(.....)` occur in the database? That would be more efficient than loading it all into memory and filtering here.

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -9,6 +9,8 @@ New Features
 ------------
 
  * Task profiling can now be enabled. This will use cProfiling on an individual task and write the profile to a directory for a given task. While this can impact performance, this can enable users to get some insight into what a task is doing or use the output to give to a developer for debugging.
+ * When ``pulp-manage-db`` is run, check if there are running processes before continuing. Users
```
bmbouter: I'm not sure a user would understand which processes are being waited on. For example we don't wait on `httpd`. Also I hope to identify this works in both standalone and clustered installs. Consider replacing it with a more explicit version like:

`pulp-manage-db will not continue if there are pulp_celerybeat, pulp_resource_manager, or pulp_workers running. This works in both standalone and clustered installations. In some cases, Pulp may wait up to 92 seconds to determine if should continue or not.`

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
...
+                            status.get_workers()]) > last_worker_time:
+                    print >> sys.stderr, _('\r\nThe following processes are still running, please'
```
werwty: the `>>` in this case is because we're printing to stderr instead of stdout

server/pulp/server/db/manage.py
werwty: Since we're filtering workers, which there will never be _that_ many of, I think moving the filter to the database would not be too much more efficient.

common/pulp/common/constants.py
```diff
@@ -66,6 +66,11 @@
 # The maximum number of seconds that will ever elapse before the scheduler looks for
 # new or changed schedules
 CELERY_MAX_INTERVAL = 90
+# The amount of time the migration script will wait to confirm that no processes are running
+# This is the 90s CELERY_TICK_DEFAULT_WAIT_TIME used in Pulp version < 2.12 and a 2s buffer.
+# This ensures that the process check feature works correctly even in cases where a user
+# forget to restart pulp_celerybeat while upgrading from Pulp 2.11 or earlier.
```
bmbouter: s/forget/forgot/

common/pulp/common/constants.py
```diff
@@ -66,6 +66,11 @@
 # The maximum number of seconds that will ever elapse before the scheduler looks for
 # new or changed schedules
 CELERY_MAX_INTERVAL = 90
+# The amount of time the migration script will wait to confirm that no processes are running
```
bmbouter: s/running/running./

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,35 @@ def main():
         options = parse_args()
         _start_logging()
         connection.initialize(max_timeout=1)
+
+        active_workers = status.get_workers()
+        if active_workers:
+            last_worker_time = max([worker['last_heartbeat'] for worker
+                                    in active_workers])
```
bmbouter: Can this be joined with the list comprehension on the previous line? I think it would be < 100 chars. If it isn't, then leave it as is.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,35 @@ def main():
...
+                print _('The following processes might still be running:')
+                for worker in active_workers:
+                    print _('\t%s' % worker['name'])
+
+                for i in range(wait_time.seconds, 0, -1):
+                    print _('\rPlease wait %s seconds while Pulp confirms this.' % i),
+                    sys.stdout.flush()
+                    time.sleep(1)
+
+                still_active_workers = [worker for worker in status.get_workers() if
```
bmbouter: You do not need to change this unless you want to, but FYI this is very similar to the usage of [filter()](https://docs.python.org/2/library/functions.html#filter).

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
...
         connection.initialize(max_timeout=1)
+
+        active_workers = status.get_workers()
+        if active_workers:
+            last_worker_time = max([worker['last_heartbeat'] for worker in active_workers])
+            time_from_last_worker = UTCDateTimeField().to_python(datetime.now()) - last_worker_time
+            wait_time = timedelta(seconds=constants.MIGRATION_WAIT_TIME) - time_from_last_worker
+
+            if wait_time > timedelta(0):
+                print _('The following processes might still be running:')
```
bmbouter: Can a `\n` be added to the beginning of this sentence. While testing I noticed this. I think the user experience would benefit if this was separated from earlier output.

server/pulp/server/db/manage.py
```diff
@@ -191,6 +196,34 @@ def main():
...
+                for i in range(wait_time.seconds, 0, -1):
+                    print _('\rPlease wait %s seconds while Pulp confirms this.' % i),
+                    sys.stdout.flush()
+                    time.sleep(1)
+
+                still_active_workers = [worker for worker in status.get_workers() if
+                                        worker['last_heartbeat'] > last_worker_time]
+
+                if still_active_workers:
+                    print >> sys.stderr, _('\r\nThe following processes are still running, please'
```
bmbouter: Is the `\r` needed here?
Also since it's not using `print as a function I think it's not including the automatic newline. Can a line of clean whitespace appear before this text? I think making `\r\n`  ->  `\n\n` would do it.

PR #1847, opened 2015-05-06
Separate view for collection actions.

jortel:
https://pulp.plan.io/issues/827

Fixing in 2.7 because all this changed with django conversion.  Did not want to fix this twice.

Changes:
- Better name for content source collection view.
- Moved post() to new view for actions on all content sources.

PR #2905, opened 2017-01-05
Update last_unit_added when unit is imported to repo.

werwty:
closes #1847
https://pulp.plan.io/issues/1847

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -810,6 +810,10 @@ def sync(repo_id, sync_config_override=None, scheduled_call_id=None):
         sync_result_collection.save(sync_result)
         # Ensure counts are updated
         rebuild_content_unit_counts(repo_obj)
+        if sync_result['added_count'] > 0:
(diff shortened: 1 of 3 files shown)


daviddavis:
What about when syncing a repo? Does or should `last_unit_added` get updated?

werwty:
ah you're right. I will update this PR. Thanks @daviddavis

wiad:
Yes, it should get updated on sync, but only if the sync actually results in any updated units - right?

daviddavis:
@wiad correct.

werwty:
ok test

wiad:
Should not 'last_unit_added' be updated when copying a rpm file from one repo to another? I'm running 2.12.0 and a `pulp-admin rpm repo copy rpm` does not modify the last_unit_added metadata, but a `pulp-admin rpm repo sync run` does.

werwty:
@wiad it should be updated when copying. I missed that use case :(
Here is the new issue I filed for this: https://pulp.plan.io/issues/2577

PR #2509, opened 2016-04-04
Adds apache_manage_sys_content_rw to pulp-streamer SELinux policy

bmbouter:
re #1771
https://pulp.plan.io/issues/1771

This was compiled and installed on the machine where the SELinux denial was encountered. After being applied no more SELinux denials were shown. The original denial was:

```
type=SYSCALL msg=audit(1459639923.760:2537): arch=c000003e syscall=21 success=no exit=-13 a0=7fe61c00e3d0 a1=4 a2=7fe644bc0fa8 a3=3 items=0 ppid=1 pid=25805 auid=4294967295 uid=48 gid=48 euid=48 suid=48 fsuid=48 egid=48 sgid=48 fsgid=48 tty=(none) ses=4294967295 comm="pulp_streamer" exe="/usr/bin/python2.7" subj=system_u:system_r:streamer_t:s0 key=(null)
type=AVC msg=audit(1459639923.760:2537): avc:  denied  { search } for pid=25805 comm="pulp_streamer" name="pulp" dev="dm-0" ino=68703432 scontext=system_u:system_r:streamer_t:s0 tcontext=system_u:object_r:httpd_sys_rw_content_t:s0 tclass=dir 
```

PR #2907, opened 2017-01-10
Reduce and simplify timings so that failover occurs within 30 seconds

dralley:
Similified and reduced the timings so that all failure detection occurs within 30 seconds. Changed resource_manager to use a timestamp-based locking mechanism.

closes #2509
https://pulp.plan.io/issues/2509

### DIFF
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -8,11 +8,19 @@ Pulp 2.12.0
 New Features
 ------------
 
- * Task profiling can now be enabled. This will use cProfiling on an individual task and write the profile to a directory for a given task. While this can impact performance, this can enable users to get some insight into what a task is doing or use the output to give to a developer for debugging.
- * ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
-   processes are running. This will prevent the user from corrupting their Pulp installation by
-   migrating with running workers. This works in both standalone and clustered installations. Pulp
-   may wait up to 92 seconds to determine if workers are running.
+* Task profiling can now be enabled. This will use `cProfile
+  <https://docs.python.org/2/library/profile.html#module-cProfile>`_ on an individual task and write
+  the profile to a directory for a given task. While this can impact performance, this enables users
+  to get some insight into what a task is doing or give the output to give to a developer for debugging.
+
+* ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
+  processes are running. This prevents the user from corrupting their Pulp installation by applying
+  migrations while workers or pulp_celerybeat are running. This works in both standalone and clustered
+  installations. Pulp may wait up to 92 seconds to determine if workers are running.
+
+* Worker failure detection and high availability failover occurs within 30 seconds. Pulp processes are
+  considered missing after 25 seconds, and heartbeats occur every 5 seconds.
+
 
 Deprecation
 -----------
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -6,13 +6,13 @@ Celery setup finishes.
 
 import logging
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 from gettext import gettext as _
 
 import mongoengine
 from celery.signals import celeryd_after_setup, worker_shutdown
 
-from pulp.common import constants
+from pulp.common import constants, dateutils
 from pulp.server import initialization
 from pulp.server.async import tasks
 from pulp.server.db.model import ResourceManagerLock, Worker
@@ -122,10 +122,17 @@ def get_resource_manager_lock(name):
     _first_check = True
 
     while True:
+
+        now = dateutils.ensure_tz(datetime.utcnow())
+        old_timestamp = now - timedelta(seconds=constants.PULP_PROCESS_TIMEOUT_INTERVAL)
+
+        ResourceManagerLock.objects(timestamp__lte=old_timestamp).delete()
+
         # Create / update the worker record so that Pulp knows we exist
         Worker.objects(name=name).update_one(set__last_heartbeat=datetime.utcnow(),
                                              upsert=True)
         try:
+            lock.timestamp = now
             lock.save()
 
             msg = _("Resource manager '%s' has acquired the resource manager lock") % name
@@ -134,11 +141,8 @@ def get_resource_manager_lock(name):
         except mongoengine.NotUniqueError:
             # Only log the message the first time
             if _first_check:
-                msg = _("Resource manager '%(name)s' attempted to acquire the the resource manager "
-                        "lock but was unable to do so. It will retry every %(interval)d seconds "
-                        "until the lock can be acquired.") % \
-                    {'name': name, 'interval': constants.CELERY_CHECK_INTERVAL}
-                _logger.info(msg)
+                _logger.info(_("Hot spare pulp_resource_manager instance '%(name)s' detected.")
+                             % {'name': name})
                 _first_check = False
 
-            time.sleep(constants.CELERY_CHECK_INTERVAL)
+            time.sleep(constants.PULP_PROCESS_HEARTBEAT_INTERVAL)
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -97,7 +97,7 @@ class CeleryProcessTimeoutMonitor(threading.Thread):
         """
         _logger.info(_('Worker Timeout Monitor Started'))
         while True:
-            time.sleep(constants.CELERY_CHECK_INTERVAL)
+            time.sleep(constants.PULP_PROCESS_HEARTBEAT_INTERVAL)
             try:
                 self.check_celery_processes()
             except Exception as e:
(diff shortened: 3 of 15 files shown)


server/test/unit/server/async/test_scheduler.py
```diff
@@ -506,7 +507,7 @@ def test_debug_logging(self, mock__logger, mock_worker, mock_delete_worker):
         scheduler.CeleryProcessTimeoutMonitor().check_celery_processes()
         mock__logger.debug.assert_has_calls([
             mock.call('Checking if pulp_workers, pulp_celerybeat, or '
-                      'pulp_resource_manager processes are missing for more than 300 seconds'),
+                      'pulp_resource_manager processes are missing for more than 25 seconds'),
```
werwty: Instead of hard-coding the timeout, can we make this PULP_PROCESS_TIMEOUT_INTERVAL

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
 New Features
 ------------
 
- * Task profiling can now be enabled. This will use cProfiling on an individual task and write the profile to a directory for a given task. While this can impact performance, this can enable users to get some insight into what a task is doing or use the output to give to a developer for debugging.
- * ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
-   processes are running. This will prevent the user from corrupting their Pulp installation by
-   migrating with running workers. This works in both standalone and clustered installations. Pulp
-   may wait up to 92 seconds to determine if workers are running.
+* Task profiling can now be enabled. This will use cProfiling on an individual task and write the
```
bmbouter: Nice wrapping fix here. thx

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  profile to a directory for a given task. While this can impact performance, this can enable
```
bmbouter: s/can enable/enables/

bmbouter: I know you didn't write this also. btw

docs/user-guide/release-notes/2.12.x.rst
bmbouter: Can s/cProfiling/cProfile/   and also can cProfile link to this page?  https://docs.python.org/2/library/profile.html#module-cProfile

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  users to get some insight into what a task is doing or use the output to give to a developer for
```
bmbouter: s/use the output to give to/give the output to/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  debugging.
+
+* ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
+  processes are running. This will prevent the user from corrupting their Pulp installation by
```
bmbouter: s/will prevent/prevents/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  migrating with running workers. This works in both standalone and clustered installations. Pulp
```
bmbouter: s/migrating with running workers/applying migrations while workers or pulp_celerybeat are running/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  may wait up to 92 seconds to determine if workers are running.
+
+* The speed of failure detection and failover has been increased dramatically. Pulp processes will
```
bmbouter: s/will be/are/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  be considered missing after 25 seconds, and heartbeats will take place every 5 seconds.
```
bmbouter: s/will take place/occur/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  the profile to a directory for a given task. While this can impact performance, this enables users
+  to get some insight into what a task is doing or give the output to give to a developer for debugging.
+
+* ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
+  processes are running. This prevents the user from corrupting their Pulp installation by applying
+  migrations while workers or pulp_celerybeat are running. This works in both standalone and clustered
+  installations. Pulp may wait up to 92 seconds to determine if workers are running.
+
+* The speed of failure detection and failover has been increased dramatically. Pulp processes are
+  considered missing after 25 seconds, and heartbeats occur place every 5 seconds. Therefore, the
```
bmbouter: s/place//

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -8,11 +8,20 @@ Pulp 2.12.0
...
+  <https://docs.python.org/2/library/profile.html#module-cProfile>`_ on an individual task and write
+  the profile to a directory for a given task. While this can impact performance, this enables users
+  to get some insight into what a task is doing or give the output to give to a developer for debugging.
+
+* ``pulp-manage-db`` will not continue if pulp_celerybeat, pulp_resource_manager, or pulp_workers
+  processes are running. This prevents the user from corrupting their Pulp installation by applying
+  migrations while workers or pulp_celerybeat are running. This works in both standalone and clustered
+  installations. Pulp may wait up to 92 seconds to determine if workers are running.
+
+* The speed of failure detection and failover has been increased dramatically. Pulp processes are
```
bmbouter: What about a less formal statement that also acknowledges our previous behavior for the first sentence?

`Worker failure detection and high availability failover occurs within 30 seconds.`

bmbouter:
One issue I ran into is:

1. have a vannila pulp box (I used vagrant with your PR checked out)
2. pstart and verify with `ps` that the expected processes are started.
3. Observe every 30 seconds the following is emitted:

```
Jan 11 21:34:33 dev pulp[28495]: pulp.server.async.scheduler:ERROR: Worker 'resource_manager@dev' has gone missing, removing from list of workers
Jan 11 21:34:33 dev pulp[28495]: pulp.server.async.tasks:ERROR: The worker named resource_manager@dev is missing. Canceling the tasks in its queue.
Jan 11 21:34:33 dev pulp[28495]: pulp.server.async.scheduler:ERROR: There are 0 pulp_resource_manager processes running. Pulp will not operate correctly without at least one pulp_resource_mananger process running.
Jan 11 21:34:35 dev pulp[28495]: pulp.server.async.worker_watcher:INFO: New worker 'resource_manager@dev' discovered
```

werwty:
@dralley @bmbouter 
The cause of the issue is here: https://github.com/pulp/pulp/pull/2907/files#diff-ac9a188d0b9425fa260a49c7def6aa0fR140
where we break out of the resource manager lock loop. Afterwards the resource manager worker timestamp isn't updated until it's timeout and then rediscovered. removing the `break` and adding a sleep fixes this issue

dralley:
@werwty I don't believe that pinpoints the root cause.  Once it breaks out of the loop it should begin heartbeating, which are picked up by celerybeat->worker_watcher which then updates the worker timestamp.

Replacing the break with a sleep only works because we're faking the heartbeat inside that loop on line 132.  But it seems like the problem would be inside the monitoring code elsewhere.

bmbouter:
The issue (as discussed on IRC) was in my environment. The timing change did not take effect for the resource_manager in my environment. I filed this issue about it:  https://pulp.plan.io/issues/2524

My environment is working normally now.

bmbouter:
In my testing I observed that when running two celerybeat's the following log line is emitted every 5 seconds:

`Jan 12 18:01:56 dev pulp[1855]: pulp.server.async.scheduler:INFO: Duplicate or new celerybeat Instance, ticking again in 5 seconds.`

We either need to switch this to a lower level like DEBUG, or eliminate the statement. What do you think?

bmbouter:
Oh one other idea is we could just have it log the first time. That is what we do in the resource manager.

bmbouter:
One other issue that I think could be improved is this:

1. start two pulp_resource_manager's in the foreground
2. Ctrl-C the one that showed the banner (and has the lock). This causes a soft shutdown which allows for self-cleanup.
3. Observe it correctly cleans up the Worker collection entry. Observe that the resource manager lock is not cleaned up. It should also clean up the lock entry to allow for quicker failover in the case of a graceful shutdown.

bmbouter:
@dralley Can the commit message be reworked so that it follows this?  http://docs.pulpproject.org/en/2.12/nightly/dev-guide/contributing/branching.html#commit-messages

bmbouter:
Commit message format looks note. In commit message s/Similified/Simplified/

server/pulp/server/async/scheduler.py
```diff
@@ -283,10 +285,14 @@ def tick(self):
...
-                               "ticking again in %(ret)s seconds.")
-                             % {'ret': ret})
+                ret = constants.PULP_PROCESS_HEARTBEAT_INTERVAL
+
+                if not self._retry_msg_logged:
+                    _logger.info(_("Duplicate or new celerybeat Instance, "
+                                   "ticking again in %(ret)s seconds.")
+                                 % {'ret': ret})
+                    # Never re-log this message
+                    self._retry_msg_logged = True
```
bmbouter: What if this variable was renamed from `_retry_msg_logged` to `_already_logged`? The if state would read more like plaintext and the comment at L#294 could be removed. Really it's the comment that got me interested in this. @ehelms taught me to reserve comment use for documenting programmer intent that is significantly non-obvious. I can rant more about the evil of comments in code some other time...

server/pulp/server/async/app.py
```diff
@@ -137,8 +144,8 @@ def get_resource_manager_lock(name):
                 msg = _("Resource manager '%(name)s' attempted to acquire the the resource manager "
```
bmbouter: Can a similar change be made to this logging statement.

`Hot spare pulp_resource_manager instance '%(resource_manager_name)s' detected.`

bmbouter:
@dralley Thank you so much for this! You've done great work, and this feature is working perfectly as far as I can tell. Pulp users are really going to benefit from this!

PR #2519, opened 2016-04-11
FileDistributor now works with files that have to be in subdirectories.

mhrivnak:
https://pulp.plan.io/issues/1824

fixes #1824

PR #2908, opened 2017-01-11
Problem: pulp_celerybeat writes hearbeat records to the database

dkliban:
Solution: All workers write their own records to the db instead of
sending them through the message bus.

This patch makes use of the Consumer blueprint that celery runs at
start time of a worker. An extra bootstep has been added to the
consumer blueprint. The new bootstep sets a 30 second timer which
runs a method that records a heartbeat in the database.

http://docs.celeryproject.org/en/master/userguide/extending.html
https://groups.google.com/d/msg/celery-users/3fs0ocREYqw/C7U1lCAp56sJ

closes #2519
https://pulp.plan.io/issues/2519

### DIFF
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -13,6 +13,9 @@ New Features
    processes are running. This will prevent the user from corrupting their Pulp installation by
    migrating with running workers. This works in both standalone and clustered installations. Pulp
    may wait up to 92 seconds to determine if workers are running.
+ * pulp_workers, pulp_resource_manager, and pulp_celerybeat record their own hearbeats to the
+   database. pulp_celerybeat does not need to be running to determine if pulp_resource_manager and
+   pulp_workers are online.
 
 Deprecation
 -----------
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from gettext import gettext as _
 
 import mongoengine
+from celery import bootsteps
 from celery.signals import celeryd_after_setup, worker_shutdown
 
 from pulp.common import constants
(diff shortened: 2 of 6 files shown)


docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -13,6 +13,9 @@ New Features
    processes are running. This will prevent the user from corrupting their Pulp installation by
    migrating with running workers. This works in both standalone and clustered installations. Pulp
    may wait up to 92 seconds to determine if workers are running.
+ * pulp_workers, pulp_resource_manager, and pulp_celerybeat record their own hearbeats to the
```
bmbouter: I think the sentence order should be reversed. The most valuable statement for users should come first. How it works is a secondary concern.

server/pulp/server/async/worker_watcher.py
```diff
@@ -78,9 +78,6 @@ def handle_worker_heartbeat(event):
         msg = _("New worker '%(worker_name)s' discovered") % event_info
```
bmbouter: I had imagined this entire functionality being distributed to where the heartbeats are recorded. That would also eliminate the need to have this thread altogether. Does that sound right?

dkliban: I like the idea of the checks being distributed between all components. Would each component log warning messages when some other component is missing?

bmbouter: I think having each one look after itself is what we need for now. That would eliminate the thread that records heartbeats from celerybeat, but the worker_watcher thread would still be there. As a yet separate story we can distribute the "watching and logging functionality" but architecturally we aren't going to eliminate pulp_celerybeat so we don't stand much to gain from distributing the worker_watcher functionality. Does ^ make any sense?

bmbouter: Specifically, looking after itself means that it logs when it heartbeats for the first time itself.

dkliban: Makes sense. I'll adjust the PR accordingly.

dralley: Once PR #2907 goes through and you rebase onto it, you will want to replace the magic number "30" with the constant ```PULP_PROCESS_HEARTBEAT_INTERVAL``` from ```pulp.common.constants```.  FYI, that interval is 5 seconds, which is the new timing for it.  In case you need to change comments/docs.

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,50 @@
...
+        self.tref = None
+
+    def start(self, consumer):
+        """
+        This method is called when the worker starts up and also whenever the AMQP connection is
+        reset (which triggers an internal restart). The timer is reset when the connection is lost,
+        so we have to install the timer again for every call to self.start.
+        """
+        self.tref = consumer.timer.call_repeatedly(
+            30.0, self.record_heartbeat, (consumer, ), priority=10,
```
dralley: Once PR #2907 goes through and you rebase onto it, you will want to replace the magic number "30" with the constant ```PULP_PROCESS_HEARTBEAT_INTERVAL``` from ```pulp.common.constants```.  FYI, that interval is 5 seconds, which is the new timing for it.  In case you need to change comments/docs.

dkliban:
I am closing this PR because  @dralley is taking over this work and will submit a new PR.

PR #2909, opened 2017-01-11
URL unquote escaped credentials in feed URL

daviddavis:
fixes #2520
https://pulp.plan.io/issues/2520

### DIFF
--- a/server/test/unit/server/controllers/test_importer.py
+++ b/server/test/unit/server/controllers/test_importer.py
@@ -201,7 +201,7 @@ class TestCleanConfigDict(unittest.TestCase):
         """
         Ensure that usernames and passwords can contain the '@' symbol.
         """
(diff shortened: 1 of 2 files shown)


server/test/unit/server/controllers/test_importer.py
```diff
@@ -201,7 +201,7 @@ def test_feed_with_basic_auth_username_password_at_symbol(self):
         """
         Ensure that usernames and passwords can contain the '@' symbol.
         """
-        mock_config = {'feed': 'http://mock@user:mock@pass@realfeed.com'}
+        mock_config = {'feed': 'http://mock%40user:mock%40pass@realfeed.com'}
```
mhrivnak: Hah, it's funny that we even had a unit test that was *so close* to testing the right thing, but just didn't quite make it. Someone was thinking in the right direction though!

PR #2913, opened 2017-01-15
Problem: RSyncFastForwardUnitPublishStep publishes incorrectly

dkliban:
Solution: Enable RSyncFastForwardUnitPublishStep to publish the same type of units into different
directories by extracting the publish path of a unit from it's symlink location.

This problem was introduced when the yum distributor was modified to publish repositories with a
new layout. This patch makes the RPM rsync distributor consistent with the yum distributor.

re #1976
https://pulp.plan.io/issues/1976

### DIFF
--- a/server/pulp/plugins/util/publish_step.py
+++ b/server/pulp/plugins/util/publish_step.py
@@ -1307,11 +1307,17 @@ class RSyncFastForwardUnitPublishStep(UnitModelPluginStep):
         storage_dir = os.path.join(pulp_config.get('server', 'storage_dir'), 'content', 'units')
         relative_content_unit_path = os.path.relpath(item.storage_path, storage_dir)
         self.parent.content_unit_file_list.append(relative_content_unit_path)
-        filename = item.get_symlink_name()
+        if self.published_unit_path:
+            filename = item.get_symlink_name()
+            published_unit_path = self.published_unit_path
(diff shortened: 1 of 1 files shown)


server/pulp/plugins/util/publish_step.py
```diff
@@ -1307,11 +1307,17 @@ def process_main(self, item=None):
         storage_dir = os.path.join(pulp_config.get('server', 'storage_dir'), 'content', 'units')
         relative_content_unit_path = os.path.relpath(item.storage_path, storage_dir)
         self.parent.content_unit_file_list.append(relative_content_unit_path)
-        filename = item.get_symlink_name()
+        if self.published_unit_path:
+            filename = item.get_symlink_name()
+            published_unit_path = self.published_unit_path
+        else:
+            filename = os.path.split(item.get_symlink_name())[1]
+            published_unit_path = os.path.split(item.get_symlink_name())[0].split('/')
```
goosemania: Just for readability, what do you think about either this approach:
```
filename = os.path.basename(item.get_symlink_name())
published_unit_path = os.path.dirname(item.get_symlink_name()).split('/')
```
or that one:
```
dirname, filename = os.path.split(item.get_symlink_name())
published_unit_path = dirname.split('/')
```
?

PR #2914, opened 2017-01-16
Adding release note for orphan removal feature

daviddavis:
refs #1268
https://pulp.plan.io/issues/1268

### DIFF
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -21,6 +21,10 @@ New Features
 * Worker failure detection and high availability failover occurs within 30 seconds. Pulp processes are
   considered missing after 25 seconds, and heartbeats occur every 5 seconds.
 
+* When removing orphan content, the number of items being that have been removed is now stored as
+  the result on the orphan removal task. This count is broken down by content unit type. In
+  pulp-admin, the result can be looked up with the ``tasks details`` command.
+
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -21,6 +21,9 @@ New Features
 * Worker failure detection and high availability failover occurs within 30 seconds. Pulp processes are
   considered missing after 25 seconds, and heartbeats occur every 5 seconds.
 
+* When removing orphan content, the number of items being that have been removed is now stored as
+  the result on the orphan removal task. This count is broken down by content unit type.
```
seandst: This is also reported by pulp-admin, right? Does it make sense to change this to something like "stored as the result on the orphan removal task and reported by pulp-admin"?

seandst:
Build failure is clearly not this PR's fault, I'm happy to see jenkies results be ignored when merging.

docs/user-guide/release-notes/2.12.x.rst
daviddavis: It depends on what you mean by "reported by pulp." The result is stored on the task which you can look up via `pulp-admin tasks details ...`. The orphan remove command doesn't report the results though (although thinking over it, maybe it should?)

daviddavis:
@seandst added note about looking up result with pulp-admin.

PR #2916, opened 2017-01-16
Explictly log failovers

dralley:
Failover events are now logged as such
at the WARNING level

closes #2525

### DIFF
--- a/docs/user-guide/release-notes/2.12.x.rst
+++ b/docs/user-guide/release-notes/2.12.x.rst
@@ -25,6 +25,10 @@ New Features
   the result on the orphan removal task. This count is broken down by content unit type. In
   pulp-admin, the result can be looked up with the ``tasks details`` command.
 
+* Failover events for instances of pulp_resource_manager and pulp_celerybeat are now logged at the
+  WARNING level. The log message explicitly states that failover has occurred, along with the
+  name of the hot spare process which has taken over as primary.
+
 
 Deprecation
 -----------
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -136,7 +136,12 @@ def get_resource_manager_lock(name):
             lock.save()
 
             msg = _("Resource manager '%s' has acquired the resource manager lock") % name
-            _logger.info(msg)
+            _logger.debug(msg)
+
+            if not _first_check:
+                msg = _("Failover occurred: '%s' is now the primary resource manager") % name
+                _logger.warning(msg)
+
             break
         except mongoengine.NotUniqueError:
             # Only log the message the first time
--- a/server/pulp/server/async/scheduler.py
+++ b/server/pulp/server/async/scheduler.py
@@ -194,7 +194,7 @@ class Scheduler(beat.Scheduler):
         """
         self._schedule = None
         self._loaded_from_db_count = 0
-        self._retry_msg_already_logged = False
+        self._first_lock_acq_check = True
 
         # Force the use of the Pulp celery_instance when this custom Scheduler is used.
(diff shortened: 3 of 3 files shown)


docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -25,6 +25,9 @@ New Features
   the result on the orphan removal task. This count is broken down by content unit type. In
   pulp-admin, the result can be looked up with the ``tasks details`` command.
 
+* Failover events are now logged at the WARNING level with a message which explicitly states that
```
bmbouter: Can this specify which components this is for. Also it would be good to use the term 'hot spare' and 'primary' somehow.

bmbouter:
I started reviewing, but I'll stop until the work in progress label is removed.

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -25,6 +25,10 @@ New Features
   the result on the orphan removal task. This count is broken down by content unit type. In
   pulp-admin, the result can be looked up with the ``tasks details`` command.
 
+* Failover events for instances of pulp_resource_manager and pulp_celerybeat are now logged at the
+  WARNING level with a message which explicitly states that failover has occurred, along with the
```
bmbouter: Can this be broken into two sentences?  I suggest s/level with a message which/level. The log message/

docs/user-guide/release-notes/2.12.x.rst
```diff
@@ -25,6 +25,10 @@ New Features
...
+  name of the hot spare which has taken over as primary.
```
bmbouter: s/spare/spare process/

server/pulp/server/async/app.py
```diff
@@ -136,7 +136,12 @@ def get_resource_manager_lock(name):
             lock.save()
 
             msg = _("Resource manager '%s' has acquired the resource manager lock") % name
-            _logger.info(msg)
+            _logger.debug(msg)
+
+            if not _first_check:
+                msg = _("Failover occurred: '%s' is now the primary resource manager") % name
```
bmbouter: Nice i8ln.

server/pulp/server/async/app.py
```diff
@@ -136,7 +136,12 @@ def get_resource_manager_lock(name):
             lock.save()
 
             msg = _("Resource manager '%s' has acquired the resource manager lock") % name
-            _logger.info(msg)
+            _logger.debug(msg)
```
bmbouter: I like how this will log every 5 seconds. Nice.

dralley: Nah it won't.  Notice the ```lock.save()``` above it - if it succeeds it will never happen again because of the ```break```.  If it fails, it skips to the exception handler.

server/pulp/server/async/app.py
```diff
@@ -135,8 +135,13 @@ def get_resource_manager_lock(name):
             lock.timestamp = now
             lock.save()
 
-            msg = _("Resource manager '%s' has acquired the resource manager lock") % name
-            _logger.info(msg)
+            msg = _("Resource manager '%s' has acquired the resource manager lock" % name)
```
bmbouter: Oh this won't i8ln correctly. The string replacement has to happen on the response from `_()` not on the argument passed into `_()`.

dralley: I wasn't sure if you were being sarcastic given the other comment on the PR lol

server/pulp/server/async/app.py
bmbouter: oh I see. OK let's leave it as is

PR #2918, opened 2017-01-19
Fix relative path issue and exit tests on missing dependencies.

sysadmind:
This relates to https://pulp.plan.io/issues/2444 and trying to get pulp
running on other operating systems and installing through pip. These changes
were tested on the docker image python:2.

Trying to run the tests for the first time, I was met with a few missing dependencies. It seems like this is something that the test runner should check instead of giving a generic no file or directory error.

### DIFF
--- a/devel/pulp/devel/test_runner.py
+++ b/devel/pulp/devel/test_runner.py
@@ -12,6 +12,27 @@ import sys
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pulp.server.webservices.settings'
 
 
+# From: http://stackoverflow.com/questions/377017/test-if-executable-exists-in-python/377028#377028
+def which(program):
+    import os
+
+    def is_exe(fpath):
+        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
+
+    fpath, fname = os.path.split(program)
+    if fpath:
+        if is_exe(program):
+            return program
+    else:
+        for path in os.environ["PATH"].split(os.pathsep):
+            path = path.strip('"')
+            exe_file = os.path.join(path, program)
+            if is_exe(exe_file):
+                return exe_file
+
+    return None
+
+
 def run_tests(packages, tests_all_platforms, tests_non_rhel5,
               flake8_paths=None, flake8_exclude=None):
     """
@@ -33,6 +54,15 @@ def run_tests(packages, tests_all_platforms, tests_non_rhel5,
     :return: the exit code from nosetests
     :rtype:  integer
     """
+
+    # Test to make sure the necessary executables exist
+    if which('flake8') is None:
+        print 'flake8 not found or is not executable! Please correct before running tests.'
+        exit(2)
+    if which('nosetests') is None:
+        print 'nosetests not found or is not executable! Please correct before running tests.'
+        exit(2)
+
     parser = argparse.ArgumentParser()
     parser.add_argument('--xunit-file')
     parser.add_argument('--with-xunit', action='store_true')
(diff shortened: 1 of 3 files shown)


dkliban:
ok test

sysadmind:
@dkliban I checked out this branch in the devel vagrant as well and it also passed all tests locally.

dkliban:
ok test

dkliban:
ok test

dkliban:
The commit message title and body are too long for the guidelines.

dkliban:
ok test

dkliban:
ok test

devel/setup.py
```diff
@@ -4,10 +4,11 @@
 
 PYTHON_MAJOR_MINOR = '%s.%s' % (sys.version_info[0], sys.version_info[1])
 
+requires = {'install_requires': ['mock']}
```
dkliban: Pulp test require mock<1.1

sysadmind:
@dkliban I modified the mock dependency. I hope that solves the build issue.

dkliban:
ok test

sysadmind:
@dkliban Fixed the setup.py test failures.

sysadmind:
@dkliban Looks like the tests pass now.

dkliban:
ok test

devel/pulp/devel/test_runner.py
```diff
@@ -12,6 +12,27 @@
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pulp.server.webservices.settings'
 
 
+# From: http://stackoverflow.com/questions/377017/test-if-executable-exists-in-python/377028#377028
```
bmbouter: Can this link be moved into a docstring for the function? Also can you add some new lines linking to the [license of the content](http://stackoverflow.com/help/licensing) and [that license itself](https://creativecommons.org/licenses/by-sa/3.0/). This snippet will need to retain that license.

Another option is to make a new implementation that would be licensed like the rest of Pulp as gplv2.

devel/pulp/devel/test_runner.py
```diff
@@ -12,6 +12,27 @@
...
+def which(program):
+    import os
```
bmbouter: Can this be moved to the imports section please.

devel/pulp/devel/test_runner.py
```diff
@@ -12,6 +12,27 @@
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pulp.server.webservices.settings'
 
 
+# From: http://stackoverflow.com/questions/377017/test-if-executable-exists-in-python/377028#377028
+def which(program):
```
bmbouter: This needs a docstring too.

devel/pulp/devel/test_runner.py
bmbouter: After getting some advice from legal, I need to ask for a new, original implementation. Since Pulp is distributed as GPLv2 and this is licensed as CC-A-SA to accept this implementation would cause Pulp to distribute under more than just GPLv2. Since its only one function, it's better to rewrite it as a GPLv2 implementation than to have Pulp distribute using a plurality of licenses.

sysadmind: Thanks for the feedback. I'll write up a new version so that it can be incorporated into pulp without license issues.

dralley:
I'd like to point out that Python 3 has an implementation of "which" in the standard library under the ```shutil``` package.

https://docs.python.org/3/library/shutil.html#shutil.which

If this doesn't end up getting pulled into the Pulp 2.x line, it's very easy to do so for Pulp 3.

bmbouter:
@sysadmind are you still interested in pushing a fix for this? We're trying to finish or close some of the PRs that have been open for a while. Thanks!

sysadmind:
I think with the information above, I'll re-evaluate this for pulp3. It's really only to help development and since that's more pulp3 now, I can see if this is even necessary on that branch.

PR #2920, opened 2017-01-20
Problem: Stale worker documents present in the db

dkliban:
Solution: A custom queryset for the Worker model allows filtering out
worker records which have not been updated in more than 25 seconds.
This queryset is used in two places:

- Status API

- Resource manager code that looks for available workers

closes #2496
https://pulp.plan.io/issues/2496

### DIFF
--- a/server/pulp/server/db/querysets.py
+++ b/server/pulp/server/db/querysets.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timedelta
 from gettext import gettext as _
 import operator
 
@@ -5,6 +6,8 @@ from mongoengine import Q
 from mongoengine.queryset import DoesNotExist, QuerySetNoCache
 from pymongo import ASCENDING
 
+from pulp.common import constants
+from pulp.common.dateutils import ensure_tz
 from pulp.server import exceptions as pulp_exceptions
 
 
@@ -114,6 +117,27 @@ class CriteriaQuerySet(QuerySetPreventCache):
             return self.get(**kwargs)
         except DoesNotExist:
             raise pulp_exceptions.MissingResource(**kwargs)
+
+
+class WorkerQuerySet(CriteriaQuerySet):
+    """
+    Custom queryset for workers
+    """
+
+    def get_online(self):
+        """
+        Returns a queryset with a subset of Worker documents.
+
+        The queryset is filtered to remove any Worker document that has not been updated in the
+        last 25 seconds.
+
+        :return: mongoengine queryset object
+        :rtype:  mongoengine.queryset.QuerySet
+        """
+        query_set = self
+        now = ensure_tz(datetime.utcnow())
+        oldest_heartbeat_time = now - timedelta(seconds=constants.PULP_PROCESS_TIMEOUT_INTERVAL)
+        return query_set.filter(last_heartbeat__gte=oldest_heartbeat_time)
 
 
 class RepoQuerySet(CriteriaQuerySet):
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -211,7 +211,7 @@ def _get_unreserved_worker():
     """
 
     # Build a mapping of queue names to Worker objects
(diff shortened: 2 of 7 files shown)


server/pulp/server/db/querysets.py
```diff
@@ -116,6 +119,27 @@ def get_or_404(self, **kwargs):
             raise pulp_exceptions.MissingResource(**kwargs)
 
 
+class WorkerQuerySet(CriteriaQuerySet):
+    """
+    Custom queryset for workers
+    """
+
+    def get_workers_online(self):
```
bmbouter: What about naming this `get_online()` or just `online()`?

If changed it would read as either `Worker.objects.get_online()` or `Worker.objects.online()`

dkliban: I prefer get_online(). I'll update the PR with that.

server/pulp/server/db/querysets.py
```diff
@@ -116,6 +119,27 @@ def get_or_404(self, **kwargs):
...
+        """
+        Returns a queryset that contains only worker documents that have been updated in the last
```
bmbouter: The first line of a multiline docstring should be a single line. Please shorten this to 1 line, or add a shorter 1 line above with a whitespace between. See the PEP here:  https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

bmbouter: s/worker/Worker/

server/pulp/server/db/querysets.py
bmbouter: thx

server/pulp/server/db/querysets.py
```diff
@@ -116,6 +119,27 @@ def get_or_404(self, **kwargs):
...
+        25 seconds.
+
+        :return: mongoengine queryset object
+        :rtype:  mongoengine.queryset.QuerySet
+        """
+        query_set = self
```
bmbouter: This read strangely to me. I think having L#138 be `self.filter....` would be fine. This is not a required change.

bmbouter: Actually, maybe this is better and should be kept. It is explicit.

bmbouter:
The commit message title and body are too long for the guidelines.

bmbouter:
This has test failure which need to be fixed.

dkliban:
ok test

dkliban:
ok test

server/pulp/server/db/querysets.py
```diff
@@ -116,6 +119,28 @@ def get_or_404(self, **kwargs):
...
+class WorkerQuerySet(CriteriaQuerySet):
+    """
+    Custom queryset for workers
+    """
+
+    def get_online(self):
+        """
+        Returns a queryset with a subset of Worker documents.
+
+        The queryset is filtered to remove any Worker document that have not been updated in the
```
bmbouter: s/have/has/

dkliban: any [one]  ... i'll push the change shortly.

server/pulp/server/db/querysets.py
```diff
@@ -116,6 +119,28 @@ def get_or_404(self, **kwargs):
...
+        The queryset is filtered to remove any Worker document that has not been updated in the
+        last 25 seconds.
+
+        :return: mongoengine queryset object
+        :rtype:  mongoengine.queryset.QuerySet
+        """
+        query_set = self
+        now = ensure_tz(datetime.utcnow())
+        oldest_heartbeat_time = now - timedelta(seconds=constants.PULP_PROCESS_TIMEOUT_INTERVAL)
+        query_set.filter(last_heartbeat__gte=oldest_heartbeat_time)
```
bmbouter: The return value of this statement needs to be returned. As it's written now the original queryset is returned. Remove L#141 and make this `return query_set.filter(last_heartbeat__gte=oldest_heartbeat_time)`

PR #2922, opened 2017-01-20
Workers write their own hearbeat records to database.

dralley:
All workers will write their own records to the database
instead of relying on pulp_celerybeat to do so for them
using celery heartbeats.

This patch makes use of the Consumer blueprint that celery
runs at the start time of a worker. An extra boot step has
been added which sets a timer to periodically update the
worker record in the database.

http://docs.celeryproject.org/en/master/userguide/extending.html
https://groups.google.com/d/msg/celery-users/3fs0ocREYqw/C7U1lCAp56sJ

closes #2519
https://pulp.plan.io/issues/2519
closes #2516
https://pulp.plan.io/issues/2516

### DIFF
--- a/docs/user-guide/release-notes/2.13.x.rst
+++ b/docs/user-guide/release-notes/2.13.x.rst
@@ -11,3 +11,23 @@ New Features
 * Log statements which are emitted from task processes now include a short task
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+* Users no longer need to have pulp_celerybeat running to know whether pulp_workers or
+  pulp_resource_manager are online. All workers now write their own heartbeat records to the
+  database, eliminating the dependence on pulp_celerybeat.
+
+Bug Fixes
+---------
+
+See the list of :fixedbugs:`2.13.0`
+
+Upgrade instructions
+--------------------
+
+Upgrade using the normal process::
+
+    $ sudo systemctl stop httpd pulp_workers pulp_resource_manager pulp_celerybeat pulp_streamer goferd
+    $ sudo yum upgrade
+    $ sudo -u apache pulp-manage-db
+    $ sudo systemctl start httpd pulp_workers pulp_resource_manager pulp_celerybeat pulp_streamer goferd
+
+``pulp_streamer`` and ``goferd`` should be omitted if those services are not installed.
--- a/docs/user-guide/troubleshooting.rst
+++ b/docs/user-guide/troubleshooting.rst
@@ -90,12 +90,12 @@ Task ID
 
 Since Pulp is a multi-process application, Pulp will attempt to log a shortened 8 character version
 of the task id if the log entry is emitted from a task process. This can be used to look up the task
-in ``pulp-admin tasks`` commands.  An example of a log statement with a task id would be:
+in ``pulp-admin tasks`` commands.  An example of a log statement with a task id would be::
 
     Jan 20 23:12:02 myhost pulp[30687]: pulp_rpm.plugins.importers.yum.sync:INFO: [338bdde4] Downloading 32 RPMs
 
 Here the shortened task id is ``338bdde4``. The full task id would be
-``338bdde4-608a-44ab-a79c-49c28b0fe037``. A statement without a task id would look like:
+``338bdde4-608a-44ab-a79c-49c28b0fe037``. A statement without a task id would look like::
 
     Jan 21 23:12:02 myhost pulp[30482]: pulp.server.async.worker_watcher:INFO: New worker 'resource_manager@myhost' discovered
 
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -9,13 +9,14 @@ import time
 from datetime import datetime, timedelta
 from gettext import gettext as _
 
-import mongoengine
+from celery import bootsteps
 from celery.signals import celeryd_after_setup, worker_shutdown
+import mongoengine
 
 from pulp.common import constants, dateutils
 from pulp.server import initialization
-from pulp.server.async import tasks
-from pulp.server.db.model import ResourceManagerLock, Worker
+from pulp.server.async import tasks, worker_watcher
+from pulp.server.db.model import Worker, ResourceManagerLock
 from pulp.server.managers.repo import _common as common_utils
 
 # This import will load our configs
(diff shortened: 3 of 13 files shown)


dralley:
ok test

daviddavis:
@dalley I think I just caused a conflict by merging a PR with `docs/user-guide/release-notes/2.13.x.rst`. Apologies!

bmbouter:
This has flake8 errors that need to be fixed.

bmbouter:
These docs build with an error because the new 2.13.rst file is not included in the index.rst. When building docs locally it says:

`/home/vagrant/devel/pulp/docs/user-guide/release-notes/2.13.x.rst:: WARNING: document isn't included in any toctree`

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,92 @@
 _logger = logging.getLogger(__name__)
 
 
+class HeartbeatStep(bootsteps.StartStopStep):
+
```
bmbouter: Can this whitespace be removed. The PEP for docstrings doesn't show it.

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,92 @@
...
+    """
+    This class is a celery "Blueprint". It extends the functionality of the celery
```
bmbouter: Can a 1-line summary description be added, and then a whitespace. See the reference for multi-line docstrings here:  https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,92 @@
...
+            self._record_heartbeat,
+            (worker, ),
+            priority=10,
+        )
+        self._record_heartbeat(worker)
+
+    def stop(self, worker):
+        """
+        This method is called every time the worker is restarted (i.e. connection is lost)
+        and also at shutdown. The worker calls stop() at shutdown only.
```
bmbouter: The last sentence here is confusing. Can this be checked for clarity/correctness.

docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -8,6 +8,26 @@ Pulp 2.13.0
 New Features
 ------------
 
-* Log statements which are emitted from task processes now include a short task
+- Log statements which are emitted from task processes now include a short task
```
bmbouter: What about switching these to `*` for consistency with all other release notes?

docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -8,6 +8,26 @@ Pulp 2.13.0
...
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+- Users will no longer need to have pulp_celerybeat running to know whether pulp_workers and
```
bmbouter: s/will//
s/and/or/

docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -8,6 +8,26 @@ Pulp 2.13.0
...
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+- Users will no longer need to have pulp_celerybeat running to know whether pulp_workers and 
+  pulp_resource_manager are online. All workers now write their own heartbeat records to the
+  database, eliminating the dependence on pulp_celerybeat.
+
+Bug Fixes
+---------
+
+See the list of :fixedbugs:`2.13.0`
```
bmbouter: This all looks great, Thanks for making this

docs/user-guide/troubleshooting.rst
```diff
@@ -97,7 +97,7 @@ in ``pulp-admin tasks`` commands.  An example of a log statement with a task id
 Here the shortened task id is ``338bdde4``. The full task id would be
 ``338bdde4-608a-44ab-a79c-49c28b0fe037``. A statement without a task id would look like:
```
bmbouter: You didn't introduce this, but when I rendered these docs the indendented portions here aren't blockquoted. It's because the `:` needs to become a `::` instead. If you render the docs locally (`make clean;make html`) you'll see what I mean. The same problem exists also on L#93.

server/pulp/server/async/app.py
```diff
@@ -10,12 +10,13 @@
 from gettext import gettext as _
 
 import mongoengine
-from celery.signals import celeryd_after_setup, worker_shutdown
+from celery import bootsteps
```
bmbouter: these should come before mongoengine for alphabetical. It's kind of ugly looking but we do mix the two import styles if they belong in the same import block.

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,93 @@
 _logger = logging.getLogger(__name__)
 
 
+class HeartbeatStep(bootsteps.StartStopStep):
+    """
+    Adds pulp heartbeat updating to celery workers.
```
bmbouter: thank you

server/pulp/server/async/app.py
```diff
@@ -30,6 +31,93 @@
...
+
+    :param worker: The worker instance (unused)
+    :type  worker: celery.apps.worker.Worker
+    """
+
+    requires = ('celery.worker.components:Timer', )
+
+    def __init__(self, worker, **kwargs):
+        """
+        The step init is called when the worker instance is created, It is called with the
```
bmbouter: Same 1-line summary needed for any of these multiline docstrings in this.

server/pulp/server/async/app.py
```diff
@@ -72,7 +166,9 @@ def initialize_worker(sender, instance, **kwargs):
     initialization.initialize()
 
     # Delete any potential old state
-    tasks._delete_worker(sender, normal_shutdown=True)
+    existing_worker = Worker.objects(name=sender).first()
+    if existing_worker:
```
bmbouter: This is susceptible to a race condition. Rather than checking and doing something conditional based on that, try to have the all of the work occur in the database. If you want a more specific idea let me know.

server/pulp/server/async/worker_utils.py
```diff
@@ -0,0 +1,59 @@
+"""
```
bmbouter: Rather than a delete and then add, can this be done with a `git mv`. I think as it is now, the git history will not be preserved as cleanly.

bmbouter: Also it's difficult for me to see which lines have been modified which causes the review to be longer.

dralley: I did actually use git mv, so I'm not sure why it's done this.  Maybe because I made substantial changes to the file?

server/test/unit/server/async/test_worker_utils.py
```diff
@@ -0,0 +1,48 @@
+import unittest
```
bmbouter: Similarly with this file (and any files that get renamed). Can this be done with `git mv`. Thanks!

server/pulp/server/async/app.py
dralley: I see what you're saying, but couldn't a new worker could just as easily start up during the execution of _delete_worker()? 

Speaking of which, would it make sense to move _delete_worker() from tasks.py to worker_utils.py, and get rid of the prepending _, since not an actual celery task, and it's being used in tons of places outside of tasks.py?

The same applies to several other free functions in tasks.py.

bmbouter: I believe the implementation in [`_delete_worker()`](https://github.com/bmbouter/pulp/blob/6782dfe8a87593dd5579f0bc4a87dea00f4235ea/server/pulp/server/async/tasks.py#L258-L269) is race condition free. By calling `delete()` on a queryset causes both the query and delete operation to occur atomically in the database. Since it's atomic it will either delete the right thing if it was there or there is nothing to delete.

dralley: _delete_worker() does more than just delete the worker record though, it cancels tasks, deletes locks, releases reserved resources, etc.

I assume we want to do all of those things, so we can't just do queryset.delete() here

bmbouter: In hindsight we should have avoided moving this code around. On Pulp3 this code has been significantly reorganized already and that work is done. I only thought of this because your idea to move non task code like `_delete_worker()` and remove the _ is exactly what we did. In thinking about this more, these moves will cause merge conflicts next week when I go to merge master -> 3.0-dev which we will then have to manually resolve. I suspect we should undo the non-required moves in this PR. We could also leave this as-is, but I'll need you to resolve the merge conflicts next week so the work is unavoidable. Which of these options (or another option) would you prefer?

server/pulp/server/async/scheduler.py
```diff
@@ -209,27 +159,29 @@ def spawn_pulp_monitor_threads(self):
...
         worker_timeout_monitor.start()
 
-    @staticmethod
     def call_tick(self, celerybeat_name):
+        """
+        Call the superclass tick method and log a debug message.
+
+        :param celerybeat_name: hostname of the celerybeat instance
+        :type  celerybeat_name: basestring
+        :return:                seconds until the next tick
```
bmbouter: I usually separate arguments from return params with a line of whitespace. Just a suggestion.

server/pulp/server/async/app.py
dralley: I suppose that if our current prime directive is stability, it would be better to do it in the PR (which, it's probably easier to do now than later anyway).

I will just put it back to the way it was before: I changed it because I figured it cause unnecessary and nominally inaccurate logging, but it's not a big deal and probably not worth the hassle.

server/pulp/server/async/worker_utils.py
bmbouter: I can see from the blame log that the history is preserved correctly. Next time I'll check that before suggesting `git mv` is used (which you already did). I looked at the different commits so please go ahead and squash them. git and github both presented the `git mv` as a delete and add so I've reached two conclustion. 1. github is just doing what git does and 2. this an area git could improve on. Thanks for clarifying this.

dralley:
ok test

server/pulp/server/async/app.py
bmbouter: Responding to 3 comments ago. Yes we can't call delete here for those reasons. I'm using deductive reasons to point out that reading in database state and having conditional logic on it is a pattern for a race condition.

dralley:
ok test

server/pulp/server/async/worker_watcher.py
```diff
@@ -17,97 +17,43 @@
...
 from pulp.server.async.tasks import _delete_worker
-from pulp.server.db.model import Worker, ResourceManagerLock
+from pulp.server.db.model import Worker
 
 
 _logger = logging.getLogger(__name__)
 
 
-def _parse_and_log_event(event):
+def handle_worker_heartbeat(worker_name):
```
bmbouter: Can a log statement be added here to log at DEBUG level with the name of the worker? This would be like the log statement in [Issue 2516](https://pulp.plan.io/issues/2516). It probably should also include the `datetime.utcnow()` timestamp that is being written to the database even though the logs already probably contain timestamps.

bmbouter:
The PR runner is complaining of flake8 errors which caused failures. Those need to be fixed.

bmbouter:
Everything tested great in terms of recording their own heartbeats, but I see one regression that I think may be introduced by this PR.

1. start with a fully stopped Pulp system
2. start the resource manager `sudo systemctl start pulp_resource_manager`
3. observe it is recording it's own heartbeat with `pulp-admin status`
4. Stop the resource manager gracefully `sudo systemctl stop pulp_resource_manager`
5. check the status again `pulp-admin status` and observe the Worker record is still there.
6. I expected that upon graceful shutdown it would remove its own record

After 30 seconds pulp-status filters out those records, but it should occur immediately for graceful shutdown. Also I do not observe this behavior for the `pulp_workers` or `pulp_celerybeat` both which do graceful cleanup nicely.

dralley:
ok test

dralley:
ok test

bmbouter:
ok test

PR #2324, opened 2016-01-12
Ensure file objects are cleaned up on error

pcreech:
Plugins save then call import_content on FileContentUnit objects.
This ensures that if an error happens on import_content, we also
clean up after ourselves.

re #[1457](https://pulp.plan.io/issues/1457)
https://pulp.plan.io/issues/1457

PR #2924, opened 2017-01-20
Appending 8 character task id to log output

daviddavis:
Appending a 8 character task id to log output if there is a relevant
task id.

fixes #2324
https://pulp.plan.io/issues/2324

TODO
- [x] Add a release note
- [x] Update troubleshooting guide
- [x] Add two examples to the troubleshooting guide
- [x] Address unit tests

### DIFF
--- a/docs/user-guide/troubleshooting.rst
+++ b/docs/user-guide/troubleshooting.rst
@@ -85,6 +85,20 @@ processes to write to the same file path, and so Pulp had to do something differ
 widely used logging protocol, and given the distributed nature of Pulp it was the most appropriate
 logging solution available.
 
+Task ID
+^^^^^^^
+
+Since Pulp is a multi-process application, Pulp will attempt to log a shortened 8 character version
+of the task id if the log entry is emitted from a task process. This can be used to look up the task
+in ``pulp-admin tasks`` commands.  An example of a log statement with a task id would be:
+
+    Jan 20 23:12:02 myhost pulp[30687]: pulp_rpm.plugins.importers.yum.sync:INFO: [338bdde4] Downloading 32 RPMs
+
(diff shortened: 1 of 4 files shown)


docs/user-guide/troubleshooting.rst
```diff
@@ -85,6 +85,20 @@ processes to write to the same file path, and so Pulp had to do something differ
 widely used logging protocol, and given the distributed nature of Pulp it was the most appropriate
 logging solution available.
 
+Task ID
+^^^^^^^
+
+Since Pulp is a multi-threaded application, Pulp will attempt to log a shortened 8 character version
```
jortel: Perhaps `multi-process application` would be more relevant since the subject is tasking.

server/pulp/server/logs.py
```diff
@@ -84,6 +88,29 @@ def stop_logging():
...
+    This formatter changes format depending on whether there is a task_id in
+    the LogRecord
+    """
+    def format(self, record):
+        if record.task_id is None:
+            self._fmt = LOG_FORMAT_STRING
+        else:
+            record.task_id = record.task_id[:8]
+            self._fmt = TASK_LOG_FORMAT_STRING
+        return super(TaskLogFormatter, self).format(record)
```
jortel: Looks like in python 2.6 (RHEL 6) the `logging.Formatter` is not a new-style object so you cannot use `super()`.

daviddavis:
@jortel thanks for the review. I made a change to the logging stuff but am not sure if that's the best way to solve the `super` issue. Let me know. Thanks.

jortel:
You just need to do it the old way.
```
return logging.Formatter.format(self, record)
```

daviddavis:
@jortel, thank you. updated!

PR #2927, opened 2017-01-23
Convert all default CA_PATH settings to work with multiple distibutio…

sysadmind:
…n default locations.

This is related to :
https://pulp.plan.io/issues/2538
https://pulp.plan.io/issues/2444

This is the first step towards building a working cross distibution pulp
package. This should start to support similar cross distibution files
for other defaults in configuration settings.

This PR removes 2 tests that were based around assumptions that are no longer valid with the change in how CA_PATH is defaulted. I would like to add new tests to counter act this removal, but I'm not sure about how best to go about it. I'm very open to suggestions on testing and the make up of this change in general. I chose to use the defaults in pulp.common.constants as it keeps this change fairly small and reasonable, but it does mean that all of these packages will require pulp.common. I assume that this is a likely scenario even without this PR, but it should be noted.

### DIFF
--- a/bindings/test/unit/test_server.py
+++ b/bindings/test/unit/test_server.py
@@ -9,6 +9,7 @@ from M2Crypto import m2, SSL
 import mock
 
 from pulp.bindings import exceptions, server
+from pulp.common.constants import DEFAULT_CA_PATH_LIST
 
 
 class TestHTTPSServerWrapper(unittest.TestCase):
@@ -210,9 +211,9 @@ class TestPulpConnection(unittest.TestCase):
         self.assertTrue(isinstance(connection.server_wrapper, server.HTTPSServerWrapper))
         self.assertEqual(connection.server_wrapper.pulp_connection, connection)
         self.assertEqual(connection.verify_ssl, True)
-        self.assertEqual(connection.ca_path, server.DEFAULT_CA_PATH)
+        self.assertTrue(connection.ca_path in DEFAULT_CA_PATH_LIST)
         # 1142376 - verify default path points to a known valid file
-        self.assertEqual(server.DEFAULT_CA_PATH, '/etc/pki/tls/certs/ca-bundle.crt')
+        self.assertTrue(server.DEFAULT_CA_PATH is not None)
 
     def test___init___ca_path_set(self):
         """
--- a/client_consumer/pulp/client/consumer/config.py
+++ b/client_consumer/pulp/client/consumer/config.py
@@ -1,7 +1,8 @@
 import os
 import socket
 
-from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL
+from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL, get_from_default_paths
+from pulp.common.constants import DEFAULT_CA_PATH_LIST
 
 
 DEFAULT = {
@@ -11,7 +12,7 @@ DEFAULT = {
         'api_prefix': '/pulp/api',
         'rsa_pub': '/etc/pki/pulp/consumer/server/rsa_pub.key',
         'verify_ssl': 'true',
-        'ca_path': '/etc/pki/tls/certs/ca-bundle.crt',
+        'ca_path': get_from_default_paths(DEFAULT_CA_PATH_LIST),
     },
     'authentication': {
         'rsa_key': '/etc/pki/pulp/consumer/rsa.key',
--- a/client_consumer/pulp/client/consumer/config.py
+++ b/client_consumer/pulp/client/consumer/config.py
@@ -1,8 +1,8 @@
 import os
 import socket
 
-from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL, get_from_default_paths
-from pulp.common.constants import DEFAULT_CA_PATH_LIST
+from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL
+from pulp.common.constants import DEFAULT_CA_PATH
 
 
 DEFAULT = {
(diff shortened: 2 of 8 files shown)


dkliban:
ok test

sysadmind:
@dkliban I was finally able to get my pulp local vagrant running and all tests pass for me.

common/pulp/common/constants.py
```diff
@@ -48,8 +48,12 @@
 PRIMARY_ID = '___/primary/___'
 
 # this is used by both platform and plugins to find the default CA path
-DEFAULT_CA_PATH = '/etc/pki/tls/certs/ca-bundle.crt'
-
+DEFAULT_CA_PATH_LIST = [
```
seandst: It might be reasonable to leave this as `DEFAULT_CA_PATH`, and set its value at run-time with `get_from_default_paths`. Otherwise, it doesn't seem useful to define this as a constant if the only useful way to access it is by calling `get_from_default_paths` on its value. Part of the motivation for this is that this constant is imported by the RPM plugin, so this change will break that plugin, maybe more.

Maybe something like this is reasonable?
```python
DEFAULT_CA_PATH_LIST = [...]
# set DEFAULT_CA_PATH based on DEFAULT_CA_PATH_LIST for backward compatibility with plugins
for path in DEFAULT_CA_PATH_LIST:
    if os.path.exists(path):
        DEFAULT_CA_PATH = path
        break
else:
    DEFAULT_CA_PATH = None
```

I've specifically avoided using `get_from_default_paths` here; just about everything in pulp imports from constants, which would create circular import problems almost anywhere the function lived.

bindings/test/unit/test_server.py
```diff
@@ -210,9 +210,6 @@ def test___init___defaults(self):
         self.assertTrue(isinstance(connection.server_wrapper, server.HTTPSServerWrapper))
         self.assertEqual(connection.server_wrapper.pulp_connection, connection)
         self.assertEqual(connection.verify_ssl, True)
-        self.assertEqual(connection.ca_path, server.DEFAULT_CA_PATH)
-        # 1142376 - verify default path points to a known valid file
-        self.assertEqual(server.DEFAULT_CA_PATH, '/etc/pki/tls/certs/ca-bundle.crt')
```
seandst: These assertions could probably be adapted rather than removed.

For line #213, it can maybe be changed to `self.assertTrue(connection.ca_path in server.DEFAULT_CA_PATH_LIST)`

For line #215, it seems important to have something verifying that the default CA path is set to a file that actually exists on a given platform. A new unit test covering the behavior of `get_from_default_paths` is necessary to ensure that that function does not return `None`, so that the tests appropriately fail if/when some distribution doesn't place the CA bundle in a location known to pulp.

sysadmind:
@seandst Thank you for taking the time to review this PR. I really appreciate it. I looked over your feedback and made some changes. I dropped the get_from_default_paths function all together in favor of your backwards compatible recommendation. I also adjusted the unit tests that I previously removed and added a new one for the pulp.common.constants module.

client_consumer/pulp/client/consumer/config.py
```diff
@@ -1,7 +1,8 @@
 import os
 import socket
 
-from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL
+from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL, get_from_default_paths
```
seandst: The jenkies tests are failing because of this import, I think it's the last one left to remove.

common/pulp/common/constants.py
```diff
@@ -48,7 +49,19 @@
...
+    '/etc/ssl/ca-bundle.pem',              # OpenSUSE
+    '/etc/pki/tls/cacert.pem',             # OpenELEC
+]
+# set DEFAULT_CA_PATH based on DEFAULT_CA_PATH_LIST for backward compatibility with plugins
+for path in DEFAULT_CA_PATH_LIST:
+    if os.path.exists(path):
+        DEFAULT_CA_PATH = path
+        break
+else:
+    DEFAULT_CA_PATH = None
```
seandst: It just occurred to me that it might be nice to have a comment on this, indicating that no CA path was found, and we probably need a new entry in the CA path list. What do you think?

sysadmind: I agree completely. I'll write something up.

seandst:
I think this looks good now, but it looks like jenkins is having some network problems. Once those clear up, I'd love to see the tests pass on all platforms before merging.

seandst:
ok test

bindings/test/unit/test_server.py
```diff
@@ -210,9 +211,9 @@ def test___init___defaults(self):
         self.assertTrue(isinstance(connection.server_wrapper, server.HTTPSServerWrapper))
         self.assertEqual(connection.server_wrapper.pulp_connection, connection)
         self.assertEqual(connection.verify_ssl, True)
-        self.assertEqual(connection.ca_path, server.DEFAULT_CA_PATH)
+        self.assertTrue(connection.ca_path in DEFAULT_CA_PATH_LIST)
         # 1142376 - verify default path points to a known valid file
-        self.assertEqual(server.DEFAULT_CA_PATH, '/etc/pki/tls/certs/ca-bundle.crt')
+        self.assertIsNotNone(server.DEFAULT_CA_PATH)
```
seandst: This doesn't work in el6 because it's using py2.6's unittest.
`self.assertTrue(server.DEFAULT_CA_PATH is not None)` is probably the most readable alternative.

There's another one over in `common/test/unit/common/test_constants.py` that will need to be fixed.

sysadmind:
@seandst Tests updated for EL6

seandst:
ok test

sysadmind:
@seandst Do you have a log of the test failures?

seandst:
"(295 failures / +121)" I don't think this one's on you... :)

seandst:
ok test

seandst:
This is still causing problems:

```python
from pulp.common.config import Config, REQUIRED, ANY, NUMBER, BOOL, OPTIONAL, get_from_default_paths
ImportError: cannot import name get_from_default_paths
```
in `client_consumer/pulp/client/consumer/config.py`

seandst:
ok test

seandst:
ok test

seandst:
ok test

seandst:
lookit the green check!

seandst:
If you haven't yet, get yourself in the authors file:
https://github.com/pulp/pulp/blob/master/AUTHORS

While you're in there, maybe add me, too. :D (just kidding)

sysadmind:
@seandst Thanks again for the review and your help to get this merged. I did add myself in a previous commit to the authors file. Much appreciated! :tada:

PR #2929, opened 2017-01-23
Fixes the cProfile Feature

bmbouter:
This feature was introduced in 2.12, and is being fixed prior to the
release of 2.12.0. This fix includes:

* Updates the server.conf which used `enable` instead of `enabled`
* Updates the config file checking which treated received a str when it
  expected a boolean.
* Provides automatic directory creation.
* Documentation updates and fix for blockquote sections which did not
  render correctly.

https://pulp.plan.io/issues/1939
closes #1939

### DIFF
--- a/docs/dev-guide/debugging.rst
+++ b/docs/dev-guide/debugging.rst
@@ -6,24 +6,36 @@ Debugging
 Task Performance Analysis
 -------------------------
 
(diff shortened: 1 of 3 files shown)


bmbouter:
Test failure is unrelated and Travis failure is linking a conf file so that is also not a true failure.

PR #2931, opened 2017-01-24
Only update last_publish if success and not canceled

daviddavis:
Only update the last_publish field is the publish was a success and not
canceled.

fixes #2263
https://pulp.plan.io/issues/2263

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1272,9 +1272,10 @@ def _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_confi
     dist = model.Distributor.objects.get_or_404(
         repo_id=repo_obj.repo_id, distributor_id=dist_id)
 
-    # Use raw pymongo not to fire the signal hander
-    model.Distributor.objects(repo_id=repo_obj.repo_id, distributor_id=dist_id).\
-        update(set__last_publish=publish_end_timestamp)
+    if not publish_report.canceled_flag:
+        # Use raw pymongo not to fire the signal hander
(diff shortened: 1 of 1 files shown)


daviddavis:
To test, here's a patch for pulp_rpm you can apply to give you time to cancel an rpm sync during publish:

https://gist.github.com/daviddavis/f05674b81d9376a4cc42eff1e498b7d3

daviddavis:
I'm a bit leery of this changes—namely I wonder if there's other logic in `_do_publish` that we should NOT be executing if success is false or cancel is true:

https://github.com/daviddavis/pulp/blob/d5c4c64066a9cb2097a5452ca499f7666a7b4aa3/server/pulp/server/controllers/repository.py#L1214

daviddavis:
ok test

server/pulp/server/controllers/repository.py
```diff
@@ -1272,9 +1272,10 @@ def _do_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_confi
     dist = model.Distributor.objects.get_or_404(
         repo_id=repo_obj.repo_id, distributor_id=dist_id)
 
-    # Use raw pymongo not to fire the signal hander
-    model.Distributor.objects(repo_id=repo_obj.repo_id, distributor_id=dist_id).\
-        update(set__last_publish=publish_end_timestamp)
+    if publish_report.success_flag and not publish_report.canceled_flag:
```
dkliban: You don't need to check the success_flag here. If it's false, an exception is raised on line 1247.

daviddavis: Ah, good catch.

PR #2934, opened 2017-01-25
Fix time_from_last_worker calculation

elyezer:
Call `datetime.utcnow()` to get the current time because the
`last_heartbeat` is stored on database on UTC timezone.
    
Fix #2545

### DIFF
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -200,7 +200,7 @@ def main():
         active_workers = status.get_workers()
         if active_workers:
             last_worker_time = max([worker['last_heartbeat'] for worker in active_workers])
-            time_from_last_worker = UTCDateTimeField().to_python(datetime.now()) - last_worker_time
+            time_from_last_worker = UTCDateTimeField().to_python(datetime.utcnow()) - last_worker_time
             wait_time = timedelta(seconds=constants.MIGRATION_WAIT_TIME) - time_from_last_worker
 
             if wait_time > timedelta(0):


elyezer:
This is likely to be cherry-picked to 2.12 branch.

seandst:
@elyezer It needs to be rebased onto 2.12.0-build

mhrivnak:
This is not a requirement for this fix (thanks @elyezer !), but this bug might have been caught much sooner by a couple of unit tests.  I would consider making a separate method that calculates how many seconds to wait. Tests could just mock ``status.get_workers()`` with various circumstances and assert the correct output for each.

Roughly something like:

```
@staticmethod
def calculate_wait_time():
         active_workers = status.get_workers()
          if active_workers:
             last_worker_time = max([worker['last_heartbeat'] for worker in active_workers])
             time_from_last_worker = UTCDateTimeField().to_python(datetime.now()) - last_worker_time
             time_from_last_worker = UTCDateTimeField().to_python(datetime.utcnow()) - last_worker_time
             return timedelta(seconds=constants.MIGRATION_WAIT_TIME) - time_from_last_worker
          return timedelta(0)
```

elyezer:
I have changed the logic, did some investigation with @werwty and we figured out that the issue was the call datetime.now, datetime.utcnow should be called instead.

elyezer:
@seandst done. I can't change the target branch, could you do that for me please?

seandst:
> done, all requests addressed.

travis reports a line failure, the updated line now crosses the 100-character boundary.

seandst:
After the rebase, I was able to correct the lint error and easily merge this onto 2.12.0-build. I've also merged it forward, and will include this in a new release candidate tomorrow.

omaciel:
Thank you @elyezer and @seandst

PR #2935, opened 2017-01-27
Streamer tries all catalog entries.

jortel:
https://pulp.plan.io/issues/2542

The streamer needed to try all available catalog entries.  This ensures that invalid entries don't prevent successful downloading when valid entries exist.  And, more robust in cases where there are multiple importers that can download a content unit and one (or more of them) pull from a remote repository that is temporarily unavailable.

I really tried not to rewrite the streamer but the flow needs to be very different.  Also, a few of the methods were doing too much and needed to be refactored.

### DIFF
--- a/streamer/pulp/streamer/__init__.py
+++ b/streamer/pulp/streamer/__init__.py
@@ -1,2 +1,2 @@
 from pulp.streamer.config import load_configuration, DEFAULT_CONFIG_FILES  # noqa
-from pulp.streamer.server import Streamer, Responder, StreamerListener  # noqa
+from pulp.streamer.server import Streamer, Responder  # noqa
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -1,20 +1,21 @@
-from gettext import gettext as _
-from httplib import NOT_FOUND, INTERNAL_SERVER_ERROR, SERVICE_UNAVAILABLE
-from urlparse import urlparse
 import logging
 
+from gettext import gettext as _
+from httplib import NOT_FOUND, INTERNAL_SERVER_ERROR
+from urlparse import urlparse
+
 from mongoengine import DoesNotExist, NotUniqueError
-from nectar import listener as nectar_listener
-import requests
+from nectar.listener import AggregatingEventListener
+from requests import Session
 from twisted.internet import reactor
-from twisted.web import resource
+from twisted.web.resource import Resource
 from twisted.web.server import NOT_DONE_YET
 
-from pulp.plugins.loader import api as plugins_api
+from pulp.plugins.loader import api as plugin_api
 from pulp.server.constants import PULP_STREAM_REQUEST_HEADER
-from pulp.server.content.sources import container as content_container
-from pulp.server.content.sources import model as content_models
-from pulp.server.db import model
+from pulp.server.content.sources.container import ContentContainer
+from pulp.server.content.sources.model import Request as ContainerRequest
+from pulp.server.db.model import DeferredDownload, LazyCatalogEntry
 from pulp.server.controllers import repository as repo_controller
 from pulp.plugins.loader.exceptions import PluginNotFound
 from pulp.streamer import adapters as pulp_adapters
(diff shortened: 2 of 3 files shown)


jortel:
PR updated to always 404 regardless of why the actual download failed.  Each failed download attempt is logged.

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
+            try:
+                downloader.config.finalize()
+            except Exception:
+                # ignored.
+                pass
+
+    @staticmethod
+    def _get_downloader(entry):
+        """
+        Get and configured downloader.
```
werwty: s/and/the

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
+
+        :param entry: A catalog entry.
+        :type  entry: LazyCatalogEntry
+        :return: The configured downloader.
+        :rtype:  nectar.downloader.Downloader
```
werwty: s/nectar.downloader.Downloader/nectar.downloaders.base.Downloader/

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
+        :raise: PluginNotFound: when plugin not found.
```
werwty: s/PluginNotFound/PluginNotFound, DoesNotExist/

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
-
-        # Build the alternate content source download request
-        unit_model = plugins_api.get_unit_model_by_id(catalog_entry.unit_type_id)
-        qs = unit_model.objects.filter(id=catalog_entry.unit_id).only(*unit_model.unit_key_fields)
+    def _on_succeeded(self, entry, request, report):
+        """
+        The download succeeded.
+
+        :param entry: A catalog entry.
+        :type  entry: LazyCatalogEntry
```
werwty: s/LazyCatalogEntry/pulp.server.db.model.LazyCatalogEntry/g

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
-                catalog_entry.url,
-                responder,
-            )
+                entry.url,
+                responder)
+            listener = downloader.event_listener
+            container = ContentContainer(threaded=False)
+            container.download(downloader, [alt_request], listener)
+            if listener.succeeded_reports:
+                return listener.succeeded_reports[0]
```
jortel: Yes.

streamer/pulp/streamer/server.py
jortel: Since `LazyCatalogEntry` has been imported, it does not need to be fully qualified.

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,182 @@ def _handle_get(self, request):
...
+        :raise: PluginNotFound: when plugin not found.
+        """
+        try:
+            importer, config, model = \
+                repo_controller.get_importer_by_id(entry.importer_id)
+            model.config = config.flatten()
+            downloader = importer.get_downloader_for_db_importer(
+                model, entry.url, working_dir='/tmp')
+            listener = DownloadListener()
+            downloader.event_listener = listener
```
jortel: Unintended.  Fixed.

streamer/pulp/streamer/__init__.py
```diff
@@ -1,2 +1,2 @@
 from pulp.streamer.config import load_configuration, DEFAULT_CONFIG_FILES  # noqa
-from pulp.streamer.server import Streamer, Responder, StreamerListener  # noqa
```
werwty: DownloadListener needs to be included here so pulp/streamer/test/unit/streamer/test_server.py can import it

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
+            msg = _('The catalog entry for {path} references invalid unit: {unit_type}:{id}')
+            logger.error(msg.format(
+                path=entry.path,
+                unit_type=entry.unit_type_id,
+                id=entry.unit_id))
+            raise
+
+    def _set_headers(self, request, report):
+        """
+        Forward response headers to the HTTP request.  This includes adding the
```
mhrivnak: I think a little more info would make this more clear. Is this the request that a client made to the streamer? Or a request the streamer is about to make to some remote source?

Likewise, what response are we talking about? I assume the ``report`` is representative of the response?

A few words on why these headers need to be copied over would also help.

streamer/pulp/streamer/server.py
```diff
@@ -278,28 +316,20 @@ def __enter__(self):
         """
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, *unused):
```
mhrivnak: This change seems a little risky, in case it got called with kwargs. I don't know if the language makes any guarantees about which way this will get called.

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
         except DoesNotExist:
-            # A catalog entry is referencing a unit that doesn't exist which is bad.
-            msg = _('The catalog entry for {path} references {unit_type}:{id}, but '
-                    'that unit is not in the database.')
-            logger.error(msg.format(path=catalog_entry.path, unit_type=catalog_entry.unit_type_id,
-                                    id=catalog_entry.unit_id))
-            request.setResponseCode(NOT_FOUND)
-        finally:
-            primary_downloader.config.finalize()
+            msg = _('The catalog entry for {path} references invalid unit: {unit_type}:{id}')
```
ipanova: what about 'references no longer valid unit'?

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
+
+        :param request: An HTTP request.
+        :type  request: twisted.web.server.Request
+        :param report: The download report.
+        :type  report: nectar.report.DownloadReport
+        :return:
+        """
+        # forward
+        for key, value in report.headers.items():
+            if key.lower() in HOP_BY_HOP_HEADERS:
```
ipanova: what about this to get rid of continue statement: 
```
if key.lower() not in HOP_BY_HOP_HEADERS:
    request.setHeader(key, value)
```

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
+            model.config = config.flatten()
+            downloader = importer.get_downloader_for_db_importer(
+                model, entry.url, working_dir='/tmp')
+            listener = DownloadListener()
+            downloader.event_listener = listener
+            downloader.session = self.session
+            return downloader
+        except (PluginNotFound, DoesNotExist):
+            msg = _('Plugin not-found: referenced by catalog entry for {path}')
+            logger.info(msg.format(path=entry.path))
```
ipanova: i would move from info level to error level

streamer/pulp/streamer/server.py
```diff
@@ -40,92 +41,30 @@
...
-                    unit_id=self.catalog_entry.unit_id,
-                    unit_type_id=self.catalog_entry.unit_type_id
-                )
-                download.save()
-            except NotUniqueError:
-                # There's already an entry for this unit.
-                pass
+        super(DownloadListener, self).download_failed(report)
+        code = report.error_report.get('response_code', NOT_FOUND)
+        logger.info(_('Download failed [{code}]: {url}').format(code=code, url=report.url))
```
ipanova: what if we move this to debug level. I am afraid it would pollute info level with a lot of messages.

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
+                path = urlparse(request.uri).path
+                q_set = LazyCatalogEntry.objects.filter(path=path)
+                q_set = q_set.order_by('-_id', '-revision')
+                count = q_set.count()
+                if not count:
+                    logger.error(_('No catalog entry found. path={p}'.format(p=path)))
+                    request.setResponseCode(NOT_FOUND)
+                    return
+                for entry in q_set.all():
+                    logger.info('Trying URL: {url}'.format(url=entry.url))
```
ipanova: Same here, i don;t think it belongs to info level, but rather to debug

streamer/pulp/streamer/server.py
```diff
@@ -174,84 +113,183 @@ def _handle_get(self, request):
...
+                    try:
+                        last_report = self._download(entry, responder)
+                        self._on_succeeded(entry, request, last_report)
+                        return
+                    except (DownloadFailed, DoesNotExist, PluginNotFound):
```
jortel: Matching catalog entries can be contributed by multiple plugins.  Basically more than 1 remote (upstream) repository *could* provide the same file.

streamer/pulp/streamer/server.py
jortel: "unknown" is probably more accurate.

streamer/pulp/streamer/server.py
jortel: 👍

streamer/pulp/streamer/server.py
jortel: 👍

streamer/pulp/streamer/server.py
jortel: I understand where you're coming from.   But, based on past experience, getting debug level logging output from customers can be difficult.  On EL6 and probably EL7, it requires users to reconfigure rsyslog and then recreate the failure.  After spending a lot of time struggling to customer cases involving lazy, I think it's better to leave this at `info` for now.

jortel:
@mhrivnak, @ipanova:  Thanks for your reviews.  I posted reply comments and pushed a new commit.

streamer/pulp/streamer/server.py
mhrivnak: But wouldn't all such repositories be provided by the same plugin? For example, if the yum importer can't be found, I assume that's what would cause that exception? And if so, what would be the point of trying more catalog entries for a file? I'm assuming that given a file, only one plugin type can provide it. Maybe that won't be true at some point in the future.

streamer/pulp/streamer/server.py
mhrivnak: We've already decided broadly to log each download that Pulp attempts at INFO. Nectar now does that by default.

streamer/pulp/streamer/server.py
```diff
@@ -40,92 +41,30 @@
...
-                    unit_id=self.catalog_entry.unit_id,
-                    unit_type_id=self.catalog_entry.unit_type_id
-                )
-                download.save()
-            except NotUniqueError:
-                # There's already an entry for this unit.
-                pass
+        super(DownloadListener, self).download_failed(report)
+        code = report.error_report.get('response_code', NOT_FOUND)
+        logger.warn(_('Download failed [{code}]: {url}').format(code=code, url=report.url))
```
mhrivnak: nectar logs this at INFO. We should make the behaviors consistent. https://github.com/pulp/nectar/pull/55/files

streamer/pulp/streamer/server.py
mhrivnak: Here is the recent nectar PR: https://github.com/pulp/nectar/pull/55/files

streamer/pulp/streamer/server.py
jortel: okay

streamer/pulp/streamer/server.py
jortel: Given the same RPM is available in 2 remote repositories  like RHEL and CentOS.  Both importers will create a catalog entry for the same file even though only one importer will actually create the unit.

streamer/pulp/streamer/server.py
ipanova: @mhrivnak thanks, did know about that

streamer/pulp/streamer/server.py
mhrivnak: We discussed this live and agreed on what the behavior is, that it's just slightly inefficient, but that it's the best behavior for now in terms of low complexity, and accommodating future use cases.

PR #2949, opened 2017-02-12
update last_unit_added when units are copied

werwty:
closes #2577
https://pulp.plan.io/issues/2577


mhrivnak:
I think this should go against 2.12-dev

PR #2950, opened 2017-02-13
Two small fixes in dev_setup guide

bmbouter:
Fixes a line length issue
Fixes a block escaping issue

### DIFF
--- a/docs/dev-guide/contributing/dev_setup.rst
+++ b/docs/dev-guide/contributing/dev_setup.rst
@@ -143,7 +143,8 @@ After preparing either the libvirt or docker prerequisites using the instruction
 
 #. You can open the Vagrantfile in your favorite editor and modify it to better fit your
    development preferences. This step is entirely optional as the default Vagrantfile should
-   work for most users. Refer to the :ref:`Advanced Vagrant <dev-setup-advanced-vagrant>` section below for some helpful tips.
+   work for most users. Refer to the :ref:`Advanced Vagrant <dev-setup-advanced-vagrant>` section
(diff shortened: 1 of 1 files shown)


docs/dev-guide/contributing/dev_setup.rst
```diff
@@ -143,7 +143,8 @@ After preparing either the libvirt or docker prerequisites using the instruction
 
 #. You can open the Vagrantfile in your favorite editor and modify it to better fit your
    development preferences. This step is entirely optional as the default Vagrantfile should
-   work for most users. Refer to the :ref:`Advanced Vagrant <dev-setup-advanced-vagrant>` section below for some helpful tips.
+   work for most users. Refer to the :ref:`Advanced Vagrant <dev-setup-advanced-vagrant>` section
+   below for some helpful tips. ::
```
seandst: Will this still render as "below for some helpful tips. :" ?

dralley: My ReST knowledge isn't the best but I remember that you do need the :: because of how : is a special character typically

dralley: I made the same fix in some other places last week

bmbouter: @seandst I didn't see your question before merging. This style will escape the section below but it will not show the colon. If it were `tips.::` then it would show the colon and escape below. It's a shorthand for putting a `::` on it's own line which I like better, but this is a shorthand used all over this document.

PR #2951, opened 2017-02-14
Add console logger option for pulp.server.

sysadmind:
This is related to https://pulp.plan.io/issues/2548

This will allow for more logging options for pulp
especially when /dev/log isn't available as is the
case in most containers.

### DIFF
--- a/docs/user-guide/installation/extra_configuration.rst
+++ b/docs/user-guide/installation/extra_configuration.rst
@@ -141,3 +141,62 @@ following steps:
    For Upstart::
 
       $ for s in httpd pulp_workers pulp_resource_manager pulp_celerybeat; do sudo service $s restart; done;
+
+Logging Configuration
+---------------------
+
+.. _logging:
+
+Logging
+-------
+
+.. note::
+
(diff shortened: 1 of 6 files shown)


sysadmind:
@mhrivnak I know you are working towards pulp on kubernetes. This may help as most containers expect to log to stdout instead of /dev/log.

sysadmind:
@bmbouter Thanks for the feedback! I updated the docs and the config file. The docs update includes some restructuring that I think will be valuable overall, but please let me know if this may cause problems.

bmbouter:
This looks great and it tests exactly as expected. 👍 🍰 🎆 I am merging it. Thank you so much @sysadmind !

I am going to also move the Redmine issue to MODIFIED and associate the commit with the issue. This can also be done with [keywords in commit messages](http://docs.pulpproject.org/en/2.12/nightly/dev-guide/contributing/branching.html#commit-messages) that move the issue to MODIFIED and associate the commit automatically.

I'm going to followup with a very small PR which makes a tiny adjustment and adds release notes.

bmbouter:
Here is my additional PR:  https://github.com/pulp/pulp/pull/2972

PR #2952, opened 2017-02-14
Set batch_size for the cursor in orphan removal tasks

goosemania:
On a busy system a cursor for iterating over all units
in a collection could be timed out during orphan removal.
The batch size is set to 100 to make sure cursor is active,
it also should not introduce any noticable performance impact.

closes #2584
https://pulp.plan.io/issues/2584


goosemania:
ok test

goosemania:
ok test

PR #2953, opened 2017-02-14
Fix streamer header forwarding.

jortel:
https://pulp.plan.io/issues/2587

Looks like moving the header management to a later stage caused squid to not cache.  Makes sense that the headers would need to be manipulated before streaming the bits.

Tested w/ smash tests mentioned in the issue:
```
[jortel@f23d ~]$ python3 -m unittest pulp_smash.tests.rpm.api_v2.test_download_policies.OnDemandTestCase.test_rpm_cache_control_header pulp_smash.tests.rpm.api_v2.test_download_policies.OnDemandTestCase.test_same_rpm_cache_header pulp_smash.tests.rpm.api_v2.test_download_policies.SwitchPoliciesTestCase.test_background_to_on_demand pulp_smash.tests.rpm.api_v2.test_download_policies.SwitchPoliciesTestCase.test_immediate_to_on_demand
/usr/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:821: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
....
----------------------------------------------------------------------
Ran 4 tests in 115.701s

OK
[jortel@f23d ~]$ 
```

### DIFF
--- a/streamer/pulp/streamer/server.py
+++ b/streamer/pulp/streamer/server.py
@@ -52,6 +52,36 @@ class DownloadListener(AggregatingEventListener):
     Nectar download listener.
     """
 
+    def __init__(self, streamer, request):
+        """
+        :param streamer: The streamer.
+        :type  streamer: Streamer
+        :param request: The original twisted client HTTP request being handled by the streamer.
+        :type  request: twisted.web.server.Request
+        """
+        super(DownloadListener, self).__init__()
+        self.streamer = streamer
+        self.request = request
+
+    def download_headers(self, report):
+        """
+        Forward response headers to the original client HTTP request.
+        This includes adding the cache-control header with the max-age
+        which is loaded from the configuration.
+
+        :param report: The download report.
+        :type  report: nectar.report.DownloadReport
+        """
+        super(DownloadListener, self).download_headers(report)
+        # forward
+        for key, value in report.headers.items():
+            if key.lower() not in HOP_BY_HOP_HEADERS:
+                self.request.setHeader(key, value)
+        # additions
+        max_age = self.streamer.config.get('streamer', 'cache_timeout')
+        cache_control = 'public, s-maxage={m}, max-age={m}'.format(m=max_age)
+        self.request.setHeader('Cache-Control', cache_control)
+
     def download_failed(self, report):
         """
         Log download failures.
@@ -102,7 +132,7 @@ class Streamer(Resource):
             * The file is downloaded using the Nectar downloader and the content
               is streamed to the client as it is received.
 
-        :param request: the request to process.
+        :param request: The original twisted client HTTP request being handled by the streamer.
         :type  request: twisted.web.server.Request
         """
         reactor.callInThread(self._handle_get, request)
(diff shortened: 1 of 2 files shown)


streamer/pulp/streamer/server.py
```diff
@@ -52,6 +52,36 @@ class DownloadListener(AggregatingEventListener):
     Nectar download listener.
     """
 
+    def __init__(self, streamer, request):
+        """
+        :param streamer: The streamer.
+        :type  streamer: Streamer
+        :param request: An HTTP request.
```
mhrivnak: As a general comment, I'm reminded that every time I look at the streamer and see a "request" or "response", I have to think carefully about where it came from. Since the streamer handles requests and makes new requests, and of course thus it receives responses and uses those to generate new responses, it gets confusing.

I wish the doc blocks were consistently explicit about whether the streamer is acting as the client or server with respect to each request and response. It would be helpful to clarify that here.

werwty:
This PR LGTM. But I'll leave it to @mhrivnak to give the approval.

streamer/pulp/streamer/server.py
```diff
@@ -129,7 +159,7 @@ def _handle_get(self, request):
                 for entry in q_set.all():
                     logger.info('Trying URL: {url}'.format(url=entry.url))
                     try:
-                        last_report = self._download(entry, responder)
+                        last_report = self._download(request, entry, responder)
```
mhrivnak: So this is a request that was made by some external client, and is being handled by the streamer, right?

What's unintuitive to me is that this object gets passed down to the download listener, which then sets headers on this request object. Am I understanding this right? Why does it make sense to set a header on a request that's already been made, and is currently being handled?

streamer/pulp/streamer/server.py
jortel: The type is a _twisted_ Request which is always the original client request being handled by the streamer.

streamer/pulp/streamer/server.py
jortel: Agreed. This is not intuitive.  My understanding is based on the original streamer code.  It seems like only the twisted request is passed to the streamer.  The twisted `Request` seems to bidirectional.  Even looking at the `Responder` it effectively writes the bits using `Request.write()`.  I'll look at twisted docs to make sure.  But, this is how the 2.10 streamer did things and it seems to work.

streamer/pulp/streamer/server.py
jortel: I'll improve the docstring.

streamer/pulp/streamer/server.py
jortel: Looking at the twisted documentation: https://twistedmatrix.com/documents/8.2.0/api/twisted.web.http.Request.html, it seems the `Request` object represents both the original client request and has methods to affect the response.  Basically, the methods that inspect the *request* pertain to the original client request and update methods like `set*` and `write` affect the response.

mhrivnak: Ok, thanks for chasing that down! This is a good time to make sure we fully understand how the streamer work, so I feel much better with this mystery solved.

PR #2957, opened 2017-02-17
Create extlinks for all the plugins in sphinx config on a platform.

goosemania:
closes #2585
https://pulp.plan.io/issues/2585

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -252,10 +252,47 @@ texinfo_documents = [
 #texinfo_show_urls = 'footnote'
 
 extlinks = {'redmine': ('https://pulp.plan.io/issues/%s', '#'),
-            'fixedbugs': ('https://pulp.plan.io/projects/pulp/issues?c%%5B%%5D=tracker&c%%5B%%5D='
-                          'status&c%%5B%%5D=priority&c%%5B%%5D=cf_5&c%%5B%%5D=subject&c%%5B%%5D=a'
-                          'uthor&c%%5B%%5D=assigned_to&c%%5B%%5D=cf_3&f%%5B%%5D=cf_4&f%%5B%%5D=tr'
-                          'acker_id&f%%5B%%5D=&group_by=&op%%5Bcf_4%%5D=%%3D&op%%5Btracker_id%%5D'
-                          '=%%3D&set_filter=1&sort=priority%%3Adesc%%2Ccf_5%%3Adesc%%2Cid%%3Adesc'
-                          '&utf8=%%E2%%9C%%93&v%%5Bcf_4%%5D%%5B%%5D=%s&v%%5Btracker_id%%5D%%5B%%5'
-                          'D=1', 'bugs fixed in ')}
+            'fixedbugs_pulp': ('https://pulp.plan.io/projects/pulp/issues?c%%5B%%5D=tracker&c%%5B%'
+                               '%5D=status&c%%5B%%5D=priority&c%%5B%%5D=cf_5&c%%5B%%5D=subject&c%%'
+                               '5B%%5D=author&c%%5B%%5D=assigned_to&c%%5B%%5D=cf_3&f%%5B%%5D=cf_4&'
+                               'f%%5B%%5D=tracker_id&f%%5B%%5D=&group_by=&op%%5Bcf_4%%5D=%%3D&op%%'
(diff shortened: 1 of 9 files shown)


seandst:
This seems tedious, but I think it's the best option. I assume that the ':fixedbugs:' string is pretty easy to find/replace safely across all the plugins.

Is your plan for the plugins to only add the "fixedbugs_<project>" directive to that project? For example, you don't plan to add fixedbugs_pulp_puppet to pulp_rpm, and only platform is going to have all the different project fixedbugs_* directives?

goosemania:
@seandst , I was going to add directive for all the plugins only in the platform because it is needed for our docs builder. Plugins have to be modified as well but they will contain only one directive for themselves.
And yes, I agree it's easy and safe to replace all the ":fixedbugs:" across the plugins.
Thanks for the feedback, it looks like we are on the same page.

docs/conf.py
```diff
@@ -251,11 +251,17 @@
...
+              '5B%%%%5D=author&c%%%%5B%%%%5D=assigned_to&c%%%%5B%%%%5D=cf_3&f%%%%5B%%%%5D=c'
+              'f_4&f%%%%5B%%%%5D=tracker_id&f%%%%5B%%%%5D=&group_by=&op%%%%5Bcf_4%%%%5D=%%%'
+              '%3D&op%%%%5Btracker_id%%%%5D=%%%%3D&set_filter=1&sort=priority%%%%3Adesc%%%%'
+              '2Ccf_5%%%%3Adesc%%%%2Cid%%%%3Adesc&utf8=%%%%E2%%%%9C%%%%93&v%%%%5Bcf_4%%%%5D'
+              '%%%%5B%%%%5D=%s&v%%%%5Btracker_id%%%%5D%%%%5B%%%%5D=1')
+
+_projects = ['pulp', 'pulp_rpm', 'pulp_puppet', 'pulp_python', 'pulp_ostree', 'pulp_docker']
+
+# create directive fixedbugs_<project name> for each project
+extlinks = {'fixedbugs_' + project: (_bugs_link % (project, '%s'), 'bugs fixed in ')
```
bmbouter: This string replacement assumes that the only difference in the urls is the project name handled by the single `%s` replacement. I suspect that isn't true for 2 reasons.

* The platform release name is a custom field that is different for some projects (python, ostree). In the future they will all likely be different as those plugins become versioned different from platform. I don't think the same GET parameters could correctly specify a platform version for these different fields with one url.
* The existing urls in all of the conf.py files are different. I compared the one in pulp_rpm versus pulp_puppet and they are different. This could just be ordering though and I didn't compile them with sphinx to find out.

Assuming ^, I think the extlinks can be written out explicitly like:

```
{
	'redmine': ('https://pulp.plan.io/issues/%s', '#'),
	'pulp': '<the url string from conf.py of platform>',
	'pulp_rpm': '<the rpm url string from pulp_rpm conf.py>',
	'pulp_ostree': '<the ostree url string from pulp_rpm conf.py>'
}
```

bmbouter:
Each plugins need two things done:

* the find and replace to change s/findbugs/pulp_xxxx/ where xxxx is the plugin name. This is @seandst is suggesting I think.
Once ^ is done, the plugins themselves will fail to build because the conf.py definitions in those files will still have `fixedbugs` so when the find and replace occurs, I think changing the extlinks dict keys in the conf.py of the plugins is also needed. This way nothing breaks.

bmbouter:
If a PR for each plugin is needed to ensure the problem is actually resolved, we should associate all commits with the issue as associated commits. To do that the others should probably all use `re #2585` in the commit message and we should merge this one last.

docs/conf.py
goosemania: Thank you, I will check the links to see if it is possible not to list almost identical links one by one... As you said, there is always an option to specify the explicitly

goosemania:
@bmbouter , I think [my comment](https://github.com/pulp/pulp/pull/2957#issuecomment-280781117) is inline with your proposal to substitute directives in plugins and fix the conf.py. Great :)

I will definitely use `re #2585` for the plugins, no worries

docs/conf.py
bmbouter: Even similar string urls now will be dis-similar when the plugins are versioned independantly with pulp3. So even if the strings are the same I think writing them out explicitly would be easy and good.

PR #2961, opened 2017-03-07
Updated the repo for pulp rhel6 installs

werwty:
Since the current pulp stable repo does not support rhel6.


werwty:
@seandst I added a note with more (hopefully enough) context.

seandst:
I'm sorry, I thought I'd already approved this. Thanks for the updates. :)

PR #2964, opened 2017-03-10
As a user I can manage Pulp using a systemwide command

BrnoPCmaniak:
closes #2421
https://pulp.plan.io/issues/2421

### DIFF
--- a/app/pulp/app/entry_points.py
+++ b/app/pulp/app/entry_points.py
@@ -1,0 +1,9 @@
+"""This module provides system wide command to interface with the pulp platform."""
+import os
+import sys
+
+
+def pulp_manager_entry_point():
+    os.environ["DJANGO_SETTINGS_MODULE"] = "pulp.app.settings"
+    from django.core.management import execute_from_command_line
+    execute_from_command_line(sys.argv)
--- a/app/setup.py
+++ b/app/setup.py
@@ -30,4 +30,9 @@ setup(
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
     ),
+    entry_points={
+        'console_scripts': [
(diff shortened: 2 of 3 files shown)


app/pulp/app/pulp_manager.py
```diff
@@ -0,0 +1,14 @@
+"""This module provides system wide command to interface with the pulp platform.
+
+In future commands that can be issued should be limitted this will probably be achieved
```
seandst: s/limitted/limited/

app/pulp/app/pulp_manager.py
```diff
@@ -0,0 +1,14 @@
...
+by implementing own ManagmentUtility similar to this one:
+https://github.com/django/django/blob/stable/1.8.x/django/core/management/__init__.py
```
seandst: Docstrings should talk about the present, not the future. If we want to implement something in the future, we should track that in Redmine, so this docstring should only be the first line.

app/pulp/app/pulp_manager.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+import os
+import sys
+
+
+def pulp_manager_entry_point():
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulp.app.settings")
```
seandst: We don't want setdefault here, since this is `pulp-manager`, not `generic-django-manager`. The story says "have all commands be passed along to the manage.py for pulp_app", which I take to mean that we need this to always use the pulp.app settings regardless of what `DJANGO_SETTINGS_MODULE` is set to.

BrnoPCmaniak:
@seandst All things fixed also I renamed the module to `entry_points.py`

PR #2968, opened 2017-03-16
Fix scheduled calls with Celery 4.x.

dralley:
Celery 4.x changes a number of implentation details which
breaks our assumptions about how the scheduling system
works. Therefore, we want to check for the celery version
and take actions to trigger the desired behavior depending
on that celery version.

closes #2615
https://pulp.plan.io/issues/2615


dralley:
ok test

PR #2971, opened 2017-03-17
Fix the arguments being passed to celery.apply_async

werwty:
Celery's Task.apply_async is expecting args and kwargs as keyword arguments.

If we want to pass any kwargs to the task, it has to be in the kwarg kwarg:
`apply_async(args, kwargs=mykwargs)`

With this PR we also assume that any args being passed in to apply_async will be passed directly to the task. Alternatively we could make this explicit and require any task args to be passed with the  `args` keyword. I am not sure which approach is more correct so any feedback is welcome.

### DIFF
--- a/tasking/pulp/tasking/tasks.py
+++ b/tasking/pulp/tasking/tasks.py
@@ -85,12 +85,11 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
     task_status = TaskStatus.objects.get(pk=inner_task_id)
     ReservedResource.objects.create(task=task_status, worker=worker, resource=resource_id)
 
-    inner_kwargs['routing_key'] = worker.name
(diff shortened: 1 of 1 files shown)


werwty:
I tested the changes by running in a django shell:
`from pulp.app.tasks import repository
import uuid
repo_uuid=str(uuid.uuid4())
repo=Repository(name=repo_uuid)
repo.save()
repository.delete.apply_async(repo_uuid)
repository.delete.apply_async_with_reservation("foo","bar",repo_uuid)`

werwty:
I looks like I was calling our apply_async with the wrong arguments, the way we would call it before is
<pre>
repository.delete.apply_async(kwargs={'repo_name':repo_uuid})
repository.delete.apply_async_with_reservation("foo","bar",kwargs={'repo_name':repo_uuid})
</pre>

I might should revert this PR

PR #2972, opened 2017-03-17
Adds release note for console feature + misc fixes

bmbouter:
* A release note for the console feature was added.
* Removal of some code handling a case that will never occur.
* Test fix because of ^
* Small flake8 fixes I saw when running tests locally

closes #2548
https://pulp.plan.io/issues/2548

### DIFF
--- a/server/test/unit/server/test_logs.py
+++ b/server/test/unit/server/test_logs.py
@@ -3,7 +3,6 @@
 This module contains tests for the pulp.server.logs module.
 """
 from cStringIO import StringIO
-import ConfigParser
 import logging
 import os
 import Queue
@@ -678,7 +677,6 @@ def console_logging_side_effect(section, key):
             return 'INFO'
         elif key == 'log_type':
             return 'console'
-    return DEFAULT
 
 
 class TestStartLogging(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


bmbouter:
@sysadmind here are the followup changes I made. You are welcome to review if you want to.

bmbouter:
ok test

server/test/unit/server/test_logs.py
```diff
@@ -765,8 +763,7 @@ def fake_getLogger(name=None):
         # We should have used the user's setting
         root_logger.setLevel.assert_called_once_with(logging.ERROR)
 
-    @mock.patch('pulp.server.logs.config.config.get',
-                side_effect=ConfigParser.NoOptionError('server', 'log_level'))
+    @mock.patch('pulp.server.logs.config.config.get', return_value='INFO')
```
sysadmind: Won't this cause the log setup to get `INFO` for `config.config.get('server', 'log_type')` and therefore not be able to set a proper logger?  I suppose it will just fail back to syslog at that point due to the `VALID_LOGGERS` check. It's probably safe to do.

bmbouter: The way that we do default checking with our unit tests is confusing to me. We've had a variety of issues with our unit tests in the past. This test was failing for me and I didn't understand it since we provide a default I never expect a `NoOptionError` to be raised. What I really want to do is to assert that the default value of `get('server', 'log_level')` is INFO.

bmbouter:
I'm going to merge it and if someone wants to look at improving the tests they can.

server/test/unit/server/test_logs.py
sysadmind: Yeah I tried to structure this test in such a way that I wasn't testing `get('server', 'log_level')` but only the `get('server', 'log_type')` option since the `log_level` is covered in another test. It seems like the unit tests mock a lot so many parts of the config get duplicated to get around that.

PR #2977, opened 2017-03-27
Merge forward 2922 into pulp3

dralley:


docs/user-guide/release-notes/2.13.x.rst
```diff
@@ -11,3 +11,23 @@ New Features
 * Log statements which are emitted from task processes now include a short task
   id of 8 characters (i.e. ``pulp.server.sync::INFO: [0ca4399c] ...``) which
   can be used to look up a task.
+* Users no longer need to have pulp_celerybeat running to know whether pulp_workers or
```
bmbouter: This is OK to be here, but it will be deleted since 2.y release notes won't be included in 3.0-dev. Consider removing.

PR #2980, opened 2017-03-29
[issue #2680]: Scheduled RPM (un)installation does not work

S1artie:
https://pulp.plan.io/issues/2680


S1artie:
Thanks for the quick response! I have added myself to the list as requested. Do you squash the additional commit with the bugfix commit on merge, or would you prefer me to create a new pull request with just one commit in it? I'd create a new branch in my fork, squash those two commits into one and then open a new pull request in that case.

As for the twitter handle, I'm @s1artie on Twitter. I pretty much only use it on conferences (and even there mostly for reading), so there's not much action there ;-). But feel free to use that handle for your purposes.

S1artie:
Nope, I found this little docker test setup script on your blog a while ago (this one: http://pulpproject.org/2015/05/21/use-docker-to-try-pulp/) and started extending and modifying that thing heavily so the environment built by it would closely resemble our specific production environment in which we're going to use Pulp, and so it would allow me to quickly tear down and set up this environment so I could start over from scratch whenever I wanted, play around with Pulp and try out various processes, and also have our systems interface with this Pulp environment via the REST API.

One day I noticed that the Docker images it pulled were a bit dated (they had a Pulp 2.6 in it), but apart from that I was pretty satisfied with the entire setup, so I forked the repo with the dockerfiles, modified them slightly to install an up-to-date Pulp version, and rebuilt the necessary Docker images (I wanted to try that Docker Cloud service some time anyway, so that was a good use case for playing with it). But that is basically still the setup we mostly use for quick testing-on-a-single-dev-machine kind of stuff.

PR #2981, opened 2017-03-30
Upgrade to celery 3.1.17

goosemania:


deps/python-celery/dist_list.txt
```diff
@@ -1 +1 @@
-el6 el7 fc23 fc24
+el7
```
dkliban: i think this is OK for now, but we may have to add el6 later for the nightlies.

PR #2982, opened 2017-03-30
Update spec file and external deps to use newer celery version

goosemania:
closes #2589
https://pulp.plan.io/issues/2589

### DIFF
--- a/deps/external_deps.json
+++ b/deps/external_deps.json
@@ -40,14 +40,9 @@
         "platform": ["el6", "el7", "fc23", "fc24"]
     },
     {
-        "name": "python-billiard",
-        "version": "3.3.0.17-2",
-        "platform": ["el6", "el7"]
-    },
-    {
         "name": "python-celery",
-        "version": "3.1.11-1",
-        "platform": ["el6", "el7", "fc23", "fc24"]
+        "version": "3.1.17-1",
+        "platform": ["el7"]
     },
     {
         "name": "python-isodate",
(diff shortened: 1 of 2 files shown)


deps/external_deps.json
```diff
@@ -40,14 +40,9 @@
         "platform": ["el6", "el7", "fc23", "fc24"]
     },
     {
-        "name": "python-billiard",
-        "version": "3.3.0.17-2",
-        "platform": ["el6", "el7"]
```
goosemania: python-billiard >= 3.3.0.19 is required for celery 3.1.17.
3.3.0.20 is in EPEL7.

bmbouter: This is great, ty.

pulp.spec
```diff
@@ -389,6 +388,12 @@ Requires: Django14
 %else
 Requires: python-django >= 1.4.0
 %endif
+# RHEL6 or RHEL7
+%if 0%{?rhel} == 6 || 0%{?rhel} == 7
+Requires: python-celery == 3.1.17
+%else
+Requires: python-celery >= 3.1.12
```
dkliban: is 3.1.12 new enough on fedora?

goosemania: Any version >= 3.1.12 will solve bootsteps issue, that's my understanding.
On Fedora celery 3.1.20 is currently available.

bmbouter: What about replacing L#391 - L#396 with `Requires: python-celery >= 3.1.17`?

goosemania: That's a good idea! It does not allow someone to use celery 3.1.12 - 3.1.16 but the probability that it is needed is very low I guess.

bmbouter:
Thanks Tanya!

PR #2984, opened 2017-04-03
Fixes issue w/ worker heartbeats on broker failure

dralley:
Starts Pulp worker heartbeat reporting when the Celery
consumer starts rather than when the Celery worker starts,
so that worker heartbeat reporting is resumed after a
disconnection and subsequent reconnection from the broker.

closes #2613
https://pulp.plan.io/issues/2613


dralley:
Ah, right

bmbouter:
Thanks @dralley I manually tested it and it works great. Thanks also for the docstring fixes. This is good to merge!

PR #2986, opened 2017-04-05
Batch save() calls to ProgressReport

werwty:
closes #2316
https://pulp.plan.io/issues/2316

### DIFF
--- a/app/pulp/app/models/progress.py
+++ b/app/pulp/app/models/progress.py
@@ -3,13 +3,18 @@ Django models related to progress reporting
 """
 from gettext import gettext as _
 import logging
+import datetime
 
 from django.db import models
 
 from pulp.tasking import util
 from pulp.app.models import Model, Task
+from django.utils import timezone
 
 _logger = logging.getLogger(__name__)
+
+# number of ms between save() calls when _using_context_manager is set
+BATCH_INTERVAL = 500
 
 
 class ProgressReport(Model):
@@ -58,6 +63,9 @@ class ProgressReport(Model):
     task = models.ForeignKey("Task", on_delete=models.CASCADE)
     suffix = models.TextField(default='')
 
+    _using_context_manager = False
+    _last_save_time = None
+
     def save(self, *args, **kwargs):
         """
         Auto-set the task_id if running inside a task
@@ -70,7 +78,16 @@ class ProgressReport(Model):
         """
         if self.task_id is None:
             self.task_id = Task.objects.get(id=util.get_current_task_id())
-        super(ProgressReport, self).save(*args, **kwargs)
+
+        now = timezone.now()
+
+        if self._using_context_manager and self._last_save_time:
+            if now - self._last_save_time >= datetime.timedelta(milliseconds=BATCH_INTERVAL):
+                super(ProgressReport, self).save(*args, **kwargs)
+        else:
+            super(ProgressReport, self).save(*args, **kwargs)
+
+        self._last_save_time = now
 
     def __enter__(self):
         """
@@ -78,6 +95,10 @@ class ProgressReport(Model):
         """
         self.state = self.RUNNING
         self.save()
+
+        # Save needs occurs immediately so it is called before _using_context_manager is set
+        self._using_context_manager = True
+        return self
 
     def __exit__(self, type, value, traceback):
         """
@@ -89,6 +110,7 @@ class ProgressReport(Model):
 
         See the context manager documentation for more info on __exit__ parameters
         """
+        self._using_context_manager = False
         if type is None:
             self.state = self.COMPLETED
             self.save()
@@ -115,7 +137,9 @@ class ProgressSpinner(ProgressReport):
         >>> metadata_progress.save()
 
     The ProgressSpinner() is a context manager that provides automatic state transitions and saving
-    for the RUNNING COMPLETED and FAILED states. Use it as follows:
+    for the RUNNING COMPLETED and FAILED states. When ProgressSpinner() is used as a context
+    manager progress reporting is rate limited to every 500 milliseconds.
+    Use it as follows:
 
         >>> spinner = ProgressSpinner('Publishing Metadata')
         >>> spinner.save() # spinner is saved as 'waiting'
@@ -157,7 +181,9 @@ class ProgressBar(ProgressReport):
 
     The ProgressBar() is a context manager that provides automatic state transitions and saving for
     the RUNNING COMPLETED and FAILED states. The increment() method can be called in the loop as
-    work is completed. Use it as follows:
+    work is completed. When ProgressBar() is used as a context manager progress reporting is rate
+    limited to every 500 milliseconds.
+    Use it as follows:
 
         >>> progress_bar = ProgressBar(message='Publishing files', total=len(files_iterator))
         >>> progress_bar.save()


app/pulp/app/models/progress.py
```diff
@@ -18,6 +20,16 @@ class ProgressReport(Model):
 
     All progress reports have a message, state, and are related to a Task.
 
+    Attributes:
+
+        batch_increments (bool): If this Flag is set save checks to make sure it only gets ran once
```
bmbouter: I didn't expect any of these would be developer configurable. I think these options will be rarely used if ever. If we keep these as simple as possible then plugin writers will have as little to think about which will make for a better plugin experience. To fix this the feature would be assumed to be always on and `batch_interval` would be hard coded probably as a constant at the top of the module.

app/pulp/app/models/progress.py
```diff
@@ -18,6 +20,16 @@ class ProgressReport(Model):
...
+            every batch_interval milliseconds. This Flag defaults to False, it's set to true if the ProgressReport
+            is being used as a Context Manager.
+        last_increment_time (datetime.datetime): The last time save was called. This is used with
```
bmbouter: I don't think the `last_increment_time` should be a public interface. A user should never set it. Can this become a private variable that is also undocumented.

app/pulp/app/models/progress.py
bmbouter: I see now that this is being used to ensure it's only run during the context manager. That is good (and required), so converting it to a private variable and calling it `using_context_manager` is what I recommend.

app/pulp/app/models/progress.py
```diff
@@ -115,7 +142,9 @@ class ProgressSpinner(ProgressReport):
         >>> metadata_progress.save()
 
     The ProgressSpinner() is a context manager that provides automatic state transitions and saving
-    for the RUNNING COMPLETED and FAILED states. Use it as follows:
+    for the RUNNING COMPLETED and FAILED states. When ProgressSpinner() is used as a context
```
bmbouter: 👍 docs

app/pulp/app/models/progress.py
```diff
@@ -157,7 +186,9 @@ class ProgressBar(ProgressReport):
 
     The ProgressBar() is a context manager that provides automatic state transitions and saving for
     the RUNNING COMPLETED and FAILED states. The increment() method can be called in the loop as
-    work is completed. Use it as follows:
+    work is completed. When ProgressBar() is used as a context manager the batch_increments flag is
```
bmbouter: 👍 docs

app/pulp/app/models/progress.py
werwty: I think we should allow plugin writer the ability to control whether or not they want to batch their save() calls. I did update the docs to call out the fact that it is up to a plugin writer has to ensure that the last save() is called. What do you think about this?

bmbouter: I don't think practically speaking a plugin writer would ever want to disable this feature. If there was a reason they would then I would want to add it. My main concern is that we want to launch with a really really simple plugin API and these docstring additions are making it more complicated with an option that I think almost no one will use.

werwty: Fair enough. I think defaulting to giving plugin writers more power is good, but I also see how we would not want to overwhelm them with too many options. (and this is something easily added if someone ever requests this feature)
I will amend this PR.

bmbouter: Thank you. Perhaps someone will request it and we could easily add it back in.

app/pulp/app/models/progress.py
```diff
@@ -3,13 +3,17 @@
...
 
 from django.db import models
 
 from pulp.tasking import util
 from pulp.app.models import Model, Task
+from django.utils import timezone
 
 _logger = logging.getLogger(__name__)
+# number of ms between save() calls when _using_context_manager is set
+_batch_interval = 500
```
bmbouter: Can this become BATCH_INTERVAL I think python constants are usually in all caps.
Also having a line of whitespace between L#15 and L#14 would be good to group the comment w/ the line it is for more clearly.

app/pulp/app/models/progress.py
```diff
@@ -70,14 +77,24 @@ def save(self, *args, **kwargs):
...
+                super(ProgressReport, self).save(*args, **kwargs)
+        else:
+            super(ProgressReport, self).save(*args, **kwargs)
 
     def __enter__(self):
         """
         Saves the progress report state as RUNNING
         """
         self.state = self.RUNNING
+        self._using_context_manager = True
```
bmbouter: I think this will cause the transition to Running to occur 0.5 seconds after it really does. I think we want the first save() to occur immediately. You could fix this by calling save() before setting `self._using_context_manager = True`.

app/pulp/app/models/progress.py
```diff
@@ -58,6 +62,9 @@ class ProgressReport(Model):
     task = models.ForeignKey("Task", on_delete=models.CASCADE)
     suffix = models.TextField(default='')
 
+    _using_context_manager = False
```
bmbouter: Is the database going to reject these attributes? Have you tried calling Save() within and outside of a context manager with this PR? I can't test it because I am working on Pulp2 stuff currently. I'm not sure if Django will ignore these attributes or not do you know?

app/pulp/app/models/progress.py
```diff
@@ -58,6 +62,9 @@ class ProgressReport(Model):
...
+    _last_increment_time = None
```
bmbouter: Since this is only set during save I think s/_last_increment_time/_last_save_time/

app/pulp/app/models/progress.py
```diff
@@ -70,14 +77,24 @@ def save(self, *args, **kwargs):
...
+        else:
+            super(ProgressReport, self).save(*args, **kwargs)
 
     def __enter__(self):
         """
         Saves the progress report state as RUNNING
         """
         self.state = self.RUNNING
+        self._using_context_manager = True
+        self._last_increment_time = timezone.now()
```
bmbouter: At this point in time it hasn't been incremented. What about having this set with every save call (regardless of the value of self._using_context_manager) and have L#81 guarded with `if self._using_context_manager and self._last_increment_time:` ?

app/pulp/app/models/progress.py
werwty: I've tested it with the following:
```
progress_bar = ProgressBar(message='Publishing files', total=2, task_id="1ad4c48b-d0a1-41e2-9917-0b8e16a761a4")
progress_bar.increment()


with ProgressBar(message='Publishing files', total=2, task_id="1ad4c48b-d0a1-41e2-9917-0b8e16a761a4") as progress_bar:
    progress_bar.increment()
    progress_bar.increment()
    progress_bar.save()
```

bmbouter: Great so it sounds like Django will ignore these attributes. Thank you.

app/pulp/app/models/progress.py
```diff
@@ -70,14 +78,26 @@ def save(self, *args, **kwargs):
         """
         if self.task_id is None:
             self.task_id = Task.objects.get(id=util.get_current_task_id())
-        super(ProgressReport, self).save(*args, **kwargs)
+
+        now = timezone.now()
+
+
```
bmbouter: Can this one line of whitespace be removed?

app/pulp/app/models/progress.py
```diff
@@ -70,14 +78,26 @@ def save(self, *args, **kwargs):
...
+        self._last_save_time = now
 
     def __enter__(self):
         """
         Saves the progress report state as RUNNING
         """
         self.state = self.RUNNING
         self.save()
+        self._using_context_manager = True
+        return self
```
bmbouter: I didn't expect this to be here since it wasn't here before. Is this needed?

bmbouter: After thinking through this some, I think we do want this. Actually it would allow the docstrings to be improved some to, but you don't have to do that now if you don't want to. Specifically with this change an example that was:

```
        >>> progress_bar = ProgressBar(message='Publishing files', total=len(files_iterator))
        >>> progress_bar.save()
        >>> with progress_bar:
        >>>     for file in progress_bar.iter(files_iterator):
        >>>         handle(file)
```

Could become:

```
        >>> with ProgressBar(message='Publishing files', total=len(files_iterator)) as progress_bar:
        >>>     for file in progress_bar.iter(files_iterator):
        >>>         handle(file)
```

bmbouter: Oh actually we already claim that we do this so you are fixing it something for sure :)

https://github.com/werwty/pulp/blob/74755293cf5165db8a41f0d2ff7c4fd5482afe1d/app/pulp/app/models/progress.py#L206

I don't think any docs changes are necessary then.

werwty: We already have both those example for ProgressBar  :grin:

PR #2990, opened 2017-04-06
Removes two unexpected Importer fields

bmbouter:
The max_download_bandwidth and max_concurrent_downloads fields were not
expected to be on the model. These features are not included in the
MVP. This PR removes those fields from the model along with their
serializers.

There isn't an issue tracking this change. It is too small.


bmbouter:
Thanks @jortel

PR #2991, opened 2017-04-06
Port sync task for Pulp 3.0

asmacdo:
closes https://pulp.plan.io/issues/2399


plugin/pulp/plugin/models/importer.py
```diff
@@ -39,29 +37,22 @@ def sync(self):
...
             1. The plugin provides an implementation called WidgetImporter which subclasses Importer
 
-            2. The user makes a call to sync widget_importer (say id=10) with some override config
+            2. The user makes a call to sync widget_importer (say id=10)
 
             3. The platform loads the saved
                 >>> wi = WidgetImporter.objects.get(id=10)
 
-            4. The platform puts the WidgetImporter into read-only mode
+            4. The platform attaches the worker's working_dir to the instance.
```
bmbouter: s/attaches the worker's working_dir to the instance/sets the working_dir attribute on the instance/

app/pulp/app/tasks/importer.py
```diff
@@ -15,3 +16,16 @@ def delete(repo_name, importer_name):
...
+
+@shared_task(base=UserFacingTask)
+def sync(importer_name):
+    """
+    Call sync on the specified importer.
+
+    Args:
+        importer_name (basestring): unique name to specify the Importer
+    """
+    importer = models.Importer.objects.get(name=importer_name).cast()
```
seandst: I just realized that this won't work, because https://github.com/pulp/pulp/blob/3.0-dev/app/pulp/app/models/repository.py#L76

Importer.name is not unique. The model's uniqueness is name + repo, not name alone.

app/pulp/app/tasks/importer.py
```diff
@@ -15,3 +18,25 @@ def delete(repo_name, importer_name):
...
     """
     models.Importer.objects.filter(name=importer_name, repository__name=repo_name).delete()
+
+
+@shared_task(base=UserFacingTask)
+def sync(repo_name, importer_name):
+    """
+    Call sync on the specified importer.
+
+    Additional "business logic" checks are included here. These checks are distinct from
```
bmbouter: What about rewriting this to be specific to this task? Something like:

`This task ensures that the 'feed_url' attribute is set before calling into the sync() method provided by the plugin.`

My thinking is that the statement written now is so general it could be applied to any function that contains business logic.

app/pulp/app/tasks/importer.py
```diff
@@ -15,3 +18,25 @@ def delete(repo_name, importer_name):
...
+    validations and are responsible for ensuring that the importer instance is able to sync.
+
+    Args:
```
bmbouter: I think this needs a:

```
 Raises:
        ValueError: When feed_url is empty.
```

app/pulp/app/tasks/importer.py
```diff
@@ -15,3 +18,25 @@ def delete(repo_name, importer_name):
...
+        repo_name (basestring): unique name to specify the repository
+        importer_name (basestring): name to specify the Importer
+    """
+    importer = models.Importer.objects.get(name=importer_name, repository__name=repo_name).cast()
+    if not importer.feed_url:
+        raise ValueError("An importer must have a 'feed_url' attribute to sync.")
```
bmbouter: This string needs to be i18ln with gettext.

plugin/pulp/plugin/models/importer.py
```diff
@@ -19,18 +18,17 @@ class Importer(PlatformImporter):
     additional persistent importer data by subclassing this object and adding Django fields. We
     defer to the Django docs on extending this model definition with additional fields.
 
-    Validation is done the Django way, so custom validation can also be added the same as any
-    Django model. We defer to the Django docs on adding custom validation to the subclassed
-    importer. If any of the model validation methods are overridden, be sure to call super() so the
-    platform can still perform its validation.
+    Validation at the API level is handled by plugin specific subclass of
```
bmbouter: s/Validation at the API level is handled by/Validation of the Importer is done at the API by the/

plugin/pulp/plugin/models/importer.py
```diff
@@ -19,18 +18,17 @@ class Importer(PlatformImporter):
...
+    :class: `pulp.plugin.serializers.repository.ImporterSerializer`. If necessary, runtime
```
bmbouter: s/runtime/additional/

plugin/pulp/plugin/models/importer.py
```diff
@@ -39,29 +37,23 @@ def sync(self):
         platform :class: `pulp.app.models.Importer` base attributes and any custom attributes
         defined by the subclass.
 
-        The model attributes were loaded from the database and then had the user specified override
-        config applied on top. Additionally the importer is read-only and prevents the saving of
-        changes to the Importer instance.
+        The importer instance should be treated as read-only and plugin writers should be cautious
```
bmbouter: Based on our recent conversations can this paragraph be removed entirely?

plugin/pulp/plugin/models/importer.py
```diff
@@ -39,29 +37,23 @@ def sync(self):
...
 
-            2. The user makes a call to sync widget_importer (say id=10) with some override config
+            2. The user makes a call to sync widget_importer (say id=10)
 
             3. The platform loads the saved
                 >>> wi = WidgetImporter.objects.get(id=10)
 
-            4. The platform puts the WidgetImporter into read-only mode
+            4. The platform creates a working directory, uses it as the working directory of the 
+            interpreter, and sets the working_dir attribute on the instance.
```
bmbouter: s/instance/instance to the full path of the working directory.

plugin/pulp/plugin/models/importer.py
```diff
@@ -39,29 +37,23 @@ def sync(self):
...
             1. The plugin provides an implementation called WidgetImporter which subclasses Importer
 
-            2. The user makes a call to sync widget_importer (say id=10) with some override config
+            2. The user makes a call to sync widget_importer (say id=10)
 
             3. The platform loads the saved
                 >>> wi = WidgetImporter.objects.get(id=10)
 
-            4. The platform puts the WidgetImporter into read-only mode
+            4. The platform creates a working directory, uses it as the working directory of the
```
bmbouter: s/uses/sets/

platform/pulp/tasking/services/storage.py
```diff
@@ -137,3 +137,13 @@ def _rmtree_fix_permissions(directory_path):
...
+
+@contextmanager
+def working_dir_context():
+    """
+    Creates a context manager for a working that needs a working directory.
+    """
+    working_dir = get_working_directory()
+    os.chdir(working_dir)
+    yield working_dir
+    delete_working_directory()
```
bmbouter: If L#148 raises an exception I think we still want the working dir to be deleted regardless. What do you think having the `yield` in a `try` block and having the `delete_working_directory()` in a `finally` block?

platform/pulp/tasking/services/storage.py
```diff
@@ -137,3 +137,13 @@ def _rmtree_fix_permissions(directory_path):
             shutil.rmtree(directory_path)
         else:
             raise
+
+@contextmanager
+def working_dir_context():
+    """
+    Creates a context manager for a working that needs a working directory.
```
bmbouter: s/working that/task that/

platform/pulp/app/settings.py
```diff
@@ -110,7 +110,7 @@
     'DEFAULT_FILTER_BACKENDS': ('rest_framework.filters.DjangoFilterBackend',),
     'DEFAULT_PAGINATION_CLASS': 'pulp.app.pagination.UUIDPagination',
     'PAGE_SIZE': 100,
-    'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',)
+    # 'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',)
```
bmbouter: Is this commented on purpose?

platform/pulp/app/tasks/importer.py
```diff
@@ -15,3 +19,27 @@ def delete(repo_name, importer_name):
...
+        importer_name (basestring): name to specify the Importer.
+
+    Raises:
+        ValueError: When feed_url is empty.
+    """
+    importer = models.Importer.objects.get(name=importer_name, repository__name=repo_name).cast()
+    if not importer.feed_url:
+        raise ValueError_("An importer must have a 'feed_url' attribute to sync.")
+
+    with storage.working_dir_context() as working_dir:
```
bmbouter: This came out so great!

PR #2992, opened 2017-04-07
Use rsync's -K to preserve symlinks on publish with rsync distributor

BrnoPCmaniak:
closes #2687
https://pulp.plan.io/issues/2687

I tried to make the title to fit the 80 characters, but not sure if it still makes sense.


dralley:
ok test

PR #2995, opened 2017-04-14
Rename result field on Task model

dralley:
closes #2675
https://pulp.plan.io/issues/2675


app/pulp/app/models/task.py
```diff
@@ -152,7 +152,7 @@ class Task(Model):
     finished_at = models.DateTimeField(null=True)
 
     non_fatal_errors = JSONField(default=list)
-    result = JSONField(null=True)
```
bmbouter: The Fields in the docstring need to be updated.

app/pulp/app/models/task.py
```diff
@@ -207,7 +206,7 @@ def set_failed(self, exc, einfo):
         """
         self.state = Task.FAILED
         self.finished_at = timezone.now()
-        self.result = exception_to_dict(exc, einfo.traceback)
```
bmbouter: The docstring of this function needs updating.

dralley:
@bmbouter Changes made

As an aside, could you help me understand the rationale for this change?  I'm not sure I understand the benefits of using 2 fields to store fatal and non-fatal errors separately, and not storing results.

bmbouter:
@dralley Thanks for the update. Those are all good questions. I'll try to give some answers.

We're removing the 'results' field because (I think) plugin writer's won't have a need to store results. Besides errors, the entire record of what happened will be recorded as Progress Report instances associated with the task. In Pulp2 the results we had were things like '25/25 rpms downloaded' and in Pulp3 that would be recorded as a progress report. So if all the information is in progress reports what would even be stored as a result.

If we keep the 'results' field we're going to have to figure out a way to display it via the CLI. Since it's an unstructured field (JSONField) and plugin writers stuff a variety of different data structures into it this will become problematic. So by removing the results field we have a derivative benefit of keeping the CLI code generic and CLI experience awesome.

Regarding having two ways to record errors, we want to record, treat, and think about fatal and non-fatal exceptions differently. There will be either 0 or 1 fatal exceptions while there can be 0 .. N non-fatal exceptions. We could both exception types into a single JSONField that has a 'fatal' and 'non-fatal' keys but having them be modeled as separate fields is a more structured way of doing it. Having more structure is better. Having them separate also makes the CLIs job easier because if there is a fatal exception it should be shown in red to the user while a non-fatal exception could not be shown by default or maybe shown in yellow.

There was a tiny bit of discussion on this [here](https://www.redhat.com/archives/pulp-dev/2017-March/msg00054.html). Ask more questions if you have them!

dralley:
Thanks for the explanation!

PR #2997, opened 2017-04-19
Don't use ssh connection sharing in rsync distributor.

ipanova:
closes #2689
https://pulp.plan.io/issues/2689

I removed -S as well since it is the location of a control socket for connection sharing( which we are removing in this PR)


ipanova:
ok test

ipanova:
ok test

dralley:
I didn't attempt to reproduce the issue but I can confirm that everything seems to work fine with this change.  @dkliban, any input?

ipanova:
@dralley i did not reproduce the issue as well, i just made sure after this change the rsync stil lwas in operational state

PR #2999, opened 2017-04-21
Support for storing units in multiple collections

goosemania:
The following was added:
 - load of auxiliary models
 - post-delete method during orphan removal
 - serialization in profiler to have all the data
   in a proper shape for applicability calculation

re #2681
https://pulp.plan.io/issues/2681

### DIFF
--- a/server/pulp/plugins/loader/manager.py
+++ b/server/pulp/plugins/loader/manager.py
@@ -13,6 +13,7 @@ from pulp.server.exceptions import PulpCodedException
 _logger = logging.getLogger(__name__)
 
 ENTRY_POINT_UNIT_MODELS = 'pulp.unit_models'
+ENTRY_POINT_AUXILIARY_MODELS = 'pulp.auxiliary_models'
 
 
 class PluginManager(object):
@@ -29,9 +30,11 @@ class PluginManager(object):
         self.profilers = _PluginMap()
         self.catalogers = _PluginMap()
         self.unit_models = dict()
+        self.auxiliary_models = dict()
 
         # Load the unit models
         self._load_unit_models()
+        self._load_auxiliary_models()
 
     def _load_unit_models(self):
         """"
@@ -40,8 +43,8 @@ class PluginManager(object):
         Attach the signals to the models here since the mongoengine signals will not be
         sent correctly if they are attached to the base class.
 
-        :raises: TypeError if a model is not a subclass of ContentUnit
-        :raises: PLP0038 if two models are defined with the same id
+        :raises TypeError: if a model is not a subclass of ContentUnit
+        :raises PulpCodedException: PLP0038 if two models are defined with the same id
         """
         _logger.debug(_("Loading Unit Models"))
         for entry_point in pkg_resources.iter_entry_points(ENTRY_POINT_UNIT_MODELS):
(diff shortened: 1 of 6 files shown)


server/pulp/plugins/loader/manager.py
```diff
@@ -66,6 +69,21 @@ def _load_unit_models(self):
 
         _logger.debug(_("Unit Model Loading Completed"))
 
+    def _load_auxiliary_models(self):
```
mhrivnak: need a doc block

server/pulp/server/db/migrations/lib/utils.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Logs header indicating start of migration.
+        Expected to be called before migration process begins.
+        """
+        stars = '*' * len(MIGRATION_INIT_MSG)
+        _logger.info(stars)
+        _logger.info(MIGRATION_INIT_MSG)
+        _logger.info(stars)
+
+    def start(self, content_type, total_units, msg=MIGRATION_START_MSG):
```
mhrivnak: This is begging to be a context manager! ;)

I think if you just create an ``__enter__`` and ``__exit__`` method that call ``start`` and ``stop`` respectively, that'll do it.

server/pulp/server/managers/content/orphan.py
```diff
@@ -237,6 +243,11 @@ def delete_orphans_by_type(content_type_id, content_unit_ids=None):
                 unit_type_id=content_type_id
             ).delete()
             content_units_collection.remove(content_unit['_id'])
+            try:
+                content_model.do_post_delete_actions(content_unit)
+            except AttributeError:
```
mhrivnak: This worries me only because ``AttributeError`` is a fairly common exception to see, especially from a basic programming error, and it would be very unfortunate if something running inside ``do_post_delete_actions`` raised an unhandled ``AttributeError`` that we never saw.

Maybe it would be better in this case to explicitly check for the method you want before calling it.

server/pulp/server/managers/content/orphan.py
```diff
@@ -261,14 +272,21 @@ def delete_orphan_content_units_by_type(type_id, content_unit_ids=None):
         """
         # get the model matching the type
         content_model = plugin_api.get_unit_model_by_id(type_id)
+        try:
+            unit_key_fields = units_controller.get_unit_key_fields_for_type(type_id)
+        except ValueError:
+            raise MissingResource(content_type_id=type_id)
+
+        fields = ('id', '_storage_path') + unit_key_fields
+        count = 0
```
mhrivnak: Is this needed? Looks like ``count`` doesn't get used until below where it gets again set to ``0``.

server/pulp/server/managers/content/orphan.py
```diff
@@ -224,10 +224,16 @@ def delete_orphans_by_type(content_type_id, content_unit_ids=None):
...
         content_units_collection = content_types_db.type_units_collection(content_type_id)
+        content_model = plugin_api.get_unit_model_by_id(content_type_id)
 
+        try:
+            unit_key_fields = units_controller.get_unit_key_fields_for_type(content_type_id)
+        except ValueError:
+            raise MissingResource(content_type_id=content_type_id)
+
+        fields = ('_id', '_storage_path') + unit_key_fields
         count = 0
```
mhrivnak: Is this needed? Looks like ``count`` doesn't get used until below where it gets again set to ``0``.

server/pulp/server/managers/content/orphan.py
```diff
@@ -294,6 +312,12 @@ def delete_orphan_content_units_by_type(type_id, content_unit_ids=None):
                     unit_type_id=str(type_id)
                 ).delete()
                 unit_to_delete.delete()
+                try:
+                    content_model.do_post_delete_actions(unit_to_delete)
+                except AttributeError:
```
mhrivnak: same here as above

dkliban:
ok test

server/pulp/server/db/migrations/lib/utils.py
goosemania: Yep, I need to re-write it a bit then, because currently it is possible to use the same instance several times (aka several start/stop-s). It is useful when multiple content types are migrated (like RPM, DRPM, SRPM). They all share the same initial message and only then specific info is printed. But that's a minor issue.

server/pulp/server/managers/content/orphan.py
goosemania: Yes, so true. +1

server/pulp/server/managers/content/orphan.py
goosemania: This one is needed , it's used on line 255.

server/pulp/server/managers/content/orphan.py
goosemania: This one is not needed, thanks

server/pulp/plugins/loader/manager.py
```diff
@@ -66,6 +69,27 @@ def _load_unit_models(self):
...
+        """
+        Load all auxiliary models from the ENTRY_POINT_AUXILIARY_MODELS.
+        Auxiliary models are plugin specific but they are not unit models themselves.
+
+        :raises: PLP0038 if two models are defined with the same id
```
mhrivnak: I think this should be

```
:raises PulpCodedException: PLP0038 if two models are defined with the same id
```

PR #3000, opened 2017-04-23
Fix 'referenced before assignment' errors

dralley:
Fix several cases where variables defined behind conditionals
are referenced elsewhere in the code from an upper scope.

closes #2722
https://pulp.plan.io/issues/2722

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1129,6 +1129,7 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
     :rtype:  pulp.server.db.model.repository.RepoPublishResult
     """
     force_full = call_config.get('force_full', False)
+    predistributor_id = call_config.get('predistributor_id')
     config_override = call_config.override_config
     last_published = conduit.last_publish()
     last_unit_removed = repo_obj.last_unit_removed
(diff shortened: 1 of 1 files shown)


dralley:
After resolving #2722 I noticed that the same problem was present with other variables in multiple other areas of the code.  I've pointed those out in the comments and resolved them as well.

server/pulp/server/controllers/repository.py
```diff
@@ -1157,13 +1143,36 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
             repo_id=repo_obj.repo_id,
             distributor_id=dist_id).update(set__last_override_config=config_override)
 
+    skip_for_predistributor = False
     # Check if a predistributor is configured and the predistributor has not published since the
     # last publish.
-    skip_for_predistributor = (predistributor_id and (published_after_predistributor or
```
dralley: This line referenced the variables ```published_after_predistributor``` and ```predistributor_last_published```, both of which were defined behind the conditional beginning on L1146.

dralley:
Also, should I set this against 2.12-dev or 2.13-dev?

dralley:
ok test

ipanova:
we had some discussion on irc wit @dralley , @dkliban we would like to hear your opinion as well http://pastebin.test.redhat.com/477926

dralley:
@ipanova I got my environment working and reproduced the issue, and also verified the fix.  This is ready for review

dralley:
^^

ipanova:
@dralley the logic seems to be correct, i also handtested the bug and did some other regular sanity checks for rsync functionality. good job, lgtm :four_leaf_clover: :lollipop: :sparkles:

PR #3002, opened 2017-04-28
Update apply_async and apply_async_with_reservations signatures

BrnoPCmaniak:
Now one should be able to tell what is passing to where.

closes #2658
https://pulp.plan.io/issues/2658

### DIFF
--- a/platform/pulp/tasking/tasks.py
+++ b/platform/pulp/tasking/tasks.py
@@ -6,20 +6,20 @@ import time
 import uuid
 from gettext import gettext as _
 
-from celery import Task as CeleryTask, task
+from celery import Task as CeleryTask
+from celery import task
 from celery.app import control
 from celery.result import AsyncResult
-from django.db import transaction, IntegrityError
 from django.conf import settings as pulp_settings
+from django.db import IntegrityError, transaction
 
-from pulp.app.models import ReservedResource, Task as TaskStatus, TaskLock, Worker
+from pulp.app.models import Task as TaskStatus
+from pulp.app.models import ReservedResource, TaskLock, Worker
 from pulp.common import TASK_FINAL_STATES, TASK_INCOMPLETE_STATES, TASK_STATES
 from pulp.exceptions import MissingResource, PulpException
 from pulp.tasking import util
-from pulp.tasking.celery_instance import celery
-from pulp.tasking.celery_instance import DEDICATED_QUEUE_EXCHANGE, RESOURCE_MANAGER_QUEUE
+from pulp.tasking.celery_instance import DEDICATED_QUEUE_EXCHANGE, RESOURCE_MANAGER_QUEUE, celery
 from pulp.tasking.services import storage
-
 
 celery_controller = control.Control(app=celery)
 _logger = logging.getLogger(__name__)
@@ -42,7 +42,7 @@ class PulpTask(CeleryTask):
 
 
 @task(base=PulpTask, acks_late=True)
-def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwargs):
+def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwargs, options):
     """
     A task that encapsulates another task to be dispatched later. This task being encapsulated is
     called the "inner" task, and a task name, UUID, and accepts a list of positional args
@@ -54,18 +54,20 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
     time. The logic deciding which queue receives a task is controlled through the
     find_worker function.
 
-    :param name:          The name of the task to be called
-    :type name:           basestring
-    :param inner_task_id: The UUID to be set on the task being called. By providing
-                          the UUID, the caller can have an asynchronous reference to the inner task
-                          that will be dispatched.
-    :type inner_task_id:  basestring
-    :param resource_id:   The name of the resource you wish to reserve for your task. The system
-                          will ensure that no other tasks that want that same reservation will run
-                          concurrently with yours.
-    :type  resource_id:   basestring
+    Args:
+        name (basestring): The name of the task to be called
+        inner_task_id (basestring): The UUID to be set on the task being called. By providing
+            the UUID, the caller can have an asynchronous reference to the inner task
+            that will be dispatched.
+        resource_id (basestring): The name of the resource you wish to reserve for your task.
+            The system will ensure that no other tasks that want that same reservation will run
+            concurrently with yours.
+        inner_args (tuple): The positional arguments to pass on to the task.
+        inner_kwargs (dict): The keyword arguments to pass on to the task.
+        options (dict): For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
 
-    :return: None
+    Returns:
+        An AsyncResult instance as returned by Celery's apply_async
     """
     while True:
         # Find a worker who already has this reservation, it is safe to send this work to them
@@ -89,14 +91,12 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
     task_status = TaskStatus.objects.get(pk=inner_task_id)
     ReservedResource.objects.create(task=task_status, worker=worker, resource=resource_id)
 
-    inner_kwargs['routing_key'] = worker.name
-    inner_kwargs['exchange'] = DEDICATED_QUEUE_EXCHANGE
-    inner_kwargs['task_id'] = inner_task_id
-
     try:
-        celery.tasks[name].apply_async(*inner_args, **inner_kwargs)
+        celery.tasks[name].apply_async(args=inner_args, task_id=inner_task_id,
+                                       routing_key=worker.name, exchange=DEDICATED_QUEUE_EXCHANGE,
+                                       kwargs=inner_kwargs, **options)
     finally:
-        _release_resource.apply_async((inner_task_id, ), routing_key=worker.name,
+        _release_resource.apply_async(args=(inner_task_id, ), routing_key=worker.name,
                                       exchange=DEDICATED_QUEUE_EXCHANGE)
 
 
@@ -109,8 +109,9 @@ def _release_resource(task_id):
     When a resource-reserving task is complete, this method releases the resource by removing the
     ReservedResource object by UUID.
 
-    :param task_id: The UUID of the task that requested the reservation
-    :type  task_id: basestring
+    Args:
+        task_id (basestring): The UUID of the task that requested the reservation
+
     """
     try:
         TaskStatus.objects.get(pk=task_id, state=TASK_STATES.RUNNING)
@@ -149,7 +150,8 @@ class UserFacingTask(PulpTask):
     # this tells celery to not automatically log tracebacks for these exceptions
     throws = (PulpException,)
 
-    def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwargs):
+    def apply_async_with_reservation(self, resource_type, resource_id, tags=[], group_id=None,
+                                     args=None, kwargs=None, **options):
         """
         This method provides normal apply_async functionality, while also serializing tasks by
         resource name. No two tasks that claim the same resource name can execute concurrently. It
@@ -163,35 +165,26 @@ class UserFacingTask(PulpTask):
         task just after calling this method, so a Task entry needs to exist for it
         before it returns.
 
-        For a list of parameters accepted by the ``*args`` and ``**kwargs`` parameters, please see
-        the docblock for the :meth:`apply_async` method.
+        Args:
+            resource_type (basestring): A string that identifies type of a resource
+            resource_id (basestring): A string that identifies some named resource, guaranteeing
+                that only one task reserving this same string can happen at a time.
+            tags (list): A list of tags (strings) to place onto the task, used for searching
+                for tasks by tag. This is an optional argument which is pulled out of kwargs.
+            group_id (uuid.UUID): The id to identify which group of tasks a task belongs to.
+                This is an optional argument which is pulled out of kwargs.
+            args (tuple): The positional arguments to pass on to the task.
+            kwargs (dict): The keyword arguments to pass on to the task.
+            options (dict): For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
 
-        :param resource_type: A string that identifies type of a resource
-        :type resource_type:  basestring
-
-        :param resource_id:   A string that identifies some named resource, guaranteeing that only
-                              one task reserving this same string can happen at a time.
-        :type  resource_id:   basestring
-
-        :param tags:          A list of tags (strings) to place onto the task, used for searching
-                              for tasks by tag. This is an optional argument which is pulled out of
-                              kwargs.
-        :type  tags:          list
-
-        :param group_id:      The id to identify which group of tasks a task belongs to. This is an
-                              optional argument which is pulled out of kwargs.
-        :type  group_id:      uuid.UUID
-
-        :return:              An AsyncResult instance as returned by Celery's apply_async
-        :rtype:               celery.result.AsyncResult
+        Returns (celery.result.AsyncResult):
+            An AsyncResult instance as returned by Celery's apply_async
         """
         # Form a resource_id for reservation by combining given resource type and id. This way,
         # two different resources having the same id will not block each other.
         resource_id = ":".join((resource_type, resource_id))
         inner_task_id = str(uuid.uuid4())
         task_name = self.name
-        tag_list = kwargs.get('tags', [])
-        group_id = kwargs.get('group_id', None)
 
         # Set the parent attribute if being dispatched inside of a Task
         parent_arg = self._get_parent_arg()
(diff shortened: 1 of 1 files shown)


bmbouter:
This needs a new push to resolve conflicts.

tasking/pulp/tasking/tasks.py
```diff
@@ -149,7 +149,7 @@ class UserFacingTask(PulpTask):
     # this tells celery to not automatically log tracebacks for these exceptions
     throws = (PulpException,)
 
-    def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwargs):
+    def apply_async_with_reservation(self, resource_type, resource_id, *async_args, **async_kwargs):
```
bmbouter: Can the docstrings be updated to include `async_args` and `async_kwargs`.

tasking/pulp/tasking/tasks.py
```diff
@@ -204,17 +205,23 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
 
         # Call the outer task which is a promise to call the real task when it can.
-        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
-                                         queue=RESOURCE_MANAGER_QUEUE)
+        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, async_args,
+                                               async_kwargs], queue=RESOURCE_MANAGER_QUEUE)
         return AsyncResult(inner_task_id)
 
-    def apply_async(self, *args, **kwargs):
+    def apply_async(self, *celery_args, tags=[], group_id=None, **celery_kwargs):
```
bmbouter: Can the docstring be updated to include `celery_args` and `celery_kwargs`.

tasking/pulp/tasking/tasks.py
bmbouter: I think this needs to be `args=None, kwargs=None` instead of `*async_args, **async_kwargs`. It needs to match the function signature of celery's apply_async. http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async

tasking/pulp/tasking/tasks.py
```diff
@@ -190,8 +191,8 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         resource_id = ":".join((resource_type, resource_id))
         inner_task_id = str(uuid.uuid4())
         task_name = self.name
-        tag_list = kwargs.get('tags', [])
-        group_id = kwargs.get('group_id', None)
+        tag_list = async_kwargs.get('tags', [])
```
bmbouter: I think tags and group_id should become explicit arguments and not expected to be pulled out of async_kwargs. This way async_kwargs will only store kwargs that will be sent to celery.

tasking/pulp/tasking/tasks.py
bmbouter: I think this also needs to be `celery_args=None` instead of `*celery_args` and `celery_kwargs=None` instead of `**celery_kwargs`.

BrnoPCmaniak:
@bmbouter I'm not sure if I get the task how it was ment. My understanding of the task was that all parameters of `apply_async` are in one `*args` and `**kwargs` and then in the function we popped some things from it. Which was what I think is the confusing part. 

So I made changes to separate these things. Also I renamed the parameters in way that I found more understandable that they are **not** the `args` and `kwargs` parameters on the celery function but rather the whole `*args` and `**kwargs` of the function. And therefore our function would become a wrapper which just takes some more arguments and then everything else is passed to the celery `apply_async`.

I get this understatement from the fact that parameters on  http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async
`apply_async` use `**options` so I thought that it's better to group all celery params into one `**kwargs` and pass them uniformly so it would be more clear, but probably it's not. 

Also I don't understand what you mean with updating the docstring. Both of the function have a paragraph containing the parameters e.g. https://github.com/pulp/pulp/pull/3002/files#diff-6233f95b689d7b89c6debe53c073b2d7R166 

But I can change it to correspond exactly to the way how celery signature looks like, if you think it's the way how it was meant.

werwty:
@BrnoPCmaniak I think we do want to duplicate celery's function signature as-is. 
The 'args' and 'kwargs' that celery takes are arguments passed onto the task, whereas the **options is used in the celery task. It would be good to make this distinction explicit in our apply_async function signature.

I envisioned our apply_async function signature to be:
`apply_async(self, tags=[], group_id=None, task_args=None, task_kwargs=None, **celery_options):`
and we would call the super with:
`async_result = super(UserFacingTask, self).apply_async(args=task_args, kwargs=task_kwargs, **celery_options)`

bmbouter:
@BrnoPCmaniak I also imagined it as @werwty wrote in her comment. We want to mimic Celery's [`apply_async()` signature](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async) but with our own keyword arguments `tags` and `group_id` to be accepted directly in the signature.

`apply_async_with_reservation(self, resource_type, resource_id, tags=[], group_id=None, args=None, kwargs=None, **options):`

Similarly for `apply_async()` it would be:

`apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options):`

and we would call the super with:

`async_result = super(UserFacingTask, self).apply_async(args=args, kwargs=kwargs, **options)`

bmbouter:
@BrnoPCmaniak Ignore my comment about the docstrings, I didn't realize we documented already like parameters even though your diff is adding them as parameters. Thanks!

BrnoPCmaniak:
@bmbouter @werwty Fixed.

platform/pulp/tasking/tasks.py
```diff
@@ -225,14 +237,26 @@ def apply_async(self, *args, **kwargs):
...
 
         :return:            An AsyncResult instance as returned by Celery's apply_async
         :rtype:             celery.result.AsyncResult
         """
-        tag_list = kwargs.pop('tags', [])
-        group_id = kwargs.pop('group_id', None)
-        async_result = super(UserFacingTask, self).apply_async(*args, **kwargs)
-        async_result.tags = tag_list
+        async_kwargs = {
+            "tags": tags,
```
werwty: we should not be passing `tags` and `group_id` to celery

platform/pulp/tasking/tasks.py
```diff
@@ -200,15 +201,26 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
+            "group_id": group_id,
+            "args": args,
+            "kwargs": kwargs,
+            "options": options
+        }
+
         # Call the outer task which is a promise to call the real task when it can.
-        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
+        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, async_args,
```
werwty: It looks like `_queue_reserved_task` wants `inner_args, inner_kwargs` not async_args and async_kwargs, so lines 208-215 can be removed

werwty: Oh nevermind, I see _queue_reserved_task passes this back to apply_async.

platform/pulp/tasking/tasks.py
```diff
@@ -225,14 +237,24 @@ def apply_async(self, *args, **kwargs):
...
         :rtype:             celery.result.AsyncResult
         """
-        tag_list = kwargs.pop('tags', [])
-        group_id = kwargs.pop('group_id', None)
-        async_result = super(UserFacingTask, self).apply_async(*args, **kwargs)
-        async_result.tags = tag_list
+        async_kwargs = {
+            "args": args,
+            "kwargs": kwargs,
+            "options": options
```
werwty: options isn't an keyword argument, it's a dict of keywork arguments that should be an unpacked before being passed in to apply_async

`async_result = super(UserFacingTask, self).apply_async(args=args, kwargs=kwargs, **options)`

werwty: Same thing here: https://github.com/BrnoPCmaniak/pulp/blob/10d85de6b3a30e06a36e7d156cb1566731b9b794/platform/pulp/tasking/tasks.py#L214

platform/pulp/tasking/tasks.py
```diff
@@ -225,14 +237,24 @@ def apply_async(self, *args, **kwargs):
 
         :param group_id:    The id that identifies which group of tasks a task belongs to
         :type group_id:     uuid.UUID
+        :param args:          The positional arguments to pass on to the task.
+        :type  args:          tuple
+        :param kwargs:        The keyword arguments to pass on to the task.
+        :type  kwargs:        dict
+        :param **options:     For all options accepted by apply_async please visit:
+                              http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
```
bmbouter: This could be combined with the previous line. It would be too long but it would be more readable, and the #noqa would let it pass with flake8

platform/pulp/tasking/tasks.py
```diff
@@ -200,15 +201,26 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
             task_status = TaskStatus.objects.create(pk=inner_task_id, state=TaskStatus.WAITING,
                                                     group=group_id, **parent_arg)
-            for tag in tag_list:
+            for tag in tags:
                 task_status.tags.create(name=tag)
 
+        # Wrap celery signature into kwargs for _queue_reserved_task
+        async_args = tuple()
+        async_kwargs = {
+            "tags": tags,
```
bmbouter: This wasn't exactly part of the story, but I don't think that tags and group_id should be kept in the async_kwargs. They are only needed in the few statements above and do not need to be passed along to the task.

platform/pulp/tasking/tasks.py
bmbouter: Oh nevermind this is just github wrapping the line. You've already done what I asked for.

BrnoPCmaniak:
@bmbouter Fixed.

platform/pulp/tasking/tasks.py
```diff
@@ -200,15 +201,26 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
+
         # Call the outer task which is a promise to call the real task when it can.
-        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
+        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, async_args,
+                                               async_kwargs],
                                          queue=RESOURCE_MANAGER_QUEUE)
         return AsyncResult(inner_task_id)
 
-    def apply_async(self, *args, **kwargs):
+    def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options):
```
bmbouter: I think queue can be removed from this docstring.

platform/pulp/tasking/tasks.py
```diff
@@ -200,15 +201,26 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
+            "tags": tags,
+            "group_id": group_id,
+            "args": args,
+            "kwargs": kwargs,
+        }
+        async_kwargs.update(options)
+
         # Call the outer task which is a promise to call the real task when it can.
-        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
+        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, async_args,
```
bmbouter: I think this should become:

`_queue_reserved_task.apply_async(args=(task_name, inner_task_id, resource_id, args,  kwargs), queue=RESOURCE_MANAGER_QUEUE)`

The change includes a switch from a list to a tuple. Also it ensures `args` and `kwargs` are passed as positional arguments to _queue_reserve_task which expects them as positional arguments.

platform/pulp/tasking/tasks.py
```diff
@@ -96,7 +96,7 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
     try:
         celery.tasks[name].apply_async(*inner_args, **inner_kwargs)
```
werwty: I believe this is correct as is. inner_kwargs in this case needs to be unpacked, since it's:
```
async_kwargs = {
            "args": args,
            "kwargs": kwargs,
}
```

BrnoPCmaniak:
@bmbouter I'm definitely in for the change into tuple. But how @werwty pointed out I really think it should be this way.

platform/pulp/tasking/tasks.py
bmbouter: I think we should avoid calling apply_async() with *something, or **something_else because our args and kwargs to the task are being mixed with the options we are passing to apply_sync itself. Say one of our tasks wants to use a keyword that is also used by apply_async like `max_retries`. Couldn't this be an issue?

It would be avoided (I think) by always having apply_async have parameters (positional and keyword) that are meant for the task itself being delivered via a keyword named `args` and `kwargs` to the apply_async().

Does that make sense? I'm still maybe a little confused. (It's complicated). :)

BrnoPCmaniak: Well the end task's parameters always live inside their own dict and tuple and never get mixed with others. For example I call this https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L152 this:
```
some_task.apply_async_with_reservation("RPM", "dnf-2.5.0", tags=[], group_id=None, args=(1, 2), kwargs={"max_retries": 5}, max_retries=10)
```
this leads to https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L45
```
_queue_reserved_task.apply_async(args=["some_task", "123e4567-e89b-12d3-a456-426655440000", "dnf-2.5.0", tuple(), {"args": (1, 2),"kwargs": {"max_retries": 5}, "max_retries": 10}], queue=RESOURCE_MANAGER_QUEUE)
```

Thanks to this method on line https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L45 will be called taht method with parameters:

```
_queue_reserved_task("some_task", "123e4567-e89b-12d3-a456-426655440000", "dnf-2.5.0", tuple(), {"args": (1, 2),"kwargs": {"max_retries": 5}, "max_retries": 10})
```

And then the apply async is finally called.
https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L220
```
some_task.apply_async(args=(1, 2), kwargs={"max_retries": 5}, max_retries=10)
```

Which then calls the task itself:
```
some_task(1,2 max_retries=5)
```

werwty: If you update async_args [here](https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L207) to be args


```
 async_args = args
 async_kwargs = {
            "kwargs": kwargs,
 }
```

Then at https://github.com/BrnoPCmaniak/pulp/blob/b6e579ca3a8ec7d50bd0127738d2403b5620f7ee/platform/pulp/tasking/tasks.py#L97
you don't have to unpack and can call apply_async with:
`celery.tasks[name].apply_async(args=inner_args, kwargs=inner_kwargs)`

BrnoPCmaniak: @werwty I can't agree with you on that. What you propose is half solution because then on the end task you would end up with `task(*args, kwargs={kwargs: {...}})`"and also you would loose this way ability to pass all additional arguments to apply_async.

werwty: @BrnoPCmaniak I'm thinking of something like: https://github.com/werwty/pulp/commit/5c8178d7ac2637d9fd0b7eb6f53f6802a36525cd

Where additional arguments are being passed through options, and you would end up with `task(args=args, kwargs=kwargs, **options)`

platform/pulp/tasking/tasks.py
```diff
@@ -64,6 +64,12 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
                           will ensure that no other tasks that want that same reservation will run
                           concurrently with yours.
     :type  resource_id:   basestring
+    :param inner_args:    The positional arguments to pass on to the task.
+    :type  inner_args:    tuple
+    :param inner_kwargs:  The keyword arguments to pass on to the task.
+    :type  inner_kwargs:  dict
+    :param **options:     For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
```
bmbouter: Since this is a positional argument in the function signature this docstring should be s/**//.

platform/pulp/tasking/tasks.py
```diff
@@ -64,6 +64,12 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
                           will ensure that no other tasks that want that same reservation will run
                           concurrently with yours.
     :type  resource_id:   basestring
+    :param inner_args:    The positional arguments to pass on to the task.
```
bmbouter: Can these docstrings be converted to Napoleon per [PUP2](https://github.com/pulp/pups/blob/master/pup-0002.md)? This can be done as a separate PR, but I think it would be good. Here is a regex to help with the conversation:  https://pulp.plan.io/issues/2347?pn=1#note-14 Thank you for anything you do in this area.

platform/pulp/tasking/tasks.py
```diff
@@ -200,15 +204,16 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
         with transaction.atomic():
             task_status = TaskStatus.objects.create(pk=inner_task_id, state=TaskStatus.WAITING,
                                                     group=group_id, **parent_arg)
-            for tag in tag_list:
+            for tag in tags:
                 task_status.tags.create(name=tag)
 
         # Call the outer task which is a promise to call the real task when it can.
-        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args, kwargs],
+        _queue_reserved_task.apply_async(args=[task_name, inner_task_id, resource_id, args,
```
bmbouter: Can this list [task_name, ... ] be converted to a tuple for more memory efficiency?

platform/pulp/tasking/tasks.py
```diff
@@ -225,14 +230,18 @@ def apply_async(self, *args, **kwargs):
 
         :param group_id:    The id that identifies which group of tasks a task belongs to
         :type group_id:     uuid.UUID
+        :param args:          The positional arguments to pass on to the task.
```
bmbouter: A few lines above L#223 - L#225 has queue as a parameter but it's not accepted as a named parameter. Please remove queue from this docstring.

platform/pulp/tasking/tasks.py
```diff
@@ -225,14 +230,18 @@ def apply_async(self, *args, **kwargs):
...
         :return:            An AsyncResult instance as returned by Celery's apply_async
         :rtype:             celery.result.AsyncResult
         """
-        tag_list = kwargs.pop('tags', [])
-        group_id = kwargs.pop('group_id', None)
-        async_result = super(UserFacingTask, self).apply_async(*args, **kwargs)
-        async_result.tags = tag_list
+
+        async_result = super(UserFacingTask, self).apply_async(args=args, kwargs=kwargs, **options)
+        async_result.tags = tags
```
bmbouter: This line should be removed. The async_result returned does not need `tags` as an attribute. This is a holdover from the Pulp2 code which also did not use it.

platform/pulp/tasking/tasks.py
```diff
@@ -54,18 +54,20 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
...
+        name (basestring): The name of the task to be called
+        inner_task_id (basestring): The UUID to be set on the task being called. By providing
+            the UUID, the caller can have an asynchronous reference to the inner task
+            that will be dispatched.
+        resource_id (basestring): The name of the resource you wish to reserve for your task.
+            The system will ensure that no other tasks that want that same reservation will run
+            concurrently with yours.
+        inner_args (tuple): The positional arguments to pass on to the task.
+        inner_kwargs (dict): The keyword arguments to pass on to the task.
+        options: For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
```
bmbouter: This should have a type also. I think `options (dict)`.

platform/pulp/tasking/tasks.py
```diff
@@ -163,35 +165,26 @@ def apply_async_with_reservation(self, resource_type, resource_id, *args, **kwar
...
+            resource_type (basestring): A string that identifies type of a resource
+            resource_id (basestring): A string that identifies some named resource, guaranteeing
+                that only one task reserving this same string can happen at a time.
+            tags (list): A list of tags (strings) to place onto the task, used for searching
+                for tasks by tag. This is an optional argument which is pulled out of kwargs.
+            group_id (uuid.UUID): The id to identify which group of tasks a task belongs to.
+                This is an optional argument which is pulled out of kwargs.
+            args (tuple): The positional arguments to pass on to the task.
+            kwargs (dict): The keyword arguments to pass on to the task.
+            options:     For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
```
bmbouter: Similar here. `options (dict):`

platform/pulp/tasking/tasks.py
```diff
@@ -300,20 +284,12 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
-        :param args:    Original arguments for the executed task.
-        :type args:     list
-
-        :param kwargs:  Original keyword arguments for the executed task.
-        :type kwargs:   dict
-
-        :param einfo:   celery's ExceptionInfo instance, containing serialized traceback.
-        :type einfo:    ???
+        Args:
+            exc: The exception raised by the task.
```
bmbouter: This could maybe be `exc (BaseException):`

PR #3003, opened 2017-05-01
Systemd unit files moved to devel repository

asmacdo:
Units will need to be created differently by the various ways to
install Pulp, so a unit file that hardcodes the path to the Pulp
dependency is not ideal.

Intended to be merged at the same time as https://github.com/pulp/devel/pull/51

https://pulp.plan.io/issues/2742


BrnoPCmaniak:
@asmacdo What about leaving the systemd files as they are and instead of that add drop-in into `pulp_*.service.d/` which would override only the specific things given the installation way.

BrnoPCmaniak:
Or you could use the environment file and change the path into something like `ExecStart=${PULP_VIRTUAl_ENV:-/usr/bin/celery} ...`
and in environment file have something like
`PULP_VIRTUAl_ENV=/home/vagrant/...`

asmacdo:
Would using an environment file like @BrnoPCmaniak suggested work for SCLs @pcreech?

pcreech:
There are couple environment specific directives that will need to be modified for an SCL version of the *.service files.  
* Another 'EnvironmentFile' directive, 
* Modifiying the ExecStart directive prepending scl enable commands.

I don't think we will be able to come up with a master service file definition to fit all use cases.

asmacdo:
@BrnoPCmaniak, If development service files are different from production, I would rather keep these in pulp/devel so we can modify them and the ansible role that uses them in one pass.

dralley:
@asmacdo could you rebase this on top of the current 3.0-dev?  I hit an error because my pulp/devel changes are expecting stuff not present

asmacdo:
@dralley, rebased

PR #3006, opened 2017-05-08
Download package and plugin API - ChangeSet 

jortel:
Introduction
============

This proposal attempts to provide tooling intended to reduce the effort required for plugin writers to contribute a plugin.  

It focus narrowly on:
 
* File downloading.
* Creating content and artifacts.
* Content association with a repository.
* Deferred (Lazy) downloading *for-free* on minimal burden to plugin writer.
* Support the streamer.
   * Twisted integration.
   * Support validation when streaming the bits.
   * Each download is autonomous and provides protocol independent interface.
* Keeping the plugin API lean-and-mean.

The proposed solution is layered.

Layer 1
-------

**Concurrent Downloading.**
Plugin writers are free to use any lib they want to download files.  However, I don't know of any that support concurrency and validation. The proposed solution is an abstract `Download` (noun) object that can also be executed concurrently within a `Batch`. A basic `HttpDownload` is provided. The `Download` hierarchy is intended to be extended to support additional protocols such asFTP or [BitTorrent](http://www.bittorrent.com/) in the future and work nicely with the `Batch`. A basic file size and *checksum* validation is included.  Additional validations can be supplied by both platform and plugin writers.

Example:
```
from pulp.plugin import Batch, HttpDownload

url = 'http://redhat.com/content/great.rpm'
path = '/tmp/working/great.rpm'
downloads = (HttpDownload(url, path) for _ in range(10))

with Batch(downloads) as batch:
    for plan in batch():
        if plan.succeeded:
            # Use the downloaded file
            pass
        else:
            # Log something
            pass
```

Note to reviewer: Start here: https://github.com/pulp/pulp/pull/3006/files#diff-30405389285d884762c0188393e28743

Layer 2
-------

**Content Changes To A Repository.**
The goal is for the plugin writer to only be concerned with determining what content needs to be added/deleted from a repository.  Once determined, a *generator* of `RemoteContent` to be added and a *generator* of `Content` to be removed is used to create a `ChangeSet`.  The term *remote content* refers to content that exists in the remote repository but does not yet exist locally (in pulp).  The `ChangeSet` is then *applied* to the repository.  The result is a *generator* of `ChangeReport`.  Each report contains an *action* (ADDED | DELETED) and the *content*.  The plugin can iterate these reports to determine overall success for failure.

The `ChangeSet` is capable of detecting the `Importer` download policy and act accordingly.  When content downloading is deferred, catalog entries are created instead of downloading artifact files.  Concurrent downloading is handled by the `ChangeSet` using the *layer1* `download` lib.

The plugin writer does not need to be concerned with whether a content (unit) already exists (or not).  The
`ChangeSet` determines this and acts appropriately.

The concept of `RemoteContent` and `RemoteArtifact` is introduced.  The term *remote* is used because it describes content and artifacts that exist in the *remote* repository but does not exist in the *local* Pulp repository.

Example:
```
# determine what needs to be added to the repository
wanted = SizedIterator(<generator>, <length>)

# determine what needs to be delete from the repository
unwanted = SizedIterator(<generator>, <length>)

# apply
changeset = ChangeSet(importer, adds=wanted, removes=unwanted)
for report in changeset.apply():
    ...
```

Note to reviewer: Start here https://github.com/pulp/pulp/pull/3006/files#diff-d7ffc4b679440e952dc514304843c770.

Also, https://github.com/pulp/pulp_file/pull/2 is a proof-of-concept PR using the File plugin to demo the `ChangeSet` API.

Discussion
===========

The deferred (Lazy) catalog cannot contain everything needed to create a Download object.
The thing that's missing is logic. It may require a specialized `Download` class
with special error handling that knows to get an authentication token (or something).
Here are a few options:

1. Add a method the importer.  Like: `Importer.get_artifact_download()`.
   The base class implementation would be sufficient for most cases and would rarely need
   to be overridden by a plugin writer. I'm leaning this way.
2. Provide a mechanism whereby a plugin writer registers a `DownloadBuilder`.
3. Other?

Both options make it the responsibility of the plugin writer but #2 may keep the Importer API cleaner.

FAQ
===

* **Question:** Is layer1 just another Nectar?
  **Answer:** Sort of but no.  I really hope not.  It is far simpler and delegates all of the
  heavy lifting to *requests* and *concurrent.futures*.

* **Question:** Why have layer1 at all?  Wny not just use *requests*?
  **Answer:** Plugin writers are free to use whatever they want.  However, *requests* does not provide
  concurrent downloading.  Also, *requests* is only HTTP.  Since downloading needs to be part of the
  plugin API, we need a formal way for plugins to participate.  Supporting *The Streamer* is one example.

* **Question:** Is the ChangeSet another `Step` framework?
  **Answer:** No. Unlike the step framework, the ChangeSet provides support for a very specific
  part of the importer synchronization workflow.

* **Question:** Does the ChangeSet do progress reporting?
  **Answer:** Yes.

* **Question:** How is the memory footprint constrained?
  **Answer:** Every component is designed to *work-with* or be a generator. Even a component
  using a `Queue` will restrict the queue size.

* **Question:** For plugins that need to import one kind of content to determine additional content 
  that needs to be imported, how would this work?
  **Answer:** The plugin could use (or chain) multiple `ChangeSet`s.  The plugin can iterate the reports
  yielded by the ChangeSet.apply() to determine the content that needs to be added using a subsequent ChangeSet.

* **Question:** For docker, downloads need to get and share auth token.  How would this work?
  **Answer:** The `Download` has a dictionary (context) for sharing information.  The `Batch` ensures
  that all downloads share the same context.  The first docker request that detects that it needs
  a new token can obtain it and put it in the context to be shared with other downloads.  The docker 
  plugin will need to use a *subclass* of `HttpDownload` with a custom `on_error()`.  But should only
  require a few lines of code.  This is why the plugin is responsible for providing the download object.

* **Question:** How are fatal errors handled within any of this machinery?
  **Answer:** Regardless of which thread or component occurs, an exception is propagated and
  raised to the caller.

* **Question:** What if my plugin is downloading with something other than regular HTTP?
  **Answer:** The `Download` is a *callable* and can be implemented to support any protocol.
  The `Batch` will provided batched, concurrent downloading for any kind of `Download`.

* **Question** What if the plugin needs to download the file before creating the `Content`.
  For example: In RPM - all RPMs are stored using a SHA256 as part of the unit key even when the
  metadata only includes SHA1.  This would mean that the `ChangeSet` would need to download the file (rpm) and generate the SHA256 before creating the DB records.  Or, provide a way for the plugin to participate in this flow.  Right?

  **Answer** Well, this may be flawed to begin with.  What about deferred (Lazy) downloading.
  When the download policy is *deferred* the file will not be available to create the SHA256.  
  I think we need to  challenge the practice of normalizing the *checksum* to SHA256 to start with.

* **Question** How would mirror lists be handled?

  **Answer** A subclass of HttpDownload would be provided by the plugin.  It would implement resolving the list of mirrors and stash it in the *context* to be shared with other downloads involved in the same batch download.  The last mirror used could also be stored in the context to support round-robin.

This includes a _sanity_  test but eventually needs some unit tests and some combination of functional and perhaps smash tests.   The _sanity_ test will not be included in the final PR as-is.

TODO
=====
[x] The `ChangeSet` only has logging roughed in.
[x] Test and optimize using real model/DB.
[ ] *Requests* (mainly the Session) and urllib3 seem to have known thread safety issues.  Look into keep the Session in a thread local to avoid using concurrently.
[x] Add DB transaction management in the ChangeSet.

[1] https://github.com/pulp/pulp/pull/3006/files#diff-30405389285d884762c0188393e28743
[2] https://github.com/pulp/pulp/pull/3006/files#diff-d7ffc4b679440e952dc514304843c770

### DIFF
--- a/platform/pulpcore/download/single.py
+++ b/platform/pulpcore/download/single.py
@@ -1,0 +1,291 @@
+from gettext import gettext as _
+from logging import getLogger
+from threading import RLock
+
+
+from .delegation import delegate
+from .error import DownloadFailed
+from .validation import ValidationError
+
+
+log = getLogger(__name__)
+
+
+class Download:
+    """
+    An ABSTRACT download object.
+    Represents a unit of work to download a file. A download is a
+    command-pattern object.
+
+    Attributes:
+        url (str): A file download URL.
+        writer (Writer): An object used to store downloaded file.
+        validations (list): Collection of Validations to be applied.
+        context (Context): The download context.
+        retries (int): The total number of retries possible.
+        delegate (Delegate): A download delegate.
+        attachment: Arbitrary object attached to the download.
+        reply: The remote server reply.
+
+    Any method decorated with `@delegate` may be forwarded to the `delegate` for
+    implementation.  The delegate must define a public method with the same name
+    (without the `_` prefix).  For example, for `_on_error()`, the delegate would
+    define a method named `on_error(self, download, error)`.  Note: The `download`
+    is injected (passed) as the first argument.
+
+    Notes:
+        The validations are applied in order listed.
+        Settings may be shared between downloads.
+    """
+
+    # Data transfer buffer size in bytes.
+    BLOCK = 102400
+    # Number of retires on failed download.
+    RETRIES = 1
+
+    __slots__ = (
+        'url',
+        'writer',
+        'validations',
+        'context',
+        'retries',
+        'delegate',
+        'attachment',
+        'reply'
+    )
+
+    def __init__(self, url, writer):
+        """
+        Args:
+            url (str): A file download URL.
+            writer (Writer): An object used to store downloaded file.
+        """
+        super(Download, self).__init__()
+        self.url = url
+        self.writer = writer
+        self.validations = []
+        self.context = Context()
+        self.retries = Download.RETRIES
+        self.delegate = None
+        self.attachment = None
+        self.reply = None
+
+    def _prepare(self):
+        """
+        Prepare to be executed.
+        """
+        log.debug(_('Prepare: %(d)s'), {'d': self})
+        self._on_prepare()
+
+    def _send(self):
+        """
+        Send the download.
+        This is the *main* method responsible for implementing the actual
+        download by sending a protocol specific download.
+
+        Raises:
+            DownloadFailed: The download failed and could not be repaired.
+
+        Notes:
+            Must be implemented by subclass.
+        """
+        raise NotImplementedError()
+
+    def _repair(self, error):
+        """
+        The download has encountered an error.
+        Attempt to repair and retry using `_on_error()` depending on remaining
+        available retries.
+
+        Args:
+            error (DownloadFailed): The raised exception.
+
+        Raises:
+            DownloadFailed: The download failed and could not be repaired.
+        """
+        log.debug(_('Repair: %(d)s'), {'d': self})
+        try:
+            retries = self.retries
+            while retries:
+                retries -= 1
+                repaired = self._on_error(error)
+                if not repaired:
+                    break
+                try:
+                    self._attempt()
+                except DownloadFailed:
+                    continue
+                else:
+                    return
+            if not retries:
+                self._on_failed()
+                raise error
+        except Exception:
+            log.exception(_('Repair failed: {d}').format(d=self))
+            self._on_failed()
+            raise error
+
+    def _write(self, bfr):
+        """
+        Write downloaded content.
+
+        Args:
+            bfr (bytes): A buffer to append.
+        """
+        self.writer.append(bfr)
+        for validation in self.validations:
+            validation.update(bfr)
+
+    def _attempt(self):
+        """
+        Attempt to download.
+
+        Raises:
+            DownloadFailed: The download failed.
+            ValidationError: On validation failed.
+        """
+        log.debug(_('Attempt: %(d)s'), {'d': self})
+        with self.writer:
+            self._send()
+            self._on_reply()
+        self._on_succeeded()
+        self._validate()
+
+    def _validate(self):
+        """
+        Apply validations.
+
+        Raises:
+            ValidationError: On validation failed.
+        """
+        for validation in self.validations:
+            log.debug(
+                _('Apply validation: %(validator)s'),
+                {
+                    'validator': validation
+                })
+            try:
+                validation.apply()
+            except ValidationError as error:
+                log.info(_('Validation failed: %(error)s'), {'error': error})
+                raise error
+            except Exception as unexpected:
+                log.exception(unexpected)
+                raise unexpected
+
+    @delegate
+    def _on_prepare(self):
+        """
+        Prepared to be executed.
+        """
+        pass
+
+    @delegate
+    def _on_reply(self):
+        """
+        A reply has been received.
+        """
+        pass
+
+    @delegate
+    def _on_succeeded(self):
+        """
+        The download has succeeded.
+        """
+        pass
+
+    @delegate
+    def _on_failed(self):
+        """
+        The download has failed.
+        All efforts to repair and retry have failed.
+        """
+        pass
+
+    @delegate
+    def _on_error(self, error):
+        """
+        The download has encountered an error.
+        This provides the (optional) delegate an opportunity to repair
+        and try the download.
+
+        Args:
+            error (DownloadFailed): The raised exception.
+
+        Returns:
+            (bool): True if repaired and may be retried.
+        """
+        return False
+
+    def __call__(self):
+        """
+        Execute the download.
+
+        Raises:
+            DownloadFailed: The download failed and could not be repaired.
+            ValidationError: On validation failed.
+        """
+        try:
+            self._prepare()
+            self._attempt()
+        except DownloadFailed as error:
+            self._repair(error)
+
+    def __str__(self):
+        _id = str(id(self))[-4:]
+        description = _(
+            '{t}: id={id} url={u} writer={w}'
+            ' | repair: retries={r}')
+        return description.format(
+            t=type(self).__name__,
+            id=_id,
+            u=self.url,
+            w=self.writer,
+            r=self.retries)
+
+
+class Context:
+    """
+    A download context.
+    Each download has a reference to a context used to safely share resources
+    such as HTTP sessions and authentication tokens.
+
+    Attributes:
+        properties (dict): Arbitrary properties.
+        _mutex (RLock): The object mutex.
+
+    Examples:
+        >>>
+        >>> def get_token(self):
+        >>>     with self.context as context:
+        >>>         try:
+        >>>             return context.token
+        >>>         except KeyError:
+        >>>             token = self.generate_token()
+        >>>             context.token = token
+        >>>             return token
+        >>>
+    """
+
+    def __init__(self, **properties):
+        """
+        Args:
+            properties (dict): Initial properties.
+        """
+        self.__dict__.update(properties)
+        self.__dict__['MUTEX'] = RLock()
+
+    @property
+    def _mutex(self):
+        return self.__dict__['MUTEX']
+
+    def __setattr__(self, key, value):
+        with self._mutex:
+            super(Context, self).__setattr__(key, value)
+
+    def __enter__(self):
+        self._mutex.acquire()
+        return self
+
+    def __exit__(self, *unused):
+        self._mutex.release()
--- a/plugin/pulpcore/plugin/changeset/iterator.py
+++ b/plugin/pulpcore/plugin/changeset/iterator.py
@@ -1,0 +1,239 @@
+import itertools
+
+from collections.abc import Iterable
+from logging import getLogger
+
+from django.db.models import Q
+
+from pulpcore.download import Download
+
+from .model import RemoteArtifact
+
+
+log = getLogger(__name__)
+
+
+class BatchIterator(Iterable):
+    """
+    Iterate large numbers of items in batches.
+
+    Attributes:
+        iterable (Iterable): An iterable to batch.
+        batch (int): The size of each batch.
+
+    Examples:
+        >>>
+        >>> numbers = [1, 2, 3, 4, 5, 6]
+        >>> for batch in BatchIterator(numbers, 3):
+        >>>     repr(batch)
+        '(1, 2, 3)'
+        '(4, 5, 6)'
+    """
+
+    # Default batch size.
+    BATCH = 1000
+
+    def __init__(self, iterable, batch=BATCH):
+        """
+        Args:
+            iterable (Iterable): An iterable to batch.
+            batch (int): The size of each batch.
+        """
+        self.iterable = iterable
+        self.batch = batch
+
+    def __iter__(self):
+        generator = (c for c in self.iterable)
+        while True:
+            batch = tuple(itertools.islice(generator, 0, self.batch))
+            if batch:
+                yield batch
+            else:
+                return
+
+
+class ContentIterator(Iterable):
+    """
+    Iterate `RemoteContent` and replace associated content models
+    With models fetched from the DB (when found).
+    """
+
+    # Number of keys in each batched query.
+    BATCH = 1024
+
+    @staticmethod
+    def _batch_q(content):
+        """
+        Build a query for the specified batch of content.
+
+        Args:
+            content (tuple): A batch of content.  Each is: RemoteContent.
+
+        Returns:
+            Q: The built query.
+        """
+        q = Q()
+        for c in content:
+            q |= Q(**c.key)
+        return q
+
+    def __init__(self, content):
+        """
+        Args:
+            content (Iterable): An Iterable of RemoteContent.
+        """
+        self.content = content
+
+    def _collated_content(self):
+        """
+        Collate each batch of content into lists by model.
+
+        Yields:
+            dict: A dictionary of {model_class: [content,]}
+                Each content is: RemoteContent.
+        """
+        for batch in BatchIterator(self.content, self.BATCH):
+            collated = {}
+            for content in batch:
+                _list = collated.setdefault(type(content.model), list())
+                _list.append(content)
+            yield collated
+
+    def _fetch(self):
+        """
+        Fetch each batch of collated content.
+
+        Yields:
+            tuple: (content, fetched).
+                The content is a list of RemoteContent.
+                The fetched is a dictionary of fetched content models keyed by natural key.
+        """
+        for collated in self._collated_content():
+            for model, content in collated.items():
+                fields = {f.name for f in model.natural_key_fields}
+                q = self._batch_q(content)
+                q_set = model.objects.filter(q)
+                q_set = q_set.only(*fields)
+                fetched = {c.natural_key(): c for c in q_set}
+                yield (content, fetched)
+
+    def _iter(self):
+        """
+        The remote content is iterated and its model is replaced with a model
+        fetched from the DB (when found). This is batched to limit the memory
+        footprint.
+
+        Yields:
+            RemoteContent: The transformed content.
+        """
+        for batch, fetched in self._fetch():
+            for content in batch:
+                natural_key = content.model.natural_key()
+                try:
+                    model = fetched[natural_key]
+                except KeyError:
+                    pass
+                else:
+                    content.update(model)
+                yield content
+
+    def __iter__(self):
+        return iter(self._iter())
+
+
+class DownloadIterator(Iterable):
+    """
+    Iterate the content artifacts and yield the appropriate download object.
+    When downloading is deferred or the artifact is already downloaded,
+    A NopDownload is yielded.
+    """
+
+    def __init__(self, content, deferred=False):
+        """
+        Args:
+            content (Iterable): An Iterable of RemoteContent.
+            deferred (bool): When true, downloading is deferred.
+        """
+        self.content = content
+        self.deferred = deferred
+
+    def _iter(self):
+        """
+        Iterate the content artifacts and yield the appropriate download object.
+        When downloading is deferred or the artifact is already downloaded,
+        A NopDownload is yielded.
+
+        Yields:
+            Download: The appropriate download object.
+        """
+        for artifact in ArtifactIterator(self.content):
+            if self.deferred or artifact.model.downloaded:
+                download = NopDownload()
+                download.attachment = artifact
+                artifact.path = None
+            else:
+                download = artifact.download
+            yield download
+
+    def __iter__(self):
+        return iter(self._iter())
+
+
+class ArtifactIterator(Iterable):
+    """
+    Iterate the content and flatten the artifacts.
+    Ensure that at least (1) artifact is yielded for each content.
+    """
+
+    def __init__(self, content):
+        """
+        Args:
+            content (Iterable): An Iterable of RemoteContent.
+        """
+        self.content = content
+
+    def _iter(self):
+        """
+        Iterate the content and flatten the artifacts.
+        Ensure that at least (1) artifact is yielded for each content. When content
+        does not have any un-downloaded artifacts, A NopArtifact is yielded.
+
+        Yields:
+            RemoteArtifact: The flattened artifacts.
+        """
+        for content in ContentIterator(self.content):
+            if content.artifacts:
+                for artifact in content.artifacts:
+                    artifact.content = content
+                    yield artifact
+            else:
+                artifact = RemoteArtifact(NopArtifact(), NopDownload())
+                artifact.content = content
+                yield artifact
+
+    def __iter__(self):
+        return iter(self._iter())
+
+
+class NopDownload(Download):
+    """
+    A no-operation (NOP) download.
+    """
+
+    def __init__(self):
+        super(NopDownload, self).__init__('', None)
+
+    def _send(self):
+        pass
+
+    def __call__(self):
+        pass
+
+
+class NopArtifact:
+    """
+    No operation (NOP) artifact model.
+    """
+
+    def __init__(self):
+        self.downloaded = True
--- a/plugin/pulpcore/plugin/changeset/main.py
+++ b/plugin/pulpcore/plugin/changeset/main.py
@@ -1,0 +1,253 @@
+from collections.abc import Sized, Iterable
+from gettext import gettext as _
+from logging import getLogger
+
+from django.db.utils import IntegrityError
+from django.db import transaction
+
+from ..download import Batch
+from ..models import ProgressBar, DownloadCatalog, RepositoryContent
+from ..tasking import Task
+
+from .iterator import BatchIterator, DownloadIterator
+from .report import ChangeReport
+
+
+log = getLogger(__name__)
+
+
+class ChangeSet:
+    """
+    A set of changes to be applied to a repository.
+
+    When applied, the following changes are made.
+
+    For each content (unit) added:
+
+    - All content artifacts are downloaded as needed (unless deferred=True).
+    - The content (model) is saved.
+    - All artifacts (models) are saved.
+    - Deferred download catalog entries are created for each artifact.
+    - The content (unit) is added to the repository.
+
+    For each content (unit) removed:
+
+    - Content removed from the repository.
+    - Deferred download catalog deleted for each artifact.
+
+    Attributes:
+        importer (pulpcore.plugin.Importer): An importer.
+        additions (SizedIterable): The content to be added to the repository.
+        removals (SizedIterable): The content IDs to be removed.
+        deferred (bool): Downloading is deferred.  When true, downloads are not performed
+            but content is still created and added to the repository.
+
+    Examples:
+        >>>
+        >>> from pulpcore.plugin.changeset import ChangeSet, ChangeFailed
+        >>>
+        >>> changeset = ChangeSet(...)
+        >>> for report in changeset.apply():
+        >>>     try:
+        >>>         report.result()
+        >>>     except ChangeFailed:
+        >>>         # failed.  Decide what to do.
+        >>>     else:
+        >>>         # be happy
+        >>>
+    """
+
+    def __init__(self, importer, additions=(), removals=()):
+        """
+        Args:
+            importer (pulpcore.plugin.Importer): An importer.
+            additions (SizedIterable): The content to be added to the repository.
+            removals (SizedIterable): The content IDs to be removed.
+
+        Notes:
+            The content to be added may already exist but not be associated
+            to the repository. Existing content is fetched and used instead
+            of provided content models as needed.
+        """
+        self.importer = importer
+        self.additions = additions
+        self.removals = removals
+        self.deferred = False
+
+    @property
+    def repository(self):
+        """
+        The repository being updated.
+
+        Returns:
+            pulpcore.plugin.models.Repository: The repository being updated.
+        """
+        return self.importer.repository
+
+    def _add_content(self, content):
+        """
+        Add the specified content to the repository.
+        Download catalog entries are created foreach artifact.
+
+        Args:
+            content (RemoteContent): The content to be added.
+        """
+        try:
+            association = RepositoryContent(
+                repository=self.importer.repository,
+                content=content.model)
+            association.save()
+        except IntegrityError:
+            # Duplicate
+            pass
+        for artifact in content.artifacts:
+            log.info(
+                _('Cataloging artifact: %(a)s'),
+                {
+                    'a', artifact
+                })
+            try:
+                entry = DownloadCatalog()
+                entry.importer = self.importer
+                entry.url = artifact.download.url
+                entry.artifact = artifact.model
+                entry.save()
+            except IntegrityError:
+                entry = DownloadCatalog.objects.filter(
+                    importer=self.importer,
+                    artifact=artifact)
+                entry.url = artifact.download.url
+                entry.save()
+
+    def _remove_content(self, content):
+        """
+        Remove content from the repository.
+        Download catalog entries are deleted foreach artifact.
+
+        Args:
+            content (pulpcore.plugin.Content): A content model to be removed.
+        """
+        q_set = DownloadCatalog.objects.filter(
+            importer=self.importer,
+            artifact__in=content.artifacts.all())
+        q_set.delete()
+        q_set = RepositoryContent.objects.filter(
+            repository=self.repository,
+            content=content)
+        q_set.delete()
+
+    def _apply_additions(self):
+        """
+        Apply additions.
+        Content listed in `additions` is created (as needed) and added to the repository.
+
+        Yields:
+            ChangeReport: A report for each content added.
+        """
+        log.info(
+            _('Apply additions: repository=%(r)s.'),
+            {
+                'r': self.repository.name
+            })
+        with Batch(DownloadIterator(self.additions, self.deferred)) as batch:
+            with ProgressBar(message=_('Add Content'), total=len(self.additions)) as bar:
+                for plan in batch():
+                    artifact = plan.download.attachment
+                    content = artifact.content
+                    try:
+                        plan.result()
+                    except Exception as error:
+                        task = Task()
+                        task.append_non_fatal_error(error)
+                        bar.increment()
+                        report = ChangeReport(ChangeReport.ADDED, content.model)
+                        report.error = plan.error
+                        yield report
+                        continue
+                    artifact.settle()
+                    content.settle()
+                    if not content.settled:
+                        continue
+                    with transaction.atomic():
+                        self._add_content(content)
+                        bar.increment()
+                    report = ChangeReport(ChangeReport.ADDED, content.model)
+                    yield report
+
+    def _apply_removals(self):
+        """
+        Apply removals.
+        Content listed in `removals` is removed from the repository.
+
+        Yields:
+            ChangeReport: A report for each removed content.
+        """
+        log.info(
+            _('Apply removals: repository=%(r)s.'),
+            {
+                'r': self.repository.name
+            })
+        with ProgressBar(message=_('Remove Content'), total=len(self.removals)) as bar:
+            for batch in BatchIterator(self.removals, 100):
+                with transaction.atomic():
+                    for content in batch:
+                        self._remove_content(content)
+                        bar.done += len(batch)
+                        bar.save()
+                        report = ChangeReport(ChangeReport.REMOVED, content)
+                        yield report
+
+    def apply(self):
+        """
+        Apply the requested changes to the repository.
+
+        Yields:
+            ChangeReport: For each change applied.
+        """
+        log.info(
+            _('Apply ChangeSet: repository=%(r)s.'),
+            {
+                'r': self.repository.name
+            })
+        for report in self._apply_additions():
+            yield report
+        for report in self._apply_removals():
+            yield report
+
+
+class SizedIterable(Sized, Iterable):
+    """
+    A sized iterable.
+
+    Attributes:
+        iterable (Iterable): An iterable.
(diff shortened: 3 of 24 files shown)


platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+from queue import Queue, Empty
+from threading import Thread, current_thread
+
+from .single import Context
+
+
+# The default queuing backlog.
+BACKLOG = 1024
+# The default concurrency.
+CONCURRENT = 10
```
mhrivnak: I think our current default of 5 is a lot and plenty to maximize available resources. Katello even reduces it to 4 via an importer setting. Beyond about 3, the returns diminish rapidly, and resource contention ramps up.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+
+    def __str__(self):
+        _id = str(id(self))[-4:]
+        return _('DownloadFeeder: id={s} shutdown={a}').format(s=_id, a=self.is_shutdown)
+
+
+class QueueIterator:
+    """
+    A Queue iterator.
+    Each item in the queue is a tuple of: (code, payload).
```
mhrivnak: You could make this a bit more formal by making a namedtuple.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+            if download:
+                return download
+
+
+class Plan:
+    """
+    Batch download execution plan.
+    The plan provides:
+      - Ensure self is returned in the future.result.
+      - Catch and stored exceptions raised by the download. This ensure that
```
mhrivnak: s/and/any/ I assume

platform/pulp/download/http.py
```diff
@@ -0,0 +1,274 @@
...
+            download: The download to send.
+            unused: Unused.
+
+        Returns:
+            Response: Always.
+        """
+        response = Response()
+        response.url = download.url
+        try:
+            path = download.url[7:]
```
mhrivnak: Trying to get rid of the scheme? It's better to use ``urllib.parse.urlparse``

platform/pulp/download/http.py
```diff
@@ -0,0 +1,274 @@
...
+            response.raw = open(path, 'rb')
+            response.status_code = HTTPStatus.OK
```
mhrivnak: I have mixed feelings about using an HTTP response code when HTTP isn't being used.

platform/pulp/download/http.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        methods = {
+            'GET': self.get,
+            'HEAD': self.head,
+        }
+        try:
+            return methods[name.upper()]
+        except KeyError:
+            _list = '|'.join(sorted(methods.keys()))
+            msg = _('method must be: ({list})'.format(list=_list))
```
mhrivnak: It's a good idea to reference the parameter name, which happens to be "name", for clarity. So I would s/method/name/

platform/pulp/download/single.py
```diff
@@ -0,0 +1,322 @@
...
+            tr=self.read_timeout,
+            U=self.user,
+            P=self.password,
+            v=self.ssl_validation,
+            a=self.ssl_ca_certificate,
+            k=self.ssl_client_key,
+            c=self.ssl_ca_certificate)
+
+
+class Context:
```
mhrivnak: If you're looking for a thread-safe dictionary, subclassing ``UserDict`` might be cleaner. Something like this might fit your needs:

```
from collections import UserDict
import threading


class Context(UserDict):
    def __init__(self, *args, **kwargs):
        self._lock = threading.RLock()
        super(Context, self).__init__(*args, **kwargs)

    def __getitem__(self, *args, **kwargs):
        with self._lock:
            super(Context, self).__getitem__(*args, **kwargs)

    def __setitem__(self, *args, **kwargs):
        with self._lock:
            super(Context, self).__setitem__(*args, **kwargs)

    def __delitem__(self, *args, **kwargs):
        with self._lock:
            super(Context, self).__delitem__(*args, **kwargs)
```

platform/pulp/download/validation.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class SizeMismatch(MismatchError):
+    """
+    The file size does not match what is expected.
+    """
+
+    def __str__(self):
+        return _(
+            'File size mismatch: expected={e} actual={a}'.format(
```
mhrivnak: It's good to put units on numbers. Is this bytes?

platform/pulp/download/validation.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        Apply the validation.
+
+        Raises:
+            ValidationError: When validation has failed.
+        """
+        if not self.enforced:
+            return
+        if self.expected != self.actual:
+            raise DigestMismatch(expected=self.expected, actual=self.actual)
```
mhrivnak: It would be helpful to include the algorithm in the error message.

platform/pulp/download/writer.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @property
+    def is_open(self):
+        """
+        Get whether the writer is open.
+
+        Returns:
+            bool: True when open.
+        """
+        return self.fp is not None
```
mhrivnak: ``and self.fp.closed is False``

platform/pulp/download/writer.py
```diff
@@ -0,0 +1,144 @@
...
+    def _mkdir(self):
+        """
+        Create the directory as needed.
+
+        Raises:
+            OSError: When the directory cannot be created.
+        """
+        _dir = os.path.dirname(self.path)
+        if not _dir:
+            return
```
mhrivnak: What would this signify? Why would it evaluate to ``False``, and why is it ok to just ``return``?

platform/pulp/download/writer.py
```diff
@@ -0,0 +1,144 @@
...
+    Downloaded text writer.
+    Used to store downloaded text in memory.
+    """
+
+    def open(self):
+        self.fp = BytesIO()
+
+    def __str__(self):
+        self.fp.seek(0)
+        return self.fp.read().decode('utf8', 'replace')
```
mhrivnak: The python docs for this method describe it as being a "convenient or concise representation". I worry that this could be an unwieldy amount of text in many cases. If the text value is the best representation we have, what do you think about limiting it to something like 40 or 80 characters of text?

platform/pulp/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,215 @@
...
+        return query
+
+    def __init__(self, content):
+        """
+        Args:
+            content (generator): A generator of RemoteContent.
+        """
+        self.content = content
+
+    def _batch_content(self, size=BATCH):
```
mhrivnak: This looks familiar. ;)

https://github.com/pulp/pulp/blob/91a1e28c9e7d3dee418d5c7680dbf25c3e7adc63/server/pulp/plugins/util/misc.py#L14

It does tend to be generally useful, so having it in a "util" kind of module isn't a bad idea.

platform/pulp/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,215 @@
...
+        Fetch each batch of collated content.
+
+        Yields:
+            tuple: (content, fetched).
+                The content is a list of RemoteContent.
+                The fetched is a dictionary of fetched content models keyed by natural key.
+        """
+        for collated in self._collated_content():
+            for model, content in collated.items():
+                fields = (f.name for f in model.natural_key_fields)
```
mhrivnak: Using a generator comprehension here strikes me as a little funny. Not wrong, but if it we me I'd use a list comprehension since it'll probably get evaluated 4 lines below anyway in the call to ``q_set.only(``, and we don't gain anything here by delaying evaluation.

platform/pulp/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,215 @@
...
+        Iterate the content artifacts and yield the appropriate download object.
+        When downloading is deferred or the artifact is already downloaded,
+        A NopDownload is yielded.
+
+        Yields:
+            Download: The appropriate download object.
+        """
+        for artifact in ArtifactIterator(self.content):
+            if self.deferred or artifact.model.downloaded:
+                download = NopDownload()
```
mhrivnak: Might be nice if you could just ``yield NopDownload(attachment=artifact)``

platform/pulp/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,215 @@
...
+        Yields:
+            RemoteArtifact: The flattened artifacts.
+        """
+        for content in ContentIterator(self.content):
+            if content.artifacts:
+                for artifact in content.artifacts:
+                    yield artifact
+            else:
+                artifact = RemoteArtifact(NopArtifact(), NopDownload())
+                artifact.content = content
```
mhrivnak: Might be nice if this could be passed to ``__init__``. It seems a little awkward to create the instance with two arguments, and then immediately add a third thing a different way.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
...
+        try:
+            association = RepositoryContent(
+                repository=self.importer.repository,
+                content=content)
+            association.save()
+        except IntegrityError:
+            # Duplicate
+            pass
+
+    def _disassociate(self, content):
```
mhrivnak: I recall discussion of disassociate vs unassociate. I don't recall how that ended. Should this all be consistent at least?

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
...
+            content (pulp.plugin.Content): A content model to be unassociated.
+        """
+        log.debug(_('Un-associate content.'))
+        RepositoryContent.objects.delete(
+            repository=self.importer.repository,
+            content=content)
+        for artifact in content.artifacts:
+            DownloadCatalog.objects.delete(
+                importer=self.importer,
+                artifact=artifact)
```
mhrivnak: I wonder if it would be better to pass the list to the DB rather than do a python loop with individual DB calls.

``artifact__in=artifacts``

should do it. The only risk I guess would be if there are a CRAZY number of artifacts.

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+            try:
+                found = known[artifact.relative_path]
+                found.model = artifact
+                self._artifacts.add(found)
+            except KeyError:
+                log.error(_('Artifact not matched.'))
+
+    def settle(self):
+        """
+        Settle matters that are prerequisite to associating the content
```
mhrivnak: Reading the description of this method, I'd expect to find code that downloads artifacts. I'd also expect that when the method returns, the "matters" would be settled. Instead it seems like this just checks to see if matters are settled, and does nothing if they aren't. I suggest re-writing the description to make it more clear what this method should be doing and how it fits into the workflow.

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+        with a repository.  This mainly pertains to the artifacts being downloaded;
+        and associated artifact models saved; the content model saved.
+
+        Notes:
+            Called whenever an artifact has been downloaded.
+        """
+        for artifact in self._artifacts:
```
mhrivnak: Just as a fun alternative, this could be

```
if all((artifact.model.downloaded for artifact in self._artifacts)):
  self.save()
  self.settled=True
```

It may not be any better, but I do like the way it conveys that the conditional is a requirement for all things to be true, and the last two lines only happen if that condition is met.

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+            try:
+                super(RemoteContent, self).save()
+            except IntegrityError:
+                is_duplicate = True
+            else:
+                for artifact in self._artifacts:
+                    artifact.model.content = self.model
+                    artifact.save()
+        if is_duplicate:
+            model = type(self.model)
```
mhrivnak: Now I'm especially getting confused about the use of "model" to refer to an instance or a class.

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+            content = model.objects.get(**self.key)
+            self.update(content)
+
+
+class RemoteArtifact(Remote):
+    """
+    Represents an artifact related to content contained within the remote
+    repository is not contained within the local (pulp) repository.
```
mhrivnak: s/is/that is/

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+    Attributes:
+        download (pulp.download.Download): An object used to download the content.
+        content (RemoteContent): The associated remote content.
+    """
+
+    def __init__(self, model, download):
+        """
+
+        Args:
+            model (pulp.plugin.Artifact): An artifact model object.
```
mhrivnak: Perhaps call this an ``artifact`` instead of the generic ``model`` ?

platform/pulp/plugin/changeset/report.py
```diff
@@ -0,0 +1,69 @@
...
+        action (str): The action taken (ADD, DEL).
+        content (pulp.plugin.Content): The affected content model.
+        errors (list): A list of reported errors.
+    """
+
+    def __init__(self, action, content):
+        """
+        Args:
+            action (str): The action taken (ADD, DEL).
+            content (pulp.plugin.Content): The affected content model.
```
mhrivnak: Here also I would avoid the use of the word "model".

platform/pulp/plugin/changeset/report.py
```diff
@@ -0,0 +1,69 @@
+class Action:
+    ADDED = 'ADDED'
+    DELETED = 'DELETED'
+
+
+class Disposition:
```
mhrivnak: Since this never gets used as an object, I think it's here just as a namespace. Is there another way to communicate the grouping without using a class perhaps?

platform/pulp/plugin/changeset/report.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Report of added content.
+    """
+
+    def __init__(self, content):
+        """
+        Args:
+            content (pulp.plugin.Content): The added content model.
+        """
+        super(AddReport, self).__init__(Action.ADDED, content)
```
mhrivnak: I wonder if you need a whole class just for this. You end up being able to identify that a ChangeReport either added or deleted in two different ways. You can check its ``action`` attribute, and you can check if it is an instance of ``AddReport`` or ``DeleteReport``. I don't think you need both of these, so I would lean toward ditching the extra classes and just use the attribute.

platform/pulp/download/writer.py
jortel: dirname() return an empty string when the path does not have a directory component.  Just a file name.

platform/pulp/download/writer.py
jortel: Fair point.  The intended usage is to support conversion to a ``str`` by doing:
```
s = str(writer)
```
more than a to _"to compute the “informal” or nicely printable string representation of an object"_ as defined in the python3 docs.  I originally named it `to_str()` but this seems even odder because is raised the question of, why not impl as ``__str__()`` which is the standard conversion to a ``str``. 

The main goal here is to provide user a convenient way to get the data captured by the TextWriter to  ``str`` and use it.  If you still feel strongly, I'm open to renaming the method.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
+from collections import Sized, Iterable
```
jortel: I'll update the import to be on safe side.  However, It would be better if the author kept them permanently (and not just for backwards compat) in the ``collections`` package interface.  I really don't understand why the python community makes such poor choices.

https://github.com/python/cpython/blob/3.6/Lib/collections/__init__.py#L22

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+        log.debug(_('Model: %(m)s - saved'), {'m': self})
+
+
+class RemoteContent(Remote):
+    """
+    Represents content contained within the remote repository but is not
+    contained within the local (pulp) repository.
+
+    Attributes:
+        model (pulp.plugin.Content): A real content model object.
```
jortel: No, it is the `Content` model instance.  I'm new to django but feel sure I've seen (perhaps incorrectly) both the model class and instance referred to as "model".  And, I could be mistaken.  I mainly wanted to make it clear that the attribute is a content model object associated with the remote content object.  I think this supports more readable code everywhere else `content.model` is used.  The alternative of naming it something like `content` would lead to wonky and confusing semantics like `content.content` or `remote_content.content` when used.

In most places in the pulp code base, I strongly support naming the model instances for what they are instead of just `model` has you suggest.

platform/pulp/app/settings.py
```diff
@@ -110,7 +110,7 @@
     'DEFAULT_FILTER_BACKENDS': ('rest_framework.filters.DjangoFilterBackend',),
     'DEFAULT_PAGINATION_CLASS': 'pulp.app.pagination.UUIDPagination',
     'PAGE_SIZE': 100,
-    'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',)
+    #'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',)
```
jortel: Accidentally included in a commit.  Will omit when I squash.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
...
+            ChangeReport: For each change applied.
+        """
+        log.info(_('Apply'))
+        for report in self._add_content():
+            yield report
+        for report in self._remove_content():
+            yield report
+
+
+class SizedIterable(Sized, Iterable):
```
goosemania: Even though this class is not an important part of the PR, I have some concerns about it, I am not sure if you agree or not :)

If I understand correctly, this class is needed so we know the number of items for the unsized iterables like generators  because in case of sequence types which are sized `__len__` is defined. 
I think it is hard to support properly `len` for unsized by design iterable.
We are claiming that we support protocol both for `Sized` and for `Iterable` aka guarantee that number of iterations will be equal to what `__len__` returns. At the same time `length` is defined once in `__init__` and `__len__` returns always this number but `self.iterable` can be changed independently or just incorrectly initialized.
I see that `SizedIterable` is currently used for progress only and in the worst case the counters will be wrong but this class is also meant to be used by plugin writers. There could be various bugs which is hard to find in case iterable and length are different even just by mistake.

If you think this class is really needed, I suggest at least to improve it to avoid honest/simple mistakes.
Maybe make it immutable to some extent (or  less accessible like `self.__iterable`) and keep track of actual number of iterations and throw some error in case of discrepancy with `self.length`?

What do you think?

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,162 @@
...
+            except IntegrityError:
+                is_duplicate = True
+            else:
+                for artifact in self.artifacts:
+                    artifact.model.content = self.model
+                    artifact.save()
+        if is_duplicate:
+            model = type(self.model)
+            content = model.objects.get(**self.key)
+            self.update(content)
```
jortel: Yes, the query needed to be outside of the transaction or an exception was raised by the django ORM.

platform/pulp/plugin/changeset/main.py
jortel: I understand your concern.  There will always be the potential for inconsistencies between the number of items yielded by an iterable and the length provided by the plugin writer.  I modeled as a `Sized`, `Iterable`  object explicitly because the combination of the interfaces places a high level of responsibility on the user (the plugin writer) to supply good data.  It formalizes the responsibility more than passing multiple related parameters to the ChangeSet.

I think that progress reporting is a best effort thing.  Nothing should ever be using progress reporting for logic or correctness.  It's information provided to the user so they can know that progress is being made and potentially make decisions like _"Will this finish before I go to lunch"_.  Given this perspective, I think logging an error may be useful but raising an error would be unproductive and not always possible.

I'm fine with making the `length` and and `iterable` protected/private to make the object immutable.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        for artifact in content.artifacts:
+            log.info(_('Cataloging: {a}').format(a=self))
+            try:
+                entry = DownloadCatalog()
+                entry.importer = self.importer
+                entry.url = artifact.download.url
+                entry.artifact = artifact.model
+                entry.save()
+            except IntegrityError:
```
jortel: I expect mirror lists will be handled by a custom `Download` object.  The `DownloadCatalog` needs to have the uniqueness constraint added.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,200 @@
...
+                        report.errors.append(plan.error)
+                        yield report
+                        continue
+                    content.settle()
+                    if not content.settled:
+                        continue
+                    bar.increment()
+                    with transaction.atomic():
+                        self._update_catalog(content)
+                        self._associate(content.model)
```
jortel: Still working on getting the DB transactions correct.

The goals for the current transaction alignment is:

1. Create a content (unit) and all its artifacts in 1 commit.
2. Add content to a repository and to the download catalog in 1 commit.

We _may_ be able to do all #2s in a single transaction.  This would also have performance advantages.  But I would need to play more with nested transactions to see how this could work with django ORM.   And, need to consider limitations of DB rollback buffer size. 

I think versioned repos will make this easier assuming the version created by a failed sync could be discarded.

platform/pulp/plugin/changeset/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    >>>
+    >>>         # ADD
+    >>>         total = # <add count>
+    >>>         adds = ChangeSet(self, adds=SizedIterable(self.find_adds(metadata), total))
+    >>>         for report in adds.apply():
+    >>>             # inspect the reports.
+    >>>             pass
+    >>>
+    >>>         # DELETE
+    >>>         deletes = ChangeSet(self, deletes=self._find_deletes(metadata)))
```
jortel: No, just an oversight.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+        Get the next future and propagate any raised exceptions.
+
+        Returns:
+            The next `Future.result()`
+
+        Raises:
+            Anything raised by the object executed.
+        """
+        future = super(FutureIterator, self).__next__()
+        exception = future.exception()
```
jortel: Good suggestion.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+        """
+        log.debug(_('%(iterator)s - input drained'), {'iterator': self})
+        while True:
+            try:
+                self.queue.get(block=False)
+            except Empty:
+                break
+        self.end()
+
+    def empty(self):
```
jortel: Not intentional.  will likely consolidate as `drain()`.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+        self.concurrent = concurrent
+        self.iterator = PlanIterator(backlog)
+        self.executor = BatchExecutor(concurrent=concurrent, backlog=backlog)
+        self.context = Context(**context)
+        self.feeder = DownloadFeeder(self)
+        self.is_shutdown = False
+        if concurrent < 2:
+            raise ValueError(_('concurrent may not be < 2'))
+        if backlog < concurrent:
+            raise ValueError(_('backlog may not be < concurrent'))
```
jortel: Agreed.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,535 @@
...
+        Notes:
+            This must be called to prevent leaking resources unless the Batch
+            is used as a context manager.
+            >>>
+            >>> with Batch(..) as batch:
+            >>>    # ...
+        """
+        if self.is_shutdown:
+            return
+        self.is_shutdown = True
```
jortel: The `is_shutdown` attributes indicate that `shutdown()` has been called and that _shutdown_ is, at least, in progress.  Mainly used internally within the package.  I can update the docstrings on those attributes or consider renaming them (though, not sure to what).  If the feeder or the executor fails to shutdown an exception will be raised and there really isn't a remedial action.

platform/pulp/plugin/changeset/model.py
```diff
@@ -0,0 +1,182 @@
...
+            model (Model): A remote (wanted) model object.
+        """
+        self.model = model
+
+    def save(self):
+        """
+        Save the model.
+        """
+        self.model.save()
+        log.debug(_('Model: %(m)s - saved'), {'m': self})
```
jortel: Fair enough.

platform/pulp/plugin/changeset/model.py
jortel: Description updated.

platform/pulp/plugin/changeset/iterator.py
jortel: Actually meant to use a set comprehension.  Updated.

platform/pulp/download/single.py
jortel: Actually, it's more than just a thread safe dictionary.  It's a dictionary that is a context manager to manage it's lock.  The goal is to support a lookup and update of the context (dictionary) while locked.  Like in the example.

platform/pulp/download/http.py
jortel: This resolving the HTTP method (example: GET) passed into the constructor as `method` to an implementation.  I think the `ValueError` raised will be clearer this way.
Example:
```
d = HttpDownload(url, method='BAD')
```
would raise a `ValueError` something like:
```
ValueError: "method must be (GET, HEAD)"
```

platform/pulp/download/http.py
jortel: Understood.

platform/pulp/plugin/changeset/main.py
mhrivnak: Yes, I'm thinking that a repo version would get some kind of boolean or timestamp to signify when it is "complete", "finalized", or something like that. Any task that wants to create a new version would first look for an existing incomplete version and remove it. Queries for usable versions, such as through the REST API, would only match those that are marked as complete. This approach should make it unnecessary to create associations in a big transaction.

platform/pulp/plugin/changeset/main.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+    def __init__(self, iterable, length):
+        """
+        Args:
+            iterable (Iterable): An iterable.
+            length (int): The number of items expected to be yielded by the iterable.
+        """
+        self.iterable = iterable
+        self.length = length
```
goosemania: +1 to at least make attributes private/protected so it won't be easy to overwrite them.

goosemania: +1 to make attributes protected/private so it won't be easy to overwrite them.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,527 @@
...
+                    {
+                        'feeder': self,
+                        'total': self.total,
+                        'url': download.url
+                    })
+                download.context = self.context
+                future = self.executor.submit(Plan(self.batch, download))
+                future.add_done_callback(self.iterator.add)
+                self.total += 1
+        except Exception as e:
```
jortel: The feeder is iterating (in most cases) a generator supplied by the plugin writer and could very easily raise exceptions.

platform/pulp/download/batch.py
```diff
@@ -0,0 +1,527 @@
...
+            log.exception(_('feeder failed.'))
+            self.iterator.raised(e)
```
jortel: Your understanding is correct.  This logic is designed to cause an exception to be raised only when a component of the _framework_ itself fails (in this example, the feeder).  The result is - the caller, who is iterating the result of the batch _execution_, will be interrupted by a raised exception ASAP.

platform/pulp/plugin/changeset/main.py
jortel: 👍

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,238 @@
...
+        >>>
+        >>> numbers = [1, 2, 3, 4, 5, 6]
+        >>> for batch in BatchIterator(numbers, 3):
+        >>>     repr(batch)
+        '(1, 2, 3)'
+        '(4, 5, 6)'
+    """
+
+    # Default batch size.
+    BATCH = 1000
```
asmacdo: This isn't really a constant, subclasses set batch size by overriding this value. How would you feel about `self.batch_size`?

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,238 @@
...
+    @staticmethod
+    def _batch_q(content):
+        """
+        Build a query for the specified batch of content.
+
+        Args:
+            content (tuple): A batch of content.  Each is: RemoteContent.
+
+        Returns:
+            Q: The built query.
```
asmacdo: s/Q/django.db.models.Q/

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            content (Iterable): An Iterable of RemoteContent.
+            deferred (bool): When true, downloading is deferred.
+        """
+        self.content = content
+        self.deferred = deferred
+
+    def _iter(self):
+        """
+        Iterate the content artifacts and yield the appropriate download object.
```
asmacdo: These are identical to class docstring

plugin/pulpcore/plugin/changeset/main.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        return self.importer.repository
+
+    def _add_content(self, content):
+        """
+        Add the specified content to the repository.
+        Download catalog entries are created foreach artifact.
+
+        Args:
+            content (RemoteContent): The content to be associated.
```
asmacdo: s/associated/added to the repository/

plugin/pulpcore/plugin/changeset/main.py
```diff
@@ -0,0 +1,253 @@
...
+                entry.url = artifact.download.url
+                entry.save()
+
+    def _remove_content(self, content):
+        """
+        Remove content from the repository.
+        Download catalog entries are deleted foreach artifact.
+
+        Args:
+            content (pulpcore.plugin.Content): A content model to be unassociated.
```
asmacdo: s/unassociated/removed/

plugin/pulpcore/plugin/changeset/main.py
```diff
@@ -0,0 +1,253 @@
...
+        q_set.delete()
+        q_set = RepositoryContent.objects.filter(
+            repository=self.repository,
+            content=content)
+        q_set.delete()
+
+    def _apply_additions(self):
+        """
+        Apply additions.
+        Content listed in `additions` created (as needed) and added to the repository.
```
asmacdo: s/created/is created/

plugin/pulpcore/plugin/changeset/main.py
```diff
@@ -0,0 +1,253 @@
...
+                        task.append_non_fatal_error(error)
+                        bar.increment()
+                        report = ChangeReport(ChangeReport.ADDED, content.model)
+                        report.error = plan.error
+                        yield report
+                        continue
+                    content.settle()
+                    if not content.settled:
+                        continue
+                    with transaction.atomic():
```
asmacdo: :atom: Cool

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -0,0 +1,190 @@
...
+        self.model = model
+        known = {a.model.relative_path: a for a in self.artifacts}
+        self.artifacts.clear()
+        for artifact in model.artifacts.all():
+            try:
+                found = known[artifact.relative_path]
+                found.model = artifact
+                self.artifacts.add(found)
+            except KeyError:
+                log.error(_('Artifact not matched.'))
```
asmacdo: This would be a tricky log message to understand

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -0,0 +1,190 @@
...
+    def save(self):
+        """
+        Save the artifact in the DB
+        Update the associated file only when actually downloaded.
+        """
+        if self.model.downloaded:
+            try:
+                self.model.file = File(open(self.download.writer.path, mode='rb'))
+            except FileNotFoundError:
+                # TODO: Revisit this.
```
asmacdo: TODO left on purpose?

plugin/pulpcore/plugin/download/__init__.py
```diff
@@ -0,0 +1,59 @@
+"""
+Provides classes to support the Importer downloading.  Primarily for downloading
+`metadata` but `may` also be used to download content when not using the `ChangeSet`.
```
asmacdo: `` is usually used only for objects, not for emphasis. *may*?

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -0,0 +1,156 @@
...
+from pulpcore.download import (DigestValidation, HttpDownload, FileDownload, FileWriter,
+                               SizeValidation)
+
+
+class Factory:
+    """
+    Factory for creating download object based on URL scheme.
+
+    Attributes:
+        importer (pulpcore.plugin.models.Importer): An importer.
```
asmacdo: pulpcore.plugin.Importer

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -0,0 +1,156 @@
...
+        >>> artifact = ..
+        >>> download = Factory(importer).build(url, artifact=artifact)
+        >>>
+    """
+
+    def __init__(self, importer):
+        """
+
+        Args:
+            importer (pulpcore.plugin.models.Importer): An importer.
```
asmacdo: same

platform/pulpcore/download/single.py
```diff
@@ -0,0 +1,291 @@
...
+        self.retries = Download.RETRIES
+        self.delegate = None
+        self.attachment = None
+        self.reply = None
+
+    def _prepare(self):
+        """
+        Prepare to be executed.
+        """
+        log.debug(_('Prepare: %(d)s'), {'d': self})
```
jortel: Agreed.  Unfortunately, logging does not support new-style string formatting.  The string should only be formatted when the log level is sufficient for the message to be logged.  In other words, we don't want to format the string when the log level is not at DEBUG.

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -0,0 +1,156 @@
...
+            pulpcore.download.Download: A download object configured using the
+                attributes of the importer.
+        """
+        if (not path) and (not artifact):
+            raise ValueError(_('Either "path" or "artifact" is required.'))
+
+        try:
+            builder = Factory.BUILDER[urlparse(url).scheme.lower()]
+        except KeyError:
+            raise ValueError(_('URL: {u} not supported.'.format(u=url)))
```
asmacdo: Scheme for url: {u} not supported.

PR #2733, opened 2016-09-05
Add units which did not pass signature filter to the task report.

ipanova:
closes #2212
https://pulp.plan.io/issues/2212

PR #3007, opened 2017-05-09
Updating our test certs to conform to v1 entitlement cert format

daviddavis:
Also, this fixes our tests on F25 which has an updated version of python-rhsm.

fixes #2733
https://pulp.plan.io/issues/2733


bmbouter:
This looks right. I'm only seeing diff in the certs and I confirmed that the unit tests are passing on F24, F25, and EL7.

bmbouter:
I *think* all the test runners are done, do you know why it says 1 pending of 2 successful checks?

daviddavis:
@bmbouter is master the right branch or should this be against 2.13-dev?

bmbouter:
Ohhhh I didn't look at the branch. 2.13-dev is right and then merge forward. At least for today :)

PR #3008, opened 2017-05-09
Cache sessions in the streamer.

jortel:
https://pulp.plan.io/issues/2677

Note: Used napoleon style in new module expected to be used in pulp3.


dkliban:
ok test

dkliban:
ok test

streamer/pulp/streamer/server.py
```diff
@@ -378,3 +376,60 @@ def write(self, data):
...
+        """
+        Get a cache key for the downloader.
+
+        :param url: The download URL.
+        :type url: basestring
+        :param downloader: A file downloader.
+        :type downloader: nectar.downloaders.base.Downloader
+        :return: A cache key based on downloader configuration.
+        """
+        url = urlparse(url)
```
mhrivnak: Rather than re-useing the name and assigning it an object of a different type, consider using a new name such as ``url_parts``, ``parsed_url`` or similar.

streamer/pulp/streamer/server.py
```diff
@@ -378,3 +376,60 @@ def write(self, data):
...
+            downloader.config.basic_auth_password,
+            downloader.config.proxy_url,
+            downloader.config.proxy_port,
+            downloader.config.proxy_username,
+            downloader.config.proxy_password,
+            downloader.config.ssl_client_cert
+        )
+        return hash(properties)
+
+    def get(self, url, downloader):
```
mhrivnak: What do you think about renaming this to something like ``get_or_create()``? That would make it obvious that the behavior is different from the super class ``get()`` method, and also makes me feel better about the function signatures being different (both in parameters and "raises"). :smile:

I also think that intuitively, anyone working with a cache object may reasonably assume that a ``get()`` method will sometimes miss, and expect they need to handle that. I like the idea of using the name to communicate that this method never misses!

streamer/pulp/streamer/server.py
```diff
@@ -232,6 +229,7 @@ def _download(self, request, entry, responder):
         finally:
             try:
                 downloader.config.finalize()
+                self.session_cache.put()
```
mhrivnak: I don't think this method exists, and that's not coming to your attention because of the lovely 2 lines below.

streamer/pulp/streamer/server.py
```diff
@@ -378,3 +376,60 @@ def write(self, data):
...
+        :rtype: Session
+        """
+        key = self.key(url, downloader)
+        try:
+            session = super(SessionCache, self).get(key)
+        except NotCached:
+            adapter = HTTPAdapter()
+            session = Session()
+            session.stream = True
+            session.mount('http://', adapter)
```
mhrivnak: Is this required if we're using the default adapter? I thought this would only be required if we were providing a customized adapter.

streamer/pulp/streamer/server.py
```diff
@@ -378,3 +376,60 @@ def write(self, data):
...
+    Extends generic cache to derive the key using the URL and
+    downloader configuration.
+    """
+
+    @staticmethod
+    def key(url, downloader):
+        """
+        Get a cache key for the downloader.
+
+        :param url: The download URL.
```
jortel: no, I'd rather not mix styles in the same module.

streamer/pulp/streamer/cache.py
```diff
@@ -0,0 +1,167 @@
...
+    def evict(self, threshold):
+        """
+        Evict all unused cached objects.
+
+        Args:
+            threshold (timedelta): Evict all cached objects that have not been
+                requested since the specified timedelta.
+
+        Returns:
+            list: The evicted objects.
```
jortel: The intention here was to support the orderly destruction of evicted objects.  For example, if the evicted session needed to be closed, the subclass in the streamer could override (call super as it should) and call Session.close() on each evicted session.  The missing test is an oversight.

streamer/pulp/streamer/cache.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        evicted = []
+        now = Item.now()
+        with self._lock:
+            for key, item in self._inventory.items():
+                duration = (now - item.last_requested)
```
jortel: Yeah, I know but it seems more readable to me :)

streamer/pulp/streamer/cache.py
mhrivnak: Fair, but are we planning to actually do any post-eviction work? For Sessions we've always been fine letting them get garbage collected as-is.

streamer/pulp/streamer/cache.py
```diff
@@ -0,0 +1,167 @@
...
+            item.touch()
+            self.evict(self.eviction_threshold)
+            return item.object
+
+    def evict(self, threshold):
+        """
+        Evict all unused cached objects.
+
+        Args:
+            threshold (timedelta): Evict all cached objects that have not been
```
jortel: Seemed useful to support cache user to be able to evict w/ specific threshold but seems pointless now that you call it out.

streamer/pulp/streamer/cache.py
```diff
@@ -0,0 +1,167 @@
...
+        evicted = []
+        now = Item.now()
+        with self._lock:
+            for key, item in self._inventory.items():
+                duration = (now - item.last_requested)
+                if item.busy or (duration < threshold):
+                    continue
+                self.purge(key)
+                evicted.append(item.object)
+        return evicted
```
jortel: Good idea.

streamer/pulp/streamer/cache.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Returns:
+            datetime: The current timestamp.
+        """
+        return datetime.utcnow()
+
+    def __init__(self, object):
+        """
+        Args:
+            object (object): The actual cached object.
```
jortel: Instead of introducing an odd semantic, I think I'll just change to `object_` as recommended by PEP8 for name collisions w/ builtin _globals_.

https://www.python.org/dev/peps/pep-0008/#id36

streamer/pulp/streamer/server.py
jortel: Yep.  meant to be `parsed_url`.

streamer/pulp/streamer/server.py
jortel: probably better.

streamer/pulp/streamer/server.py
jortel: Oh, good catch.  Leftover from when items needed to be released back to the cache.

streamer/pulp/streamer/server.py
mhrivnak: :+1:

streamer/pulp/streamer/server.py
jortel: Logging the finalize() exception only when DEBUG too.

streamer/pulp/streamer/server.py
jortel: nope.  seems not.

jortel:
@mhrivnak, thanks for the review.  Ready for re-review.

streamer/pulp/streamer/server.py
```diff
@@ -378,3 +374,57 @@ def write(self, data):
...
+        :type url: str
+        :param downloader: A nectar downloader.
+        :type downloader: nectar.downloaders.core.Downloader
+        :return: A cached session.
+        :rtype: Session
+        """
+        key = self.key(url, downloader)
+        try:
+            session = super(SessionCache, self).get(key)
```
mhrivnak: This can become just ``self.get()`` since you're not overriding it anymore.

streamer/test/unit/streamer/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_get(self, now):
+        t1 = Mock()
+        now.side_effect = [1, 2, 3]
+        key = 't1'
+        cache = Cache()
+        cache.add(key, t1)
+        gotten = cache.get(key)
+        self.assertEqual(cache._inventory[key].last_requested, 2)
+        self.assertEqual(gotten, t1)
+        self.assertRaises(NotCached, cache.get, 'xx')
```
mhrivnak: This is a good candidate for being a separate test.

streamer/test/unit/streamer/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @patch(MODULE + '.Item.now')
+    def test_evict(self, now):
+        now.side_effect = [1, 2, 3, 4]
+        key = 't1'
+        cache = Cache()
+        cache.add(key, Mock(key=key))
+        cache.eviction_threshold = 3
```
mhrivnak: Very minor suggestion here, but you could pass this when instantiating the instance 2 lines above.

streamer/test/unit/streamer/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+        self.assertEqual([obj.key for obj in evicted], [key])
+
+    @patch(MODULE + '.Item.now')
+    def test_evict_busy(self, now):
+        now.side_effect = [1, 2, 3, 4]
+        key = 't1'
+        t1 = Mock()  # hold ref to make it busy.
+        cache = Cache()
+        cache.add(key, t1)
+        cache.eviction_threshold = 0
```
mhrivnak: Very minor suggestion here, but you could pass this when instantiating the instance 2 lines above.

streamer/test/unit/streamer/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+        cache.evict()
+        self.assertTrue('t1' in cache)
```
mhrivnak: If this is an intended use case, it's worth implementing ``__contains__``:

https://docs.python.org/2/reference/datamodel.html#object.__contains__

As-is, this line would cause an iteration through all of the keys in the cache until ``t1`` is found. Not a big deal with small collections, but can make a huge difference as size increases.

PR #3016, opened 2017-05-11
The importer serializer's download_policy field should be a ChoiceField.

mhrivnak:
It was previously a MultipleChoiceField, which caused problems. It should only
ever have one value.

### DIFF
--- a/platform/pulp/app/serializers/repository.py
+++ b/platform/pulp/app/serializers/repository.py
@@ -112,9 +112,9 @@ class ImporterSerializer(MasterModelSerializer):
         write_only=True,
         required=False,
     )
-    download_policy = serializers.MultipleChoiceField(
+    download_policy = serializers.ChoiceField(
         help_text='The policy for downloading content.',
-        allow_empty=False,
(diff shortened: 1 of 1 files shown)


platform/pulp/app/serializers/repository.py
```diff
@@ -112,9 +112,9 @@ class ImporterSerializer(MasterModelSerializer):
         write_only=True,
         required=False,
     )
-    download_policy = serializers.MultipleChoiceField(
+    download_policy = serializers.ChoiceField(
```
asmacdo: I must have git stashed this fix out of existence. This is the problem that caused full_clean to raise an exception early in the validation discussion.

platform/pulp/app/serializers/repository.py
```diff
@@ -112,9 +112,9 @@ class ImporterSerializer(MasterModelSerializer):
...
         help_text='The policy for downloading content.',
-        allow_empty=False,
+        allow_blank=False,
```
asmacdo: nice catch

PR #3018, opened 2017-05-11
Add headers as an optional alternate content source descriptor

werwty:
closes #1282
https://pulp.plan.io/issues/1282

### DIFF
--- a/common/pulp/common/config.py
+++ b/common/pulp/common/config.py
@@ -60,9 +60,11 @@ OPTIONAL = 0
 ANY = None
 NUMBER = '^\d+$'
 BOOL = '(^YES$|^TRUE$|^1$|^NO$|^FALSE$|^0$)', re.I
+HEADER = '^\S*=\S*$'
 
 # Regular expression to test if a value is a valid boolean type
 BOOL_RE = re.compile(*BOOL)
+HEADER_RE = re.compile(HEADER)
 
 
 class ValidationException(Exception):
@@ -140,6 +142,27 @@ def parse_bool(value):
         raise Unparsable()
 
     return value.upper() in ('YES', 'TRUE', '1')
+
+def parse_header(headers):
+    """
+       Parses the given header into its dictionary representation.
+       @param headers: headers to parse
+       @type value: str
+       @return: dictionary of headers
+       @rtype: dict
+       @raise Unparsable: if the value is not one of the accepted values for
+              indicating a header field
+       """
+    headers_dict = {}
+    for header in re.split(r'\s+', headers):
+        header = header.strip()
+        if header == '\\':
+            continue
+        if not HEADER_RE.match(header):
+            raise Unparsable()
+        field, value = header.split("=")
+        headers_dict.update({field: value})
+    return headers_dict
 
 
 class Config(dict):
--- a/docs/user-guide/content-sources.rst
+++ b/docs/user-guide/content-sources.rst
@@ -15,7 +15,7 @@ finally downloaded from the original (primary) source.
 Defining A Content Source
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Content sources are defined in ``/etc/pulp/content/sources/conf.d``. Each file with a .conf suffix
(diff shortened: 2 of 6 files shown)


werwty:
This nectar PR needs to be merged with this PR: https://github.com/pulp/nectar/pull/58

common/pulp/common/config.py
```diff
@@ -141,6 +143,28 @@ def parse_bool(value):
...
+       @rtype: dict
+       @raise Unparsable: if the value is not one of the accepted values for
+              indicating a header field
+       """
+    headers_dict = {}
+    for header in re.split(r'\s+', headers):
+        header = header.strip()
+        if header == '\\':
+            continue
+        else:
```
jortel: Don't need the `else` here.

docs/user-guide/content-sources.rst
```diff
@@ -98,6 +103,7 @@ The pulp-admin client can be use to list all defined content sources as follows:
   Base URL:       http://content-world/content/
   Enabled:        1
   Expires:        3d
+  Headers:        X-RHUI-ID=1234 \ X-Csrf-Token=31415926535897
```
jortel: Suggest you remove the '\'  here.  Also, I think we want to encourage each header on a separate line for clarity and consistency with previous example.

docs/user-guide/content-sources.rst
```diff
@@ -98,6 +103,7 @@ The pulp-admin client can be use to list all defined content sources as follows:
...
   Max Concurrent: 2
   Name:           Content World
   Paths:          f18/x86_64/os/ f18/i386/os/ f19/x86_64/os f19/i386/os
```
jortel: Actually, while you're at it (and I know you did not edit this part), I would appreciate the paths be on separate lines as well.

docs/user-guide/release-notes/2.14.x.rst
```diff
@@ -0,0 +1,30 @@
...
+* An additional `header` property has been added to the alternate content source descriptor.
+  This allows users to specify additional HTTP headers when using an alternate content source.
+
+
+Bug Fixes
+---------
+
+See the list of :fixedbugs_pulp:`2.14.0`
+
+Upgrade instructions
```
jortel: Are these upgrade instructions needed?  Different than usual in some way for just this version?  If not, suggest it be removed.

werwty: The "normal process" upgrade instructions have been in every release note 2.11+
Personally I think it has its merits since when the new version comes out the user does not have to dig for upgrade instructions.

docs/user-guide/content-sources.rst
werwty: This is the way pulp-admin formats the content source when `pulp-admin content sources list` is called. 
Let me know if you want me to update `pulp-admin content sources list` formatting to show this on separate lines. 
But as far as the documentation goes I think it should match the returned format

jortel: If on the same line, pulp-admin should do either:
1. strip the '\'
2. show on separate lines.

If not too much trouble, I would appreciate your fixing this.  Not a blocker for merge.

werwty: The trailing slash is unnecessary per configparser's [long header field](https://tools.ietf.org/html/rfc822.html#section-3.1.1)
I removed the `\` from the example, but did not remove the checks for `\` at https://github.com/pulp/pulp/blob/master/server/pulp/server/content/sources/model.py#L253 and
https://github.com/pulp/pulp/pull/3018/files#diff-83a560fd6552be2c1c0b06503804c190R159
for users who are using the old config format.

As far as showing the sources on separate lines, the newline character is deliberately replaced by a `' '` by pulp-admin at https://github.com/pulp/pulp/blob/master/client_lib/pulp/client/extensions/core.py#L366
I've updated this to a v = v.replace('\n', '\n\t\t') to keep the newline and tab it for formatting

werwty:
@jortel back to you.

PR #3019, opened 2017-05-11
Fixing flake8 in travis and flake8 errors

daviddavis:
Error:

```
$ git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg
fatal: ambiguous argument '*.py': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
The command "git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg" exited with 0.
```

Also, I recommend against running flake8 against the diff instead of all files. Doing so may not pick up some flake8 errors. For example, if a function call gets removed and it's the last one in the file, then the unused import warning won't get raised as the import line isn't part of the diff.


.travis.yml
```diff
@@ -20,8 +20,8 @@ before_script:
     - psql -U postgres -c "CREATE USER pulp WITH SUPERUSER LOGIN;"
     - psql -U postgres -c "CREATE DATABASE pulp OWNER pulp;"
 script:
-    # flake8 the diff
-    - "git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg"
```
asmacdo: We originally did it this way because most of pulp failed flake8. I'm happy with this change assuming this passes when merged.

platform/pulp/app/tasks/importer.py
```diff
@@ -38,7 +37,7 @@ def sync(repo_name, importer_name):
     """
     importer = models.Importer.objects.get(name=importer_name, repository__name=repo_name).cast()
     if not importer.feed_url:
-        raise ValueError_("An importer must have a 'feed_url' attribute to sync.")
+        raise ValueError(_("An importer must have a 'feed_url' attribute to sync."))
```
asmacdo: Thank you for this! You just saved me some time!

pulp-dev.py
```diff
@@ -362,7 +362,7 @@ def install(opts):
         os.system('chmod 3775 /var/cache/pulp')
```
asmacdo: Pretty sure this whole file is unused in Pulp 3. Want to remove in a separate commit on the same PR?

asmacdo: As evidence, this Python 2 would have failed.

server/pulp/server/controllers/consumer.py
```diff
@@ -1,3 +1,5 @@
+# TODO: fix flake8 undefined name errors on this file
```
asmacdo: You can remove this since you ignored server/*

server/pulp/server/controllers/distributor.py
```diff
@@ -6,7 +6,7 @@
 import celery
 
 from pulp.common import tags
-from pulp.common.error_codes import PLP0002, PLP0003
+from pulp.common.error_codes import PLP0002
```
asmacdo: Undo this change as well. I think we should not add any changes to server/* even to make it flake8. It might imply that this code is not dead in the 3.0-dev line.

rel-eng/lib/tools.py
```diff
@@ -35,7 +35,7 @@ def increment(version):
 
 def main():
```
asmacdo: @pcreech, @dkliban, can we remove the rel-eng directory from pulp 3?

platform/pulpcore/tasking/tasks.py
```diff
@@ -64,7 +64,7 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
             concurrently with yours.
         inner_args (tuple): The positional arguments to pass on to the task.
         inner_kwargs (dict): The keyword arguments to pass on to the task.
-        options (dict): For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  #NOQA
+        options (dict): For all options accepted by apply_async please visit: http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async  # NOQA
```
asmacdo: since we use lowercase `#noqa` elsewhere, it would be better to keep it consistent.

PR #3021, opened 2017-05-11
Problem: install guide ommits THP recommendation

dkliban:
Solution: add a note about disabling Transparent Huge Pages

closes #2746

### DIFF
--- a/docs/user-guide/installation/index.rst
+++ b/docs/user-guide/installation/index.rst
@@ -42,6 +42,11 @@ Prerequisites
   `serious limitations <http://docs.mongodb.org/manual/faq/fundamentals/#what-are-the-32-bit-limitations>`_
   on 32-bit architectures. It is strongly advised that you run MongoDB on a 64-bit architecture.
 
+.. note::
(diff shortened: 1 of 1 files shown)


docs/user-guide/installation/index.rst
```diff
@@ -42,6 +42,10 @@ Prerequisites
   `serious limitations <http://docs.mongodb.org/manual/faq/fundamentals/#what-are-the-32-bit-limitations>`_
   on 32-bit architectures. It is strongly advised that you run MongoDB on a 64-bit architecture.
 
+.. note::
+  `MongoDB documentation <https://docs.mongodb.com/manual/tutorial/transparent-huge-pages/>`_
+  suggests disabling Transparent Huge Pages to ensure best performance with MongoDB.
```
bmbouter: What about putting the link on the text `disabling Transparent Huge Pages` instead of `MongoDB documentation`?

dkliban: sure. will push that change shortly.

dkliban:
I updated the link and built the docs locally. They built without error and render as expected.

PR #2741, opened 2016-09-07
Small Django models fixes

bmbouter:
- removes lingering unique_together
- removes MasterModelQuerySet
- removes MasterModelManager
- fixes typo

PR #3025, opened 2017-05-15
Supporting django 1.6-1.11

daviddavis:
fixes #2741
https://pulp.plan.io/issues/2741

### DIFF
--- a/server/pulp/server/content/web/urls.py
+++ b/server/pulp/server/content/web/urls.py
@@ -2,11 +2,6 @@ from django.conf.urls import patterns, url
 
 from pulp.server.content.web.views import ContentView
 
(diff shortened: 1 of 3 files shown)


daviddavis:
@bmbouter I tested against 1.9.13 and 1.10.7. I tested some basic functionality (repo creation, syncing, etc) and then downloading content from pulp.

daviddavis:
@goosemania great find. If you look at the versions I tested against in my comment above (1.9.13 and 1.10.7), in both cases, the patch level value (13 and 7) are greater than 6 which is why it worked for me. Anyway, will update.

daviddavis:
@goosemania updated. Thanks again for the review.

PR #3026, opened 2017-05-18
custom arguments for rsync distributor

midnightercz:
As user I want to be able to provide custom rsync arguments to rsync …

closes #2730

### DIFF
--- a/server/pulp/plugins/rsync/configuration.py
+++ b/server/pulp/plugins/rsync/configuration.py
@@ -128,6 +128,12 @@ def validate_config(repo, config, config_conduit):
     """
     _config = config.flatten()  # now config is {}
     err_list = []
+
+    if "rsync_extra_args" in _config:
+        (valid, err) = TypeValidation([list])(_config["rsync_extra_args"], _config)
+        if not valid:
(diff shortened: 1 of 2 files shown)


server/pulp/plugins/rsync/configuration.py
```diff
@@ -128,6 +128,12 @@ def validate_config(repo, config, config_conduit):
     """
     _config = config.flatten()  # now config is {}
     err_list = []
+
+    if "rsync_extra_args" in _config:
+        (valid, err) = TypeValidation([list])(_config["rsync_extra_args"])
```
dralley: Upon testing, I got a few errors.  This line needs to be

```(valid, err) = TypeValidation([list])(_config["rsync_extra_args"], _config)```

instead.  After this change, it works great.

dralley: All it's missing is ```, _config)``` at the end there to fulfill the second parameter.

dralley:
I agree with all the changes @bmbouter requested, I'm just approving based on the aforementioned technical feedback being resolved.

PR #3027, opened 2017-05-18
Update worker viewset to handle the . character

werwty:
closes #2755
https://pulp.plan.io/issues/2755


werwty:
DRF default regex: https://github.com/encode/django-rest-framework/blob/8f6173cd8afadd503dde124a616355ea80b5f6fe/rest_framework/routers.py#L233

werwty:
This fix takes care of the drf API not loading. I filed a separate issue to deal with the '@' character in the url path: https://pulp.plan.io/issues/2775

PR #3028, opened 2017-05-18
Fix progress bar.

jortel:
Fix:
```
psycopg2.ProgrammingError: can't adapt type 'Task'
```

### DIFF
--- a/platform/pulp/app/models/progress.py
+++ b/platform/pulp/app/models/progress.py
@@ -77,7 +77,7 @@ class ProgressReport(Model):
         kwargs (dict): keyword arguments to be passed on to the real save
         """
         if self.task_id is None:
-            self.task_id = Task.objects.get(id=util.get_current_task_id())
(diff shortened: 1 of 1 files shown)


platform/pulp/app/models/progress.py
```diff
@@ -77,7 +77,7 @@ def save(self, *args, **kwargs):
         kwargs (dict): keyword arguments to be passed on to the real save
         """
         if self.task_id is None:
```
asmacdo: This line needs to be changed also.

jortel: If changed to `if self.task is None` then 

```
<RelatedObjectDoesNotExist> ProgressReport has no task.
```

is raised.

asmacdo:
`self.task_id` is related to `self.task` https://docs.djangoproject.com/en/dev/ref/models/fields/#database-representation.

PR #3030, opened 2017-05-22
Use sphinx<1.6.1 until ImportError in sphinx is fixed

goosemania:
This is a temporary solution until the following issue is fixed:
https://github.com/sphinx-doc/sphinx/issues/3779


docs/requirements.txt
```diff
@@ -1,2 +1,2 @@
-sphinx
+sphinx<1.6.1
```
bmbouter: Can a comment be added with a URL to your Sphinx PR? That way we know when to remove the workaround.

PR #3031, opened 2017-05-22
TaskStatus records now error correctly

bmbouter:
For both apply_async() and apply_async_with_reservation() if the task
fails to dispatch due to an error the task status record is updated to
show 'error'.

https://pulp.plan.io/issues/2770
closes #2770


bmbouter:
Thanks for the review and testing @werwty

PR #3039, opened 2017-05-30
Fix rsync distributor skipping units

dralley:
Corrects the rsync distributor fast-forward logic so that it
uses the proper date range to determine which units to include.
This fixes an issue where units were not being published if the
association, predistributor publish and rsync publish occurred
in a certain order.

closes #2532
https://pulp.plan.io/issues/2532

### DIFF
--- a/server/pulp/plugins/rsync/publish.py
+++ b/server/pulp/plugins/rsync/publish.py
@@ -258,6 +258,27 @@ class RSyncPublishStep(PublishStep):
             raise PulpCodedException(message=output)
 
 
+class UpdateLastPredistDateStep(PublishStep):
+    """
+    After a publish of the Predistributor completes, store the date in the scratchpad.
+    """
+    def __init__(self, distributor, predist_pub_date):
+        super(UpdateLastPredistDateStep, self).__init__("UpdateLastPredistDate")
+        self.distributor = distributor
+        self.date = predist_pub_date
+
+    def process_main(self):
+        """
+        Save last_predist_last_published.
+        """
+        if "scratchpad" not in self.distributor:
+            self.distributor["scratchpad"] = {}
+
+        self.distributor["scratchpad"]["last_predist_last_published"] = self.date
+
+        self.get_conduit().set_scratchpad(self.distributor["scratchpad"])
+
+
 class Publisher(PublishStep):
     """
     RSync publisher class that provides the common code for publishing to remote server. Each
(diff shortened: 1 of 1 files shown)


server/pulp/plugins/rsync/publish.py
```diff
@@ -258,6 +258,36 @@ def process_main(self):
             raise PulpCodedException(message=output)
 
 
+class UpdateLastPredistDateStep(PublishStep):
```
ipanova: so i was thinking whether it is the correct place for the Step. Since this step is used just by rpm plugin, i thought maybe it would be better to keep in plugins/pulp_rpm/plugins/distributors/?
i looked into docker plugin and i saw some docker specific steps located and subclassed in plugins/pulp_docker/plugins/distributors/publish_steps.py
What do you think of this? @dralley

dralley: So, it's currently only used in the RPM plugin, but should I go add this to docker as well? I haven't tested to see whether the docker rsync distributor is vulnerable to the same unit skipping issue.

ipanova: docker does not have the concept of a predistributor, it has postdistributor. That said the docker has different logic that should not be impacted by this issue.

dralley: So is it likely that the only plugin that will need this code is the RPM plugin?

server/pulp/plugins/rsync/publish.py
```diff
@@ -258,6 +258,36 @@ def process_main(self):
...
+    def __init__(self, distributor, predist_pub_date):
+        super(UpdateLastPredistDateStep, self).__init__("UpdateLastPredistDate")
+        self.distributor = distributor
+        self.date = predist_pub_date
+
+    def process_main(self):
+        """
+        Save last_predist_last_published.
+        """
+        if not self.distributor["scratchpad"]:
```
ipanova: you will get the keyerror here case of scratchpad missing

server/pulp/plugins/rsync/publish.py
```diff
@@ -258,6 +258,36 @@ def process_main(self):
...
+            self.distributor["scratchpad"] = {}
+
+        self.distributor["scratchpad"]["last_predist_last_published"] = self.date
+
+        try:
```
ipanova: would not it be better to use already existing set_scratchpad()

server/pulp/plugins/rsync/publish.py
dralley: I do think it should be left where it is.  Considering that it's not used just by the rsync distributor but also the ISO rsync distributor, there's not an obvious place to put it at a lower level.

server/pulp/plugins/rsync/publish.py
```diff
@@ -351,9 +381,8 @@ def is_fastforward(self):
         force_full = force_full | config_force_full
         delete = self.get_config().get("delete", False)
 
-        return last_published and ((last_deleted and last_published > last_deleted) or
-                                   not last_deleted) and not force_full and\
-            not delete
+        return not force_full and not delete and not self.last_predist_last_published and \
+            ((last_deleted and last_published and last_published > last_deleted) or not last_deleted)
```
dkliban: last_published is twice in here

PR #2784, opened 2016-10-11
Fix AVC denials while restoring SELinux context as part of publishing.

jortel:
https://pulp.plan.io/issues/2326
https://pulp.plan.io/issues/2277

While reproducing this issue, the relabeling during publishing was failing with permission denied.

Found this:
```
[jortel@f23s ~]$ sudo audit2allow -Ral

require {
	type celery_t;
}

#============= celery_t ==============
selinux_set_enforce_mode(celery_t)
selinux_validate_context(celery_t)
```

After adding these 2 rules to the policy and installing, the publish succeeded and GET on files worked:

```
[jortel@f23s ~]$ sudo setenforce 1

[jortel@f23s ~]$ pulp-admin rpm repo sync run --repo-id=zoo 
+----------------------------------------------------------------------+
                     Synchronizing Repository [zoo]
+----------------------------------------------------------------------+

This command may be exited via ctrl+c without affecting the request.

Downloading metadata...
[\]
... completed

Downloading repository content...
[-]
[==================================================] 100%
RPMs:       0/0 items
Delta RPMs: 0/0 items

... completed

Downloading distribution files...
[==================================================] 100%
Distributions: 0/0 items
... completed

Importing errata...
[-]
... completed

Importing package groups/categories...
[-]
... completed

Cleaning duplicate packages...
[-]
... completed

Task Succeeded

Task Succeeded
```

```
[jortel@f23s ~]$  wget --no-check-certificate https://f23s.redhat.com/pulp/repos/repos/pulp/pulp/fixtures/rpm/bear-4.1-1.noarch.rpm
--2016-10-11 15:25:15--  https://f23s.redhat.com/pulp/repos/repos/pulp/pulp/fixtures/rpm/bear-4.1-1.noarch.rpm
Resolving f23s.redhat.com (f23s.redhat.com)... 192.168.122.170
Connecting to f23s.redhat.com (f23s.redhat.com)|192.168.122.170|:443... connected.
WARNING: cannot verify f23s.redhat.com's certificate, issued by ‘CN=f23s.redhat.com,O=Default Company Ltd,L=Default City,C=XX’:
  Self-signed certificate encountered.
HTTP request sent, awaiting response... 200 OK
```

PR #3040, opened 2017-05-31
Update spec files Source0

zjhuntin:
The Source0 currently does not lead to the correct source file.
This updates the spec file and renames the source to what would be
the expected source name.

ref #2784
https://pulp.plan.io/issues/2784


pcreech:
ok test

PR #3042, opened 2017-06-01
Problem: tasks API ommits error field in response

dkliban:
Solution: update the serializer for Task model by renaming 'result' field to 'error'

closes #2787


asmacdo:
https://pulp.plan.io/issues/2787

PR #3043, opened 2017-06-05
Fix rsync distributor publishing units not published by predistributor

dralley:
Fixes an issue with rsync logic where some units which have 
not yet been published by the predistributor are included
with rsync publishes.

closes #2791
https://pulp.plan.io/issues/2791

### Don't merge until after #3039


dralley:
@ipanova This currently includes the commit from the other PR because it relies on it, but I will rebase once the other PR gets merged.

The only commit that matters here is the 2nd one.

PR #3045, opened 2017-06-05
Update plugin api docs

werwty:
closes #2180
https://pulp.plan.io/issues/2180


docs/contributing/3.0-development/plugin-layout.rst
```diff
@@ -84,3 +84,12 @@ of a plugin app, and so on.
...
 plugin components, particularly for both models and viewsets.
+
+
+Packaging
+---------
+
+The Plugin API is available from PyPI as pulpcore-plugin. A plugin writer needs to specify the
+version of pulpcore-plugin their plugin is dependent on. A plugin writer does not need to be
+concerned which version of pulpcore would work with their plugin since pulpcore-plugin will
+resolve the pulpcore dependency.
```
bmbouter: Do you think we should indicate that they can expect the pulpcore will be within the 3.y line? I think plugin writers who indirectly receive core would like to know which X version of the REST API they will receive. I'm not entirely sure how to express this. Maybe we should even leave it out. Whatever you think is best I'm good with.

docs/contributing/plugin_api/index.rst
```diff
@@ -1,6 +1,11 @@
 Plugin API
 ==========
 
+The Pulp Plugin API is versioned separately from Pulp Core. It is governed by `semantic
```
bmbouter: I'm wondering where we declare this version? Is it in the setup.py maybe? How does that get declared in these docs? I'm thinking the API verison is `0.0.1a1dev0` or something like that.

docs/release_notes/plugin_api/0.1.x.rst
```diff
@@ -0,0 +1,8 @@
+Plugin API 0.1.alpha Release Notes
```
bmbouter: I'm not entirely sure if we're doing a x.y.z or x.y version scheme for the plugin api. If it's x.y then this is right but the first should be called `0.1.rst` if it's x.y.z this should probably be 0.0.1.alpha instead. What do you think?

bmbouter: Actually we need to have a hyphen separating the alpha part:  http://semver.org/#spec-item-9

docs/release_notes/plugin_api/0.1.x.rst
```diff
@@ -0,0 +1,8 @@
...
+==================================
+
+The Plugin API is currently in the alpha stage. Backwards incompatible changes might be made until
```
bmbouter: I think it's s/is currently in the alpha stage/is not yet declared as stable/.

docs/contributing/plugin_api/index.rst
werwty: I think the user would see the latest version in the release notes. I will add a link for easier access

docs/release_notes/plugin_api/0.1.x.rst
werwty: Yeah, since we're going with x.y.z I'll rename the first to `0.0.x.rst` (or should it be `0.0.z.rst`?).
I don't think it should be `0.0.1.alpha.rst` because we want to group all the z releases together

docs/contributing/plugin_api/index.rst
bmbouter: Yeah that sounds good.

docs/contributing/3.0-development/plugin-layout.rst
werwty: Since this documentation is in the Pulp3 docs, I don't think it's necessary to declare it here. Maybe a line in the release notes? Something like: The PluginAPI currently works with this version of Platform?

docs/release_notes/plugin_api/0.1.x.rst
bmbouter: I think `0.0.z.rst` is right.  I'm generally +0 on using `x.y.z` versus `x.y` only. bugfixes may be infrequent unlike platform. Maybe having the z bit is also good. I'm not sure. Let's probably leave it at x.y.z for now.

docs/contributing/3.0-development/plugin-layout.rst
bmbouter: Yeah maybe the release notes can articulate the version of platform that the plugin API version is compatible with. It's a release by release thing so that probably is the place for it.

werwty:
@bmbouter I've pushed some changes, can you re-review?

docs/contributing/3.0-development/plugin-layout.rst
```diff
@@ -84,3 +84,13 @@ of a plugin app, and so on.
 
 This matching of module names is required for the Pulp Platform to be able to auto-discover
 plugin components, particularly for both models and viewsets.
+
+
+Packaging
+---------
+
+The Plugin API is available from PyPI as pulpcore-plugin. A plugin writer needs to specify the
+version of pulpcore-plugin their plugin is dependent on. A plugin writer does not need to be
```
bmbouter: s/version/minimum version/

docs/contributing/plugin_api/index.rst
```diff
@@ -1,6 +1,12 @@
...
+versioning <http://semver.org/>`_. Backwards incompatible changes may be made while the Plugin
+API is in the alpha stage. The Plugin API will be considered stable when it reaches v1.0.
```
bmbouter: Semantic versioning doesn't define an alpha stage. We define an alpha stage as a code maturity statement, but I think semver only has "stable"( >= 1.0) and "not yet stable" ( < 1.0).

I think the sentence `Backwards incompatible changes may be made while the Plugin API is in the alpha stage. The Plugin API will be considered stable when it reaches v1.0.` should be rewritten to something like:

`Backwards incompatible changes may be made until the Plugin API reaches stability with v1.0`.

docs/release_notes/plugin_api/0.0.z.rst
```diff
@@ -0,0 +1,18 @@
+============================
+Plugin API 0.0 Release Notes
+============================
+
+The Plugin API is not yet declared as stable. Backwards incompatible changes might be made until
+stable version 1.0 is reached.
+
+The PluginAPI currently supports version 3.y of Pulp Core.
```
bmbouter: s/PluginAPI/Plugin API/

docs/release_notes/plugin_api/0.0.z.rst
```diff
@@ -0,0 +1,18 @@
...
+
+See :doc:`Plugin API <../../contributing/plugin_api/index>` and
+:doc:`Plugin Development <../../contributing/3.0-development/plugin-layout>`
+
+
+Plugin API 0.0.1-alpha
```
bmbouter: This release note is in preparation for the GA release so I think this should be s/0.0.1-alpha/0.0.1/

PR #3046, opened 2017-06-06
Dispatch a task to update a publisher.

asmacdo:
closes #2409

### DIFF
--- a/platform/pulpcore/app/tasks/publisher.py
+++ b/platform/pulpcore/app/tasks/publisher.py
@@ -1,10 +1,37 @@
-from gettext import gettext as _
-
 from celery import shared_task
+from django.http import QueryDict
 
 from pulpcore.app import models
+from pulpcore.app.apps import get_plugin_config
 from pulpcore.tasking.services import storage
 from pulpcore.tasking.tasks import UserFacingTask
+
+
+@shared_task(base=UserFacingTask)
+def update(publisher_id, app_label, serializer_name, data=None, partial=False):
+    """
+    Update an instance of a :class:`~pulpcore.app.models.Publisher`
+
+    Args:
+        publisher_id (str): id of the publisher instance
+        app_label (str): the Django app label of the plugin that provides the publisher
+        serializer_name (str): name of the serializer class for this publisher
+        data (dict): Data to update on the publisher. keys are field names, values are new values.
+        partial (bool): When true, update only the specified fields. When false, omitted fields
+            are set to None.
+    Raises:
+        :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
+            due to validation error. This theoretically should never occur since validation is
+            performed before the task is dispatched.
+    """
+    instance = models.Publisher.objects.get(id=publisher_id).cast()
+    data_querydict = QueryDict("", mutable=True)
+    data_querydict.update(data)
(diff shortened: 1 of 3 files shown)


platform/pulpcore/app/tasks/publisher.py
```diff
@@ -3,11 +3,20 @@
 from celery import shared_task
 
 from pulpcore.app import models
+from pulpcore.app import serializers
 from pulpcore.tasking.services import storage
 from pulpcore.tasking.tasks import UserFacingTask
 
 
 @shared_task(base=UserFacingTask)
+def update(publisher_id, data, *args, **kwargs):
```
asmacdo: i need to add `partial` here

platform/pulpcore/app/tasks/publisher.py
```diff
@@ -3,11 +3,20 @@
...
+    instance = models.Publisher.objects.get(id=publisher_id)
+    serializer = serializers.PublisherSerializer(instance, data=data, partial=partial)
```
werwty: partial needs to be `kwargs.pop('partial')` or added as a keyword argument in the function definition

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -140,6 +140,19 @@ class PublisherViewSet(NamedModelViewSet):
...
     filter_class = PublisherFilter
 
+    def update(self, request, *args, **kwargs):
+        partial = kwargs.pop('partial', True)
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        async_result = tasks.publisher.update.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, instance.repository.name,
+            kwargs={'publisher_id': instance.id,
```
werwty: publisher_id and data should be passed with a args keyword argument (since that's what pulpcore.app.tasks.publisher.update() is expecting)

```
tasks.publisher.update.apply_async_with_reservation(
    tags.RESOURCE_REPOSITORY_TYPE, instance.repository.name,
    args=(publisher_id, data, ),
    kwargs={'partial': partial}
)
```

platform/pulpcore/app/tasks/publisher.py
werwty: Should data be a required argument? Is it allowable for the user to call PUT without passing in data?

werwty: Docstrings please :)

bmbouter: Can `*args` and `*kwargs` be converted to named positional and keyword parameters. We should never use `*args, and **kwargs` unless we don't know what will be passed in. Since we control the signature and the arguments sent when the task is dispatched we know the arguments in this case.

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -140,6 +140,19 @@ class PublisherViewSet(NamedModelViewSet):
     queryset = Publisher.objects.all()
     filter_class = PublisherFilter
 
+    def update(self, request, *args, **kwargs):
```
bmbouter: Same for here, we should know all positional and keyword arguments since we are defining the API. Please convert to named positional and keyword arguments.

PR #3048, opened 2017-06-07
Problem: importer updates are performed syncronously

dkliban:
Solution: override the update method in the Importer viewset

The update method now dispatches a task to update the importer.

closes #2402
https://pulp.plan.io/issues/2402


platform/pulpcore/app/tasks/importer.py
```diff
@@ -1,14 +1,29 @@
...
+    """
+    Update an :class:`~pulp.app.models.Importer`
+
+    Args:
+        repo_name (str): the name of a repository
+        importer_name (str): the name of an importer
+        data (dict): dictionary of key:value pairs to update the importer with
+    """
+    instance = models.Importer.objects.get(id=importer_id)
+    serializer = ImporterSerializer(instance, data=data, partial=True)
```
asmacdo: `update` accepts `partial` as an optional kwarg, which should be used here.

platform/pulpcore/app/tasks/importer.py
```diff
@@ -38,7 +53,7 @@ def sync(repo_name, importer_name):
     """
     importer = models.Importer.objects.get(name=importer_name, repository__name=repo_name).cast()
     if not importer.feed_url:
-        raise ValueError_("An importer must have a 'feed_url' attribute to sync.")
+        raise ValueError(_("An importer must have a 'feed_url' attribute to sync."))
```
asmacdo: I would have sworn this was fixed.

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -114,6 +114,18 @@ class ImporterViewSet(NamedModelViewSet):
...
     filter_class = ImporterFilter
 
+    def update(self, request, *args, **kwargs):
+        partial = kwargs.pop('partial', True)
+        importer = self.get_object()
+        serializer = self.get_serializer(importer, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        async_result = tasks.importer.update.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, importer.repository.name,
+            kwargs={'importer_id': importer.id,
```
werwty: apply_async_with_reservation has a args keyword argument which is for the position arguments to pass on to the task. I think that's what should be used here, since that is what update() is expecting.

```
async_result = tasks.importer.update.apply_async_with_reservation(
             tags.RESOURCE_REPOSITORY_TYPE, importer.repository.name,
             args=(importer.id, ), kwargs={'data': request.data} )
```

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -114,6 +114,19 @@ class ImporterViewSet(NamedModelViewSet):
     endpoint_name = 'importers'
     filter_class = ImporterFilter
 
+    def update(self, request, *args, **kwargs):
```
bmbouter: I'm hoping that we can avoid `*args, **kwargs` from the codebase entirely. We should only use it when we specifically don't know what args will be passed. If you see any else consider changing those too, although maybe as a separate PR.

platform/pulpcore/app/apps.py
```diff
@@ -17,6 +17,23 @@ def pulp_plugin_configs():
...
+    Args:
+        plugin_label (basestring): Django app label of the pulp plugin
+
+    Returns:
+        :class:`pulpcore.app.apps.PulpPluginAppConfig`: The app config of the Pulp plugin.
+    """
+    for config in pulp_plugin_configs():
+        if config.label == plugin_app_label:
+            return config
+    raise RuntimeError("Plugin with Django app label %s is not installed." % plugin_app_label)
```
bmbouter: Can this be i8ln with gettext?

platform/pulpcore/app/apps.py
```diff
@@ -17,6 +17,23 @@ def pulp_plugin_configs():
         if isinstance(app_config, PulpPluginAppConfig):
             yield app_config
 
+def pulp_plugin_config(plugin_app_label):
```
asmacdo: Could we rename this? `get_plugin_config` would be my vote. The object names in this module already feel a little overloaded.

platform/pulpcore/app/apps.py
```diff
@@ -17,6 +17,23 @@ def pulp_plugin_configs():
...
+    """
+    A getter of specific pulp plugin config
+
+    This makes it easy to retrieve a config for a specific Pulp plugin when looking for a
+    registered model.
+
+    Args:
+        plugin_label (basestring): Django app label of the pulp plugin
+
+    Returns:
```
bmbouter: This needs a `Raises` section in the google style docstring for the RuntimeError

platform/pulpcore/app/apps.py
```diff
@@ -17,6 +17,23 @@ def pulp_plugin_configs():
...
 
+def pulp_plugin_config(plugin_app_label):
+    """
+    A getter of specific pulp plugin config
+
+    This makes it easy to retrieve a config for a specific Pulp plugin when looking for a
+    registered model.
+
+    Args:
+        plugin_label (basestring): Django app label of the pulp plugin
```
asmacdo: s/plugin_label (basestring)/plugin_app_label (str)/

platform/pulpcore/app/apps.py
bmbouter: Should this become a coded exception? Expecting to have something installed and it's not installed I think is something we want to raise in many places. What do you think?

platform/pulpcore/app/apps.py
```diff
@@ -38,13 +55,38 @@ def __init__(self, app_name, app_module):
...
 
     def ready(self):
         # register signals here as suggested in Django docs
         # https://docs.djangoproject.com/en/1.8/topics/signals/#connecting-receiver-functions
         import pulpcore.app.signals  # noqa
 
         self.import_viewsets()
+        self.import_serializers()
+
+    def import_serializers(self):
```
asmacdo: It would be nice if this could be combined with `import_viewsets` since the code is nearly identical.

platform/pulpcore/app/tasks/importer.py
```diff
@@ -1,14 +1,37 @@
...
 
 @shared_task(base=UserFacingTask)
+def update(importer_id, app_label, serializer_name, data=None, partial=False):
+    """
+    Update an :class:`~pulp.app.models.Importer`
+
+    Args:
+        importer_id (basestring): the id of the importer
+        app_label (str): the Django app label of the plugin that provides the importer
+        serializer_class_name (str): name of the serializer class for the importer
```
asmacdo: s/serializer_class_name/serializer_name/

platform/pulpcore/app/viewsets/repository.py
asmacdo: I think `def update(self, request, partial=False)` is what we need.

platform/pulpcore/app/tasks/importer.py
```diff
@@ -1,14 +1,37 @@
...
 from pulpcore.tasking.tasks import UserFacingTask
 
 
 @shared_task(base=UserFacingTask)
+def update(importer_id, app_label, serializer_name, data=None, partial=False):
+    """
+    Update an :class:`~pulp.app.models.Importer`
+
+    Args:
+        importer_id (basestring): the id of the importer
```
asmacdo: s/basestring/str/

platform/pulpcore/app/apps.py
```diff
@@ -38,13 +55,38 @@ def __init__(self, app_name, app_module):
...
+        # circular import avoidance
+        from pulpcore.app.serializers import ModelSerializer
+
+        self.named_serializers = {}
+        if module_has_submodule(self.module, SERIALIZERS_MODULE_NAME):
+            # import the viewsets module and track any interesting viewsets
```
bmbouter: s/interesting/discovered/

platform/pulpcore/app/tasks/importer.py
```diff
@@ -1,14 +1,37 @@
...
+        app_label (str): the Django app label of the plugin that provides the importer
```
bmbouter: Is it s/str/basestring/ ? I'm not sure. If this field correctly supports unicode objects then it should be basestr. Similar throughout.

bmbouter:
We needs docs on these 'discovery by name' conventions. Can that either come with this PR or a new ticket made for those docs? I think adding docs (even simple ones) along with features is my preference.

platform/pulpcore/app/tasks/importer.py
asmacdo: str for Python 3 https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit

platform/pulpcore/app/apps.py
dkliban: done!

platform/pulpcore/app/apps.py
dkliban: added

platform/pulpcore/app/apps.py
dkliban: It probably should. I'd rather make that change as part of anotehr PR though.

dkliban: I did add the i8ln

platform/pulpcore/app/viewsets/repository.py
dkliban: i updated the args to be explicit.

platform/pulpcore/exceptions/plugin.py
```diff
@@ -0,0 +1,22 @@
...
+    Missing plugin exception.
+
+    Exception that is raised when a requested plugin is not installed.
+    """
+    def __init__(self, plugin_app_label):
+        """
+        :param resources: keyword arguments of resource_type=resource_id
+        :type resources: dict
+        """
+        super(MissingPlugin, self).__init__("PLP0002")
```
asmacdo: Will these numbers have a scheme or are we just going to increment?

bmbouter: Yeah I think just incrementally added over time.

PR #3050, opened 2017-06-08
Add repository update task

werwty:
closes #2756
https://pulp.plan.io/issues/2756

### DIFF
--- a/platform/pulpcore/app/tasks/repository.py
+++ b/platform/pulpcore/app/tasks/repository.py
@@ -1,16 +1,36 @@
 from celery import shared_task
 
 from pulpcore.app import models
+from pulpcore.app import serializers
 from pulpcore.tasking.tasks import UserFacingTask
 
 
 @shared_task(base=UserFacingTask)
-def delete(repo_name):
+def delete(repo_id):
     """
     Delete a :class:`~pulpcore.app.models.Repository`
 
-    :param repo_name:       the name of a repository
-    :type  repo_name:       str
+    Args:
+        repo_id (UUID): The name of the repository to be deleted
     """
 
-    models.Repository.objects.filter(name=repo_name).delete()
+    models.Repository.objects.filter(pk=repo_id).delete()
+
+
+@shared_task(base=UserFacingTask)
+def update(repo_id, partial=True, data=None):
+    """
+    Updates a :class:`~pulpcore.app.models.Repository`
+
+    Args:
+        repo_id (UUID): The id of the repository to be updated
+        partial (bool): Boolean to allow partial updates. If set to False, values for all
+                        required fields must be passed or a validation error will be raised.
+                        Defaults to True
+        data (QueryDict): dict of attributes to change and their new values; if None, no attempt to
+                     update the repository object will be made
+    """
+    instance = models.Repository.objects.get(pk=repo_id)
+    serializer = serializers.RepositorySerializer(instance, data=data, partial=partial)
+    serializer.is_valid(raise_exception=True)
+    serializer.save()
--- a/platform/pulpcore/app/viewsets/repository.py
+++ b/platform/pulpcore/app/viewsets/repository.py
@@ -51,10 +51,28 @@ class RepositoryViewSet(NamedModelViewSet):
         serializer = ImporterSerializer(page, many=True, context={'request': request})
         return paginator.get_paginated_response(serializer.data)
 
+    def update(self, request, name, partial=False):
+        """
+        Generates a Task to update a :class:`~pulpcore.app.models.Repository`
+        """
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        async_result = tasks.repository.update.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, str(instance.id),
+            args=(instance.id, ),
+            kwargs={'data': request.data,
+                    'partial': partial}
+        )
+        return OperationPostponedResponse([async_result])
+
     def destroy(self, request, name):
+        """
+        Generates a Task to delete a :class:`~pulpcore.app.models.Repository`
+        """
         repo = self.get_object()
         async_result = tasks.repository.delete.apply_async_with_reservation(
-            tags.RESOURCE_REPOSITORY_TYPE, repo.name, kwargs={'repo_name': repo.name})
+            tags.RESOURCE_REPOSITORY_TYPE, str(repo.id), kwargs={'repo_id': repo.id})
         return OperationPostponedResponse([async_result])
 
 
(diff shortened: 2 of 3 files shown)


werwty:
There's an issue that comes up when updating with crispy forms:
`raised unexpected: ValidationError({'scratchpad': ['Expected a dictionary of items but got type "str".'], 'notes': ['Expected a dictionary of items but got type "str".']},)`

Because crispy forms passes in an empty string for scratchpad and notes. This issue does not come up when testing with curl
`curl -X PUT -F name=test -F description=123456 http://127.0.0.1:3000/api/v3/repositories/test/`

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -51,6 +51,20 @@ def importers(self, request, name):
         serializer = ImporterSerializer(page, many=True, context={'request': request})
         return paginator.get_paginated_response(serializer.data)
 
+
+    def update(self, request, *args, **kwargs):
```
bmbouter: The other viewsets in this file avoid using *args and **kwargs. Is it possible for us to only used named positional and keyword arguments for this signature? I'm not sure.

werwty: I took my function signature cues from https://github.com/pulp/pulp/pull/3046/files and https://github.com/pulp/pulp/pull/3048/files 
I like having a consistent update function signature, but since neither of those are merged yet, I guess I can set the example :)

platform/pulpcore/app/tasks/repository.py
```diff
@@ -9,8 +10,27 @@ def delete(repo_name):
...
+def update(repo_name, partial=True, data=None):
+    """
+    Updates a :class:`~pulpcore.app.models.Repository`
+
+    Args:
+        repo_name (str): The name of the repository to be updated
+        partial (bool): Boolean to allow partial updates. If set to False, values for all
+                        required fields must be passed or a validation error will be raised.
+                        Defaults to True
+        data (dict): dict of attributes to change and their new values; if None, no attempt to
```
asmacdo: `data` is a `QueryDict`

platform/pulpcore/app/tasks/repository.py
```diff
@@ -9,8 +10,27 @@ def delete(repo_name):
...
-    :type  repo_name:       str
+    Args:
+        repo_name (str): The name of the repository to be deleted
     """
 
     models.Repository.objects.filter(name=repo_name).delete()
+
+
+@shared_task(base=UserFacingTask)
+def update(repo_name, partial=True, data=None):
```
asmacdo: s/repo_name/repository_id/

If the user changes the name of the repo in another, the id will still be the same.

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -51,6 +51,19 @@ def importers(self, request, name):
         serializer = ImporterSerializer(page, many=True, context={'request': request})
         return paginator.get_paginated_response(serializer.data)
 
+
+    def update(self, request, partial=True):
```
asmacdo: I don't know how we want to document this type of thing. The ViewSet definitely will need a docstring, but it seems silly and fragile to write the same docstring twice.

asmacdo: It would be nice to mention that `partial=True` for PATCH and `partial=false` for PUT

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -51,6 +51,19 @@ def importers(self, request, name):
...
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        async_result = tasks.repository.update.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, instance.name,
+            args=(instance.name, ),
```
asmacdo: instance.id

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -51,6 +51,19 @@ def importers(self, request, name):
         serializer = ImporterSerializer(page, many=True, context={'request': request})
         return paginator.get_paginated_response(serializer.data)
 
+
+    def update(self, request, partial=True):
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        async_result = tasks.repository.update.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, instance.name,
```
asmacdo: I think we should start reserving by repo id instead of repo name. This is necessary if repo names can be changed, which we haven't discussed disallowing yet. Thoughts, @bmbouter?

platform/pulpcore/app/viewsets/repository.py
asmacdo: DRF defaults to `partial=False`. I think we ought to do the same. Partial updates must be PATCH.

https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L66

platform/pulpcore/app/tasks/repository.py
```diff
@@ -9,8 +10,27 @@ def delete(repo_name):
     """
     Delete a :class:`~pulpcore.app.models.Repository`
 
-    :param repo_name:       the name of a repository
-    :type  repo_name:       str
+    Args:
```
asmacdo: screenshot from 2017-06-22 16-59-17

platform/pulpcore/app/viewsets/repository.py
bmbouter: Yes what we lock on needs to be immutable. Allowing people to change the repo name sounds valuable also. Together that would suggest we should lock on something other than the repo name. The one thing that I think is always immutable is the id field which is a uuid.

One thing to be aware of with ^ change is that users liked that we used the repo name so they could search for all outstanding tasks for a given repo. We need to allow that use case somehow.

This is a broad issue affecting all of our locks; we should look at all existing `apply_async_with_reservation` calls and change all of them. This is probably worth of a check-in thread on pulp-dev.

platform/pulpcore/app/viewsets/repository.py
werwty: Since the PATCH partial default is true for all the update calls, what do you think about documenting this in the contributing/platform_api/app/viewsets.rst?

werwty: Actually now that I think about it PUT/PATCH RFCs already note that PUT should include all parameter by design [0]. Since this is a part of the protocol design, I don't think we need extra documentation on it.

[0] https://stackoverflow.com/questions/28459418/rest-api-put-vs-patch-with-real-life-examples

asmacdo: I'd like to mention it somewhere in our docs since PATCH is a newish verb and we used PUTs to for partial updates in Pulp 2. Leaving it out here is fine though.

PR #2808, opened 2016-11-01
Add Publisher serializer

goosemania:
re #2298
https://pulp.plan.io/issues/2298

PR #3053, opened 2017-06-09
Update docs to clarify use of pulp-consumer

kdelee:
~~Make all examples in docs use `$ sudo pulp-consumer ...` to make it clear
that the user must run pulp-consumer as root or with sudo.~~

(see below)

closes #2808
https://pulp.plan.io/issues/2808

Hello! This is my first pull request to pulp/pulp, mostly just trying to learn the process! 
I found this part of the documentation confusing, so here are the changes that I think would make it easier to understand.


kdelee:
Added a couple more notes notifying users that `pulp-consumer` at times needs root level permissions.

kdelee:
@bmbouter 
I had forgotten to add my name, but it is in there now.
Thanks! 👍

PR #3054, opened 2017-06-12
Associate worker foreign key with task

dralley:
Ensure the worker field on the Task model gets populated.

closes #2768
https://pulp.plan.io/issues/2768

### DIFF
--- a/platform/pulpcore/tasking/tasks.py
+++ b/platform/pulpcore/tasking/tasks.py
@@ -89,6 +89,8 @@ def _queue_reserved_task(name, inner_task_id, resource_id, inner_args, inner_kwa
         time.sleep(0.25)
 
     task_status = TaskStatus.objects.get(pk=inner_task_id)
+    task_status.worker = worker
+    task_status.save()
     ReservedResource.objects.create(task=task_status, worker=worker, resource=resource_id)
 
     try:
@@ -243,14 +245,25 @@ class UserFacingTask(PulpTask):
 
     def __call__(self, *args, **kwargs):
         """
-        Set the :class:`pulpcore.app.models.Task` object in the running state and log some output.
+        Called on the worker just before the task begins executing.
 
-        Skip the status updating if the task is called synchronously.
+        Set the :class:`pulpcore.app.models.Task` object in the running state,
+        associate the worker, and log a message.
         """
-        if not self.request.called_directly:
-            task_status = TaskStatus.objects.get(pk=self.request.id)
-            task_status.set_running()
-        _logger.debug("Running task : [%s]" % self.request.id)
+        task_status = TaskStatus.objects.get(pk=self.request.id)
+        task_status.set_running()
+
+        try:
+            worker = Worker.objects.get(name=self.request.hostname)
+        except Worker.DoesNotExist:
+            _logger.warning(_("Failed to find worker with hostname {host}").format(
+                host=self.request.hostname))
+        else:
+            if not task_status.worker:
+                task_status.worker = worker
+                task_status.save()
+
+        _logger.debug(_("Running task : [%s]") % self.request.id)
 
         if pulp_settings.PROFILING['enabled'] is True:
             self.pr = cProfile.Profile()


platform/pulpcore/app/models/task.py
```diff
@@ -163,6 +163,23 @@ class Task(Model):
     parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks")
     worker = models.ForeignKey("Worker", null=True, related_name="tasks")
 
+    def set_worker(self, worker):
+        """
+        Set the worker Foreign Key on this Task, save it, and log output in warning cases.
+
+        This updates the :attr:`worker`.
```
bmbouter: This could be the full path to the attribute. I don't think Sphinx will link this as is. If it had the full classpath it would link it I think. Maybe rendering the docs would be a good way to check the docstrings.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,13 +243,22 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
 
     def __call__(self, *args, **kwargs):
         """
-        Set the :class:`pulpcore.app.models.Task` object in the running state and log some output.
+        Set the :class:`pulpcore.app.models.Task` object in the running state, set the worker,
```
bmbouter: This somehow needs to be condensed to a single line. The first line of a multi-line docstring should be a single line with a line of whitespace after to adhere to [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

platform/pulpcore/app/models/task.py
dralley: I was assuming it was correct as other attributes on the same model are referenced the same way, but I will try to render the docs to make sure.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,13 +243,22 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
...
-            task_status = TaskStatus.objects.get(pk=self.request.id)
-            task_status.set_running()
+        task_status = TaskStatus.objects.get(pk=self.request.id)
+        task_status.set_running()
+
+        worker = None
+        try:
+            worker = Worker.objects.get(name=self.request.hostname)
+        except Worker.DoesNotExist:
+            _logger.warn("Failed to find worker with hostname {host}".format(
```
bmbouter: This would benefit from using `gettext` for i8ln

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,13 +243,22 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
...
+                host=self.request.hostname))
+
+        task_status.set_worker(worker)
```
bmbouter: This only needs to be set if L#256 does not raise an exception. In that case, L#254 can be deleted and this line L#261 can be moved to an `else:` block as part of the `try/except` block.

platform/pulpcore/app/models/task.py
bmbouter: cool thx

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,14 +245,25 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
 
     def __call__(self, *args, **kwargs):
         """
-        Set the :class:`pulpcore.app.models.Task` object in the running state and log some output.
+        Init code for our task.
+
+        Set the :class:`pulpcore.app.models.Task` object in the running state,
+        associate the worker, and log a message.
 
         Skip the status updating if the task is called synchronously.
```
bmbouter: This needs to be removed since it no longer does this.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,14 +245,25 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
...
         """
-        if not self.request.called_directly:
-            task_status = TaskStatus.objects.get(pk=self.request.id)
-            task_status.set_running()
-        _logger.debug("Running task : [%s]" % self.request.id)
+        task_status = TaskStatus.objects.get(pk=self.request.id)
+        task_status.set_running()
+
+        try:
+            worker = Worker.objects.get(name=self.request.hostname)
```
bmbouter: We should only update if the worker exists and the `worker` field is None. I think this needs an additional `worker=None` option on this `get()` call.

platform/pulpcore/tasking/tasks.py
dralley: Does it not do it?  Not in the common case, but I thought we still need this code for other scenarios.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,14 +245,25 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
 
     def __call__(self, *args, **kwargs):
         """
-        Set the :class:`pulpcore.app.models.Task` object in the running state and log some output.
+        Init code for our task.
```
bmbouter: Can this be:  "Called on the worker just before the task begin executing."

platform/pulpcore/tasking/tasks.py
dralley: I think an easier solution is to put ```task_status.set_worker()``` behind a conditional.   This is a query for workers, not a query for task statuses, so we can't just put worker=None here.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,14 +245,26 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
...
+        task_status.set_running()
+
+        try:
+            worker = Worker.objects.get(name=self.request.hostname)
+        except Worker.DoesNotExist:
+            _logger.warning(_("Failed to find worker with hostname {host}").format(
+                host=self.request.hostname))
+        else:
+            if not task_status.worker:
+                task_status.set_worker(worker)
```
bmbouter: Can set_worker() be removed entirely and have it's two lines of function be moved here instead?

platform/pulpcore/tasking/tasks.py
bmbouter: Yeah you are right, I misread this to be a query for Tasks

platform/pulpcore/tasking/tasks.py
bmbouter: This does need to be removed since L#250 was been removed.

platform/pulpcore/tasking/tasks.py
```diff
@@ -243,14 +245,25 @@ def apply_async(self, tags=[], group_id=None, args=None, kwargs=None, **options)
 
     def __call__(self, *args, **kwargs):
         """
-        Set the :class:`pulpcore.app.models.Task` object in the running state and log some output.
+        Called on the worker just before the task begin executing.
```
bmbouter: s/begin/begins/

dralley:
Thanks @bmbouter , I fixed that issue and I'll merge once I test one more time.

PR #3056, opened 2017-06-13
Remove SECRET_KEY from Django settings

werwty:
closes #2416
https://pulp.plan.io/issues/2416

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ before_script:
 script:
     # flake8 the diff
     - "git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg"
+    # add secret key to django settings.py
+    - echo "SECRET_KEY= '$(cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(-_=+)' | head -c 50)'" >> platform/pulpcore/app/settings.py
     # tests can't run without migrations being made
     - "python manage.py makemigrations pulp_app"
     # chain these so we don't try to run tests if the db reset fails
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,34 @@
+==================
+Installation Guide
+==================
+
+PyPI
+----
+
+CentOS, RHEL, Fedora
+--------------------
+
+Source
+------
+
+
+Configuration Files
+-------------------
+
+Pulp's server configuration file is located at `/etc/pulp/server.yaml`
+
+SECRET_KEY
+    In order to get a pulp server up and running a `Django SECRET_KEY
+    <https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-SECRET_KEY>`_ must be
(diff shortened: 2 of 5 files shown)


werwty:
Should be merged with https://github.com/pulp/devel/pull/59

.travis.yml
```diff
@@ -23,6 +23,8 @@ before_script:
 script:
     # flake8 the diff
     - "git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg"
+    # add secret key to django settings.py
+    - echo "SECRET_KEY= '$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 100 | head -n 1)'" >> platform/pulpcore/app/settings.py
```
dralley: From what I can see, Django secret keys also contain special characters.  Would the lack thereof here compromise the integrity of the key?

I believe the full set of possible characters is 

abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)

werwty: This line ensures that travis can run `makemigrations` and `manage.py test` without Django complaining. 
I don't think having a bad secret for our travis testing is an issue, since I really doubt anyone would use this for their pulp environment and the travis instance does not live for very long.

dralley: 👍

.travis.yml
```diff
@@ -23,6 +23,8 @@ before_script:
 script:
     # flake8 the diff
     - "git diff HEAD^ '*.py' | flake8 --diff --config flake8.cfg"
+    # add secret key to django settings.py
+    - echo "SECRET_KEY= '$(cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(-_=+)' | head -c 50)'" >> platform/pulpcore/app/settings.py
```
bmbouter: Nice one liner. I tested it. My only question is: should this this be added as an entry to server.yaml instead? It's Travis so it probably doesn't matter, but that is what we recommend the users to do. What do you think?

docs/installation.rst
```diff
@@ -0,0 +1,34 @@
...
+Pulp's sever configuration file is located at `/etc/pulp/server.yaml`
+
+SECRET_KEY
+    In order to get a pulp server up and running a `Django SECRET_KEY
+    <https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-SECRET_KEY>`_ must be
+    provided in server.yaml.
+
+    The following code snippet can be used to generate a random SECRET_KEY
+
+.. code-block:: python
```
bmbouter: nice

.travis.yml
werwty: I tried making a /etc/pulp/server.yaml file on travis, but it threw a 'exit code 1' error that I assumed to be due to lack of permission. This was the next best thing that worked :)

bmbouter: Oh that makes sense. Thanks.

PR #3058, opened 2017-06-16
Added status api endpoint

daviddavis:
fixes #2804
https://pulp.plan.io/issues/2804


platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns True if pulp is connected to the database
+
+    Returns:
+        bool: True if there's a db connection. False otherwise.
+    """
+    try:
+        Worker.objects.count()
+        return True
+    except:
+        return False
```
mhrivnak: It's probably worth logging something here about what went wrong. Maybe just ``_logger.exception()``

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def get_workers():
+    """
+    Gets a list of workers from the db
+
+    Returns:
+        list: a list of worker records
```
mhrivnak: I think this will return a ``QuerySet`` object if the conditional expression is True.

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,35 @@
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from pulpcore.app.serializers.task import WorkerSerializer
+import status
```
mhrivnak: I'm surprised this worked for you. It should be a fully-qualified import, I think: ``from pulpcore.app import status``

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,35 @@
...
+        version = {'pulpcore_version': status.get_version()}
+
+        workers = status.get_workers()
+        context = {'request': request}
+        worker_status = [WorkerSerializer(worker, context=context).data for worker in workers]
+
+        broker_status = {'connected': status.get_broker_conn_status()}
+        db_status = {'connected': status.get_db_conn_status()}
+
+        response = {
```
mhrivnak: I wonder if this should be done by a Serializer. Perhaps that would better enable DRF to introspect the returned structure, but I'm fuzzy on those details of DRF. Thoughts?

@asmacdo  ?

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Returns:
+        bool: True if pulp can connect to the broker. False otherwise.
+    """
+    try:
+        conn = celery.connection()
+        conn.connect()
+        conn.release()
+        return True
+    except Exception:
```
mhrivnak: I think it's better to omit the ``Exception`` word here and just make it a bare ``except`` clause like the one above.

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,35 @@
...
+
+from pulpcore.app.serializers.task import WorkerSerializer
+import status
+
+
+class StatusView(APIView):
+    """
+    Returns status information about the application
+    """
+    authentication_classes = []
```
mhrivnak: I think it's worth a quick comment to explain why these are being set to empty lists.

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,35 @@
...
+    permission_classes = []
+
+    def get(self, request, format=None):
+        """
+        Returns app information including the version, known workers, database connection status,
+        and messaging connection status
```
mhrivnak: Obvious as it may be, I'm a fan of documenting the parameters, even if that means copy-pasting from the parent class. Since parent classes can change, that at least captures what your expectations were at the time of writing the subclass. It also helps the reader immensely.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+def get_broker_conn_status():
+    """
+    Returns True if pulp can connect to the message broker
+
+    Returns:
+        bool: True if pulp can connect to the broker. False otherwise.
+    """
+    try:
+        conn = celery.connection()
+        conn.connect()
```
mhrivnak: The connection has a ``connected`` attribute. Would it be better to use that?

http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.connection.html#kombu.connection.Connection.connected

If this function momentarily connects to the broker, but then disconnects, and the rest of Pulp is also disconnected from the broker, then this could be misleading.

platform/pulpcore/app/views/status.py
mhrivnak: @dkliban maybe you have thoughts on this too?

platform/pulpcore/app/status.py
bmbouter: We can't use the connected attribute because it is not true in many cases. We started off originally with that but then we ran into this:  https://github.com/celery/kombu/issues/432

I don't know of a better way than to form a connection and then disconnect. I wish there was one.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+def get_db_conn_status():
+    """
+    Returns True if pulp is connected to the database
+
+    Returns:
+        bool: True if there's a db connection. False otherwise.
+    """
+    try:
+        Worker.objects.count()
+        return True
```
bmbouter: This could be moved to an `else` block because we're not trying to handle exceptions generated from this statement.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns True if pulp can connect to the message broker
+
+    Returns:
+        bool: True if pulp can connect to the broker. False otherwise.
+    """
+    try:
+        conn = celery.connection()
+        conn.connect()
+        conn.release()
+        return True
```
bmbouter: This could also be moved to an `else` block for similar reasons as above.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+
+def get_workers():
+    """
+    Gets a list of workers from the db
+
+    Returns:
+        list: a list of worker records
+    """
+    if get_db_conn_status():
+        workers = Worker.objects.all()
```
bmbouter: This could just be `return Worker.objects.all()`

platform/pulpcore/app/status.py
bmbouter: I don't have a pulp3 env to test this out, but I suspect @mhrivnak is right.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+        return workers
+    else:
+        return []
```
bmbouter: If we can't connect to the DB we probably shouldn't return []. This suggests that the db has no workers in it, but it may so we would be reporting slightly incorrect info in this case.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns:
+        bool: True if pulp can connect to the broker. False otherwise.
+    """
+    try:
+        conn = celery.connection()
+        conn.connect()
+        conn.release()
+        return True
+    except Exception:
+        _logger.exception('Connection to broker failed during status check!')
```
bmbouter: Can this be i8ln with gettext please?

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        version = {'pulpcore_version': status.get_version()}
+        broker_status = {'connected': status.get_broker_conn_status()}
+        db_status = {'connected': status.get_db_conn_status()}
+
+        try:
+            workers = status.get_workers()
+        except:
+            worker_status = None
+        else:
```
bmbouter: This looks really great @daviddavis

platform/pulpcore/app/views/status.py
```diff
@@ -0,0 +1,47 @@
...
+            context = {'request': request}
+            worker_status = [WorkerSerializer(worker, context=context).data for worker in workers]
```
asmacdo: I think you can use `multi=True` to avoid a list comprehension

platform/pulpcore/app/views/status.py
mhrivnak: Was there any additional discussion about this?

daviddavis: I talked to @asmacdo briefly on IRC and it sounded like he was ambivalent on using a serializer here. It would be consistent with other views, or viewsets rather, but it didn't sound like a Serializer would buy us much. I'm happy to create a Serializer but I don't see much benefit other than getting the chance to learn about Serializers.

platform/pulpcore/app/status.py
```diff
@@ -0,0 +1,65 @@
+"""
+Module to return status information about a running Pulp instance
```
jortel: Why the helper functions in a separate module?  Why not just implement in the StatusView?  I like the modularity but seems these functions should be protected (and perhaps static) methods on the StatusView class itself.

daviddavis: This module was copied over from server/pulp/server/managers/status.py, which separates the helper functions from the view (server/pulp/server/webservices/views/status.py). I have no problem with moving these functions into the view though.

bmbouter: +1 to moving them into the view. This code is dedicated to that view and not used otherwise.

daviddavis:
I added a Status serializer. I think all the items have been addressed.

platform/pulpcore/app/serializers/status.py
```diff
@@ -0,0 +1,27 @@
+from gettext import gettext as _
+from rest_framework import serializers
+
+from pulpcore.app.serializers.task import WorkerSerializer
+
+
+class StatusSerializer(serializers.Serializer):
+    """
+    Serializer for the status information of the app
```
mhrivnak: Can we be more specific for each of the fields about what it will return? Maybe the ``WorkerSerializer`` will take care of itself, since that's a known data structure that DRF may be able to auto-document as-is. That would be good to verify though.

The rest would benefit from more detail, so an integrator doesn't need to just try to get the value, see what is returned, hope they interpret it correctly, and hope it stays consistent. Ideally someone should be able to write tests based on the API docs, which I assume comes from this class.

mhrivnak: Not to go serializer-crazy, but maybe the right answer is to make sub-serializers for the other fields? You could have a VersionsSerializer for example. Maybe that's a silly idea, but I thought it's worth considering.

daviddavis:
@mhrivnak thanks for the review. I added sub-Serializers for the nested fields and changed "pulpcore_version" to just "pulpcore".

platform/pulpcore/app/serializers/status.py
```diff
@@ -0,0 +1,57 @@
+from gettext import gettext as _
+from rest_framework import serializers
+
+from pulpcore.app.serializers.task import WorkerSerializer
+
+
+class VersionsSerializer(serializers.Serializer):
```
mhrivnak: I was imagining there would be two fields: ``component`` and ``version``, or something like that. Then below in the ``StatusSerializer``, it would render this as a collection. I'm not sure how to do that exactly, but maybe ``many=True`` is enough? Then you can serialize any number of component versions into that list.

daviddavis: What about if we change the structure and instead of using a dictionary/hash like:

```json
{"pulpcore": "3.0.0"}
```

We could do something like this?

```json
[ {"component": "pulpcore", "version": "3.0.0"} ]
```

daviddavis:
@mhrivnak updated.

PR #3075, opened 2017-07-10
sync and publish should use the POST method in the REST API

mhrivnak:


platform/pulpcore/app/viewsets/repository.py
```diff
@@ -141,7 +141,7 @@ def destroy(self, request, repository_name, name):
         )
         return OperationPostponedResponse([async_result], request)
 
-    @decorators.detail_route()
+    @decorators.detail_route(methods=('post',))
```
asmacdo: Internally to drf (and in their examples) this is a list rather than a tuple. I'm sure this works, but I would use a list, safety first :) 

I won't block on this though. Change if you want.

mhrivnak: Heh, "safety first" is why I used an immutable tuple!

mhrivnak:
It appears that the docs builders are failing. Since I didn't change any docs, I'm going ahead with the merge despite that job not having run.

PR #3076, opened 2017-07-11
Add python-debpkgr as a dependency for pulp_deb

pcreech:
Add python-debpkgr as a dependency for pulp_deb

closes #[2803](https://pulp.plan.io/issues/2803)


deps/python-debpkgr/dist_list.txt
```diff
@@ -0,0 +1 @@
+el7 fc24 fc25
```
werwty: remove fc24 and add fc26

pcreech: fc24/fc26 will be updated when the build infrastructure gets moved to f26

PR #3077, opened 2017-07-11
Fix import in views lookup function

goosemania:


platform/pulpcore/app/serializers/base.py
```diff
@@ -61,7 +61,7 @@ def view_name_for_model(model_obj, view_action):
         LookupError: if no ViewSet is found for the Model
     """
     # Import this here to prevent out-of-order plugin discovery
-    from urls import root_router, nested_routers
+    from ..urls import root_router, nested_routers
```
dralley: Could you use an absolute import e.g.

```from pulpcore.app.urls import root_router, nested_routers```

PR #3078, opened 2017-07-11
Fix lookup fields for some viewsets

goosemania:


platform/pulpcore/app/viewsets/base.py
```diff
@@ -6,7 +6,8 @@
 
 class NamedModelViewSet(viewsets.ModelViewSet):
     """
-    A customized ModelViewSet that understands how to register itself with the Pulp API router.
+    A customized named ModelViewSet that understands how to register itself with the Pulp API router.
+    This viewset is discoverable by its name.
```
asmacdo: (Non-blocking comment) Would you mind adding this to the next paragraph? A small detail of google-style is that the first section is a summary limited to 1 line. http://docs.pulpproject.org/en/3.0/nightly/contributing/style_guide.html#example-docstring

goosemania: Sure, will do.

PR #3080, opened 2017-07-12
Problem: Artifacts can't be created via REST API

dkliban:
Solution: add a viewset for Artifacts

The REST API user can now POST to /api/v3/artifacts endpoint with a file,
size, and any number of checksums. During the upload of the file, each
possible checksum is calculated. If user provides size or any checksums,
they are verified after the upload finishes. If one of the values does not
match a ValidationError is raised.

The user can also list all the the Artifacts by performing a GET request
to /api/v3/artifacts. An Artifact can be deleted by making a DELETE
request to /api/v3/artifacts/<pk>.

The user can not update Artifacts via REST API after one is created.

The 'django.core.files.uploadhandler.MemoryFileUploadHandler' is disabled so
all uploads are handled by 'pulpcore.app.upload.PulpFileUploadHandler' and
are written to '/var/lib/pulp/tmp' as the file is being uploaded.

closes #2843
https://pulp.plan.io/issues/2843

### DIFF
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -1,12 +1,10 @@
 """
 Content related Django models.
 """
-import hashlib
-
+from django.core.files.storage import default_storage
 from django.db import models
 
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation
-from pulpcore.app.models.storage import ArtifactLocation
 
 
 class Content(MasterModel):
@@ -39,35 +37,6 @@ class Content(MasterModel):
         """
         return tuple(getattr(self, f.name) for f in self.natural_key_fields)
 
-    def natural_key_digest(self):
-        """
-        Get the SHA-256 digest of the natural key.
-        The digest is updated with each field name followed by its value.
-        The field names are only necessary to preserve backward compatibility
-        with digests generated in pulp 2.
-
-        :return: The hex digest.
-        :rtype: str
-        """
-        h = hashlib.sha256()
-        for name in sorted(f.name for f in self.natural_key_fields):
-            h.update(name.encode(encoding='utf-8'))
-            value = getattr(self, name)
-            if isinstance(value, str):
-                h.update(value.encode(encoding='utf-8'))
-            else:
-                h.update(value)
-        return h.hexdigest()
-
-    def __str__(self):
-        cast = self.cast()
-
-        # so here's a pretty good case for making namedtuples for content types
-        # the third replacement string results in "natural_key_fieldname_0=repr(field_value), ..."
-        natural_key_zip = zip(cast.natural_key_fields, cast.natural_key())
-        natural_key_string = ', '.join(('='.join((t[0].name, repr(t[1]))) for t in natural_key_zip))
-        return '<{} (type={}): {}>'.format(self._meta.object_name, cast.TYPE, natural_key_string)
-
 
 class Artifact(Model):
     """
@@ -76,13 +45,6 @@ class Artifact(Model):
     Fields:
 
         file (models.FileField): The stored file.
-        downloaded (models.BooleanField): The associated file has been successfully downloaded.
-        requested (models.BooleanField): The associated file has been requested by a client at
-            least once.
-        relative_path (models.TextField): The artifact's path relative to the associated
-            :class:`Content`. This path is incorporated in the absolute storage path of the file
-            and its published path relative to the root publishing directory. At a minimum the path
-            will contain the file name but may also include sub-directories.
         size (models.IntegerField): The size of the file in bytes.
         md5 (models.CharField): The MD5 checksum of the file.
         sha1 (models.CharField): The SHA-1 checksum of the file.
@@ -90,25 +52,22 @@ class Artifact(Model):
         sha256 (models.CharField): The SHA-256 checksum of the file.
         sha384 (models.CharField): The SHA-384 checksum of the file.
         sha512 (models.CharField): The SHA-512 checksum of the file.
-
-    Relations:
-
-        content (models.ForeignKey): The associated content.
     """
+    def storage_path(self, name):
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
 
-    file = models.FileField(db_index=True, upload_to=ArtifactLocation(), max_length=255)
-    downloaded = models.BooleanField(db_index=True, default=False)
-    requested = models.BooleanField(db_index=True, default=False)
-    relative_path = models.TextField(db_index=True, blank=False, default=None)
-    size = models.IntegerField(blank=True, null=True)
-    md5 = models.CharField(max_length=32, blank=True, null=True)
-    sha1 = models.CharField(max_length=40, blank=True, null=True)
-    sha224 = models.CharField(max_length=56, blank=True, null=True)
-    sha256 = models.CharField(max_length=64, blank=True, null=True)
-    sha384 = models.CharField(max_length=96, blank=True, null=True)
-    sha512 = models.CharField(max_length=128, blank=True, null=True)
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                sha256 checksum is used to determine a file path instead.
+        """
+        return default_storage.get_artifact_path(self.sha256)
 
-    content = models.ForeignKey(Content, related_name='artifacts', on_delete=models.CASCADE)
-
-    class Meta:
-        unique_together = ('content', 'relative_path')
+    file = models.FileField(db_index=True, upload_to=storage_path, max_length=255)
+    size = models.IntegerField(blank=False, null=False)
+    md5 = models.CharField(max_length=32, blank=False, null=False, unique=False, db_index=True)
+    sha1 = models.CharField(max_length=40, blank=False, null=False, unique=False, db_index=True)
+    sha224 = models.CharField(max_length=56, blank=False, null=False, unique=False, db_index=True)
+    sha256 = models.CharField(max_length=64, blank=False, null=False, unique=True, db_index=True)
+    sha384 = models.CharField(max_length=96, blank=False, null=False, unique=True, db_index=True)
+    sha512 = models.CharField(max_length=128, blank=False, null=False, unique=True, db_index=True)
--- a/platform/pulpcore/app/serializers/content.py
+++ b/platform/pulpcore/app/serializers/content.py
@@ -1,9 +1,14 @@
 from gettext import gettext as _
+import hashlib
 
 from rest_framework import serializers
+from rest_framework.validators import UniqueValidator
 
 from pulpcore.app import models
 from pulpcore.app.serializers import base, fields, generic
+
+
+UNIQUE_ALGORITHMS = ['sha256', 'sha384', 'sha512']
 
 
 class ContentSerializer(base.MasterModelSerializer):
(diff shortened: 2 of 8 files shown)


platform/pulpcore/app/models/content.py
```diff
@@ -79,27 +79,21 @@ class Artifact(Model):
...
         sha384 (models.CharField): The SHA-384 checksum of the file.
         sha512 (models.CharField): The SHA-512 checksum of the file.
-
-    Relations:
-
-        content (models.ForeignKey): The associated content.
     """
+    def storage_path(self, name):
+        digest = self.sha256
+        return "%s/artifacts/%s/%s" % (settings.MEDIA_ROOT, digest[0:2], digest[2:])
```
jortel: Suggest `os.path.join()` would be better.
Also, directories are implicitly plural so `artifact` would be better and more consistent with linux directory naming conventions.

jortel: Also, as we discussed, the storage path should be calculated by something coupled to the actual storage mechanism.  The current implementation makes an attempt at this by having `ArtifactLocation` defined in the _storage_ module but falls a little short of the goal.  I suggest adding a method to the `FileStorage` that can be used to get a _deconstructable_ object that can calculate all paths appropriately for that storage mechanism.   Then in `Artifact.storage_path()` instantiate `DefaultStorage`, get the path calculator and use it to calculate and return the storage path.  We can do the same for the certificate storage in another PR.

platform/pulpcore/app/serializers/content.py
```diff
@@ -38,55 +43,71 @@ class ArtifactSerializer(base.ModelSerializer):
...
+        Args:
+            data (dict): the id of the importer
+
+
+        Raises:
+            :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
+                due to validation error. This theoretically should never occur since validation is
+                performed before the task is dispatched.
+        :param data:
+        :return:
```
jortel: `:param_data:` and `:return:` probably left over from cut-n-paste.

platform/pulpcore/app/serializers/content.py
```diff
@@ -38,55 +43,71 @@ class ArtifactSerializer(base.ModelSerializer):
...
         required=False
     )
 
+    def validate(self, data):
+        """
+        Validate file by size and all checksums which values are provided.
+
+        Args:
+            data (dict): the id of the importer
+
```
jortel: Extra spaces here.

platform/pulpcore/app/upload.py
```diff
@@ -0,0 +1,33 @@
...
+class PulpTemporaryUploadedFile(TemporaryUploadedFile):
+    """
+    A file uploaded to a temporary location in Pulp.
+    """
+
+    def __init__(self, name, content_type, size, charset, content_type_extra=None):
+        self.hashers = {}
+        for hasher in hashlib.algorithms_guaranteed:
+            self.hashers[hasher] = getattr(hashlib, hasher)()
+        super(PulpTemporaryUploadedFile, self).__init__(name, content_type, size, charset, content_type_extra)
```
jortel: Line length.

platform/pulpcore/app/upload.py
```diff
@@ -0,0 +1,33 @@
...
+class PulpFileUploadHandler(TemporaryFileUploadHandler):
+    """
+    Upload handler that streams data into a temporary file.
+    """
+    def new_file(self, file_name, *args, **kwargs):
+        """
+        Create the file object to append to as data is coming in.
+        """
+        super(PulpFileUploadHandler, self).new_file(file_name, *args, **kwargs)
+        self.file = PulpTemporaryUploadedFile(self.file_name, self.content_type, 0, self.charset, self.content_type_extra)
```
jortel: Line length.

platform/pulpcore/app/models/content.py
mhrivnak: Good points. I was generally thinking that this type of logic is a good fit for being handled within the domain of the FileField object, and I think that lines up with your specific suggestions.

platform/pulpcore/app/serializers/content.py
```diff
@@ -23,9 +24,13 @@ class Meta:
...
 
 class ArtifactSerializer(base.ModelSerializer):
+    _href = serializers.HyperlinkedIdentityField(
+        view_name='artifacts-detail',
+    )
+
     file = serializers.FileField(
         help_text=_("The stored file."),
-        read_only=True
+        required=True
```
mhrivnak: What does this actually mean for a ``FileField``? That bits must be present? Note that for on-demand content, we will need an ``Artifact`` record so that we can publish it and also have ``Catalog`` entries, but the file may not yet be present.

platform/pulpcore/app/serializers/content.py
```diff
@@ -38,55 +43,71 @@ class ArtifactSerializer(base.ModelSerializer):
...
-        read_only=True,
         required=False
     )
 
+    def validate(self, data):
+        """
+        Validate file by size and all checksums which values are provided.
+
+        Args:
+            data (dict): the id of the importer
```
mhrivnak: copy-and-paste gone wrong?

platform/pulpcore/app/upload.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    def __init__(self, name, content_type, size, charset, content_type_extra=None):
+        self.hashers = {}
+        for hasher in hashlib.algorithms_guaranteed:
+            self.hashers[hasher] = getattr(hashlib, hasher)()
+        super(PulpTemporaryUploadedFile, self).__init__(name, content_type, size, charset, content_type_extra)
+
+
+class PulpFileUploadHandler(TemporaryFileUploadHandler):
```
mhrivnak: To be more descriptive, consider something like ``HashingFileUploadHandler`` as the name.

platform/pulpcore/app/upload.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Upload handler that streams data into a temporary file.
+    """
+    def new_file(self, file_name, *args, **kwargs):
+        """
+        Create the file object to append to as data is coming in.
+        """
+        super(PulpFileUploadHandler, self).new_file(file_name, *args, **kwargs)
```
mhrivnak: In an act of mercy, Python 3 no longer requires us to pass stuff into ``super()``. :smile:

platform/pulpcore/app/serializers/content.py
dkliban: Good point. I am making the file not required.

platform/pulpcore/app/upload.py
dkliban: that's a better name for sure. thanks!

platform/pulpcore/app/models/content.py
```diff
@@ -76,39 +76,30 @@ class Artifact(Model):
...
-    sha512 = models.CharField(max_length=128, blank=True, null=True)
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                sha256 checksum is used to determine a file path instead.
+        """
+        return default_storage.get_artifact_path(self.sha256)
 
-    content = models.ForeignKey(Content, related_name='artifacts', on_delete=models.CASCADE)
+    file = models.FileField(db_index=True, upload_to=storage_path, max_length=255)
+    size = models.IntegerField(blank=True, null=False)
```
mhrivnak: If the file itself is required, why not require this as well?

jortel:
LGTM.  Suggest removing: `Content.natural_key_digest()`.

platform/pulpcore/app/upload.py
```diff
@@ -0,0 +1,34 @@
...
+
+class HashingFileUploadHandler(TemporaryFileUploadHandler):
+    """
+    Upload handler that streams data into a temporary file.
+    """
+    def new_file(self, file_name, *args, **kwargs):
+        """
+        Create the file object to append to as data is coming in.
+        """
+        super().new_file(file_name, *args, **kwargs)
```
mhrivnak: This will create a file whose reference I think will be overwritten by the next line. I also think we don't need or want that file at all. Is there anything that calling the superclass ``new_file()`` gets us that we need? Or could that be left out?

https://docs.djangoproject.com/en/1.11/_modules/django/core/files/uploadhandler/#TemporaryFileUploadHandler

jortel:
It seems reasonable to expect that the content (unit) has all metadata needed to publish.  Both `size` and `digest` (sha256) are properties obtained from the (remote) manifest.  So, adding both to `FileContent` seems reasonable.  When publishing with other digest(s) (like in the RPM checksum case), the publisher already requires the file be downloaded and Artifact created in the DB so that the additional digest is available.

platform/pulpcore/app/upload.py
dkliban: we get this: https://docs.djangoproject.com/en/1.11/_modules/django/core/files/uploadhandler/#FileUploadHandler.new_file

jortel:
Although, after thinking through this further -

Storing artifact related information (fields) on the Content (unit) such as size and digest seems wrong.  Instead, I think we should explore keeping the artifact as required on `ContentArtifact` and create them even when the file is not downloaded.  Also, keep `Artifact.downloaded`.  This puts things back to _What is the natural key for an artifact?_   The natural key is used to look up an artifact when it already exists to support:

- de-duplication
- download an artifact (file) via API.

I suggest the digest fields are the natural key.  An artifact could be located by any one of it's _strong_ digest fields.

Artifacts created with upload (or sync with immediate download policy) will always have at least (1) digest because the file exists and they can all be calculated.  no problem there.

Artifacts created during sync (when downloading is deferred) of content types supported today probably all would have a digest because a digest is included in the metadata supplied by the remote repository.

So, what about edge cases where downloading is deferred and the remote repository does not provide a digest.  Perhaps in that case, Artifacts are created without any digest fields populated.  The artifact cannot be fetched using a natural key and duplicates _could_ be created.  Maybe that is not so bad.  As files are downloaded and digests calculated and set, pulp could perform lazy de-duplication.

Anyway.  I think we should discuss this further.

dkliban:
Artifacts are provided by the core as a way for plugins to store files in Pulp. If a plugin doesn't have  a file to store it doesn't need to create an Artifact.  

If a plugin needs to be able to publish content for which there are no artifacts, it is the plugin's responsibility to store all information required for publishing.

dkliban:
@mhrivnak @jortel In the on-demand scenario where will we store the URL for the remote artifact that the streamer needs to download? The same table in the database should probably store the expected size and checksums.

platform/pulpcore/app/models/content.py
```diff
@@ -76,39 +45,30 @@ class Artifact(Model):
...
-            :class:`Content`. This path is incorporated in the absolute storage path of the file
-            and its published path relative to the root publishing directory. At a minimum the path
-            will contain the file name but may also include sub-directories.
         size (models.IntegerField): The size of the file in bytes.
         md5 (models.CharField): The MD5 checksum of the file.
         sha1 (models.CharField): The SHA-1 checksum of the file.
         sha224 (models.CharField): The SHA-224 checksum of the file.
         sha256 (models.CharField): The SHA-256 checksum of the file.
         sha384 (models.CharField): The SHA-384 checksum of the file.
         sha512 (models.CharField): The SHA-512 checksum of the file.
```
jortel: The _digest_ fields need indexes with sha256 and stronger being unique.

I'm assuming that (eventually) `Artifact` and `RemoteArtifact` will have a common abstract base model and all (3) models defined here, correct?  If so, let refactor that here now since you know you'll be adding the `RemoteArtifact` eventually.

jortel:
Adding another LGTM but leaving approval to someone that knows more about serializer and viewsets.

platform/pulpcore/app/viewsets/content.py
```diff
@@ -20,8 +23,34 @@ class Meta:
...
 class ContentViewSet(NamedModelViewSet):
     endpoint_name = 'content'
     queryset = Content.objects.all()
     serializer_class = ContentSerializer
     filter_class = ContentFilter
+
+
+class ArtifactViewSet(NamedModelViewSet):
+    endpoint_name = 'artifacts'
+    lookup_field = 'pk'
```
goosemania: This `lookup_field` value is a default for `NamedModelViewset`, so it's not needed.
[Docstring for `NamedModelViewset`](https://github.com/pulp/pulp/blob/3.0-dev/platform/pulpcore/app/viewsets/base.py#L16) can be fixed though since it says about the old, no longer valid, default value. My bad.

platform/pulpcore/app/models/content.py
mhrivnak: We may have different validation rules, and I don't know if that might prevent us from having an ABC. For example, on an ``Artifact`` all fields will be required.

mhrivnak:
I didn't re-review, but I'm happy with the changes I previously requested. If there's anything specific you'd like me to look at, let me know, but otherwise I'm happy to leave it to the others.

PR #3082, opened 2017-07-13
Adding pep8speaks config file

daviddavis:
refs #2870
https://pulp.plan.io/issues/2870

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,10 @@
+pycodestyle:
+    max-line-length: 100  # Default is 79 in PEP8
+    ignore:  # Errors and warnings to ignore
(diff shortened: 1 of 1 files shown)


daviddavis:
@bmbouter good point. I'm also wondering if we should open a PR against the latest release branch (2.13-dev) too?

bmbouter:
Yeah do you want to rebase back to 2.13-dev and merge forward to master to introduce in both of those places (and then cherry pick to 3.0-dev) ?

daviddavis:
Yea. I opened [a separate PR for 3.0-dev](https://github.com/pulp/pulp/pull/3083) since the config is slightly different.

Will merge forward this change after we merge it.

PR #3091, opened 2017-07-14
Problem: Changeset is not aware of new content related models

dkliban:
Solution: Update changesets to use Artifact, DeferredArtifact, and ContentArtifact

### DIFF
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -1,8 +1,10 @@
 """
 Content related Django models.
 """
+from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import models
+
 
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation
 
@@ -90,6 +92,33 @@ class ContentArtifact(Model):
         """
         return default_storage.get_artifact_path(self.sha256)
 
-    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE)
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE, null=True)
     content = models.ForeignKey(Content, on_delete=models.CASCADE)
     relative_path = models.CharField(max_length=64)
+
+
+class DeferredArtifact(Model):
+    """
+    A URL for fetching a file associated with a piece of content.
+
+    Fields:
+
+        url (models.TextField): The URL where the artifact can be retrieved.
+        size (models.IntegerField): The expected size of the file in bytes.
+        md5 (models.CharField): The expected MD5 checksum of the file.
+        sha1 (models.CharField): The expected SHA-1 checksum of the file.
+        sha224 (models.CharField): The expected SHA-224 checksum of the file.
+        sha256 (models.CharField): The expected SHA-256 checksum of the file.
+        sha384 (models.CharField): The expected SHA-384 checksum of the file.
+        sha512 (models.CharField): The expected SHA-512 checksum of the file.
+    """
+    url = models.TextField(blank=True, validators=[validators.URLValidator])
+    size = models.IntegerField(blank=True, null=True)
+    md5 = models.CharField(max_length=32, blank=True, null=True)
+    sha1 = models.CharField(max_length=40, blank=True, null=True)
+    sha224 = models.CharField(max_length=56, blank=True, null=True)
+    sha256 = models.CharField(max_length=64, blank=True, null=True)
+    sha384 = models.CharField(max_length=96, blank=True, null=True)
+    sha512 = models.CharField(max_length=128, blank=True, null=True)
+    content_artifact = models.ForeignKey(ContentArtifact, on_delete=models.CASCADE)
+    importer = models.ForeignKey('Importer', on_delete=models.CASCADE)
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -7,37 +7,6 @@ from django.db import models
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation
 
 
-class Content(MasterModel):
-    """
-    A piece of managed content.
-
-    Attributes:
-
-        natural_key_fields (tuple): Natural key fields.  Must be models.Field subclasses.
-
-    Relations:
-
-        notes (GenericKeyValueRelation): Arbitrary information stored with the content.
-    """
-    TYPE = 'content'
-
-    natural_key_fields = ()
-
-    notes = GenericKeyValueRelation(Notes)
-
-    class Meta:
-        verbose_name_plural = 'content'
-
-    def natural_key(self):
-        """
-        Get the model's natural key based on natural_key_fields.
-
-        :return: The natural key.
-        :rtype: tuple
-        """
-        return tuple(getattr(self, f.name) for f in self.natural_key_fields)
-
-
 class Artifact(Model):
     """
     A file associated with a piece of content.
@@ -71,3 +40,56 @@ class Artifact(Model):
     sha256 = models.CharField(max_length=64, blank=False, null=False, unique=True, db_index=True)
     sha384 = models.CharField(max_length=96, blank=False, null=False, unique=True, db_index=True)
     sha512 = models.CharField(max_length=128, blank=False, null=False, unique=True, db_index=True)
+
+
+class Content(MasterModel):
+    """
+    A piece of managed content.
+
+    Attributes:
+
+        natural_key_fields (tuple): Natural key fields.  Must be models.Field subclasses.
+
+    Relations:
+
+        notes (GenericKeyValueRelation): Arbitrary information stored with the content.
+    """
+    TYPE = 'content'
+
+    natural_key_fields = ()
+
+    notes = GenericKeyValueRelation(Notes)
+    artifacts = models.ManyToManyField(Artifact, through='ContentArtifact')
+
+    class Meta:
+        verbose_name_plural = 'content'
+
+    def natural_key(self):
+        """
+        Get the model's natural key based on natural_key_fields.
+
+        :return: The natural key.
+        :rtype: tuple
+        """
+        return tuple(getattr(self, f.name) for f in self.natural_key_fields)
+
+
+class ContentArtifact(Model):
+    """
+    A relationship between a Content and an Artifact.
+
+    Serves as a through model for the 'artifacts' ManyToManyField in Content.
+    """
+    def storage_path(self, name):
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                sha256 checksum is used to determine a file path instead.
+        """
+        return default_storage.get_artifact_path(self.sha256)
+
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE)
+    content = models.ForeignKey(Content, on_delete=models.CASCADE)
+    relative_path = models.CharField(max_length=64)
--- a/platform/pulpcore/app/viewsets/content.py
+++ b/platform/pulpcore/app/viewsets/content.py
@@ -1,9 +1,10 @@
+from django.db import transaction
 from django_filters.rest_framework import filterset
 from rest_framework import status
 from rest_framework.response import Response
 
 
-from pulpcore.app.models import Artifact, Content
+from pulpcore.app.models import Artifact, Content, ContentArtifact
 from pulpcore.app.serializers import ArtifactSerializer, ContentSerializer
 from pulpcore.app.viewsets import NamedModelViewSet
 
@@ -43,6 +44,28 @@ class ContentViewSet(NamedModelViewSet):
     queryset = Content.objects.all()
     serializer_class = ContentSerializer
     filter_class = ContentFilter
+
+    @transaction.atomic
+    def create(self, request):
+        """
+        Create a new Content instance from
+        :param request:
+        :return:
+        """
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        artifacts = serializer.validated_data.pop('artifacts')
+        content = serializer.save()
+
+        for artifact, relative_path in artifacts.items():
+            ca = ContentArtifact(artifact=artifact, content=content, relative_path=relative_path)
+            ca.save()
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def update(self, request, pk):
+        return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
 
 
 class ArtifactViewSet(NamedModelViewSet):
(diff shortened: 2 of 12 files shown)


platform/pulpcore/app/models/content.py
```diff
@@ -69,46 +104,13 @@ def __str__(self):
         return '<{} (type={}): {}>'.format(self._meta.object_name, cast.TYPE, natural_key_string)
 
 
-class Artifact(Model):
+class ContentArtifact(Model):
```
jortel: Need a unique index on (content, artifact, relative_path).

platform/pulpcore/app/models/content.py
```diff
@@ -3,10 +3,44 @@
...
+    file = models.FileField(db_index=True, upload_to=storage_path, max_length=255)
+    downloaded = models.BooleanField(db_index=True, default=False)
+    requested = models.BooleanField(db_index=True, default=False)
+    size = models.IntegerField(blank=True, null=True)
+    md5 = models.CharField(max_length=32, blank=True, null=True)
+    sha1 = models.CharField(max_length=40, blank=True, null=True)
+    sha224 = models.CharField(max_length=56, blank=True, null=True)
+    sha256 = models.CharField(max_length=64, blank=True, null=True)
+    sha384 = models.CharField(max_length=96, blank=True, null=True)
+    sha512 = models.CharField(max_length=128, blank=True, null=True)
```
jortel: I assume an artifact cannot be created without at least 1 digest, correct?
And, the query to find an existing artifact is by matching one of the digests?

like:
```
where md5 = <md5> or,
      sha1 = <sha1> or,
      sha224 = <sha224> or,
      sha256 = <sha256> or,
      ...
```
only for non-null digests.

If so, won't we need indexes on all of the digests?

jortel:
The FK in `DownloadCatalog` to Artifact will need to be changed to ContentArtifact.

platform/pulpcore/app/models/content.py
```diff
@@ -63,11 +92,33 @@ def storage_path(self, name):
...
-    sha224 = models.CharField(max_length=56, blank=False, null=False, unique=False, db_index=True)
-    sha256 = models.CharField(max_length=64, blank=False, null=False, unique=True, db_index=True)
-    sha384 = models.CharField(max_length=96, blank=False, null=False, unique=True, db_index=True)
-    sha512 = models.CharField(max_length=128, blank=False, null=False, unique=True, db_index=True)
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE, null=True)
+    content = models.ForeignKey(Content, on_delete=models.CASCADE)
+    relative_path = models.CharField(max_length=64)
+
+
+class DeferredArtifact(Model):
```
jortel: Suggest another name.  Perhaps `PendingArtifact` might be better since they are created even when downloading is not deferred.

platform/pulpcore/app/models/content.py
```diff
@@ -1,12 +1,49 @@
...
+    """
+    def storage_path(self, name):
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                sha256 checksum is used to determine a file path instead.
+        """
+        return default_storage.get_artifact_path(self.sha256)
```
jortel: It has been my experience that for this to work, the Artifact must be a `deconstructable`.  Otherwise, it just fails silently.  Did you validate the file is actually written to the storage location?

Also, I thought we agreed a cleaner approach would be to add a path calculator to the storage module and add as an attribute to the `FileStorage`.  See:  https://github.com/pulp/pulp/pull/3100.

platform/pulpcore/download/writer.py
```diff
@@ -92,6 +93,9 @@ def __init__(self, path):
         """
         super(FileWriter, self).__init__()
         self.path = path
+        self.hashers = {}
+        for hasher in hashlib.algorithms_guaranteed:
+            self.hashers[hasher] = getattr(hashlib, hasher)()
```
jortel: For better separation of concerns, a `Writer` should not be generating digests.  This is also redundant with digests being calculated by `Validation`.  There are cleaner ways to hook this in.  For example: adding a buffer event listener to calculate all the digests to be used when creating an `Artifact`.

platform/pulpcore/app/viewsets/content.py
```diff
@@ -44,6 +45,28 @@ class ContentViewSet(NamedModelViewSet):
     serializer_class = ContentSerializer
     filter_class = ContentFilter
 
+    @transaction.atomic
+    def create(self, request):
```
jortel: I haven't seen this kind of _helper_ method on other models.  So far, we've made a deliberate effort to keep the Pulp domain model and the data model separated.  It seems like this functionality belongs somewhere else.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -166,13 +169,21 @@ def _iter(self):
         Yields:
             Download: The appropriate download object.
         """
-        for artifact in ArtifactIterator(self.content):
-            if self.deferred or artifact.model.downloaded:
+        for remote_artifact in ArtifactIterator(self.content):
```
jortel: The variable named `artifact` refers to a `RemoteArtifact` consistently in the code base, specifically naming this `remote_artifact` introduces confusion.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -199,15 +210,15 @@ def _iter(self):
         does not have any un-downloaded artifacts, A NopArtifact is yielded.
 
         Yields:
-            RemoteArtifact: The flattened artifacts.
+            DelayedArtifact: The flattened artifacts.
```
jortel: I don't think you mean `DelayedArtifact` here.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -131,13 +135,42 @@ def save(self):
...
+                for remote_artifact in self.artifacts:
+                    if not remote_artifact.content_artifact.artifact_id:
+                        if not deferred:
+                            artifact = self.validate_and_create_artifact(remote_artifact.model,
+                                                                         remote_artifact.download)
+                            remote_artifact.content_artifact.artifact = artifact
+                        remote_artifact.model.save()
+                        remote_artifact.content_artifact.save()
+
+    def validate_and_create_artifact(self, remote_artifact, download):
```
jortel: Each download is configured with digest validation.  If the validation fails, the download fails.  We don't need/want to do it here as well.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -147,41 +180,47 @@ class RemoteArtifact(Remote):
...
-        >>> from pulpcore.plugin.models import Artifact
+        >>> from pulpcore.plugin.models import ContentArtifact, DeferredArtifact
         >>>
-        >>> model = Artifact(...)  # DB model instance.
+        >>> model = DeferredArtifact(...)
+        >>> content_artifact = ContentArtifact(...)
         >>> download = ...
         >>> ...
-        >>> artifact = RemoteArtifact(model, download)
+        >>> artifact = RemoteArtifact(model, content_artifact, download)
```
jortel: Seems like we can just add relative_path to `RemoteArtifact` and simplify the interface by not having the plugin writer deal with it specifically.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -147,41 +180,47 @@ class RemoteArtifact(Remote):
...
         >>>
     """
 
     __slots__ = (
         'content',
+        'content_artifact',
```
jortel: Let's make this `relative_path` and manage the `ContentArtifact` for the plugin writer.

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -106,8 +108,9 @@ def _https(self, url, path=None, artifact=None):
         Returns:
             HttpDownload: An https download.
         """
-        if artifact:
-            _path = artifact.relative_path
+        if remote_artifact:
+            import os
+            _path = os.path.join("/var/lib/pulp/tmp", remote_artifact.sha256)
```
jortel: Why are we doing this?  We should not be hard coding paths.  the downloader should be configured to download into the working_dir which is the CWD.

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -118,7 +121,6 @@ def _https(self, url, path=None, artifact=None):
         download.user.name = self.importer.basic_auth_user
         download.user.password = self.importer.basic_auth_password
         download.proxy_url = self.importer.proxy_url
-        self._add_validation(download, artifact)
```
jortel: Why is the call to add validation removed?

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -166,13 +169,21 @@ def _iter(self):
...
             Download: The appropriate download object.
         """
-        for artifact in ArtifactIterator(self.content):
-            if self.deferred or artifact.model.downloaded:
+        for remote_artifact in ArtifactIterator(self.content):
+            q = Q()
+            for algorithm in hashlib.algorithms_guaranteed:
+                if getattr(remote_artifact.model, algorithm):
+                    kwargs = {algorithm: getattr(remote_artifact.model, algorithm)}
+                    q |= Q(**kwargs)
```
jortel: Django already has a convention for getting the natural key for a model.  The `natural_key()` method normally returns a tuple which is obviously not sufficient for this.  But, I think we should follow the same pattern and have a method on `RemoteArtifact` that can be used to fetch it from the DB.  Instead of here.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -166,13 +169,21 @@ def _iter(self):
...
+            try:
+                artifact = Artifact.objects.get(q)
```
jortel: This query needs to be done in batches.  For performance, we cannot be hitting the DB for every artifact.  Other code in the _changeset_ package does this.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -166,13 +169,21 @@ def _iter(self):
...
+            except Artifact.DoesNotExist:
+                artifact = None
+            if artifact or self.deferred:
                 download = NopDownload()
-                download.attachment = artifact
-                artifact.path = None
+                download.attachment = remote_artifact
```
jortel: Setting the `Artifact.path` is how it's know that a file was actually downloaded for the artifact.  This should not change.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -166,13 +169,21 @@ def _iter(self):
...
             else:
-                download = artifact.download
```
jortel: As mentioned earlier.  The convention in the package is that:

`content` is `RemoteContent`
and
`artifact` is `RemoteArtifact`

And that `model` was used for DB models.  If you want to change this convention, you need to do a mass rename.  Though it seems outside the scope of this work.

plugin/pulpcore/plugin/changeset/iterator.py
jortel: This should be `RemoteArtifact`.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -147,41 +180,47 @@ class RemoteArtifact(Remote):
...
         'download',
         'path'
     )
 
-    def __init__(self, model, download):
+    def __init__(self, model, content_artifact, download):
```
jortel: content_artifact removed.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -91,17 +95,17 @@ def update(self, model):
             model (pulpcore.plugin.Content): A fetched content model object.
         """
         self.model = model
-        known = {a.model.relative_path: a for a in self.artifacts}
+        known = {a.content_artifact.relative_path: a for a in self.artifacts}
         self.artifacts.clear()
-        for artifact in model.artifacts.all():
+        for content_artifact in ContentArtifact.objects.filter(content=model):
```
jortel: Unfortunate the django does not support query through the explicit thru table :/

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -131,13 +135,42 @@ def save(self):
...
+            data['size'] = remote_artifact.size
+        for hasher in hashlib.algorithms_guaranteed:
+            if getattr(remote_artifact, hasher, None):
+                data[hasher] = getattr(remote_artifact, hasher, None)
+
+        serializer_class = get_plugin_config("pulp_app").named_serializers["ArtifactSerializer"]
+        serializer = serializer_class(data=data)
+        serializer.validate(data=data)
+        serializer.is_valid(raise_exception=True)
+        return serializer.save()
```
jortel: Seems this logic should be moved to the `RemoteArtifact.save()`.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -131,13 +135,42 @@ def save(self):
...
+                    if not remote_artifact.content_artifact.artifact_id:
+                        if not deferred:
+                            artifact = self.validate_and_create_artifact(remote_artifact.model,
+                                                                         remote_artifact.download)
+                            remote_artifact.content_artifact.artifact = artifact
+                        remote_artifact.model.save()
+                        remote_artifact.content_artifact.save()
+
+    def validate_and_create_artifact(self, remote_artifact, download):
+        """
```
jortel: I don't know what decisions were made with the epic _validation_ discussion on pulp-dev but seem wrong for platform _core_ components to be using the serializer to do validation and creation.  I don't think we do this anywhere else.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -131,13 +135,42 @@ def save(self):
...
                 is_duplicate = True
             else:
-                for artifact in self.artifacts:
-                    artifact.model.content = self.model
-                    artifact.save()
-        if is_duplicate:
-            model = type(self.model)
-            content = model.objects.get(**self.key)
-            self.update(content)
+
```
jortel: extra blank line.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -131,13 +135,42 @@ def save(self):
...
+                for remote_artifact in self.artifacts:
+                    if not remote_artifact.content_artifact.artifact_id:
+                        if not deferred:
```
jortel: The decision to create the artifact should be based on decisions made by the `ArtifactIterator` already.  When the `RemoteArtifact.path` is set, the file has been downloaded and real `Artifact` should be created.

plugin/pulpcore/plugin/changeset/model.py
jortel: What about race conditions whereby the artifact does not exist when the artifact iterator checks but another thread has created it before this code tries to insert it.  The dup needs to be handled.

mhrivnak:
It sounds like this is getting replaced by this new work: https://pulp.plan.io/issues/2942

To avoid confusion (which confused me just now), I'm closing this PR. Obviously please re-open if I'm misinterpreting.

PR #2833, opened 2016-11-14
Add Publisher filter class

goosemania:
closes #2310
https://pulp.plan.io/issues/2310

PR #3097, opened 2017-07-19
Get location of server.yaml from environmental variable

werwty:
closes #2833
https://pulp.plan.io/issues/2833


werwty:
devel change that goes with this: https://github.com/pulp/devel/pull/72

PR #3098, opened 2017-07-19
Delegation replaced with event model.

jortel:
The event model is more flexible than delegation.  Pulp will likely have multiple 3rd parties interested in participating in a download.  Each may be interested in the same or different aspects (events).

Also added an event to support progress reporting.


platform/pulpcore/download/single.py
```diff
@@ -173,36 +201,31 @@ def _validate(self):
...
         The download has failed.
         All efforts to repair and retry have failed.
         """
-        pass
+        self.router.send(Failed(self))
 
-    @delegate
     def _on_error(self, error):
         """
         The download has encountered an error.
```
asmacdo: No longer a delegate.

asmacdo: The gh interface wouldnt let me comment on right line, but the next line mentions a delegate.

jortel: Fixed.

PR #3100, opened 2017-07-21
Add publication models.

jortel:
https://pulp.plan.io/issues/2893
https://pulp.plan.io/issues/2895

Add publication models and test using file plugin.

This PR currently includes (2) files for the content app.  May split out into separate PR before merging.

Testing requires creating a `Distribution` in the DB with `psql`.

### DIFF
--- a/platform/pulpcore/app/models/publication.py
+++ b/platform/pulpcore/app/models/publication.py
@@ -1,0 +1,102 @@
+from django.db import models
+from django.core.files.storage import DefaultStorage
+
+from pulpcore.app.models import Model
+
+
+class Publication(Model):
+    """
+    Fields:
+        created (models.DatetimeField): When the publication was created UTC.
+
+    Relations:
+        publisher (models.ForeignKey): The publisher that created the publication.
+    """
+
+    created = models.DateTimeField(auto_now_add=True)
+
+    publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
+
+
+class PublishedFile(Model):
+    """
+    A file included in Publication.
+
+    Fields:
+        relative_path (models.CharField): The (relative) path component of the published url.
+
+    Relations:
+        publication (models.ForeignKey): The publication in which the artifact is included.
+
+    """
+    relative_path = models.CharField(max_length=255)
+
+    publication = models.ForeignKey(Publication, on_delete=models.CASCADE)
+
+    class Meta:
+        abstract = True
+
+
+class PublishedArtifact(PublishedFile):
+    """
+    An artifact that is part of a publication.
+
+    Relations:
+        content_artifact (models.ForeignKey): The referenced content artifact.
+    """
+    content_artifact = models.ForeignKey('ContentArtifact', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('publication', 'content_artifact')
+        default_related_name = 'published_artifact'
+
+
+class PublishedMetadata(PublishedFile):
+    """
+    Metadata file that is part of a publication.
+
+    Fields:
+        file (models.FileField): The stored file.
+    """
+
+    def _storage_path(self, name):
+        return DefaultStorage().published_metadata_path(self, name)
+
+    file = models.FileField(upload_to=_storage_path, max_length=255)
+
+    class Meta:
+        unique_together = ('publication', 'file')
+        default_related_name = 'published_metadata'
+
+
+class Distribution(Model):
+    """
+    A distribution defines how a publication is distributed by pulp.
+
+    Fields:
+        name (models.CharField): The name of the distribution.
+            Examples: "rawhide" and "stable".
+        base_path (models.CharField): The base (relative) path component of the published url.
+        auto_updated (models.BooleanField): The publication is updated automatically
+            whenever the publisher has created a new publication.
+        http (models.BooleanField): The publication is distributed using HTTP.
+        https (models.BooleanField): The publication is distributed using HTTPS.
+
+    Relations:
+        publisher (models.ForeignKey): The associated publisher.
+        publication (models.ForeignKey): The current publication associated with
+            the distribution.  This is the publication being served by Pulp through
+            this relative URL path and settings.
+    """
+
+    name = models.CharField(max_length=255)
+    base_path = models.CharField(max_length=255, unique=True)
+    auto_updated = models.BooleanField(default=True)
+    http = models.BooleanField(default=True)
+    https = models.BooleanField(default=True)
+
+    publication = models.ForeignKey(Publication, null=True, on_delete=models.SET_NULL)
+    publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('publisher', 'name')
--- a/platform/pulpcore/app/models/storage.py
+++ b/platform/pulpcore/app/models/storage.py
@@ -223,6 +223,25 @@ class FileSystem(Storage):
         """
         return os.path.join(settings.MEDIA_ROOT, 'artifact', sha256digest[0:2], sha256digest[2:])
 
+    @staticmethod
+    def published_metadata_path(model, name):
+        """
+        Get the storage path for published metadata.
+
+        Args:
+            model (pulpcore.app.models.PublishedMetadata): A model instance.
+            name (str): The file name.
+
+        Returns:
+            str: The absolute storage path.
+        """
+        return os.path.join(
+            settings.MEDIA_ROOT,
+            'published',
+            'metadata',
+            str(model.pk),
+            name)
+
     def get_available_name(self, name, max_length=None):
         """
         Get the available absolute path based on the name requested.
--- a/platform/pulpcore/app/tasks/publisher.py
+++ b/platform/pulpcore/app/tasks/publisher.py
@@ -1,10 +1,18 @@
+from datetime import datetime
+from gettext import gettext as _
+from logging import getLogger
+
 from celery import shared_task
+from django.db import transaction
 from django.http import QueryDict
 
 from pulpcore.app import models
 from pulpcore.app.apps import get_plugin_config
 from pulpcore.tasking.services import storage
 from pulpcore.tasking.tasks import UserFacingTask
+
+
+log = getLogger(__name__)
 
 
 @shared_task(base=UserFacingTask)
@@ -59,9 +67,32 @@ def publish(repo_name, publisher_name):
         repo_name (str): unique name to specify the repository.
         publisher_name (str): name to specify the Publisher.
     """
-    publisher = models.Publisher.objects.get(name=publisher_name,
-                                             repository__name=repo_name).cast()
+    publisher = models.Publisher.objects.get(
+        name=publisher_name,
+        repository__name=repo_name).cast()
+    log.warn(
+        _('Publishing: repository=%(r)s, publisher=%(p)s'),
+        {
+            'r': repo_name,
+            'p': publisher_name
+        })
+    with transaction.atomic():
+        publication = models.Publication(publisher=publisher)
+        publisher.publication = publication
+        publication.save()
+        with storage.working_dir_context() as working_dir:
+                publisher.working_dir = working_dir
+                publisher.publish()
+                publisher.last_published = datetime.utcnow()
+                publisher.save()
+                distributions = models.Distribution.objects.filter(
+                    publisher=publisher,
+                    auto_updated=True)
+                distributions.update(publication=publication)
+    log.warn(
+        _('Publication: %(p)s created'),
+        {
+            'p': publication.pk
+        })
 
-    with storage.working_dir_context() as working_dir:
-        publisher.working_dir = working_dir
-        publisher.publish()
+    # TODO: Replace WARN w/ INFO or DEBUG.
--- a/platform/pulpcore/app/urls.py
+++ b/platform/pulpcore/app/urls.py
@@ -7,7 +7,7 @@ from django.conf.urls import url, include
 from rest_framework_nested import routers
 
 from pulpcore.app.apps import pulp_plugin_configs
-from pulpcore.app.views.status import StatusView
+from pulpcore.app.views import ContentView, StatusView
 
 root_router = routers.DefaultRouter(
     schema_title='Pulp API',
@@ -63,6 +63,7 @@ except AppRegistryNotReady as ex:
 
 
 urlpatterns = [
+    url(r'^{}/'.format(ContentView.BASE_PATH), ContentView.as_view()),
     url(r'^api/v3/', include(root_router.urls)),
     url(r'^api/v3/status/', StatusView.as_view()),
 ]
--- a/platform/pulpcore/app/views/content.py
+++ b/platform/pulpcore/app/views/content.py
@@ -1,0 +1,188 @@
+import os
+
+from gettext import gettext as _
+from logging import getLogger, DEBUG
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.http import HttpResponseNotFound, StreamingHttpResponse
+from django.views.generic import View
+
+from wsgiref.util import FileWrapper
+
+from pulpcore.app.models import Distribution
+
+
+log = getLogger(__name__)
+log.level = DEBUG
+
+
+class ContentView(View):
+    """
+    Content endpoint.
+
+    URL matching algorithm.
+
+    http://redhat.com/content/cdn/stage/files/manifest
+                              |-------------||--------|
+                                     (1)          (2)
+
+    1. Match: Distribution.base_path
+    2. Match: PublishedFile.relative_path
+    """
+
+    BASE_PATH = 'content'
+
+    @staticmethod
+    def _base_paths(path):
+        """
+        Get a list of base paths used to match a distribution.
+
+        Args:
+            path (str): The path component of the URL.
+
+        Returns:
+            list: Of base paths.
+
+        """
+        tree = []
+        while True:
+            base = os.path.split(path.strip('/'))[0]
+            if not base:
+                break
+            tree.append(base)
+            path = base
+        return tree
+
+    def _match_distribution(self, path):
+        """
+        Match a distribution using a list of base paths.
+
+        Args:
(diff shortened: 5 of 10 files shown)


platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class PublishedArtifact(PublishedFile):
+    """
+    An artifact that is part of a publication.
+
+    Relations:
+        artifact (models.ForeignKey): The referenced artifact.
+    """
+    artifact = models.ForeignKey('Artifact', blank=True, null=True, on_delete=models.CASCADE)
```
mhrivnak: This will get a second FK relation to a ``DeferredArtifact``/``RemoteArtifact`` or whatever we call it, right?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+    Metadata file that is part of a publication.
+
+    Fields:
+        file (models.FileField): The stored file.
+    """
+
+    file = models.FileField(upload_to=path_calculator.published_metadata, max_length=255)
+
+    class Meta:
+        unique_together = ('publication', 'file')
```
mhrivnak: What does this mean with respect to a file field?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class Distribution(Model):
+    """
+    A distribution defines how a publication is distributed by pulp.
+
+    Fields:
+        name (models.TextField): The name of the distribution.
```
mhrivnak: Surely a charfield would suffice here?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+            Examples: "rawhide" and "stable".
+        base_path (models.TextField): The base (relative) path component of the published url.
```
mhrivnak: same

platform/pulpcore/app/tasks/publisher.py
```diff
@@ -59,9 +66,30 @@ def publish(repo_name, publisher_name):
...
+        publication = models.Publication(publisher=publisher)
+        publisher.publication = publication
+        publication.save()
+        with storage.working_dir_context() as working_dir:
+                publisher.working_dir = working_dir
+                publisher.publish()
+                distributions = models.Distribution.objects.filter(
+                    publisher=publisher,
+                    auto_updated=True)
+                distributions.update(publication=publication)
```
mhrivnak: This works out nicely. Django makes it so simple.

platform/pulpcore/app/views/content.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _match(self, path):
+        """
+        Match either a PublishedArtifact or PublishedMetadata.
+
+        Args:
+            path (str): The path component of the URL.
+
+        Returns:
+            str: The storage path of the matched object.
```
mhrivnak: It would be helpful to explain when/why it returns ``None``

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+class Publication(Model):
+    """
+    Fields:
+        created (models.DatetimeField): When the publication was created UTC.
+
+    Relations:
+        publisher (models.ForeignKey): The publisher that created the publication.
+    """
+
+    created = models.DateTimeField(blank=True, null=True, auto_now_add=True)
```
bmbouter: With `auto_now_add=True` this field should never be blank. Shouldn't `blank=True, null=True` be removed?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+    A file included in Publication.
+
+    Fields:
+        relative_path (models.CharField): The (relative) path component of the published url.
+
+    Relations:
+        publication (models.ForeignKey): The publication in which the artifact is included.
+
+    """
+    relative_path = models.CharField(blank=False, null=False, max_length=255)
```
bmbouter: Is this a required field? I expected every publication is required to have a `relative_path` set.

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+
+class PublishedArtifact(PublishedFile):
+    """
+    An artifact that is part of a publication.
+
+    Relations:
+        content_artifact (models.ForeignKey): The referenced content artifact.
+    """
+    content_artifact = models.ForeignKey(
+        'ContentArtifact', blank=True, null=True, on_delete=models.CASCADE)
```
bmbouter: In what cases will a `PublishedArtifact` not require this `content_artifact` field to be set?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+    Relations:
+        publisher (models.ForeignKey): The associated publisher.
+        publication (models.ForeignKey): The current publication associated with
+            the distribution.  This is the publication being served by Pulp through
+            this relative URL path and settings.
+    """
+
+    name = models.CharField(max_length=255)
+    base_path = models.CharField(max_length=255, blank=False, null=False, unique=True)
+    auto_updated = models.BooleanField(default=True)
```
bmbouter: This is a pull-style design. I still wonder about having publishers configured to point at a distribution instead. What are your thoughts about that alternative?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Relations:
+        publisher (models.ForeignKey): The associated publisher.
+        publication (models.ForeignKey): The current publication associated with
+            the distribution.  This is the publication being served by Pulp through
+            this relative URL path and settings.
+    """
+
+    name = models.CharField(max_length=255)
+    base_path = models.CharField(max_length=255, blank=False, null=False, unique=True)
```
bmbouter: In what cases would this field field be empty?

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,103 @@
...
+    auto_updated = models.BooleanField(default=True)
+    http = models.BooleanField(default=True)
```
bmbouter: I think `default=True` is a better default. I don't think we want to enable content delivery that can be MITM by default.

platform/pulpcore/app/models/storage.py
```diff
@@ -223,6 +223,25 @@ def get_artifact_path(sha256digest):
...
         return os.path.join(settings.MEDIA_ROOT, 'artifact', sha256digest[0:2], sha256digest[2:])
 
+    @staticmethod
+    def published_metadata_path(model, name):
+        """
+        Get the storage path for published metadata.
+
+        Args:
+            model (pulpcore.app.models.PublishedMetadata): A model instance.
+            name: The file name.
```
bmbouter: what type is name?

platform/pulpcore/app/models/publication.py
jortel: Yes.  I will remove.

platform/pulpcore/app/models/publication.py
jortel: Yes, it is required.

platform/pulpcore/app/models/publication.py
jortel: None, it should always be set.  Removing blank=True, null=True since `False` is the default for both.

platform/pulpcore/app/models/publication.py
jortel: None, it should always be set.  But removing blank=False, null=False since `False` is the default for both.

platform/pulpcore/app/models/publication.py
jortel: Sorry, I don't follow.  It already is `default=True`.  Can you clarify what your asking for?

platform/pulpcore/app/models/publication.py
jortel: Each distribution is owned by a publisher.  After a new publication is created, each distribution with auto_updated=True is associated with the new publication.  This workflow is _auto promotion_ (of publications).  The _manual promotion_ workflow works with distributions with auto_updated=False.  In this, case the user would promote the publication to a specific distribution using the API.

For example:

**Created**:
Distribution(name=rawhide, auto_updated=True, publication_id=NULL)
Distribution(name=stable, auto_updated=False, publication_id=NULL)

**Publish**:
1. Publication(id=1) created.
2. _auto-promotion_ Distribution(name=rawhide, **publication_id=1**)

(Tested)

**Manual Promotion**
1. API: Distribution(name=stable, **publication_id=1**)

Can you elaborate on the change "publishers configured to point at a distribution instead" is suggesting?

platform/pulpcore/app/views/content.py
```diff
@@ -0,0 +1,151 @@
+import os
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.servers.basehttp import FileWrapper
```
dkliban: I couldn't create migrations.

`ImportError: cannot import name 'FileWrapper'`

FileWrapper is not available in Django 1.9 according to [this on StackOverflow](https://stackoverflow.com/questions/35817597/django-core-servers-basehttp-filewrapper-disappears-in-django-1-9)

jortel: Fixed.

dkliban:
I don't see a viewset for creating a Distribution. 

I ran a publish and the last_published date on the publisher did not get updated.

jortel:
@dkliban: The viewset and serializer is: https://pulp.plan.io/issues/2916.  To test, you just need create a Distribution with psql.

platform/pulpcore/app/views/content.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Raises:
+            ObjectDoesNotExist: The referenced object does not exist.
+            None: When not matched.
+
+        """
+        base_paths = self._base_paths(path)
+        distribution = Distribution.objects.get(base_path__in=base_paths)
+        publication = distribution.publication
+        rel_path = path.lstrip(distribution.base_path)
```
dkliban: lstrip leaves the / on the front of the path so the relative path cannot be matched on line 70

jortel: What is your value of `base_path`?

The `base_path` is a _relative_ path and must not contain a leading '/'.  This is noted in the docstring for the Distribution model and hopefully would be enfored by the viewset.

@dkliban: FYI, I pushed an update that improves handling of 404 conditions.

jortel:
I'm testing with:
```
pulp=# select id, name, base_path  from pulp_app_distribution;
                  id                  |  name   | base_path  
--------------------------------------+---------+------------
 ea57688a-1c64-47d0-afb2-06c5ba942e63 | rawhide | redhat/cdn
```
using the file plugin and url:
```
[jortel@f25a pulp]$ http GET http://f25a.redhat.com:8000/content/redhat/cdn/PULP_MANIFEST
[17/Aug/2017 22:51:21] "GET /content/redhat/cdn/PULP_MANIFEST HTTP/1.1" 200 841
HTTP/1.0 200 OK
Content-Disposition: attachment; filename=PULP_MANIFEST
Content-Length: 841
Content-Type: text/html; charset=utf-8
Date: Thu, 17 Aug 2017 22:51:21 GMT
Server: WSGIServer/0.2 CPython/3.5.2
X-Frame-Options: SAMEORIGIN

zoo/tiger.iso, ada32934a80a567f025dc7cd49e74fc2669bb387dcbf6afe09fa328aa04fa0dd, 210
zoo/dog.iso, 34796992f2afedfea9c43f63b7d4fd568c3b39772b84759753bffa16fd305c24, 191
zoo/cat.iso, aa82c86bb4388a756c90064ce2c3956802894da017f65a59e3a4250eb5c4d858, 5
zoo/monkey.iso, 1b132f04d2b17378f3bc9c9a0a4c36d03588a92bbb13308aebe1e2eb7d0e8330, 34
zoo/lion.iso, f17a599e4bf624a7c014806fd4afe8324152d3d2e308d6f39b0e16bb0f715908, 43
zoo/zebra.iso, c20a13a30b62b2c77188d298f783898abe7fd8b84813665c54eb65bba75145a3, 28
zoo/dragon.iso, 6d192ccff23768e8e0b890e0cda493617a41d5a3aa3dd2b812b2862aeb679f0e, 43
zoo/rhino.iso, 3b772a2d92f1a3d76d52683a7a52e4361a5c18093ba986a3d3077a6105428a23, 39
zoo/rabbit.iso, 5ff6e76882a4af8f879052266929d2e0724bb323370b930c8a1b5b89b0a94f0b, 32
zoo/elephant.iso, 5d78cbc9f42a241d5305ea294c145258a510fc5be81356ab456eddc18e5c46dc, 6
```

dkliban:
@jortel the content is served correctly when the distribution's base path is 'zoo/cool' however it is still not working when the base path is just 'zoo'. Is this expected?

jortel:
@dkliban: The base_path of 'zoo' is intended to work.  I'll dig into that.

jortel:
@dkliban: path matching problem you found, fixed.

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+    base_path = models.CharField(max_length=255, unique=True)
+    auto_updated = models.BooleanField(default=True)
+    http = models.BooleanField(default=True)
+    https = models.BooleanField(default=True)
+
+    publication = models.ForeignKey(Publication, null=True, on_delete=models.SET_NULL)
+    publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('publisher', 'name')
```
dkliban: The base_path should probably be unique also. 

I just created a distribution with a duplicate base_path and the content view blew up because 2 distributions patched the base_path.

platform/pulpcore/app/urls.py
```diff
@@ -63,6 +63,7 @@ def router_for_nested_viewset(viewset):
 
 
 urlpatterns = [
+    url(r'^content/', ContentView.as_view()),
```
asmacdo: Is it possible to use `ContentView.BASE_URL` instead? Seems undesirable to define a constant but hardcode here.

platform/pulpcore/app/models/publication.py
jortel: The base_path needs to be unique across all Distributions.  The Distribution.base_path has unique=True but we still need to add a check for nested base paths.  I think this should be a follow on (separate) issue.

https://pulp.plan.io/issues/2987

platform/pulpcore/app/urls.py
jortel: Agreed.

asmacdo:
@jortel, do we still want `Publisher.last_published`? I think it might be more useful to just track a publisher's latest publication. `Publisher.latest_publish.created`

jortel:
@asmacdo, good observation.  Probably not but might be good to deal with that question separately.

platform/pulpcore/tasking/services/scheduler.py
```diff
@@ -15,6 +15,11 @@
 from pulpcore.tasking.services import worker_watcher
 
 
+# The import below is not used in this module, but it needs to be kept here. This module is the
+# first and only Pulp module to be imported by celerybeat, and by importing pulpcore.app.logs, it
+# configures the celerybeat logging to log as Pulp does.
+import pulpcore.app.logs  # noqa
```
mhrivnak: This got removed on 3.0-dev. Perhaps it is only here because of a rebase mishap?

jortel:
Rebased on 3.0-dev and squashed.

platform/pulpcore/app/views/content.py
jortel: No longer returns none.

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+    Relations:
+        publisher (models.ForeignKey): The associated publisher.
+        publication (models.ForeignKey): The current publication associated with
+            the distribution.  This is the publication being served by Pulp through
+            this relative URL path and settings.
+    """
+
+    name = models.CharField(max_length=255)
+    base_path = models.CharField(max_length=255, unique=True)
+    auto_updated = models.BooleanField(default=True)
```
bmbouter: I think this would be better in active voice. s/auto_updated/auto_update/

platform/pulpcore/app/models/progress.py
```diff
@@ -60,7 +60,7 @@ class ProgressReport(Model):
     total = models.IntegerField(null=True)
     done = models.IntegerField(default=0)
 
-    task = models.ForeignKey("Task", related_name='progress_reports', on_delete=models.CASCADE)
```
dkliban: I think you unintentionally made this change during the rebase to 3.0-dev. The related_name is needed there.

jortel: Weird.  Nice catch.

jortel: Looks like a few other unintended changes happened in the rebase.  Reviewed the diff and think I found them all.

jortel:
@dkliban, rebased to 3.0-dev.

platform/pulpcore/app/models/publication.py
```diff
@@ -0,0 +1,102 @@
...
+        https (models.BooleanField): The publication is distributed using HTTPS.
+
+    Relations:
+        publisher (models.ForeignKey): The associated publisher.
+        publication (models.ForeignKey): The current publication associated with
+            the distribution.  This is the publication being served by Pulp through
+            this relative URL path and settings.
+    """
+
+    name = models.CharField(max_length=255)
```
dkliban: The name should be required. I created a distribution without a name and then I was trying to create another one without a name I got a uniqueness constraint violation.

jortel: The default is `null=False` and `blank=False` but the `blank` is not enforced unless clean() is called.

https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.clean_fields

PR #3102, opened 2017-07-21
Resolve race condition during applicability generation

goosemania:
Race condition was possible when applicability generation was running
in parallel for the same consumer profiles.

closes #2874
https://pulp.plan.io/issues/2874


beav:
this looks more pythonic 🐍 better to ask for forgiveness than permission!

goosemania:
ok test

PR #2783, opened 2016-10-11
Add Repository Filters

asmacdo:
Finish the Repository ViewSet by implementing a FilterSet and creating
some custom filters.

https://pulp.plan.io/issues/2307

PR #3103, opened 2017-07-25
Copy mtime for copytree operation

werwty:
closes #2783
https://pulp.plan.io/issues/2783

### DIFF
--- a/server/pulp/server/util.py
+++ b/server/pulp/server/util.py
@@ -256,8 +256,10 @@ def copytree(src, dst, symlinks=False, ignore=None):
             elif os.path.isdir(srcname):
                 copytree(srcname, dstname, symlinks, ignore)
             else:
-                # Don't need to copy attributes
+                # Copy mtime and atime attributes
                 copy(srcname, dstname)
+                st = os.stat(srcname)
+                os.utime(dstname, (st.st_atime, st.st_mtime))
             # XXX What about devices, sockets etc.?
         except (IOError, os.error) as why:
             errors.append((srcname, dstname, str(why)))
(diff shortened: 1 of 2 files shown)


server/pulp/server/util.py
```diff
@@ -258,6 +258,8 @@ def copytree(src, dst, symlinks=False, ignore=None):
             else:
                 # Don't need to copy attributes
                 copy(srcname, dstname)
+                st = os.stat(srcname)
+                os.utime(dstname, (st.st_atime, st.st_mtime))
```
mhrivnak: How about just using ``shutil.copy2`` ? Are there any attributes we specifically do not want to copy?

server/pulp/server/util.py
```diff
@@ -258,6 +258,8 @@ def copytree(src, dst, symlinks=False, ignore=None):
             else:
                 # Don't need to copy attributes
```
mhrivnak: This is no longer correct.

server/pulp/server/util.py
werwty: afaik we don't want to copy over the selinux labels (otherwise we could use the shutil.copytree function)

server/pulp/server/util.py
```diff
@@ -256,8 +256,10 @@ def copytree(src, dst, symlinks=False, ignore=None):
             elif os.path.isdir(srcname):
                 copytree(srcname, dstname, symlinks, ignore)
             else:
-                # Don't need to copy attributes
+                # Copy only the mtime attribute
```
mhrivnak: and "atime"

PR #3104, opened 2017-07-25
Problem: Content can not be created using REST API

dkliban:
Solution: Add viewset and serializers to serve Content creation API

This change also introduces a DeferredArtifact that will be used by importers that
want to support deferred download policies.

closes #2872
https://pulp.plan.io/issues/2872

### DIFF
--- a/platform/pulpcore/app/models/__init__.py
+++ b/platform/pulpcore/app/models/__init__.py
@@ -6,7 +6,7 @@ from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyVa
                       GenericKeyValueModel, Config, Notes, Scratchpad)  # noqa
 
 from .consumer import Consumer, ConsumerContent  # noqa
-from .content import Content, Artifact  # noqa
+from .content import Content, Artifact, ContentArtifact, DeferredArtifact  # noqa
 from .repository import Repository, Importer, Publisher, RepositoryContent  # noqa
 from .storage import FileContent  # noqa
 from .task import ReservedResource, Worker, Task, TaskTag, TaskLock  # noqa
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -1,41 +1,12 @@
 """
 Content related Django models.
 """
+from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import models
 
+
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation
-
-
-class Content(MasterModel):
-    """
-    A piece of managed content.
-
-    Attributes:
-
-        natural_key_fields (tuple): Natural key fields.  Must be models.Field subclasses.
-
-    Relations:
-
-        notes (GenericKeyValueRelation): Arbitrary information stored with the content.
-    """
-    TYPE = 'content'
-
-    natural_key_fields = ()
-
-    notes = GenericKeyValueRelation(Notes)
-
-    class Meta:
-        verbose_name_plural = 'content'
-
-    def natural_key(self):
-        """
-        Get the model's natural key based on natural_key_fields.
-
-        :return: The natural key.
-        :rtype: tuple
-        """
-        return tuple(getattr(self, f.name) for f in self.natural_key_fields)
 
 
 class Artifact(Model):
@@ -63,7 +34,7 @@ class Artifact(Model):
         """
         return default_storage.get_artifact_path(self.sha256)
 
-    file = models.FileField(db_index=True, upload_to=storage_path, max_length=255)
+    file = models.FileField(blank=False, null=False, upload_to=storage_path, max_length=255)
     size = models.IntegerField(blank=False, null=False)
     md5 = models.CharField(max_length=32, blank=False, null=False, unique=False, db_index=True)
     sha1 = models.CharField(max_length=40, blank=False, null=False, unique=False, db_index=True)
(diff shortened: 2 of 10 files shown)


platform/pulpcore/app/models/__init__.py
```diff
@@ -6,7 +6,7 @@
                       GenericKeyValueModel, Config, Notes, Scratchpad)  # noqa
 
 from .consumer import Consumer, ConsumerContent  # noqa
-from .content import Content, Artifact  # noqa
+from .content import Content, Artifact, ContentArtifact, DeferredArtifact  # noqa
```
jortel: I'm not sold on the name `DeferredArtifact` but not a blocker.  We can revisit.

platform/pulpcore/app/models/content.py
```diff
@@ -1,12 +1,49 @@
...
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                sha256 checksum is used to determine a file path instead.
+        """
+        return default_storage.get_artifact_path(self.sha256)
+
+    file = models.FileField(db_index=True, upload_to=storage_path, max_length=255)
```
jortel: Probably should add blank=False and null=False.  Right?

platform/pulpcore/app/viewsets/content.py
```diff
@@ -44,6 +45,28 @@ class ContentViewSet(NamedModelViewSet):
...
+        content = serializer.save()
+
+        for relative_path, artifact in artifacts.items():
+            ca = ContentArtifact(artifact=artifact, content=content, relative_path=relative_path)
+            ca.save()
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def update(self, request, pk):
```
mhrivnak: It would be better if we can find a way for this to not show up as an option in the API's metadata, such as when a user makes a METHODS request to this endpoint, or just DRF's knowledge of the API.

Sadly it looks like the options are limited. I think there may be two:

1. override the ``as_view()`` method and remove the "update" value from the "actions" argument before passing it through to the superclass.
2. delete the method from the class. This isn't great, but that seems to be how DRF detects capabilities. So under the class definition, I think ``del ContentViewSet.update`` would work.

Maybe you can come up with a better idea. I'm a little surprised DRF doesn't give us an easy way to control this. Their guidance is to use mixings to assemble only the capabilities you want, but that doesn't work so well when we're extending a ViewSet with our own boilerplate.

dkliban: All these methods are added by inheriting from a mixin for each. So the solution is to explicitly iherit from the CreateMixin and DeleteMixin.

platform/pulpcore/app/models/content.py
dkliban: @mhrivnak I don't think this is right. it shouldn't be indexed.

platform/pulpcore/app/models/content.py
```diff
@@ -63,11 +92,33 @@ def storage_path(self, name):
...
-    sha384 = models.CharField(max_length=96, blank=False, null=False, unique=True, db_index=True)
-    sha512 = models.CharField(max_length=128, blank=False, null=False, unique=True, db_index=True)
+    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE, null=True)
+    content = models.ForeignKey(Content, on_delete=models.CASCADE)
+    relative_path = models.CharField(max_length=64)
+
+
+class DeferredArtifact(Model):
+    """
+    A URL for fetching a file associated with a piece of content.
```
dkliban: Yeah .... i'll add more details

platform/pulpcore/app/models/content.py
```diff
@@ -63,11 +92,33 @@ def storage_path(self, name):
...
+        url (models.TextField): The URL where the artifact can be retrieved.
+        size (models.IntegerField): The expected size of the file in bytes.
+        md5 (models.CharField): The expected MD5 checksum of the file.
+        sha1 (models.CharField): The expected SHA-1 checksum of the file.
+        sha224 (models.CharField): The expected SHA-224 checksum of the file.
+        sha256 (models.CharField): The expected SHA-256 checksum of the file.
+        sha384 (models.CharField): The expected SHA-384 checksum of the file.
+        sha512 (models.CharField): The expected SHA-512 checksum of the file.
+    """
+    url = models.TextField(blank=True, validators=[validators.URLValidator])
```
dkliban: @mhrivnak URLField seems appropriate. I used the TextField based on the comment in the DownloadCatalog

https://github.com/pulp/pulp/blob/3.0-dev/platform/pulpcore/app/models/catalog.py#L29

platform/pulpcore/app/serializers/fields.py
```diff
@@ -30,3 +35,89 @@ def to_internal_value(self, data):
...
+        ContentArtifact models related to this Content.
+
+        Args:
+            instance (:class:`pulpcore.app.models.Content`): An instance of Content being
+                serialized.
+
+        Returns:
+            A list of ContentArtifact models related to the instance of Content.
+        """
+        return models.ContentArtifact.objects.filter(content=instance)
```
dkliban: i think you are right. i will try to do that.

platform/pulpcore/app/serializers/fields.py
```diff
@@ -30,3 +35,89 @@ def to_internal_value(self, data):
...
+
+        Returns a dict mapping relative paths inside the Content to the corresponding Artifact
+        URLs.
+
+        Args:
+            value (list of :class:`pulpcore.app.models.ContentArtifact`): A list of all the
+                ContentArtifacts related to the Content model being serialized.
+
+        Returns:
+            A dict mapping Artifact URLs to the corresponding relative path of the
```
dkliban: Yeha, line 104 is correct and this is wrong. i'll update.

platform/pulpcore/app/models/content.py
```diff
@@ -38,36 +76,47 @@ def natural_key(self):
...
-        file (models.FileField): The stored file.
-        size (models.IntegerField): The size of the file in bytes.
-        md5 (models.CharField): The MD5 checksum of the file.
-        sha1 (models.CharField): The SHA-1 checksum of the file.
-        sha224 (models.CharField): The SHA-224 checksum of the file.
-        sha256 (models.CharField): The SHA-256 checksum of the file.
-        sha384 (models.CharField): The SHA-384 checksum of the file.
-        sha512 (models.CharField): The SHA-512 checksum of the file.
+    class Meta:
+        unique_together = ('artifact', 'content', 'relative_path')
```
goosemania: @dkliban , as per our discussion, let's not add `artifact` to the unique index, `content` and `relative_path` should be enough.
Because `artifact` can be `null` and is a part of this unique index, we can end up with with the duplicated (`relative_path`, `content`) in case of the sync of the same content with different download policies.

Example:
- Create repo with on_demand policy. Sync.
- Create another repo with the same feed and immediate policy. Sync.
- Check DB: https://paste.fedoraproject.org/paste/rbkT1gjqKzNGBgcRvzgLQQ

platform/pulpcore/app/models/content.py
```diff
@@ -24,6 +61,7 @@ class Content(MasterModel):
     natural_key_fields = ()
 
     notes = GenericKeyValueRelation(Notes)
+    artifacts = models.ManyToManyField(Artifact, through='ContentArtifact')
```
goosemania: Docstring should be updated to reflect this change

platform/pulpcore/app/models/content.py
```diff
@@ -38,36 +76,47 @@ def natural_key(self):
...
+    url = models.TextField(blank=True, validators=[validators.URLValidator])
+    size = models.IntegerField(blank=True, null=True)
+    md5 = models.CharField(max_length=32, blank=True, null=True)
+    sha1 = models.CharField(max_length=40, blank=True, null=True)
+    sha224 = models.CharField(max_length=56, blank=True, null=True)
+    sha256 = models.CharField(max_length=64, blank=True, null=True)
+    sha384 = models.CharField(max_length=96, blank=True, null=True)
+    sha512 = models.CharField(max_length=128, blank=True, null=True)
+    content_artifact = models.ForeignKey(ContentArtifact, on_delete=models.CASCADE)
+    importer = models.ForeignKey('Importer', on_delete=models.CASCADE)
```
goosemania: Relations should be added to the docstring (as per [style guide](http://docs.pulpproject.org/en/3.0/nightly/contributing/style_guide.html#exceptions-and-clarifications)).

platform/pulpcore/app/serializers/fields.py
```diff
@@ -30,3 +35,89 @@ def to_internal_value(self, data):
 
     def to_representation(self, value):
         return str(value)
+
+
+class ContentArtifactsField(serializers.DictField):
+    """
+    A serializer field for the 'artifacts' ManyToManyField on the Content model.
```
goosemania: Class attributes can be specified with [sphinx `:attribute:` directive](http://www.sphinx-doc.org/en/stable/domains.html?highlight=%3Aattr%3A#directive-py:attribute) to improve our docs.

platform/pulpcore/app/models/content.py
mhrivnak: Looking at some docs, django stores a CharField as a VARCHAR, and postgres treats a non-size-limited VARCHAR field the same as a TEXT field. So we should get the same behavior and performance out of a ``CharField`` as a ``TextField``.

PR #3105, opened 2017-07-25
Support calculating all artifact digests during download.

jortel:
Was thinking we could use something that could be shared by the artifact upload, changeset and plugin writers (not using changeset).  Basically anything creating artifacts.

This is the intended way to _hook_ into download processing.  Anything using a `Download` directly can simply register this event handler and then use the digest dictionary to set the digest fields on an Artifact.  This can also be used by the _upload_ code as well by calling `DigestHandler.update()` directly.


jortel:
Mainly to share the concept.  Will re-introduce in another PR.

PR #3109, opened 2017-07-27
Add JWT token auth

BrnoPCmaniak:
closes #2359
https://pulp.plan.io/issues/2359

### DIFF
--- a/docs/contributing/platform_api/app/auth.rst
+++ b/docs/contributing/platform_api/app/auth.rst
@@ -1,0 +1,9 @@
+pulp.app.auth
+=================
+
+.. automodule:: pulpcore.app.auth
+
+pulp.app.auth.jwt
+-------------------------
+
+.. automodule:: pulpcore.app.auth.jwt
--- a/docs/integration_guide/rest_api/authentication.rst
+++ b/docs/integration_guide/rest_api/authentication.rst
@@ -1,0 +1,154 @@
+Authentication
+==============
+
+All calls to REST API endpoints must be authenticated as a particular User.
+
+.. tip::
+  The password for the "admin" user can be set using two methods.
+
+      ``pulp-manager reset-admin-password``
+
+  The above command prompts the user to enter a new password for "admin" user.
+
+      ``pulp-manager reset-admin-password --random``
+
+  The above command generates a random password for "admin" user and prints it to the screen.
+
+.. tip::
+  If you are using django rest framework browsable API these browser addons may come handy:
+
+  * Chrome `ModHeader <https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj>`_
+  * Firefox `Modify Headers <https://addons.mozilla.org/cs/firefox/addon/modify-headers/>`_
+
+Basic Authentication
+--------------------
+
+Any call to the REST API may use
+`HTTP basic authentication <http://tools.ietf.org/html/rfc1945#section-11.1>`_ to provide
+a username and password.
+
+JWT Authentication
+------------------
+
+Alternatively you can use `JSON Web Tokens authentication <https://tools.ietf.org/html/rfc7519>`_.
+
+Token Structure
+^^^^^^^^^^^^^^^
+
+The structure of the token consists of ``username`` and ``exp``. The tokens are signed by each
+user's individual secret by HMAC SHA-256 ("HS256") algorithm.
+
+Example:
+::
+    {
+      "username": "admin",
+      "exp": 1501172472
+    }
+
+Obtaining token from server
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can "login" to Pulp by supplying your credentials in POST request and obtain a JWT token
+with expiration time set by ``JWT_AUTH.JWT_EXPIRATION_DELTA``
+
+
+ * **method:** ``post``
+ * **path:** ``api/v3/jwt/``
+ * response list:
+    * **code:** ``200`` - credentials were accepted
+    * **code:** ``400`` - credentials are wrong
+
+ **Sample POST request:**
+ ::
+  {
+    "username": "admin",
+    "password": "admin_password"
+  }
+
+
+ **Sample 200 Response Body:**
+ ::
+    {
+        "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTAyMzgzMDExfQ.3ZpcclxV6hN8ui2HUbwXLJsHl2lhesiCPeDVV2GIbJg"
+    }
+
+Using a token
+^^^^^^^^^^^^^
+
+For using JWT tokens you have to set ``Authorization`` header as follows:
+::
+  Authorization: JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTAyMzgzMDExfQ.3ZpcclxV6hN8ui2HUbwXLJsHl2lhesiCPeDVV2GIbJg
+
+User secret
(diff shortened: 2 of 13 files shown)


docs/integration_guide/rest_api/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+If not you can implement the above function like this:
+
+.. code-block:: python
+
+   import jwt  # pip install pyjwt
+   from datetime import datetime, timedelta
+
+
+   def generate_token_offline(username, jwt_secret, exp_delta=timedelta(days=14)):
```
bmbouter: The requirement we heard from users is that they want to generate tokens without having Pulp installed, so I think this function should go away and we need docs describing from a high level about how users can generate the JWT using non-pulp tools. Does that make any sense? I could also not be understanding this.

bmbouter: ohhhhh this is in a Python codeblock and isn't actual code. This is perfect. Thank you!

mikeadamz:
Let me know if you need help with this, I've recently spent a bunch of time implementing JWT elsewhere.

bmbouter:
@mikeadamz Some help testing would be ideal. Overall we're looking for this PR to fulfill the use cases of [the authentication portion of the MVP](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Authentication). Getting a Pulp3 vagrant environment going is what I recommend to test out pulp. Also here are some [Pulp3 dev notes](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Developer_Notes) we maintain on the wiki.

BrnoPCmaniak:
The doc failure will be solved when https://github.com/pulp/pulp_packaging/pull/428 is merged.

docs/integration_guide/rest_api/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+    {
+        "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTAyMzgzMDExfQ.3ZpcclxV6hN8ui2HUbwXLJsHl2lhesiCPeDVV2GIbJg"
+    }
+
+Using a token
+^^^^^^^^^^^^^
+
+For using JWT tokens you have to set ``Authorization`` header as follows:
+::
+  Authorization: JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTAyMzgzMDExfQ.3ZpcclxV6hN8ui2HUbwXLJsHl2lhesiCPeDVV2GIbJg
```
daviddavis: Why are you using `JWT` here? On [the jwt.io site](https://jwt.io/introduction/#how-do-json-web-tokens-work-) and elsewhere I see `Bearer` being used.

daviddavis:
A couple observations from testing. First, I assumed I could get a jwt token by using basic auth. However, it looks like I have to POST username and password. I think it would be handy to be able to use basic auth to get a jwt (but maybe not?). Maybe we can open a plan.io issue and worry about it later.

Also, I get a 401 response when using an outdated jwt token (which I expect) but also a weird "Error decoding signature" error in the response body:

```
$ http -v --auth-type jwt get :8000/api/v3/repositories/                                                                                                                              
GET /api/v3/repositories/ HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Authorization: JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTA2Mzc1NTE3fQ.JoS2KFN7rj22PSZJAjf7F1tIKF4RroJWhFLf7zvL7ro
Connection: keep-alive
Host: localhost:8000
User-Agent: HTTPie/0.9.9



HTTP/1.0 401 Unauthorized
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 38
Date: Mon, 11 Sep 2017 21:56:45 GMT
Server: WSGIServer/0.2 CPython/3.5.4
Vary: Accept
WWW-Authenticate: JWT realm="api"
X-Frame-Options: SAMEORIGIN

{
    "detail": "Error decoding signature."
}
```

For basic auth when I use a bad username/password, I get something more expected:

```
{
    "detail": "Invalid username/password."
}
```

Other than that, everything seems to work well. 👍

docs/integration_guide/rest_api/authentication.rst
BrnoPCmaniak: @daviddavis It's the default value from the library they decided to use `JWT` instead of `Bearer`, which they originally used, to maintain compatibility with `django-oauth2-provider` whihc also uses it. Source: https://github.com/GetBlimp/django-rest-framework-jwt/pull/4

BrnoPCmaniak:
@daviddavis It was in the original MVP, but then we decided not to include it to maximize the usage of the library vs rewriting it. I wasn't present the decision so I'm not exactly sure. You can read more on it here: https://pulp.plan.io/issues/2359#note-11

BrnoPCmaniak:
ok test

BrnoPCmaniak:
@daviddavis @bmbouter I changed the error message and even fixed a bug which I didn't know about (when a user that no longer exists try to login), But I have to add few things so if you could re-review it would be great.

daviddavis:
Retested and it works 👍 

Any clue on the docs failure?

bmbouter:
The docs builder error is that the `jwt` library is not installed in the docs building environment. That environment is bootstrapped on each Jenkins slave [here](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/docs.yaml#L54).

It looks like python3 is being used to build all docs environments. Adding jwt in as a dependency to install in that environment is probably what we need to do. The best thing to do is to have the setup.py build the environment so that we can stay DRY on these deps, but there are several issues preventing that from being done including:

1. That job runs both pulp2 and pulp3 jobs and only pulp3 has the setup.py that names its required Python deps
2. We would need the docs builder that builds nightly to build from source, but we would need the docs builder for PRs to build from the branch of the PR. Only that branch would have the updated setup.py change needed to have the docs builder pass.

For all of ^ reasons, I recommend:

1. Add and merge `jwt` as a dependency for all docs build environments [here](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/docs.yaml#L54).
2. rebuild these docs by force pushing a meaningless change.
3. Merge this after docs builders are happy.

bmbouter:
I posted a PR that should fix the doc builders here: https://github.com/pulp/pulp_packaging/pull/434

bmbouter:
ok test

docs/contributing/platform_api/app/auth.rst
```diff
@@ -0,0 +1,9 @@
+pulp.app.auth
```
bmbouter: I think these need to be s/pulp/pulpcore/ right? Also the indention needs to line up too (iirc).

docs/contributing/platform_api/app/auth.rst
```diff
@@ -0,0 +1,9 @@
...
+=================
+
+.. automodule:: pulpcore.app.auth
+
+pulp.app.auth.jwt
```
bmbouter: Same comments here as above.

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
ok test

bmbouter:
I applied the changes to pulp_packaging and pushed them to Jenkins with JJB. One more test will show if it all works for realz.

bmbouter:
ok test

PR #2906, opened 2017-01-10
Added boilerplate release notes for 2.11.1

dralley:

PR #3110, opened 2017-07-31
Pulp logger should log at an INFO level to syslog

werwty:
closes #2906
https://pulp.plan.io/issues/2906


mhrivnak:
@bmbouter can you elaborate on why we don't need the ``CompliantSysLogHandler`` anymore? I don't recall the details, but there were important gaps in the standard library's behavior that required creating this handler. Perhaps python 3 fixed that? That would certainly be nice!

bmbouter:
@mhrivnak Actually I was going to ask you the same question. Can you describe why we are using a custom logger in the first place?

werwty:
I think the previous `CompliantSysLogHandler` did a couple of things:
1. It encoded the message to utf-8 per RFC5424. This has been fixed in `logging.SyslogHandler` for a while. 
2. Break up long messages per RFC5426[0] recommendation.
3. Strip out newline character to comply with old syslog receivers that do not support newline characters

IMHO we should not be concerned about 1 and 3 (and I can remove these parts from `CompliantSysLogHandler`). I am unsure about 2.

[0] https://tools.ietf.org/html/rfc5426#section-3.2

bmbouter:
@werwty thanks for the info! For item 2 I remember something about not splitting the snowman utf-8 symbol by ensuring multibyte utf-8 data was not split in the middle.

+1 to removing behaviors 1 and 3 at least.

I also think the benefit of 2 is not worth the effort of maintaining this logger. We can use Python's loggers. I imagine if this linesplitting feature is not important enough for the Python community to fix, it's probably not important enough for Pulp to maintain a separate implementation for.

^ is my reasoning to disinclude it completely, but whatever @werwty and @mhrivnak can agree to is fine with me. What do you both want to do?

werwty:
FWIW `SysLogHandler` can't send long messages is closed as NOTABUG in python[0] with a recommendation of writing your own handler. 
I don't know how often pulp logs things over 1024 bytes, but dropping messages over that length is not expected behavior.

[0] http://bugs.python.org/issue11907

werwty:
Upon further testing, the `SysLogHandler` can't handle long messages error appears to only be affecting unix systems with smaller syslog buffers. This is not an issue for fedora or rhel.
I've updated the PR to remove the custom sysloghandler, since the builtin logger is fine for our use.

docs/installation/configuration-files.rst
```diff
@@ -24,3 +24,19 @@ SECRET_KEY
...
    print(''.join(random.choice(chars) for i in range(50)))
 
+logging
+    By default Pulp logs at an INFO level to syslog. Pulp also comes with a `console` handler.
+    Additional handlers can be defined and used like so:
+
+.. code-block:: yaml
+   :linenos:
+
+   logging:
```
bmbouter: What about moving this to the server.yaml file? We have all of the defaults there as commented items, so there probably should be a 'handlers' section to the commented things. This example of how to configure I think would be more useful in the commented server.yaml file so they can see the example right when they go to configure it.

PR #3111, opened 2017-08-01
Start using pip to install pulp in travis builds

BrnoPCmaniak:
Due to the fact that using setup.py for installation is no longer
suggested practice in the python community. We start using
`pip install -e .` instead of `python setup.py develop`.


bmbouter:
@BrnoPCmaniak Changing our install to `pip install -e` I believe is right. I'm going to skip adding a LGTM though so @werwty who can identify other places where this change needs to be made. Feel free to merge whenever @werwty lgtms. Thanks @BrnoPCmaniak

PR #3113, opened 2017-08-02
Fix sphinx nitpicky warnings

goosemania:
Add intersphinx extension to support reference to Python and Django docs.
Fix broken references in docstrings and in docs.

re #2878
https://pulp.plan.io/issues/2878


goosemania:
There are some issues which require additional investigation and discussion, they are mostly related to the `napoleon_django` extension which adds Fields/Relations doc fields for Django models in a way that makes nitpicky Sphinx unhappy.

dralley:
What is the status of this PR?

goosemania:
Ah, we can close it for now and maybe get back to it when we decide to work on the issue again and if this PR is not too outdated by then.
Thanks, @dralley

PR #2954, opened 2017-02-15
Improved logging of worker abnormal termination.

jortel:
https://pulp.plan.io/issues/2551

PR #3114, opened 2017-08-05
Turn on task_reject_on_worker_lost to prevent lost tasks

daviddavis:
Turn on task_reject_on_worker_lost (aka CELERY_REJECT_ON_WORKER_LOST) to prevent the loss of tasks when a worker dies. This option is only available in Celery 4+.

fixes #2954
https://pulp.plan.io/issues/2954


dralley:
LGTM

PR #3116, opened 2017-08-08
Update changeset for change in content/artifact models.

jortel:
https://pulp.plan.io/issues/2942

File plugin counterpart: https://github.com/pulp/pulp_file/pull/9

Update the _changeset_ package to align with changes to the content/artifact model.

With `Artifact` being shared, the changeset needed to find and reuse artifacts much like it does for de-duplicating content.  It also needs to manage the `ContentArtifact`.  This effort revealed a few opportunities for better division of responsibilities.  But, the changeset paradigm remains the same.  A plugin writer will only need to deal with `Content` and `Artifact` DB models.

 This effort also revealed opportunities for minor improvements in API clarity.  It's possible that the `DeferredArtifact` DB model could be renamed to something like `RemoteArtifact` which would be confusing to plugin writers.  With this in mind, I decided to try a change to the idiom in the _changeset_ API from _Remote_ to _Pending_ for classes plugin writers use to express _additions_.  After working with the changeset for a while, _pending_ seems more appropriate and should be less likely to conflict with django models exposed in the plugin API.

### DIFF
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -43,6 +43,37 @@ class Artifact(Model):
     sha384 = models.CharField(max_length=96, blank=False, null=False, unique=True, db_index=True)
     sha512 = models.CharField(max_length=128, blank=False, null=False, unique=True, db_index=True)
 
+    # All digest fields ordered by algorithm strength.
+    DIGEST_FIELDS = (
+        'sha512',
+        'sha384',
+        'sha256',
+        'sha224',
+        'sha1',
+        'md5',
+    )
+
+    # Reliable digest fields ordered by algorithm strength.
+    RELIABLE_DIGEST_FIELDS = DIGEST_FIELDS[:-3]
+
+    def is_equal(self, other):
+        """
+        Is equal by matching digest.
+
+        Args:
+            other (Artifact): A artifact to match.
+
+        Returns:
+            bool: True when equal.
+        """
+        for field in Artifact.RELIABLE_DIGEST_FIELDS[:-1]:
+            digest = getattr(self, field)
+            if not digest:
+                continue
+            if digest == getattr(other, field):
+                return True
+        return False
+
 
 class Content(MasterModel):
     """
@@ -126,5 +157,9 @@ class DeferredArtifact(Model):
     sha256 = models.CharField(max_length=64, blank=True, null=True)
     sha384 = models.CharField(max_length=96, blank=True, null=True)
     sha512 = models.CharField(max_length=128, blank=True, null=True)
+
     content_artifact = models.ForeignKey(ContentArtifact, on_delete=models.CASCADE)
     importer = models.ForeignKey('Importer', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('content_artifact', 'importer')
--- a/platform/pulpcore/download/validation.py
+++ b/platform/pulpcore/download/validation.py
@@ -208,25 +208,6 @@ class DigestValidation(Validation):
         'md5',
     )
 
-    @staticmethod
-    def _find_algorithm(name):
-        """
-        Find the hash algorithm by name in hashlib.
-
-        Args:
-            name: The algorithm name.
-
-        Returns:
-            hashlib.Algorithm: The algorithm object.
-
-        Raises:
-            ValueError: When not found.
-        """
-        try:
-            return getattr(hashlib, name.lower())()
-        except AttributeError:
-            raise ValueError(_('Algorithm {n} not supported').format(n=name))
-
     def __init__(self, algorithm, digest, enforced=True):
         """
         Args:
@@ -238,7 +219,7 @@ class DigestValidation(Validation):
             ValueError: When `algorithm` not supported by hashlib.
         """
         super(DigestValidation, self).__init__(enforced)
-        self.algorithm = self._find_algorithm(algorithm)
+        self.algorithm = hashlib.new(algorithm)
         self.expected = digest
         self.actual = None
 
@@ -268,7 +249,8 @@ class DigestValidation(Validation):
             log.warn(str(error))
 
     def __str__(self):
-        return _(
-            'DigestValidation: alg={al} expected={e} actual={a}').format(al=self.algorithm,
-                                                                         e=self.expected,
-                                                                         a=self.actual)
+        description = _('DigestValidation: alg={al} expected={e} actual={a}')
+        return description.format(
+            al=self.algorithm,
+            e=self.expected,
+            a=self.actual)
--- a/plugin/pulpcore/plugin/changeset/iterator.py
+++ b/plugin/pulpcore/plugin/changeset/iterator.py
@@ -5,9 +5,10 @@ from logging import getLogger
 
 from django.db.models import Q
 
+from pulpcore.app.models import Artifact
 from pulpcore.download import Download
 
-from .model import RemoteArtifact
+from .model import PendingArtifact
 
 
 log = getLogger(__name__)
@@ -54,12 +55,12 @@ class BatchIterator(Iterable):
 
 class ContentIterator(Iterable):
     """
-    Iterate `RemoteContent` and replace associated content models
-    With models fetched from the DB (when found).
-    """
+    Iterate `PendingContent` and foreach, replace the DB model instance
+    with an instance fetched from the DB (when found).
 
-    # Number of keys in each batched query.
-    BATCH = 1024
+    Attributes:
+        content (iterable): An iterable of PendingContent.
+    """
 
     @staticmethod
     def _batch_q(content):
@@ -67,7 +68,7 @@ class ContentIterator(Iterable):
         Build a query for the specified batch of content.
 
         Args:
-            content (tuple): A batch of content.  Each is: RemoteContent.
+            content (tuple): A batch of content.  Each is: PendingContent.
 
         Returns:
             Q: The built query.
@@ -80,7 +81,7 @@ class ContentIterator(Iterable):
     def __init__(self, content):
         """
         Args:
-            content (Iterable): An Iterable of RemoteContent.
+            content (iterable): An iterable of PendingContent.
         """
         self.content = content
 
@@ -90,9 +91,9 @@ class ContentIterator(Iterable):
 
         Yields:
             dict: A dictionary of {model_class: [content,]}
-                Each content is: RemoteContent.
+                Each content is: PendingContent.
         """
-        for batch in BatchIterator(self.content, self.BATCH):
+        for batch in BatchIterator(self.content, 1024):
             collated = {}
             for content in batch:
                 _list = collated.setdefault(type(content.model), list())
@@ -105,7 +106,7 @@ class ContentIterator(Iterable):
 
         Yields:
             tuple: (content, fetched).
-                The content is a list of RemoteContent.
+                The content is a list of PendingContent.
                 The fetched is a dictionary of fetched content models keyed by natural key.
         """
         for collated in self._collated_content():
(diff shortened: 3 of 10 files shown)


plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -158,30 +203,85 @@ class RemoteArtifact(Remote):
...
+        download.register(Event.SUCCEEDED, succeeded)
+        self.monitor = DownloadMonitor(download)
+        return download
+
+    def artifact_q(self, artifact=None):
+        """
+        Get a query for the actual artifact.
+
+        Args:
+            artifact: A artifact model.
```
mhrivnak: It would be good to document that this is optional, and what happens when it's not provided.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -198,10 +298,51 @@ def save(self):
...
+            with transaction.atomic():
+                deferred_artifact = DeferredArtifact(
+                    url=self.url,
+                    importer=self.importer,
+                    content_artifact=content_artifact,
+                    size=self.model.size,
+                    **{f: getattr(self.model, f) for f in Artifact.DIGEST_FIELDS})
+                deferred_artifact.save()
+        except IntegrityError:
+            # TODO: update?
```
mhrivnak: What's the plan for this?

plugin/pulpcore/plugin/download/monitor.py
```diff
@@ -0,0 +1,80 @@
...
+        algorithms (dict): Dictionary of validations keyed by algorithm.
+            Used only to leverage digest calculation.
+        size (int): The total file size in bytes.
+
+    Examples:
+        >>> download = ...
+        >>> monitor = DownloadMonitor(download)
+        >>> download()
+        >>> monitor.dict()
+            {'size': 1109, 'sha1': aFc12, 'sha256': '837e9ab1', ...}
```
mhrivnak: The first theoretical digest value is missing its theoretical quotes. ;)

plugin/pulpcore/plugin/download/monitor.py
```diff
@@ -0,0 +1,80 @@
...
+class DownloadMonitor:
+    """
+    Monitor a download and collect information:
+     - Total number of bytes downloaded.
+     - Standard set of file digests.
+
+    Attributes:
+        algorithms (dict): Dictionary of validations keyed by algorithm.
+            Used only to leverage digest calculation.
+        size (int): The total file size in bytes.
```
mhrivnak: It looks like this is the total number of bytes that have been received **thus far**, right?

platform/pulpcore/app/models/content.py
```diff
@@ -43,6 +43,34 @@ def storage_path(self, name):
...
+    DIGEST_FIELDS = (
+        'sha512',
+        'sha384',
+        'sha256',
+        'sha224',
+        'sha1',
+        'md5',
+    )
+
+    def is_equal(self, other):
```
jortel: Considered it but concerned it could affect django ORM internals.

platform/pulpcore/app/models/content.py
```diff
@@ -43,6 +43,34 @@ def storage_path(self, name):
...
+        """
+        Is equal by matching digest.
+
+        Args:
+            other (Artifact): A artifact to match.
+
+        Returns:
+            bool: True when equal.
+        """
+        for field in Artifact.DIGEST_FIELDS:
```
jortel: I think it needs to include sha1.

platform/pulpcore/download/validation.py
```diff
@@ -209,7 +209,7 @@ class DigestValidation(Validation):
     )
 
     @staticmethod
-    def _find_algorithm(name):
+    def find_algorithm(name):
```
jortel: Yes!  Didn't know about `hashlib.new()` .

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -146,15 +147,19 @@ class DownloadIterator(Iterable):
...
+    def __init__(self, artifacts, deferred):
         """
         Args:
-            content (Iterable): An Iterable of RemoteContent.
-            deferred (bool): When true, downloading is deferred.
+            artifacts (iterable): An iterable of PendingArtifact.
+            deferred (bool): Downloading is deferred.
         """
-        self.content = content
+        self.input = artifacts
```
jortel: It seemed clearer in the code to operate on the collection of content that is the _input_ to the iterator.  I don't feel strongly.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -181,34 +184,102 @@ def __iter__(self):
...
+        Build a query for the specified batch of artifacts.
+
+        Args:
+            artifacts (tuple): A batch of artifacts.  Each is: PendingArtifact.
+
+        Returns:
+            Q: The built query.
+        """
+        q = Q()
+        for artifact in (a for a in artifacts if a.model):
```
jortel: Sure, the `NopPendingArtifact` has a model = None and should not be included in the query.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -181,34 +184,102 @@ def __iter__(self):
...
+            q |= artifact.artifact_q()
+        return q
+
+    def _batch_artifacts(self):
+        """
+        Build a flattened collection of pending artifacts.
+        A NopPendingArtifact is yielded when the content has no artifacts.
+
+        Returns:
+            BatchIterator: Flattened artifacts.
```
jortel: 👍

plugin/pulpcore/plugin/changeset/model.py
jortel: Not sure.  What do you think?  Update the existing entry or should they be immutable?

plugin/pulpcore/plugin/download/monitor.py
jortel: Yes, but when the download is finished it will be all of the bytes.  I can reword.

platform/pulpcore/app/models/content.py
jortel: What do you think?

mhrivnak: Ah, good point. They document their approach here: https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.__eq__

And I agree it may be better to leave that alone.

plugin/pulpcore/plugin/changeset/iterator.py
mhrivnak: How about testing that more directly with ``if not isinstance(a, NopPendingArtifact)`` ?

jortel: Done.

plugin/pulpcore/plugin/changeset/model.py
mhrivnak: Looking at the ``DeferredArtifact`` model, it doesn't have any uniqueness constraints beyond the default ``pk`` one. So given that, I don't think you'd ever get an IntegrityError, nor would you need to do this in a transaction.

That doesn't seem intuitive or correct though. We need to think about uniqueness first on the model, and then come back here to think about collisions.

I suspect it should be unique based on the importer and the content_artifact, and that if any other attributes change, the DB record should be updated.

Thoughts @jortel and @dkliban ?

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -81,6 +112,19 @@ def key(self):
         """
         return {f.name: getattr(self.model, f.name) for f in self.model.natural_key_fields}
 
+    def bind(self, changeset):
+        """
+        Bind to a changeset.
```
jortel: Done

plugin/pulpcore/plugin/changeset/model.py
jortel: I hadn't noticed the unique on (importer, content_artifact) was missing and think it should be added.  Seems reasonable to update on IntegrityError.  The main use case for getting IntegrityError is when the importer specifies content in the _additions_ and it's already in the repository.  Basically a flaw in the importer logic.

platform/pulpcore/app/models/content.py
```diff
@@ -43,6 +43,37 @@ def storage_path(self, name):
...
+        'sha512',
+        'sha384',
+        'sha256',
+        'sha224',
+        'sha1',
+        'md5',
+    )
+
+    # Reliable digest fields ordered by algorithm strength.
+    RELIABLE_DIGEST_FIELDS = DIGEST_FIELDS[:-1]
```
jortel: @mhrivnak: Looks like we only have unique=True on sha256 field and higher.  The `RELIABLE_DIGEST_FIELDS` should probably match, right?

mhrivnak: Yep, that sounds reasonable.

plugin/pulpcore/plugin/changeset/model.py
jortel: @mhrivnak, Any objection to adding the constraint in this PR?

mhrivnak: Seems reasonable to me.

There is a potential use case for an importer deliberately doing this. Consider that the layout of a yum repo changed, but the same content is present. A smart importer would recognize this and update the URLs on any DeferredArtifact entries. Putting each such content unit in the iterator of additions would be a reasonable way to ensure it got updated in the DB.

jortel: Actually decided to file a separate issue to add the constraint.

dkliban: @jortel @mhrivnak The DeferredArtifact should be unique for each ContentArtifact associated with an Importer. I think an update on collision is appropriate, but i am not 100% sure.

jortel: Thanks @dkliban!

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -198,10 +295,58 @@ def save(self):
...
-        if self.path:
-            self.model.file = File(open(self.path, mode='rb'))
-            self.model.downloaded = True
-        super(RemoteArtifact, self).save()
+        artifact = None
+
+        if not self.fetched:
+            if self._path:  # downloaded
+                try:
+                    with transaction.atomic():
```
mhrivnak: In each of these three cases, do you think the work being done in the ``except`` block should be done in the transaction? I've been out of the "transaction mindset" for a while now, so I'm not recalling the expected behavior.

If an ``IntegrityError`` were raised on line 307, then the model got deleted by some other connection, and then line 310 tried to fetch the model, that would be bad. Would doing it all in one transaction prevent that?

mhrivnak: For that matter, why not use one transaction for this entire method?

jortel: All of this is being done in 1 main transaction started by the `PendingContent.save()` which invokes each `PendingArtifact.save()`.  The nested transactions here are really acting as _checkpoints_.  Without them, the `IntegrityError` would put the main transaction in an error state and the `get()` in the _except_ block would be rejected by django.

The code here is just dealing with duplicates created by race conditions.  To properly deal with the situation you described with a referenced row being deleted by another process (also a race condition), this code would need a way to differentiate between the two errors.  Unfortunately, except for parsing the description of the IntegrityError (ugh), I don't see a way to do that.  Django should raise a specific exception for dups but best I can tell (and googled), it does not.  If it did, I think we could deal with that by catching an exception in `PendingContent.save()` and retrying the entire transaction.

jortel:
squashed.

PR #3118, opened 2017-08-10
removes an import of logging code that no longer exists

mhrivnak:
re #2906


mhrivnak:
This bug report came in after the fix: https://pulp.plan.io/issues/2973

PR #3120, opened 2017-08-14
Adding a default for notes field on repository

daviddavis:
fixes #2947
https://pulp.plan.io/issues/2947


daviddavis:
Originally I tried to define a default on the field class here:

https://github.com/pulp/pulp/blob/3.0-dev/platform/pulpcore/app/serializers/generic.py#L16

But that didn't seem to work.

PR #3123, opened 2017-08-15
Port pulp2 caching of python-requests sessions.

jortel:
Mitigates issues in python-requests and urlib3.  Here is the pulp2 issue:  https://pulp.plan.io/issues/2677

The `Cache` class copied almost exactly from pulp2 but moved to context.py.  The `Context` class just moved from single.py to context.py.

### DIFF
--- a/platform/pulpcore/download/context.py
+++ b/platform/pulpcore/download/context.py
@@ -1,0 +1,216 @@
+import sys
+
+from gettext import gettext as _
+from logging import getLogger
+from threading import RLock
+from datetime import datetime, timedelta
+
+log = getLogger(__name__)
+
+
+class Cache:
+    """
+    Generic object cache.
+
+    Attributes:
+        eviction_threshold (timedelta): How long an unrequested item will be cached.
+        _lock (RLock): The object mutex.
+        _inventory (dict): The inventory of cached objects.
+            Each value is an Item.
+    """
+
+    def __init__(self, eviction_threshold=None):
+        """
+        Args:
+            eviction_threshold (timedelta): How long an unrequested item will be cached.
+
+        """
+        self.eviction_threshold = eviction_threshold or timedelta(hours=4)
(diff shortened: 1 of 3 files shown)


platform/pulpcore/download/cache.py
```diff
@@ -0,0 +1,176 @@
+import sys
```
jortel: cache.py copied exactly from pulp2.

platform/pulpcore/download/context.py
```diff
@@ -0,0 +1,223 @@
...
+
+from gettext import gettext as _
+from logging import getLogger
+from threading import RLock
+from datetime import datetime, timedelta
+
+log = getLogger(__name__)
+
+
+class NotCached(Exception):
```
mhrivnak: This could use ``LookupError`` as a base. It's more specific, and I think well-suited.

platform/pulpcore/download/http.py
```diff
@@ -118,13 +120,25 @@ def session(self):
...
+        properties = (
+            Session,
+            parsed_url.hostname,
+            parsed_url.port,
+            self.user.name,
+            self.user.password,
+            self.proxy_url,
+            self.ssl.client_certificate
+        )
+        key = hash(properties)
```
jortel: The real protection from race conditions is provided by the `Context` context manager which acquires the mutex for the get-or-create logic.  When porting this I forgot the `with self.context as context:`.  I added it back and pushed the change.

As for eliminating the `SessionCache`, it seemed more appropriate that the user of the cache be responsible for constructing the key and doing the get or create/add logic.  Just like we don't subclass a `dict ` and add methods to building the key and get_or_create().

platform/pulpcore/download/context.py
jortel: Good idea.

PR #3127, opened 2017-08-23
Fix some tasks failing to cancel after pulp restart

dralley:
Fixes an issue where tasks which go through apply_async()
directly fail to be associated with a worker, and thus
are not properly cancelled when the worker is restarted
mid-execution and performs cleanup.

closes #2734
https://pulp.plan.io/issues/2734


beav:
thanks! have an 🍨 or 🍺

dralley:
@pcreech I made this PR to master, but I really should have made it for 2.14-dev.  Sorry : /

This should go into the next 2.14.z release

PR #3128, opened 2017-08-23
Adding patch to kombu to make messages durable

daviddavis:
Fixing problem where messages in qpid were being lost when qpid gets
restarted.

fixes #2927
https://pulp.plan.io/issues/2927

Successful koji build: http://koji.katello.org/koji/taskinfo?taskID=21835


daviddavis:
@bmbouter is this good to merge or are we waiting on an upstream release instead?

PR #3129, opened 2017-08-24
Introducing asyncio downloaders

bmbouter:
Adds a HttpDownloader
- handles both synchronous and asynchronous downloads
- designed for subclassing to make custom downloader
- size validation works
- digest validation works
- has a well defined return interface DownloadResult

Reworks the Factory
- Updates the Factory to use HttpDownloader
- Fully configures HttpDownloader with all importer settings including:
  ssl client and server options, basic auth, proxy, proxy_auth
- Can be reused with custom downloaders

Adds a GroupDownloader
- Downloads files in parallel but only return back to the use when any
  group of files are fully downloaded.
- Drivable with or with generators
- Constrains in-memory objects through generator use

General Updates
- Updates the importer to use the Factory
- Updates the docs to use docs from the new code

https://pulp.plan.io/issues/2951
closes #2951

### DIFF
--- a/docs/plugins/plugin-api/asyncio.rst
+++ b/docs/plugins/plugin-api/asyncio.rst
@@ -1,0 +1,154 @@
+.. _asyncio-docs:
+
+pulpcore.plugin.download.asyncio
+================================
+
+The module implements downloaders that solve many of the common problems plugin writers have while
+downloading remote data. A high level list of features provided by these downloaders include:
+
+* auto-configuration from importer settings (auth, ssl, proxy)
+* efficient parallel downloading
+* digest and size validation computed during download
+* grouping downloads together to return to the user when all files are downloaded
+* customizable download behaviors via subclassing
+
+All classes documented here should be imported directly from the
+``pulpcore.plugin.download.asyncio`` namespace.
+
+
+Overview
+--------
+
+For basic synchronous or parallel downloading of files, the easiest usage is using the
+:ref:`DownloaderFactory <downloader-factory>` and scheduling those downloads with asyncio.
+
+For parallel downloading where multiple files need to be downloaded before a content unit can be
+saved, use the :ref:`GroupDownloader <group-downloader>`. The GroupDownloader manages the asyncio
+loop for you, which some users may prefer even for basic usage.
+
+.. _downloader-factory:
+
+Basic Downloading
+-----------------
+
+The DownloaderFactory constructs a downloader for any given url. It contains built-in support for
+`http://`, `https://` and `file://`. It creates :ref:`HttpDownloader <http-downloader>` objects for
+`http` and `https` and :ref:`FileDownloader <file-downloader>` objects for `file`. All downloaders
+are auto-configured from the core settings saved on an importer.
+
+All downloaders can be run in parallel using asyncio. Each downloader has a `run()` method which
+returns a coroutine object that asyncio can run.
+
+:ref:`HttpDownloader <http-downloader>` objects produced by an instantiated DownloaderFactory share
+a session, which contains a connection pool inside, connection reusage and keep-alives.
+
+Size and/or digest based validation can be configured using arguments provided to the
+:func:`~pulpcore.plugin.download.asyncio.DownloaderFactory.build`. method.
+
+.. autoclass:: pulpcore.plugin.download.asyncio.DownloaderFactory
+    :members:
+
+.. _download-result:
+
+Download Results
+----------------
+
+The download result contains all the information about a completed download and is returned from a
+the downloader's `run()` method when the download is complete.
+
+.. autoclass:: pulpcore.plugin.download.asyncio.DownloadResult
+    :no-members:
+
+.. _group-downloader:
+
+Downloading Groups of Files
+---------------------------
+
+Motivation
+##########
+
+A content unit that requires multiple files to be all downloaded before the content unit can be
+saved is a common problem. Consider a content unit foo, that requires three files, A, B, and C. One
+option is to use the :ref:`DownloaderFactory <downloader-factory>` to generate a downloader for each URL
+(A, B, C) and wait for those downloads to complete and then save the content unit foo and its
+associated Artifacts. The issue with this approach is that you also want to download other content
+units, e.g. a unit named bar with files (D, E, and F) and while waiting on A, B, and C you are not
+also downloading D, E, and F in parallel.
+
+GroupDownloader Overview
+########################
+
+The GroupDownloader allows you to schedule a :class:`~pulpcore.plugin.download.asyncio.Group`
+of downloads in a way that results are returned when the entire Group is ready instead of
+download-by-download. This is significant because multiple downloads from multiple groups still run
+in parallel. See the examples below.
+
+.. autoclass:: pulpcore.plugin.download.asyncio.GroupDownloader
+    :members:
+
+.. autoclass:: pulpcore.plugin.download.asyncio.Group
+    :members:
+
+.. _exception-handling:
+
+Exception Handling
+------------------
+
+All downloaders are expected to handle recoverable errors automatically. When an unrecoverable error
+occurs it can be one of two types of errors, i.e. one of the
+:ref:`validation exceptions <validation-exceptions>`, or a protocol specific error. A validation
+error example would be if a download size is expected to be 1945 bytes but it is actually 1990
+bytes, a :class:`~pulpcore.plugin.download.asyncio.SizeValidationError` would be raised. An example
+of a protocol specific error would be an HTTP 403 response.
+
+In both cases these are fatal exceptions and should likely be recorded with the
+:meth:`~pulpcore.plugin.tasking.Task.append_non_fatal_error` interface. A fatal exception on a single
+download likely does not cause an entire sync to fail, so a downloader's fatal exception is recorded
+as a non-fatal exception on the task. Plugin writers can also opt to halt the entire task by
+allowing the exception be uncaught and propogate up.
+
+.. _http-downloader:
+
+HttpDownloader
+--------------
+
+This downloader is an asyncio-aware parallel downloader which is the default downloader produced by
+the :ref:`downloader-factory` and the :ref:`group-downloader` For urls starting with `http://` or
+`https://`.
+
+.. autoclass:: pulpcore.plugin.download.asyncio.HttpDownloader
+    :members:
+
+.. _file-downloader:
+
+FileDownloader
+--------------
+
+This downloader is an asyncio-aware parallel file reader which is the default downloader produced by
+the :ref:`downloader-factory` and the :ref:`group-downloader` for urls starting with file://
+
+.. autoclass:: pulpcore.plugin.download.asyncio.FileDownloader
+    :members:
+
+.. _base-downloader:
+
+BaseDownloader
+--------------
+
+This is an abstract downloader that is meant for subclassing. All downloaders are expected to be
+descendants of BaseDownloader.
+
+.. autoclass:: pulpcore.plugin.download.asyncio.BaseDownloader
+    :members:
+
+.. autoclass:: pulpcore.plugin.download.asyncio.attach_url_to_exception
+    :members:
+
+.. _validation-exceptions:
+
+Validation Exceptions
+---------------------
+
+.. autoclass:: pulpcore.plugin.download.asyncio.DigestValidationError
+.. autoclass:: pulpcore.plugin.download.asyncio.SizeValidationError
+.. autoclass:: pulpcore.plugin.download.asyncio.DownloaderValidationError
--- a/plugin/pulpcore/plugin/download/asyncio/base.py
+++ b/plugin/pulpcore/plugin/download/asyncio/base.py
@@ -1,0 +1,203 @@
+from collections import namedtuple
+import hashlib
+import os
+import tempfile
+
+from pulpcore.app.models import Artifact
+from .exceptions import DigestValidationError, SizeValidationError
+
+
+DownloadResult = namedtuple('DownloadResult', ['url', 'artifact_attributes', 'path', 'exception'])
+"""
+Args:
+    url (str): The url corresponding with the download.
+    path (str): The absolute path to the saved file
+    artifact_attributes (dict): Contains keys corresponding with
+        :class:`pulpcore.plugin.models.Artifact` fields. This includes the computed digest values
+        along with size information.
+    exception (Exception): Any downloader exception emitted while the
+        :class:`~pulpcore.plugin.download.asyncio.GroupDownloader` is downloading. Otherwise this
+        value is `None`.
+"""
+
+
+def attach_url_to_exception(func):
+    """
+    A decorator that attaches the `url` to any exception emitted by a downloader's `run()` method.
+
+    >>> class MyCustomDownloader(BaseDownloader)
+    >>>     @attach_url_to_exception
+    >>>     async def run(self):
+    >>>         pass  # downloader implementation of run() goes here
+
+    The url is stored on the exception as the `_pulp_url` attribute. This is used by the
+    :class:`~pulpcore.plugin.download.asyncio.GroupDownloader` to know the url a given exception
+    is for.
+
+    Args:
+        func: The method being decorated. This is expected to be the `run()` method of a subclass of
+            :class:`~pulpcore.plugin.download.asyncio.BaseDownloader`
+
+    Returns:
+        A function that will attach the `url` to any exception emitted by `func`
+    """
+    async def wrapper(downloader):
+        try:
+            return await func(downloader)
+        except Exception as error:
+            error._pulp_url = downloader.url
+            raise error
+    return wrapper
+
+
+class BaseDownloader:
+    """
+    The base class of all downloaders. This is an abstract class and is meant to be subclassed.
+
+    This provides data digest calculation and validation and the writing to a file.
+
+    All subclassed downloaders should pass all downloaded data the into
+    :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data` to allow the file digest to
+    be computed while data is written to disk. This avoids having to re-read the data later. The
+    digests computed are required to save the file as an Artifact, so we need to compute them.
+
+    The :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data` method by default
+    writes to a random file in the current working directory or you can pass in your own file
+    object. See the ``custom_file_object`` keyword argument for more details. Allowing the download
+    instantiator to define the file to receive data allows the streamer to receive the data instead
+    of having it written to disk.
+
+    Attributes:
+        url (str): The url to download.
+        expected_digests (dict): Keyed on the algorithm name provided by hashlib and stores the
+            value of the expected digest. e.g. {'md5': '912ec803b2ce49e4a541068d495ab570'}
+        expected_size (int): The number of bytes the download is expected to have.
+        path (str): The full path to the file containing the downloaded data if no
+            ``custom_file_object`` option was specified, otherwise None.
+    """
+
+    def __init__(self, url, custom_file_object=None, expected_digests=None, expected_size=None):
+        """
+        Create a BaseDownloader object. This is expected to be called by all subclasses.
+
+        Args:
+            url (str): The url to download.
+            custom_file_object (file object): An open, writable file object that downloaded data
+                can be written to by
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+            expected_digests (dict): Keyed on the algorithm name provided by hashlib and stores the
+                value of the expected digest. e.g. {'md5': '912ec803b2ce49e4a541068d495ab570'}
+            expected_size (int): The number of bytes the download is expected to have.
+        """
+        self.url = url
+        if custom_file_object:
+            self._writer = custom_file_object
+            self.path = None
+        else:
+            fd, self.path = tempfile.mkstemp(dir=os.getcwd())
+            self._writer = os.fdopen(fd, mode='wb')
+        self.expected_digests = expected_digests
+        self.expected_size = expected_size
+        self._digests = {n: hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
+        self._size = 0
+
+    def handle_data(self, data):
+        """
+        Write data to the file object and compute its digests.
+
+        All subclassed downloaders are expected to pass all data downloaded to this method. Similar
+        to the hashlib docstring, repeated calls are equivalent to a single call with
+        the concatenation of all the arguments: m.handle_data(a); m.handle_data(b) is equivalent to
+        m.handle_data(a+b).
+
+        Args:
+            data (bytes): The data to be handled by the downloader.
+        """
+        self._writer.write(data)
+        self._record_size_and_digests_for_data(data)
+
+    def _record_size_and_digests_for_data(self, data):
+        """
+        Record the size and digest for an available chunk of data.
+
+        Args:
+            data (bytes): The data to have its size and digest values recorded.
+        """
+        for algorithm in self._digests.values():
+            algorithm.update(data)
+        self._size += len(data)
+
+    @property
+    def artifact_attributes(self):
+        """
+        A property that returns a dictionary with size and digest information. The keys of this
+        dictionary correspond with :class:`pulpcore.plugin.models.Artifact` fields.
+        """
+        attributes = {'size': self._size}
+        for algorithm in Artifact.DIGEST_FIELDS:
+            attributes[algorithm] = self._digests[algorithm].hexdigest()
+        return attributes
+
+    def validate_digests(self):
+        """
+        Validate all digests validate if ``expected_digests`` is set
+
+        Raises:
+            :class:`~pulpcore.plugin.download.asyncio.DigestValidationError`: When any of the
+                ``expected_digest`` values don't match the digest of the data passed to
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+        """
+        if self.expected_digests:
+            for algorithm, expected_digest in self.expected_digests.items():
+                if expected_digest != self._digests[algorithm].hexdigest():
+                    raise DigestValidationError()
+
+    def validate_size(self):
+        """
+        Validate the size if ``expected_size`` is set
+
+        Raises:
+            :class:`~pulpcore.plugin.download.asyncio.SizeValidationError`: When the
+                ``expected_size`` value doesn't match the size of the data passed to
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+        """
+        if self.expected_size:
+            if self._size != self.expected_size:
+                raise SizeValidationError()
+
+    async def run(self):
+        """
+        Run the downloader.
+
+        This is a coroutine that asyncio can schedule to complete downloading. This is required to
+        be implemented by subclasses.
+
+        It is expected that the subclass implementation call
+        :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.validate_size` and
+        :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.validate_digests` after the last
+        call to :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data` which will
+        validate the data.
+
+        It is also expected that the subclass implementation return a
+        :class:`~pulpcore.plugin.download.asyncio.DownloadResult` object. The
+        ``artifact_attributes`` value of the
+        :class:`~pulpcore.plugin.download.asyncio.DownloadResult` is usually set to the
+        :attr:`~pulpcore.plugin.download.asyncio.BaseDownloader.artifact_attributes` property value.
+
+        It is also expected that the subclass implementation be decorated with the
+        :class:`~pulpcore.plugin.download.asyncio.attach_url_to_exception` decorator. This is
+        required to allow the :class:`~pulpcore.plugin.download.asyncio.GroupDownloader` to properly
+        record the exceptions emitted from subclassed downloaders.
+
+        Returns:
+            :class:`~pulpcore.plugin.download.asyncio.DownloadResult`
+
+        Raises:
+            :class:`~pulpcore.plugin.download.asyncio.DigestValidationError`: When any of the
+                ``expected_digest`` values don't match the digest of the data passed to
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+            :class:`~pulpcore.plugin.download.asyncio.SizeValidationError`: When the
+                ``expected_size`` value doesn't match the size of the data passed to
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+        """
+        raise NotImplementedError('Subclasses must define a run() method that returns a coroutine')
--- a/plugin/pulpcore/plugin/download/asyncio/exceptions.py
+++ b/plugin/pulpcore/plugin/download/asyncio/exceptions.py
@@ -1,0 +1,34 @@
+from gettext import gettext as _
+
+from pulpcore.exceptions import PulpException
+
+
+class DownloaderValidationError(PulpException):
+    """
+    A base class for all Download Validation Errors.
+    """
+    pass
+
+
+class DigestValidationError(DownloaderValidationError):
+    """
+    Raised when a file fails to validate a digest checksum.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0003")
+
+    def __str__(self):
+        return _("A file failed validation due to checksum.")
+
+
+class SizeValidationError(DownloaderValidationError):
+    """
+    Raised when a file fails to validate a size checksum.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0004")
+
+    def __str__(self):
+        return _("A file failed validation due to size.")
--- a/plugin/pulpcore/plugin/download/asyncio/factory.py
+++ b/plugin/pulpcore/plugin/download/asyncio/factory.py
@@ -1,0 +1,157 @@
+import aiohttp
+import atexit
+import copy
+from gettext import gettext as _
+import ssl
+from urllib.parse import urlparse
+
+from .http import HttpDownloader
+from .file import FileDownloader
+
+
+PROTOCOL_MAP = {
+    'http': HttpDownloader,
+    'https': HttpDownloader,
+    'file': FileDownloader
+}
+
+
+class DownloaderFactory:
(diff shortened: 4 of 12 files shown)


bmbouter:
These docs build  correctly. I published a built version [here](http://file.rdu.redhat.com/~bbouters/plugins/plugin-api/download.html). I think the docs builder error is a docs builder problem.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+    All subclassed downloaders should pass all downloaded data the into
+    :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data` to allow the file digest to
+    be computed while data is written to disk. This avoids having to re-read the data later. The
+    digests computed are required to save the file as an Artifact, so we need to compute them.
+
+    The :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data` method by default
+    writes to a random file in the current working directory or you can pass in your own file
+    object. See the ``custom_file_object`` keyword argument for more details. Allowing the download
+    instantiator to define the file to receive data allows the streamer to receive the data instead
+    of having it written to disk.
```
jortel: Please document class `Attributes:`.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+        Args:
+            custom_file_object (file object): An open, writable file object that downloaded data
+                can be written to by
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+            expected_digests (dict): Keyed on the algorithm name provided by hashlib and stores the
+                value of the expected digest. e.g. {'md5': '912ec803b2ce49e4a541068d495ab570'}
+            expected_size (int): The number of bytes the download is expected to have.
+        """
+        if custom_file_object:
+            self._writer = custom_file_object
```
jortel: Looks like `self.path` only exists when `custom_file_object` is not passed to `__init__()`.  Suggest adding: `self.path = None` to ensure the instance always has the attribute.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+        for algorithm in Artifact.DIGEST_FIELDS:
+            attributes[algorithm] = self._digests[algorithm].hexdigest()
+        return attributes
+
+    def validate_digests(self):
+        """
+        Validate all digests validate if ``expected_digests`` is set
+
+        Raises:
+            :class:`~pulpcore.plugin.download.asyncio.DigestValidationError`: When any of the
```
jortel: What value does the `:class:` syntax add?  Seems autodoc already links the class by just specifying the type.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+        validate the data.
+
+        It is also expected that the subclass implementation return a
+        :class:`~pulpcore.plugin.download.asyncio.DownloadResult` object. The
+        ``artifact_attributes`` value of the
+        :class:`~pulpcore.plugin.download.asyncio.DownloadResult` is usually set to the
+        :attr:`~pulpcore.plugin.download.asyncio.BaseDownloader.artifact_attributes` property value.
+
+        Returns:
+            :class:`~pulpcore.plugin.download.asyncio.DownloadResult`
```
jortel: This method actually returns a `coroutine` not a `DownloadResult` and It's the coroutine that actually returns a `DownloadResult`, right?  If so, I'm not sure how to document that but this seems misleading.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+        Raises:
+            :class:`~pulpcore.plugin.download.asyncio.SizeValidationError`: When the
+                ``expected_size`` value doesn't match the size of the data passed to
+                :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.handle_data`.
+        """
+        if self.expected_size:
+            if self._size != self.expected_size:
+                raise SizeValidationError()
+
+    async def run(self):
```
jortel: Assuming this returns a `coroutine` and does not actually *run* the download, I suggest another name like `get_coroutine()` would be more accurate.

plugin/pulpcore/plugin/download/asyncio/exceptions.py
```diff
@@ -0,0 +1,34 @@
+from gettext import gettext as _
+
+from pulpcore.exceptions import PulpException
+
+
+class DownloaderValidationError(PulpException):
```
jortel: Considering your change in tasking to log non-fatal `PulpException` with stack traces - these exceptions extending `PulpException` will cause validation exceptions to be logged with a stack trace.  This seems undesirable.  I think we only want to log stack traces when unhandled exceptions indicate something is broken.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
+import aiohttp
+import atexit
+from gettext import gettext as _
+import ssl
+from urllib.parse import urlparse
+
+from .http import HttpDownloader
+
+
+class DownloadFactory(object):
```
jortel: All objects are new-style in python3.  Don't need the `(object)`.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+DownloadResult = namedtuple('DownloadResult', ['url', 'artifact_attributes', 'path'])
+"""
+Args:
+    url (str): The url corresponding with the download.
+    path (str): The absolute path to the saved file
+    artifact_attributes (dict): Contains keys corresponding with pulpcore.plugin.Artifact fields.
+        This includes the computed digest values along with size information.
+"""
+
+class BaseDownloader(object):
```
jortel: All objects are new-style in python3.  Don't need the `(object)`.

plugin/pulpcore/plugin/download/asyncio/exceptions.py
```diff
@@ -0,0 +1,34 @@
...
+class DigestValidationError(DownloaderValidationError):
+    """
+    Raised when an Artifact fails to validate a digest checksum.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0003")
+
+    def __str__(self):
+        return _("An Artifact failed validation due to checksum.")
```
jortel: What if the file being downloaded w/ validation is not an artifact?  Misleading?  Also, I think including the url would be helpful.

plugin/pulpcore/plugin/download/asyncio/exceptions.py
```diff
@@ -0,0 +1,34 @@
...
+class SizeValidationError(DownloaderValidationError):
+    """
+    Raised when an Artifact fails to validate a size checksum.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0004")
+
+    def __str__(self):
+        return _("An Artifact failed validation due to size.")
```
jortel: What if the file being downloaded w/ validation is not an artifact?  Misleading?  Also, I think including the url would be helpful.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    A factory for creating configured downloader objects that all share a single session.
+
```
jortel: Please document the instance `Attributes:`.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+        >>> the_factory = DownloadFactory(importer)
+        >>> downloader_a = the_factory.build(url_a)
+        >>> downloader_b = the_factory.build(url_b)
+        >>> downloads_list = [downloader_a, downloader_b]
+        >>> done, not_done = loop.run_until_complete(asyncio.wait(downloads_list))
+        >>> for task in done:
+        >>>     result = task.result()  # 'result' is a DownloadResult
+    """
+
+    def __init__(self, importer, download_class=HttpDownloader):
```
jortel: Shouldn't the default `download_class` depend on protocol?

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+                settings.
+            download_class (pulpcore.plugin.download.HttpDownloader): The asyncio-aware downloader
+                class this factory will use when instantiating downloaders. This option allows the
+                factory to work with custom downloaders. This is an optional parameter and defaults
+                to the HttpDownloader.
+        """
+        self._importer = importer
+        self._download_class = download_class
+        self._session = self._make_session_from_importer()
+        self._protocol_map = {'https': self._http_or_https, 'http': self._http_or_https}
```
jortel: Shouldn't the `_protocol_map` support `file:///` too?

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+                    self._importer.ssl_client_key.name,
+                    self._importer.ssl_client_certificate.name
+                )
+        else:
+            if self._importer.ssl_client_key.name and self._importer.ssl_client_certificate.name:
+                sslcontext = ssl.create_default_context()
+                sslcontext.load_cert_chain(
+                    self._importer.ssl_client_key.name,
+                    self._importer.ssl_client_certificate.name
+                )
```
jortel: It's common for the certificate to contain the client cert and key.  I think `self._importer.ssl_client_key` could be `None`.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+            importer (pulpcore.plugin.models.Importer): The importer used to populate downloader
+                settings.
+            download_class (pulpcore.plugin.download.HttpDownloader): The asyncio-aware downloader
+                class this factory will use when instantiating downloaders. This option allows the
+                factory to work with custom downloaders. This is an optional parameter and defaults
+                to the HttpDownloader.
+        """
+        self._importer = importer
+        self._download_class = download_class
+        self._session = self._make_session_from_importer()
```
jortel: So, every downloader created by this factory will share the same session?  So, the streamer will need to cache factories?

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+        auth_options = {}
+        if self._importer.basic_auth_user and self._importer.basic_auth_password:
+            auth_options['auth'] = aiohttp.BasicAuth(
+                username=self._importer.basic_auth_user,
+                password=self._importer.basic_auth_password
+            )
+
+        return aiohttp.ClientSession(connector=conn, **auth_options)
+
+    def build(self, url, expected_digests=None, expected_size=None):
```
jortel: Did you mean to add `**kwargs` to `build()` and pass them to builder methods?

plugin/pulpcore/plugin/download/asyncio/file.py
```diff
@@ -0,0 +1,46 @@
...
+    async def run(self):
+        """
+        Read, validate, and compute digests on the `uri`.
+
+        This method provides the same return object type and documented in
+        :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.run`.
+        """
+        with open(self._path, 'r') as f_handle:
+            while True:
+                chunk = await f_handle.read(1024)
```
jortel: Suggest 1MB instead of 1KB.  I don't think this will perform well.

plugin/pulpcore/plugin/download/asyncio/file.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        Download files from a uri that starts with `file://`
+
+        Args:
+            uri (str): The uri to the file. This is expected to begin with `file://`
+            kwargs (dict): This accepts the parameters of
+                :class:`~pulpcore.plugin.download.asyncio.BaseDownloader`.
+        """
+        self.uri = uri
+        kwargs['custom_file_object'] = open(os.devnull, 'w')
```
jortel: Why are you writing the downloaded file to /dev/null?

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: Fixed in next push and throughout PR.

plugin/pulpcore/plugin/download/asyncio/factory.py
bmbouter: Fixed in next push and throughout PR.

plugin/pulpcore/plugin/download/asyncio/file.py
```diff
@@ -0,0 +1,46 @@
...
+        p = urlparse.urlparse(uri)
+        self._path = os.path.abspath(os.path.join(p.netloc, p.path))
```
jortel: Why not use `BaseDownloader.path`?

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: fixed in next push.

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: Yeah good catch. Fixed in next push along w/ some doc lines explaining the cases of None versus when its actually a path.

plugin/pulpcore/plugin/download/asyncio/base.py
```diff
@@ -0,0 +1,153 @@
...
+    def _record_size_and_digests_for_data(self, data):
+        """
+        Record the size and digest for an available chunk of data.
+
+        Args:
+            data (bytes): The data to have its size and digest values recorded.
+        """
+        for algorithm in self._digests.values():
+            algorithm.update(data)
+        self._size = self._size + len(data)
```
jortel: Suggest:
```
self._size += len(data)
```

plugin/pulpcore/plugin/download/asyncio/file.py
jortel: Why not just use `p.path`?

plugin/pulpcore/plugin/download/asyncio/file.py
```diff
@@ -0,0 +1,46 @@
...
+        This method provides the same return object type and documented in
+        :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.run`.
+        """
+        with open(self._path, 'r') as f_handle:
+            while True:
+                chunk = await f_handle.read(1024)
+                if not chunk:
+                    self._validate_size()
+                    self._validate_digests()
+                    break  # the reading is done
```
jortel: I think you meant: 
```
self._validate_size()
self._validate_digests()
```
to be:
```
self.validate_size()
self.validate_digests()
```

plugin/pulpcore/plugin/download/asyncio/http.py
```diff
@@ -0,0 +1,99 @@
...
+    http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientSession
+
+    The `aiohttp.ClientSession` can additionally be configured for SSL configuration by passing in a
+    `aiohttp.TCPConnector`. For information on configuring either server or client certificate based
+    identity verification, see the aiohttp documentation:
+    http://aiohttp.readthedocs.io/en/stable/client.html#ssl-control-for-tcp-sockets
+
+    For more information on `aiohttp.BasicAuth` objects, see their docs:
+    http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.BasicAuth
+
```
jortel: Please document instance `Attributes:`.

plugin/pulpcore/plugin/download/asyncio/file.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, uri, **kwargs):
+        """
+        Download files from a uri that starts with `file://`
+
+        Args:
+            uri (str): The uri to the file. This is expected to begin with `file://`
+            kwargs (dict): This accepts the parameters of
+                :class:`~pulpcore.plugin.download.asyncio.BaseDownloader`.
+        """
+        self.uri = uri
```
jortel: Suggest adding `BaseDownloader.url` and using it instead.

plugin/pulpcore/plugin/download/asyncio/http.py
```diff
@@ -0,0 +1,99 @@
...
+                Authorization.
+            headers_ready_callback (callable): An optional callback that accepts a single dictionary
+                as its argument. The callback will be called when the response headers are
+                available. The dictionary passed has the header names as the keys and header values
+                as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            kwargs (dict): This accepts the parameters of
+                :class:`~pulpcore.plugin.download.asyncio.BaseDownloader`.
+        """
+        self.session = session
+        self.url = url
```
jortel: Suggest adding BaseDownloader.url` and using it instead.

plugin/pulpcore/plugin/download/asyncio/http.py
```diff
@@ -0,0 +1,99 @@
...
+            response (aiohttp.ClientResponse): The response to handle.
+
+        Returns:
+             DownloadResult: Contains information about the result. See the DownloadResult docs for
+                 more information.
+        """
+        if self.headers_ready_callback:
+            self.headers_ready_callback(response.headers)
+        while True:
+            chunk = await response.content.read(1024)
```
jortel: Suggest 1MB instead of 1KB.  I don't think this will perform well.

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: fixed in next push

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: I think this is how Sphinx needs it. When I compile the docs without the `:class:` they don't link or look correctly (like Sphinx recognizes them). This is the syntax also they recommend which has `:class:`, `:attr:`, `:meth:`, etc. I'm leaving this as-is.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+class GroupDownloader(object):
+    """
+    Download groups of files, with all downloads from all groups occurring in parallel.
+
+    This supports groups to be downloaded by adding them explicitly or via a iterator such as a
+    generator. The generator case is useful for limiting the number of in-memory objects.
+
+    Basic Usage:
+        >>> downloader = GroupDownloader(importer)
+        >>> artifact_a = DeferredArtifact(url=url_a md5='912ec803b2ce49e4a541068d495ab570')
```
jortel: `DeferredArtifact` renamed to `RemoteArtifact`.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+        >>>    yield (id, [DeferredArtifact(url=url_a), DeferredArtifact(url=url_b)])
+        >>>
+        >>> downloader = GroupDownloader(importer)
+        >>> downloader.schedule_from_iterator(group_generator)
+        >>> for id, group_results in downloader:
+        >>>     print(id)  # id is set to 'my_id'
+        >>>     # group_results is a dict of DownloadResult objects keyed by DeferredArtifact
+    """
+
+    def __init__(self, importer, download_class=HttpDownloader):
```
jortel: Suggest using `Importer.get_asyncio_downloader()` instead using `DownloadFactory` and `download_class`

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
+import asyncio
+from collections import defaultdict
+from contextlib import suppress
+from itertools import chain
+
+from .http import HttpDownloader
+from .factory import DownloadFactory
+
+
+class GroupDownloader(object):
```
jortel: `(object)` not needed.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Download groups of files, with all downloads from all groups occurring in parallel.
+
+    This supports groups to be downloaded by adding them explicitly or via a iterator such as a
+    generator. The generator case is useful for limiting the number of in-memory objects.
+
```
jortel: Please document instance `Attributess:`.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+        Args:
+            importer (pulpcore.plugin.models.Importer): The importer used to configure downloaders
+                with.
+            download_class (pulpcore.plugin.download.HttpDownloader): The asyncio-aware downloader
+                class the GroupDownloader will use when instantiating downloaders. This option
+                allows the GroupDownloader to work with custom downloaders.
+        """
+        self.importer = importer
+
+        self.download_factory = DownloadFactory(self.importer, download_class=download_class)
```
jortel: Suggest using `Importer.get_asyncio_downloader()` instead using `DownloadFactory` and `download_class`

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: In looking at other docs (like aiohttp which has lots of coroutines) the method names of co-routines are the method names you would expect with or without the co-routine pattern. This method runs the downloader so I want to keep the name `run`. With that reasoning I'm going to leave as-is.

Interestingly as Sphinx improves so will our docs once this is merged and released:  https://github.com/sphinx-doc/sphinx/pull/1826 I am adding a line to each `run()` method so that until this Sphinx improvement is merged it will still be clear it's a co-routine in the docs. In the code it should be totally clear with the `async` keyword in the method definition already.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+            # This is not the first call to schedule an iterable
+            self.group_iterator = chain(self.group_iterator, group_iterator)
+            return
+
+        self.group_iterator = group_iterator
+        for i in range(parallel_group_limit):
+            with suppress(StopIteration):
+                self.schedule_group(**next(self.group_iterator))
+
+    def schedule_group(self, id, deferred_artifacts):
```
jortel: Suggest this signature be `schedule_group(group)` for better symmetry with `schedule_from_iterator()`.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+                if self.group_iterator:
+                    self.schedule_group(**next(self.group_iterator))
+            for task in done_this_time:
+                download_result = task.result()
+                for group in self.urls[download_result.url]:
+                    group.handle_download_result(download_result)
+                group = self._find_and_remove_done_group()
+                if group:
+                    return group.finished_downloads
+        content_unit = self._find_and_remove_done_group()
```
jortel: The *content_unit* is a group, right?  Naming *group* would be much clearer.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+        if content_unit:
+            return content_unit.finished_downloads
+        else:
+            raise StopIteration()
+
+
+class Group(object):
+
```
jortel: Please add class docstring including `Attributres:`.

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        self.finished_urls.append(download_result.url)
+        self.downloaded_files[download_result.url] = download_result
+
+    @property
+    def done(self):
+        return len(self.urls) == len(self.finished_urls)
+
+    @property
+    def finished_downloads(self):
```
jortel: Please add docstring with `Returns:`.

jortel: I *think* it returns `(id, DownloadResult)`.  Right?

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+                self.groups_not_done.pop(index)
+                return content_unit
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        """
+        Returns:
+            tuple: The first element is the `id` first specified by the user. The second element is
```
jortel: I *think* this actually returns `(id, {DeferredArtifact: DownloadResult})`.  Right?

jortel: It's a little odd that the input is Group(s) and output is something *group-like*.  What do you think about returning the finished `Group` and letting the user inspect the group.

Something like:
```
for group in downloader:
    for deferred_artifact, download_result in group.finished_downloads:
        # do something.
```
Or,  may just go ahead and create the Artifact for them since the group downloader is artifact oriented.

Something like:
```
for group in downloader:
    for deferred_artifact, downloaded_artifact in group.downloaded_artifacts:
        # do something.
```

plugin/pulpcore/plugin/download/asyncio/group.py
```diff
@@ -0,0 +1,190 @@
...
+                class the GroupDownloader will use when instantiating downloaders. This option
+                allows the GroupDownloader to work with custom downloaders.
+        """
+        self.importer = importer
+
+        self.download_factory = DownloadFactory(self.importer, download_class=download_class)
+        self.group_iterator = None
+        self.downloads_not_done = set()
+        self.groups_not_done = []
+        self.urls = defaultdict(list)  # dict with url as the key and a lists of Groups as the value
```
jortel: Using the `defaultdict` here is cool.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+from .http import HttpDownloader
+
+
+class DownloadFactory:
+    """
+    A factory for creating configured downloader objects that all share a single session.
+
+    Usage:
+        >>> import asyncio
+        >>> loop = asyncio._get_running_loop()
```
jortel: Did you mean `_get_running_loop()`?  It's named to be non-public.

plugin/pulpcore/plugin/download/asyncio/base.py
bmbouter: I observed the same thing. I'm following what the `aiohttp` library doe,s and they have the return value of coroutines documented as the final return value emitted from the coroutine itself. Of the two choices only this one lets me express what the ultimate result will be. Otherwise, I'm not sure how to declare the result of the coroutine if I use the return value to declare that it is a coroutine. With that thinking I'm going to leave as is. I am going to check the three `run()` implementations to make sure their return values are consistent also.

bmbouter: Oh yeah the other subclasses only refer to this docstring.

plugin/pulpcore/plugin/download/asyncio/factory.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if self._importer.ssl_validation:
+            tcp_conn_opts['verify_ssl'] = self._importer.ssl_validation
+
+        conn = aiohttp.TCPConnector(**tcp_conn_opts)
+
+        auth_options = {}
+        if self._importer.basic_auth_user and self._importer.basic_auth_password:
+            auth_options['auth'] = aiohttp.BasicAuth(
+                username=self._importer.basic_auth_user,
```
jortel: s/username/login
`class BasicAuth(namedtuple('BasicAuth', ['login', 'password', 'encoding'])):`

plugin/pulpcore/plugin/download/asyncio/exceptions.py
bmbouter: The downloaders can raise these exceptions, but they are designed to be caught by the user and not to propogate all the way up. For direct usage of the downloaders the user will need to catch and handle these types of exceptions. I wrote some about that [here](https://github.com/pulp/pulp/pull/3129/files#diff-a947592032b960576ac9523a9a255e67R97).

In the event that the exception isn't caught, I think having a stack trace is what we want. Celery typically does log the stack trace of a fatal exception, but for some reason it wasn't logging PulpException based exceptions. I think we want to also log those stacktraces since they are fatal task exceptions and to be consistent with celery's logging. What do you think given all ^?

plugin/pulpcore/plugin/download/asyncio/exceptions.py
bmbouter: This is a leftover from a name change. Next push will update with s/An Artifact/A file/

We can't include any useful data (it sucks) until we resolve https://pulp.plan.io/issues/2988

plugin/pulpcore/plugin/download/asyncio/exceptions.py
bmbouter: same fix here as above.

plugin/pulpcore/plugin/download/asyncio/factory.py
bmbouter: The streamer will use `get_downloader()` right? If so, the importer instance will provide the session management. That way all the downloaders generated from the same importer instance will share their session. If this doesn't sound good, let me know.

jortel: My concern is that the streamer will fetch the importer from the DB for each call.  So, if the importer stores the factory as an instance attribute, the factory will not be shared across streamer requests.  Unless I'm missing something.

plugin/pulpcore/plugin/download/asyncio/factory.py
bmbouter: What would you recommend here? The way its written now, if `ssl_client_key` is None then these things won't be configured at all. I'm thinking this can be fixed later.

plugin/pulpcore/plugin/download/asyncio/file.py
bmbouter: It's because I didn't want to have the file importer make a copy of the file. I realize now that it needs to make a copy because otherwise when the original file is moved into place it will destroy the original file:// data. I'm going to remove this code now so that it makes a copy instead that way the original data can be imported multiple times.

plugin/pulpcore/plugin/download/asyncio/file.py
bmbouter: Yup, this was a leftover. Thank you! fixed in next push.

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: Fixed in next push.

jortel:
Testing the group downloader:
```
gd = GroupDownloader(TestImporter())
gd.schedule_group(_id, [artifact, artifact])
for x in gd:  <------------ RAISED HERE
    print(x)
```
I observed some undesirable behavior.  A failed download emitted an exception in a way that I could not catch and continue iteration.  Also, I would expect that all groups be emitted giving an indication of success for each group.

**(blocker)**

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: Yes that is what it returns. This is added in the next push along with one for the `done` property too.

plugin/pulpcore/plugin/download/asyncio/factory.py
bmbouter: yeah this was wrong. I updated it with the next push.

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: Agreed. Fixed in next push.

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: fixed in next push

plugin/pulpcore/plugin/download/asyncio/factory.py
bmbouter: Thanks. Fixed in next push.

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: I'm switching it to return `Group` and also to have it be `schedule_group(group)`.

bmbouter: This is a good idea in making the switch a good amount of code was removed from the PR.

plugin/pulpcore/plugin/download/asyncio/group.py
bmbouter: This is a good idea. Fixed in next push.

plugin/pulpcore/plugin/download/asyncio/http.py
bmbouter: fixed in next push.

bmbouter:
I pushed a small update regarding exception raising with the `GroupDownloader`. Now it does not raise exceptions, but instead records them on the `DownloadResult` and on the `Task.append_non_fatal_exception` interface.

docs/plugins/plugin-api/asyncio.rst
```diff
@@ -0,0 +1,148 @@
...
+
+This downloader is an asyncio-aware parallel file reader which is the default downloader produced by
+the :ref:`download-factory` and the :ref:`group-downloader` for urls starting with file://
+
+.. autoclass:: pulpcore.plugin.download.asyncio.FileDownloader
+    :members:
+
+.. _base-downloader:
+
+BaseDownloader
```
jortel: IIRC, in your review of the _futures_ download PR, you explicitly requested that the `Download` base class be removed from the documentation because plugin writers were not expected to create customer downloaders or override behavior.  Do you still feel that way?  If so should `BaseDownloader` be removed?  Or, should I submit a PR to include the _futures_ base class as well?

bmbouter: Yes, thanks for bringing this up. I think opening a PR to include the futures base class would be ideal. I think the use cases that require subclassing are in the must-have list of requirements and since they are both ready we should include them sooner (like now). Sound ok?

jortel: Sounds good.

PR #2758, opened 2016-09-16
pulp reports details about import failure

midnightercz:
closes #2264
https://pulp.plan.io/issues/2264

PR #3132, opened 2017-08-28
Documentation on Pulp's storage logic

ammaritiz:
Added an FAQ section to the User Guide to address
some pulp storage logic queries.

closes #2758
https://pulp.plan.io/issues/2758

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ to Pulp and your name is not on this list, please add yourself.
 Joe Adams (adams10301@gmail.com)
 Partha Aji (parthaa@gmail.com)
 Daniel Alley (dalley@redhat.com)
+Muhammad Ammar Ansari (mansari@redhat.com)
 Randy Barlow (rbarlow@redhat.com)
 Shubham Bhawsinka(sbhawsin@redhat.com)
 Sander Bos (www.sbosnet.nl)
--- a/docs/user-guide/content-storage-faq.rst
+++ b/docs/user-guide/content-storage-faq.rst
@@ -1,0 +1,97 @@
+Content Storage FAQ
+===================
+
+**How Pulp determines uniqueness when storing units?**
+
+Uniqueness of each content type is determined by it's unit key. A unit key is
+a combination of content attributes. The following combinations of attributes
+represents the unit key for some of the content type supported by Pulp.
+
+**DRPM**:
+
+- epoch
+- version
+- release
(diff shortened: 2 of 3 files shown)


asmacdo:
I am not fond of documenting plugin things in the core pulp docs. Also, I don't think we should mention a model attribute `unit_key_fields` in our user docs. Instead, it might make more sense to say that each plugin defines its own uniqueness by its "natural key". It might be awkward to do this though. Some plugin docs have a nice place for this http://docs.pulpproject.org/plugins/pulp_python/reference/python-type.html but others don't. 

@dkliban, what do you think?

docs/user-guide/faq.rst
```diff
@@ -0,0 +1,92 @@
...
+**What happens when a source repository changes the checksum type that is
+published in the repository?**
+
+Some of the content types, especially RPM, supports many checksum types while
+publishing a repository. Since, this checksum is a part of the
+'unit_key_fields', it is possible for a repository to have multiple copies of
+the same content unit. Therefore, you need to be careful when changing the
+checksum type while publishing.
+
+**How Pulp keeps track of units that belong to a particular repository?**
```
dkliban: Each repository is stored as a document in the 'repos' MongoDB collection. Each content type is stored in a collection with a prefix of 'units_'. Relationships between content and repositories are stored in the 'repo_content_units' collection.

docs/user-guide/faq.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Pulp uses MongoDB collections to keep track of units in a particular
+repository.
+
+**How symlinks are generated during a repository publish?**
```
dkliban: How are symlinks generated during a repository publish?

Pulp deduplicate content when possible. As a result all content units are stored in one place. Published content is actually a symlink to a content unit elsewhere on disk. When publishing a repository Pulp uses the relationships stored in the 'repo_content_units' for a particular repository to determine which symlinks need to be published.

docs/user-guide/faq.rst
```diff
@@ -0,0 +1,92 @@
+FAQs
```
dkliban: let's rename this to 'Content storage FAQ'

ammaritiz:
@dkliban Thanks for reviewing. I have pushed the necessary changes.

ammaritiz:
@dkliban Are we good to merge this?

PR #3133, opened 2017-08-28
Rename DeferredArtifact and remove DownloadCatalog

daviddavis:
fixes #2968
https://pulp.plan.io/issues/2968


jortel:
@dkliban originally named it `DeferredArtifact` so would like him to weigh in on that part at least.

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -36,7 +36,7 @@ def build(self, url, path=None, artifact=None):
         Args:
             url (str): The download URL.
             path (str): The optional absolute path to where the downloaded file is to be stored.
-            artifact (pulpcore.app.models.DeferredArtifact): An optional artifact.
+            artifact (pulpcore.app.models.RemoteArtifact): An optional artifact.
```
dkliban: should probably rename this to remote_artifact

platform/pulpcore/app/models/content.py
```diff
@@ -122,7 +122,7 @@ class Meta:
         unique_together = ('content', 'relative_path')
 
 
-class DeferredArtifact(Model):
+class RemoteArtifact(Model):
     """
     Stores information about an artifact that has not been retrieved yet.
```
jortel: Probably want to reword the description to something like:
```
Represents an artifact that exists in a remote repository.
```

daviddavis:
@dkliban @jortel addressed feedback and also tested against removals (which worked).

PR #3135, opened 2017-09-06
minor enhancements.

jortel:
Few minor API changes: better naming; cloning to better support mirror lists.

### DIFF
--- a/plugin/pulpcore/plugin/download/futures/event.py
+++ b/plugin/pulpcore/plugin/download/futures/event.py
@@ -162,6 +162,20 @@ class EventRouter:
         """
         self.handler.setdefault(event, []).append(handler)
 
+    def unregister(self, event, handler):
+        """
+        Unregister an event handler.
+
+        Args:
+            event (str): The event name.
+            handler (callable): A callable registered to receive the event.
+
+        Raises:
+            ValueError: when handler not registered for the
+                specified event.
+        """
+        self.handler.get(event, []).remove(handler)
+
     def send(self, event):
         """
         Send an event to all registered handlers.
@@ -179,3 +193,9 @@ class EventRouter:
             except Exception:
                 log.exception(str(event))
         return event
+
+    def clear(self):
+        """
+        Clear all registered handlers.
+        """
+        self.handler.clear()
--- a/plugin/pulpcore/plugin/download/futures/single.py
+++ b/plugin/pulpcore/plugin/download/futures/single.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from gettext import gettext as _
 from logging import getLogger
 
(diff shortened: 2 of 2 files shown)


platform/pulpcore/download/single.py
```diff
@@ -101,6 +102,25 @@ def register(self, event, handler):
...
+        """
+        Clone the download.
+        The clone has cleared:
+        - event handlers
+        - validations
+        - attachment
+        - reply
+
+        Returns:
+            Download: A cloned download.
```
mhrivnak: Is this method intending to do the equivalent of:

```
clone = Download(self.url, self.writer)
clone.context = deepcopy(self.context)
return clone
```
?

I'm not suggesting that's better necessarily, but I'm trying to fully understand the intent. By doing a deepcopy, you'll end up with a copy of each attribute as well. You're dropping/resetting most of them, but leaving context as-is, which is an exact copy of the original.

platform/pulpcore/download/single.py
```diff
@@ -101,6 +102,25 @@ def register(self, event, handler):
         """
         self.router.register(event, handler)
 
+    def clone(self):
```
mhrivnak: Did you consider implementing this as ``__deepcopy__``?

See the last paragraph here for details: https://docs.python.org/3/library/copy.html#copy.deepcopy

platform/pulpcore/download/single.py
jortel: Not quite the same thing.

The clone() impl will return a copy of the *concrete* class with all of the attributes except those cleared or reset.  For example:
```
download = HttpDownload(..)
clone = download.clone()
```

The *clone* is an instance of HttpDownload with all of the certificates, proxy url etc copied. Actually, `clone()` needs to be adjusted to do a shallow copy of the *context*.

platform/pulpcore/download/single.py
jortel: Yes, but I think `clone()` has the following advantages:
* seems more explicit
* subclasses not required to impl anything.
* one less import for *user*.

Thoughts?

platform/pulpcore/download/single.py
mhrivnak: Gotcha, that makes sense. I figured you'd want the context to be shared, not copied, so thought I'd better get clarification.

The reminder that this is supposed to work with subclasses also makes a lot of sense.

platform/pulpcore/download/single.py
mhrivnak: +1 That makes sense, especially the part about making it work for subclasses.

jortel:
@mhrivnak, updated so context shared not (deep) copied.  Anything else blocking?

PR #3136, opened 2017-09-07
2916 add distribution viewset and serializer

asmacdo:
This was based off of a @jortel branch. It will need to be rebased, cleaned up and tested, but demonstrates the POC for Nested Serialization.

This PR does a few things, which I plan to clean up into separate commits.
1. Allows nested serializers to populate a field that represents an objects "url parent". For example, the publisher created at repositories/r1/publishers/file/p1/ should obviously use r1 as its repository.
2. Allow doubly nested viewsets to be registered with the router in the correct order. 
3. Create a Distribution ViewSet and Serializer.

This (at least right now) seems like the best approach, because it avoids creating lots of Mixins that handle the url creation. It also avoids needing to use `reverse` or url manipulation.

There are a lot more comments necessary to explain how this should work, and a few tweaks, but in the end it should be simple. For a serializer to have a writable field that represents its url parent, the viewset must declare its parent_viewset, parent_lookup_kwargs, and maybe (TBD) a nest_prefix.

### DIFF
--- a/platform/pulpcore/app/serializers/base.py
+++ b/platform/pulpcore/app/serializers/base.py
@@ -61,13 +61,12 @@ def view_name_for_model(model_obj, view_action):
         LookupError: if no ViewSet is found for the Model
     """
     # Import this here to prevent out-of-order plugin discovery
-    from pulpcore.app.urls import root_router, nested_routers
+    from pulpcore.app.urls import all_routers
 
     viewset = viewset_for_model(model_obj)
 
     # return the complete view name, joining the registered viewset base name with
     # the requested view method.
-    all_routers = nested_routers + (root_router,)
     for router in all_routers:
         for pattern, registered_viewset, base_name in router.registry:
             if registered_viewset is viewset:
@@ -142,6 +141,39 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
         for field_name, mapping in field_mappings.items():
             field = getattr(instance, field_name)
             field.mapping.replace(mapping)
+
+
+class NestedModelSerializer(ModelSerializer):
+    """
+    This class supports WritableNestedUrlRelated fields, which represents nested parent objects
+    that are implied by the url of a request.
+
+    Note:
+        For asynchronous writes, during the initial validation before creating a task,
+        allow a possible Http404 bubble up. When running validation from a task,
+        catch the Http404.
+    """
+
+    def to_internal_value(self, data):
+        """
+        After letting DRF populate and validate all writable fields, populate and validate
+        `href_writable` fields.
+
+        Args:
+            data (dict): Data to be converted from primitive values to native objects.
+        Returns:
+            OrderedDict: Data that is ready to be written to the db.
+        Raises:
+            django.http.Http404: When the parent specified in the url does not exist.
+        """
+        # super().to_internal_value validates all writable fields
+        validated_data = super().to_internal_value(data)
+        for field in self.fields.values():
+            # href_writable is a special attribute of WritableNestedUrlRelatedFields
+            if (getattr(field, 'href_writable', False)):
+                parent_field, parent = self.context['view'].get_parent_field_and_object()
+                validated_data[parent_field] = parent
+        return validated_data
 
 
 class MasterModelSerializer(ModelSerializer):
@@ -295,4 +327,23 @@ class DetailNestedHyperlinkedIdentityField(_DetailFieldMixin, NestedHyperlinkedI
     """
     For use with nested viewsets of master/detail models
     """
+    pass
+
+
+class WritableNestedUrlRelatedField(NestedHyperlinkedRelatedField):
+    """
+    For use with a NestedModelSerializer.
+
+    This field supports a special attribute, `href_writable`. Fields that carry this flag represent
+    a nested parent, which is the model determined by the url parameters instead of request body
+    parameters.
+    """
+    def __init__(self, *args, **kwargs):
+        # rest_framework.fields.Field.__init__ sets self.read_only, so we have to inject this kwarg.
+        kwargs['read_only'] = True
+        self.href_writable = True
+        super().__init__(*args, **kwargs)
+
+
+class DetailWritableNestedUrlRelatedField(_DetailFieldMixin, WritableNestedUrlRelatedField):
     pass
--- a/platform/pulpcore/app/urls.py
+++ b/platform/pulpcore/app/urls.py
@@ -1,73 +1,117 @@
 """pulp URL Configuration"""
-
-import warnings
-
-from django.core.exceptions import AppRegistryNotReady
 from django.conf.urls import url, include
 from rest_framework_nested import routers
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.views import ContentView, StatusView
 
+import logging
+log = logging.getLogger(__name__)
+
+
+class ViewSetNode:
+    """
+    Each node is a tree that can register nested ViewSets with DRF nested routers.
+
+    The structure of the tree becomes the url heirarchy when the ViewSets are registered.
+
+    Example Structure:
+        RootNode
+          ├─ some-non-nested viewset
+          └─ RepositoryViewSet (non-nested)
+                ├─ PluginPublisherViewSet
+                │   └─ DistributionViewSet
+                ├─ AnotherPluginPublisherViewSet
+                │   └─ DistributionViewSet (This node is attached to all Publisher Detail parents)
+                └─ FileImporterViewSet
+    """
+    def __init__(self, viewset=None):
+        """
+        Create a new node.
+
+        Args:
+            viewset (NamedModelViewSet): If provided, represent this viewset. If not provided, this
+                                         is the root node.
+        """
+        self.viewset = viewset
+        self.children = []
+
+    def add_decendent(self, node):
+        """
+        Add a VSNode to the tree. If node is not a direct child, attempt to add the to each child.
+
+        Some nodes must be added more than once if they have a Master/Detail parent. Using
+        Distributions as an example, DistributionViewset.parent_viewset is PublisherViewSet, which
+        is a MasterViewset. Each of the publisher detail viewsets like FilePublisherViewSEt will
+        have its own router, and the DistributionViewSet must be registered with each.
+
+        Args:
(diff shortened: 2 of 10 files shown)


asmacdo:
@werwty, some extra stuff is on here because I based this on a jortel branch. Hopefully it will be obvious which are my changes until I get my git together.

jortel:
The publishing model PRs have been merged to 3.0-dev.  This can be rebased.

asmacdo:
This adds the Distribution ViewSet and Serializer. This required some secondary changes as well:
1) I reworked viewset registration to allow for arbitrarily deep nesting. 
2) I created the WritableNestedUrlRelatedField, because users should not specify which publisher in the post body, it is already implicit in the url.
3) I added a Distributions field to publishers.

platform/pulpcore/app/serializers/base.py
```diff
@@ -144,6 +143,39 @@ def _populate_generic_fields(self, instance, field_mappings):
             field.mapping.replace(mapping)
 
 
+class NestedModelSerializer(ModelSerializer):
+    """
+    This class allows the serializer to have WritableNestedUrlRelated fields, which represent
```
jortel: s/allows/supports/

platform/pulpcore/app/serializers/base.py
```diff
@@ -144,6 +143,39 @@ def _populate_generic_fields(self, instance, field_mappings):
...
+    nested parent objects that are implied by the url of a request.
+    """
+
+    def to_internal_value(self, *args, **kwargs):
```
jortel: In most cases it's a much better practice to match the specific signature when overriding methods.
```
def to_internal_value(self, data)
```

platform/pulpcore/app/serializers/base.py
```diff
@@ -144,6 +143,39 @@ def _populate_generic_fields(self, instance, field_mappings):
...
+        """
+        After letting DRF populate and validate all writable fields, we populate and validate
+        `href_writable` fields.
+        """
+        # super().to_internal_value will validate all writable fields and populate serializer.data
+        validated_data = super().to_internal_value(*args, **kwargs)
+        # nested parent fields should be read_only (so they don't accept args that are ignored)
+        # instead, validate the parent based on the parent url and add it to the validated data.
+        validated_data.update(self.writable_nested_parent_url_to_internal())
```
jortel: Why is `writable_nested_parent_url_to_internal()` a separate method instead of having that code in the body here?
This method does almost nothing else.

platform/pulpcore/app/serializers/base.py
```diff
@@ -144,6 +143,39 @@ def _populate_generic_fields(self, instance, field_mappings):
...
+        """
+        Returns an ordered dict containing each `href_writable` field and its internal value.
+        """
+        url_parents = OrderedDict()
+        for field in self.fields.values():
+            # Writable nested parent objects must be read only (parent is regrieved using kwargs)
+            if (getattr(field, 'href_writable', False)):
+                # Use the original viewset to retrieve the parent object implied by the request url
+                # TODO(asmacdo) catch the 404
+                parent_field, parent_obj = self.context['view'].get_parent_field_and_object()
```
jortel: Everything is an object in python, what is the the `_obj` suffix trying to communicate?

platform/pulpcore/app/serializers/base.py
```diff
@@ -144,6 +143,39 @@ def _populate_generic_fields(self, instance, field_mappings):
...
+        return validated_data
+
+    def writable_nested_parent_url_to_internal(self):
+        """
+        Returns an ordered dict containing each `href_writable` field and its internal value.
+        """
+        url_parents = OrderedDict()
+        for field in self.fields.values():
+            # Writable nested parent objects must be read only (parent is regrieved using kwargs)
+            if (getattr(field, 'href_writable', False)):
```
jortel: Should `href_writable` be a constant?

jortel: Can you help me understand this *special* attribute?

platform/pulpcore/app/serializers/base.py
```diff
@@ -296,3 +328,20 @@ class DetailNestedHyperlinkedIdentityField(_DetailFieldMixin, NestedHyperlinkedI
     For use with nested viewsets of master/detail models
     """
     pass
+
+
+class WritableNestedUrlRelatedField(DetailNestedHyperlinkedRelatedField):
+    """
+    For use with a NestedModelSerializer.
+
+    This field is writable, but does not accept request parameters. Instead, the parent object is
```
jortel: What is `object` here?  model, viewset, serialzer?

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
-
-nested_routers = (
-    routers.NestedDefaultRouter(root_router, 'repositories', lookup='repository'),
-)
+import logging
+log = logging.getLogger(__name__)
 
 
-def router_for_nested_viewset(viewset):
+class VSNode(object):
```
jortel: Just FYI, don't need the `(object)` in python3.

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
-        viewset (pulpcore.app.viewsets.NamedModelViewSet): a viewset for which a nested router
-            should exist
-
-    Returns:
-        routers.NestedDefaultRouter: the nested router whose parent_prefix corresponds to the
-            viewset's nest_prefix attribute.
+    This is a tree that can register nested ViewSets with DRF routers.
+    """
+    def __init__(self, viewset=None):
+        self.vs = viewset
```
jortel: `self.viewset` might be more readable.

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
+        self.children = []
+
+    def add_node(self, node):
+        """
+        Add a VSNode to the tree.
```
jortel: Document node argument:
```
Args:
    node:
```

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
+        if not node.vs.parent_viewset:
+            self.children.append(node)
+        # link to all parent viewsets
+        elif self.vs and issubclass(self.vs, node.vs.parent_viewset):
+            self.children.append(node)
+        else:
+            for child in self.children:
+                child.add_node(node)
+
+    def register_with(self, router, all_routers=None):
```
jortel: Suggest:
```
def register_with(self, *router):
```

can be called with:
```
node.register_with(router)
node.register_with(routerA, routerB, ...)
node.register_with(*list_of_router)
```

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
-    Find the router for a nested viewset.
-
-    Args:
-        viewset (pulpcore.app.viewsets.NamedModelViewSet): a viewset for which a nested router
-            should exist
-
-    Returns:
-        routers.NestedDefaultRouter: the nested router whose parent_prefix corresponds to the
-            viewset's nest_prefix attribute.
+    This is a tree that can register nested ViewSets with DRF routers.
```
jortel: You describe this as a tree (and seems to behave like a tree).  Should this be named: `VSTree`?

platform/pulpcore/app/viewsets/base.py
```diff
@@ -146,6 +126,29 @@ class NamedModelViewSet(mixins.CreateModelMixin,
...
+    A Nested Viewset that can determine parent objects from the url and write them.
+    """
+
+    def get_parent_field_and_object(self):
+        """
+        Assumes that attr `parent_viewset` has been set and that it is one level up in the url.
+
+        TODO(asmacdo) document extra fields and fail if they arent there!
+        Raises:
+            404 TODO(asmacdo)
```
jortel: Returns:

platform/pulpcore/app/viewsets/repository.py
```diff
@@ -195,6 +201,17 @@ def publish(self, request, repository_name, name):
         return OperationPostponedResponse([async_result], request)
 
 
+class DistributionViewSet(NestedNamedModelViewSet):
```
jortel: I love how simple this is!

platform/pulpcore/app/urls.py
```diff
@@ -1,73 +1,92 @@
...
     url(r'^{}/'.format(ContentView.BASE_PATH), ContentView.as_view()),
-    url(r'^api/v3/', include(root_router.urls)),
     url(r'^api/v3/status/', StatusView.as_view()),
 ]
 
-
-for nrouter in nested_routers:
-    urlpatterns.append(url(r'^api/v3/', include(nrouter.urls)))
+all_routers = vs_tree.register_with(root_router)
+for each in all_routers:
```
werwty: s/each/router/

platform/pulpcore/app/serializers/base.py
```diff
@@ -296,3 +328,20 @@ class DetailNestedHyperlinkedIdentityField(_DetailFieldMixin, NestedHyperlinkedI
     For use with nested viewsets of master/detail models
     """
     pass
+
+
+class WritableNestedUrlRelatedField(DetailNestedHyperlinkedRelatedField):
```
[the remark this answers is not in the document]
asmacdo: This should be usable for the repository field of importers and publishers.

platform/pulpcore/app/urls.py
jortel: 👍

asmacdo:
> Right now, if I pass in a publisher as an argument to distribution creation it just gets ignored. I think it should instead return a 400 error. This can be done by checking to see if the validated_data in the NestedModelSerializer already contains the href_writable field

This problem is more general, all extra arguments are ignored. Maybe we can address that more generally.

platform/pulpcore/app/urls.py
asmacdo: I updated the docstring, because this is a little confusing. The node is registered with `router` only, `all_routers` is just keeping track of all the new routers we create.

platform/pulpcore/app/urls.py
asmacdo: @jortel and I discussed this and decided on ViewSetNode

PR #3137, opened 2017-09-09
Fix issue causing worker to be left in a broken state

dralley:
Mitigates an issue where workers can be left in a broken
state if they are restarted while processing a task.

Fixes only the SIGTERM shutdown case, the most common one.

re #2835
https://pulp.plan.io/issues/2835

### DIFF
--- a/server/pulp/server/async/app.py
+++ b/server/pulp/server/async/app.py
@@ -10,7 +10,7 @@ from datetime import datetime, timedelta
 from gettext import gettext as _
 
 from celery import bootsteps
-from celery.signals import celeryd_after_setup, worker_shutdown
+from celery.signals import celeryd_after_setup
 import mongoengine
 
 from pulp.common import constants, dateutils
@@ -93,6 +93,18 @@ class HeartbeatStep(bootsteps.StartStopStep):
             self.timer_ref.cancel()
             self.timer_ref = None
 
+    def shutdown(self, consumer):
+        """
+        Called when the consumer is shut down.
+
+        So far, this just cleans up the database by removing the worker's record in
+        the workers collection.
+
+        :param consumer: The consumer instance
+        :type  consumer: celery.worker.consumer.Consumer
+        """
+        tasks._delete_worker(consumer.hostname, normal_shutdown=True)
+
     def _record_heartbeat(self, consumer):
         """
         This method creates or updates the worker record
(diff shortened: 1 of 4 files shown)


dralley:
@pcreech, be aware that I accidentally pushed this to pulp/2.14-dev instead of dralley/2.14-dev, noticed it, and then reverted that commit.

dralley:
re: @pep8speaks , I've already fixed those two issues in a different PR which should have gone into 2.14-dev, but I accidentally made it for master instead, and it was already merged.

It will be resolved when that commit makes it to 2.14-dev

server/pulp/server/async/app.py
```diff
@@ -40,36 +40,45 @@ class HeartbeatStep(bootsteps.StartStopStep):
...
-        Create a timer which periodically runs the heartbeat routine.
+        Create a timer which periodically runs the heartbeat routine. Delete any stale worker data.
 
         This method is called when the worker starts up and also whenever the AMQP connection is
         reset (which triggers an internal restart). The timer is reset when the connection is lost,
         so we have to install the timer again for every call to self.start.
 
-        :param worker: The consumer instance
-        :type  worker: celery.worker.consumer.Consumer
+        We are using this to deleting any stale worker data from improperly killed workers.
```
bmbouter: s/deleting/delete/

server/pulp/server/async/app.py
```diff
@@ -40,36 +40,45 @@ class HeartbeatStep(bootsteps.StartStopStep):
...
+        We need to do this here, because if this happens any earlier in the process, the task
+        revokation inside of _delete_worker() will cause the process to lock up. However, since
```
bmbouter: What causes the task to lockup? I'm curious.

server/pulp/server/async/tasks.py
```diff
@@ -655,7 +655,7 @@ def cancel(task_id):
         consumer_id = tag_dict.get(tags.RESOURCE_CONSUMER_TYPE)
         agent_manager.cancel_request(consumer_id, task_id)
     else:
-        controller.revoke(task_id, terminate=True)
+        controller.revoke(task_id)
```
bmbouter: I think with `terminate=True` it will cause a long-running task to be terminated immediately versus terminating after some time or when the task has completed. That is a behavior we need. Has a cancel been tested w/ this code to ensure that cancelling a running task cancels immediately and not later?

server/pulp/server/async/app.py
```diff
@@ -93,6 +93,18 @@ def stop(self, consumer):
             self.timer_ref.cancel()
             self.timer_ref = None
 
+    def shutdown(self, consumer):
```
bmbouter: So this shutdown() will be called by the consumer bootstep when it is shutting down. Can you describe specifically which situations cause this codepath to be called? I want to make sure we're only deleting the worker (and its reservations) in the right situations and none of the wrong situations.

dralley: Only when it is being torn down entirely, i.e. this won't happen merely when the broker goes down.  That's when stop() gets called.

bmbouter: OK this sounds completely right. Thanks @dralley

ipanova:
@dralley even if pr is merged, the issue is still in post, please take a look and update the status accordingly.

dralley:
@ipanova the merged fix doesn't solve the less common case where the worker
dies to OOM or pkill, only the more common systemctl stop case.

That's why I left it open for now.

PR #3138, opened 2017-09-12
Disabling %check for python-kombu

pcreech:
Disabling %check for python-kombu

Fedora's python-kombu doesn't have this enabled, and it's causing build
failures for rhel7 for python-six version requirements.

### DIFF
--- a/deps/external_deps.json
+++ b/deps/external_deps.json
@@ -51,7 +51,7 @@
     },
     {
         "name": "python-kombu",
-        "version": "3.0.33-7.pulp",
+        "version": "3.0.33-8.pulp",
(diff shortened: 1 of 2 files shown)


pcreech:
It's worth noting that this is what was done in epel as well

http://pkgs.fedoraproject.org/cgit/rpms/python-kombu.git/tree/python-kombu.spec?h=epel7#n125

And fedora does not have a %check section in their spec files.

http://pkgs.fedoraproject.org/cgit/rpms/python-kombu.git/tree/python-kombu.spec?h=f25
http://pkgs.fedoraproject.org/cgit/rpms/python-kombu.git/tree/python-kombu.spec?h=f24

PR #3143, opened 2017-09-13
Storage refined.

jortel:
There can be (rare) race conditions where an artifact is copied into storage when it already exists.  This is because the file is stored before the DB commit can raise the constraint violation.  Storing artifacts is a concurrent operation.  In all cases, an existing file will not be overwritten.  The atomic copy prevents partial files from being stored.

### DIFF
--- a/platform/pulpcore/app/models/storage.py
+++ b/platform/pulpcore/app/models/storage.py
@@ -4,6 +4,7 @@ import io
 import stat
 
 from datetime import datetime
+from uuid import uuid4
 
 from django.conf import settings
 from django.core.files import File
@@ -108,24 +109,6 @@ class FileSystem(Storage):
     """
 
     @staticmethod
-    def mkdir(path):
-        """
-        Make the directory (and parent directories) at the specified path.
-        No exception is raised if the directory already exists.
-
-        Args:
-            path (str): Absolute directory path.
-        """
-        try:
-            os.makedirs(path)
-        except OSError as e:
-            if e.errno == errno.EEXIST and os.path.isdir(path):
-                # ignored.
-                pass
-            else:
-                raise
-
-    @staticmethod
     def unlink(path):
         """
         Delete the link at the specified path.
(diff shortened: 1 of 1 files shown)


asmacdo:
New flow also approved.

platform/pulpcore/app/models/storage.py
```diff
@@ -115,13 +116,35 @@ def mkdir(path):
...
             os.makedirs(path)
-        except OSError as e:
-            if e.errno == errno.EEXIST and os.path.isdir(path):
+        except FileExistsError:
+            if os.path.isdir(path):
                 # ignored.
                 pass
+
+    @staticmethod
+    def link(path, content):
```
asmacdo: I like these as as static methods.

platform/pulpcore/app/models/storage.py
```diff
@@ -133,12 +156,42 @@ def unlink(path):
...
+        tmp_path = '.'.join((path, str(uuid4()), 'tmp'))
+        try:
+            with open(tmp_path, 'wb+') as fp:
+                while True:
+                    bfr = content.read(1024000)
+                    if bfr:
+                        fp.write(bfr)
+                    else:
+                        break
+            os.link(tmp_path, path)
```
asmacdo: WIll this raise `errno.EXDEV`? This only runs if we are on a remote system, right? I'm guessing we want `os.rename`.

asmacdo: A comment could indicate that this is the moment that the race condition is resolved.

platform/pulpcore/app/models/storage.py
```diff
@@ -115,13 +116,35 @@ def mkdir(path):
 
         Args:
             path (str): Absolute directory path.
+
+        Raises:
+            OSError: On error.
         """
         try:
             os.makedirs(path)
```
asmacdo: `exist_ok=True` does what we want here. It raises `FileExistsError` on files, but not on directories.

platform/pulpcore/app/models/storage.py
jortel: No, EXDEV will never be raised here.  The tmp_path and path will always be on the same filesystem.  

As it turns out, `os.rename()` has some nasty behavior in that if the target already exists, it replaces it instead of raising `FileExistsError` so it cannot be used.

jortel: Using `shutil.move()` trades 2 lines of code for 2 lines of code and adds a dependency.  Just using the standard lib for this seems better.

platform/pulpcore/app/models/storage.py
jortel: Nice.

PR #3144, opened 2017-09-13
Set the href_writable on RepositoryRelatedFields

werwty:
The parent repo will now be determined by the url instead of by
request parameters.

closes #2985
https://pulp.plan.io/issues/2985

### DIFF
--- a/platform/pulpcore/app/serializers/fields.py
+++ b/platform/pulpcore/app/serializers/fields.py
@@ -16,13 +16,15 @@ class ContentRelatedField(DetailRelatedField):
     queryset = models.Content.objects.all()
 
 
-class RepositoryRelatedField(serializers.HyperlinkedRelatedField):
+class HrefWritableRepositoryRelatedField(serializers.HyperlinkedRelatedField):
     """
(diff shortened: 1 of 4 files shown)


platform/pulpcore/app/serializers/fields.py
```diff
@@ -16,13 +16,14 @@ class ContentRelatedField(DetailRelatedField):
...
+class HrefWritableRepositoryRelatedField(serializers.HyperlinkedRelatedField):
     """
-    A serializer field with the correct view_name and lookup_field to link to a repository.
+    A serializer field for a repository that is the parent in a nested relationship.
+    It has the href_writable field set so the repository is determined by url parameters.
     """
     view_name = 'repositories-detail'
     lookup_field = 'name'
-    queryset = models.Repository.objects.all()
+    href_writable = True
```
asmacdo: I guess we don't need to override `__init__` if `read_only=True` is passed as a kwarg. I like this better, particularly because I don't want to run into mro problems with the mixins. Would you mind adding a comment about setting read_only?

For consistency, (and only if it works easily) would you mind making this change in DetailWritableNestedUrlRelatedField also?

PR #3147, opened 2017-09-15
Simpler batch implementation.

jortel:
The `Batch` greatly simplified.

Fixed Download._repair logging stack trace for DownloadError (like 404s)

### DIFF
--- a/plugin/pulpcore/plugin/download/futures/batch.py
+++ b/plugin/pulpcore/plugin/download/futures/batch.py
@@ -1,15 +1,14 @@
+from collections import Iterator
 from concurrent.futures import ThreadPoolExecutor
 from gettext import gettext as _
-from logging import getLogger, DEBUG
-from queue import Queue, Empty
-from threading import Thread, current_thread
+from logging import getLogger
+from queue import Queue
+from threading import RLock, current_thread
 
 from .single import Context
 
 
-# The default queuing backlog.
-BACKLOG = 1024
-# The default concurrency.
+# The default number of downloads to be executed concurrently.
 CONCURRENT = 4
 
 
@@ -22,15 +21,9 @@ class Batch:
 
     Attributes:
         downloads (collections.abc.Iterable): An Iterable of downloads.
-        concurrent (int): The number of downloads to execute in concurrently.
+        concurrent (int): The number of downloads to be executed concurrently.
         iterator (PlanIterator): Used to iterate downloads as they complete.
         context (SharedContext): A shared download context.
-        feeder (DownloadFeeder): Used to feed submit downloads to the executor.
-        _is_shutdown (bool): Batch has been shutdown.
-
-    Notes:
-        * The batch should be used as a context manager.
-        * Or, `shutdown()` must be called manually.
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


plugin/pulpcore/plugin/download/futures/batch.py
```diff
@@ -52,33 +46,46 @@ class Batch:
...
+    def completed(self, plan):
         """
-        return self._is_shutdown
+        A plan has completed.
+
+        Args:
+            plan (Plan): The completed plan.
+        """
+        self.outstanding -= 1
+        for download in self.downloads:
```
mhrivnak: I don't think accessing an iterator is necessarily thread-safe, so it probably needs to acquire a lock here.

jortel: True.  good catch.

plugin/pulpcore/plugin/download/futures/batch.py
```diff
@@ -89,29 +96,13 @@ def download(self):
                 The iterator will yield the download `Plan` in the order completed.
         """
         log.debug(_('%(batch)s - download started'), {'batch': self})
-        self.feeder.start()
+        for download in self.downloads:
+            plan = Plan(download)
+            self.submit(plan)
+            if self.outstanding == (self.concurrent + 1):
```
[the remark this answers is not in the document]
jortel: I like the proposed changes here and in your PR re: moving outstanding to the iterator.

Although, I don't see the value in re-introducing BACKLOG.  What's the value of having more items in the pipeline than we have threads in the executor?  If you're concerned about momentary starvation, we could limit the pipeline to `2 x concurrent`.  It's unlikely but what if BACKLOG = 1024 and concurrent = 2048?  I know it's not very realistic but demonstrates the added complexity of having 2 values.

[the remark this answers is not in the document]
mhrivnak: Good point. The starvation concern is limited to the fact that calling ``next`` on the iterator might itself be expensive. Consider:

```
def xml_iterator():
  for item in somehow_parse_an_xml_file():
    artifacts = do_expensive_processing(item)
    for artifact in artifacts:
      yield artifact.as_http_download()
```

If that iterator is being used to feed the downloads, it's beneficial to do some processing in advance and maintain a buffer of available downloads that are ready to use.

I like the idea of making the buffer size a multiple of the concurrency number. For a very small amount of memory, we could make it 4x, 5x, 10x the concurrency level. It's hard to prove what will work best or how much it matters, but my gut says we should have at least a few downloads ready to go for each thread at any given time.

plugin/pulpcore/plugin/download/futures/batch.py
```diff
@@ -89,29 +96,13 @@ def download(self):
                 The iterator will yield the download `Plan` in the order completed.
         """
         log.debug(_('%(batch)s - download started'), {'batch': self})
-        self.feeder.start()
+        for i, download in enumerate(self.downloads):
```
mhrivnak: If getting downloads out of this iterator is even slightly expensive (as described in another comment), it's possible that the ``completed()`` method above might get called before this iteration is finished. So I think the lock should be acquired here as well.

jortel: Agreed.  Done.

PR #3155, opened 2017-09-19
GroupDownloader now validates

bmbouter:
The GroupDownloader was not correctly setting the expected_size and
expected_digests even though it could. This fixes the GroupDownloader so
that it sets those expected values causing validation to happen
correctly.

https://pulp.plan.io/issues/2951
re #2951


bmbouter:
Thank you!

PR #3156, opened 2017-09-19
Update Ansible install docs

werwty:


docs/installation/instructions.rst
```diff
@@ -19,10 +19,18 @@ Pulp 3 Ansible Galaxy Installation
...
    $ ansible-galaxy install -r requirements.yml -p ./roles
 
-5. Download ``deploy-pulp3.yml`` playbook and run it by specifying which broker is desired rabbitmq|qpidd to be installed::
+5. Configure your Ansible `inventory <http://docs.ansible.com/ansible/latest/intro_inventory.html>`_
+
+6. Download the ``deploy-pulp3.yml`` playbook and run it::
 
    $ wget https://raw.githubusercontent.com/pulp/devel/3.0-dev/ansible/deploy-pulp3.yml
-   $ ansible-playbook deploy-pulp3.yml --extra-vars "pulp3_broker=rabbitmq"
+   $ ansible-playbook deploy-pulp3.yml
```
mhrivnak: I suspect a lot of people will want to run this on the same machine that they're installing on, in which case they can just append `` -i localhost, -c local -K`` and skip step 5 above. Or we could provide an example inventory file with just that host and the connection type set. I think it's a good idea to make this clear as an option one way or the other, since it makes the experience so easy.

werwty:
@mhrivnak fixed

docs/installation/instructions.rst
```diff
@@ -19,10 +19,24 @@ Pulp 3 Ansible Galaxy Installation
...
-5. Download ``deploy-pulp3.yml`` playbook and run it by specifying which broker is desired rabbitmq|qpidd to be installed::
+5. To install Pulp locally, download the ``deploy-pulp3.yml`` playbook and run::
 
    $ wget https://raw.githubusercontent.com/pulp/devel/3.0-dev/ansible/deploy-pulp3.yml
-   $ ansible-playbook deploy-pulp3.yml --extra-vars "pulp3_broker=rabbitmq"
+   $ ansible-playbook deploy-pulp3.yml -i localhost, -c local -K
+
+.. note::
+
+    To do a non-local install of Pulp configure your Ansible
```
mhrivnak: s/Pulp/Pulp,/

PR #3158, opened 2017-09-19
Add docs for plugin writers

goosemania:


docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,106 @@
...
+* :class:`~pulpcore.plugin.models.PublishedArtifact` refers to :class:`~pulpcore.plugin.models.ContentArtifact` which is published and belongs to a certain :class:`~pulpcore.app.models.Publication`.
+* :class:`~pulpcore.plugin.models.PublishedMetadata` is a repository metadata which is published, located in ``/var/lib/pulp/published`` and belongs to a certain :class:`~pulpcore.app.models.Publication`.
+* :class:`~pulpcore.plugin.models.Publisher` knows specifics of the plugin :class:`~pulpcore.plugin.models.Content` to make it available outside of Pulp. :class:`~pulpcore.plugin.models.Publisher` defines how to publish content available in Pulp.
+* :class:`~pulpcore.app.models.Publication` is a result of publish operation of a specific :class:`~pulpcore.plugin.models.Publisher`.
```
jortel: Need to document :class:`~pulpcore.app.models.Distribution`

Each publisher has associated distributions which define how a publication is distributed.

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,106 @@
...
+
+  * in case of downloaded content, create instance of :class:`~pulpcore.plugin.models.Artifact` which refers to a downloaded file on a filesystem and contains calculated checksums for it.
+  * in case of downloaded content, update :class:`~pulpcore.plugin.models.ContentArtifact` with reference to the created :class:`~pulpcore.plugin.models.Artifact`.
+  * create and save instance of :class:`~pulpcore.plugin.models.RepositoryContent` to associate content to a repository.
+  * save all created artifacts and content: ``ExampleContent``, :class:`~pulpcore.plugin.models.ContentArtifact`, :class:`~pulpcore.plugin.models.RemoteArtifact`.
+
+ * Use :class:`~pulpcore.plugin.models.ProgressBar` to report progress of some steps if needed.
+
+Check our Example plugin implementation of importers: `one uses asyncio <link to pulp_example>`_ as a solution for downloading, `the other one follows concurrent.futures approach <link to pulp_example>`_.
+
```
jortel: I'm concerned that this does not include the File plugin which uses the `ChangeSet`.  The ExamplePlugin does a good job at illustrating all the details of what needs to be done but I really think we should be encouraging plugin writers to use the `ChangeSet`.

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,106 @@
...
+ * If a remote content should be downloaded right away (aka ``immediate`` download policy), use one of the suggested download solutions (:ref:`concurrent.futures <futures-docs>` or :ref:`asyncio <asyncio-docs>` approach) to start downloading. If content should be downloaded later (aka ``on_demand`` or ``background`` download policy), feel free to skip this step.
+ * Save all artifact and content data in one transaction:
+
+  * in case of downloaded content, create instance of :class:`~pulpcore.plugin.models.Artifact` which refers to a downloaded file on a filesystem and contains calculated checksums for it.
+  * in case of downloaded content, update :class:`~pulpcore.plugin.models.ContentArtifact` with reference to the created :class:`~pulpcore.plugin.models.Artifact`.
+  * create and save instance of :class:`~pulpcore.plugin.models.RepositoryContent` to associate content to a repository.
+  * save all created artifacts and content: ``ExampleContent``, :class:`~pulpcore.plugin.models.ContentArtifact`, :class:`~pulpcore.plugin.models.RemoteArtifact`.
```
jortel: We need to document that `Content` is shared between repositories and `Artifact` is shared between Content.  The plugin writer needs to understand that either may already exist and the importer needs to fetch and use them when they do.

docs/plugins/plugin-writer/first-plugin.rst
goosemania: The initial idea was to cover only models which plugin writers may use at some point.
But I already mentioned `Publication` so I guess it makes sense to add `Distribution` as well. Will do.

docs/plugins/plugin-writer/first-plugin.rst
goosemania: Good point! I'll add it.

docs/plugins/plugin-writer/first-plugin.rst
goosemania: Added

docs/plugins/plugin-writer/basics.rst
```diff
@@ -61,8 +65,34 @@ of a plugin app, and so on.
 This matching of module names is required for the Pulp Core to be able to auto-discover
 plugin components, particularly for both models and viewsets.
 
-Subclassing Importer, Publisher
--------------------------------
+Take a look at `the structure of our Example plugin <https://github.com/pulp/pulp_example/tree/master/pulp_example/app>`_.
```
bmbouter: Overall can we /our Example plugin/pulp_example/ ? And then near the top have 1 sentence which says that we have an example plugin called pulp_example? What do others think about this?

goosemania: Sounds good to me, no objections.

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,109 @@
 Writing Your First Plugin
 =========================
+
+For a complete list of what should be done to have a working plugin, check :doc:`checklist`.
+In this section key parts of plugin implementation are covered in more detail to help you as a plugin writer to get started.
+
+Understanding models
+--------------------
+There are models which are expected to be used in plugin implementation, so understanding what they are designed for is useful for a plugin writer.
```
bmbouter: Would it be possible to limit the lines to 100 chars and wrap the lines? We don't have a docs style guide, but it prevents the code from wrapping when its read on github. If not, we can fix it afterwards too.

goosemania: I can do it for the line like you pointed here but I'm not sure how to wrap lines which are part of list (aka start with `*`). If I try to do so, lists are no longer rendered correctly.

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,109 @@
...
+ * For every content which should be added to Pulp create but do not save yet:
+
+  * instance of ``ExampleContent`` which will be later associated to a repository.
+  * instance of :class:`~pulpcore.plugin.models.ContentArtifact` to be able to create relations with the artifact models.
+  * instance of :class:`~pulpcore.plugin.models.RemoteArtifact` to store information about artifact from remote source and to make a relation with  :class:`~pulpcore.plugin.models.ContentArtifact` created before.
+
+ * If a remote content should be downloaded right away (aka ``immediate`` download policy), use one of the suggested download solutions (:ref:`concurrent.futures <futures-docs>` or :ref:`asyncio <asyncio-docs>` approach) to start downloading. If content should be downloaded later (aka ``on_demand`` or ``background`` download policy), feel free to skip this step.
+ * Save all artifact and content data in one transaction:
+
+  * in case of downloaded content, create instance of :class:`~pulpcore.plugin.models.Artifact` which refers to a downloaded file on a filesystem and contains calculated checksums for it.
```
bmbouter: s/create/create an/

bmbouter: Actually should this be:  s/create/create and save an/ ? We want them to save this here right?

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,109 @@
...
+  * in case of downloaded content, update :class:`~pulpcore.plugin.models.ContentArtifact` with reference to the created :class:`~pulpcore.plugin.models.Artifact`.
```
bmbouter: s/update/update the/
s/with/with a/

docs/plugins/plugin-writer/first-plugin.rst
bmbouter: I think you indent the second line to start underneath the first character of the bulleted line. Like this?

```
* this is a really long line
  that won't end
```

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,135 @@
...
+  * create and save instance of :class:`~pulpcore.plugin.models.RepositoryContent` to associate content to a repository.
+  * save all created artifacts and content: ``ExampleContent``, :class:`~pulpcore.plugin.models.ContentArtifact`, :class:`~pulpcore.plugin.models.RemoteArtifact`.
+
+ * Use :class:`~pulpcore.plugin.models.ProgressBar` to report progress of some steps if needed.
```
bmbouter: s/report/report the/

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,135 @@
...
+
+
+There are several important aspects relevant to importer implementation which were briefly mentioned in :ref:`understanding-models` section:
```
bmbouter: s/in/in the/

docs/plugins/plugin-writer/first-plugin.rst
```diff
@@ -1,2 +1,135 @@
...
+ * :class:`~pulpcore.plugin.models.ContentArtifact` associates :class:`~pulpcore.plugin.models.Content` and :class:`~pulpcore.plugin.models.Artifact`. If :class:`~pulpcore.plugin.models.Artifact` is not downloaded yet, :class:`~pulpcore.plugin.models.ContentArtifact` contains ``NULL`` value for :attr:`~pulpcore.plugin.models.ContentArtifact.artifact`. It should be updated whenever corresponding :class:`~pulpcore.plugin.models.Artifact` is downloaded.
+
+Check our Example plugin implementation of importers: `one uses asyncio <https://github.com/pulp/pulp_example/blob/master/pulp_example/app/models.py#L529-L833>`_ as a solution for downloading, `the other one follows concurrent.futures approach <https://github.com/pulp/pulp_example/blob/master/pulp_example/app/models.py#L184-L526>`_.
+
+The importer implementation suggestion above allows plugin writer to have an understanding and control at a low level.
+Pulp team also offers plugin writers to try out a higher level, more simplified, API which introduces the concept of :class:`~pulpcore.plugin.changeset.ChangeSet`.
```
bmbouter: What about s/Pulp team also offers plugin writers to try out/The plugin API has?

PR #3159, opened 2017-09-20
Fix task results lost due to race condition

rohanpm:
A task might have completed or failed by the time apply_async attempts
to upsert a TaskStatus record.  We have to avoid touching the fields
set on success/failure (such as finish_time, result) during update,
otherwise we'll overwrite them with empty values.

closes #2959
https://pulp.plan.io/issues/2959

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -483,10 +483,12 @@ class Task(PulpTask, ReservedTaskMixin):
             task_id=async_result.id, task_type=self.name,
             state=constants.CALL_WAITING_STATE, worker_name=routing_key, tags=tag_list,
             group_id=group_id)
-        # To avoid the race condition where __call__ method below is called before
-        # this change is propagated to all db nodes, using an 'upsert' here and setting
-        # the task state to 'waiting' only on an insert.
-        task_status.save_with_set_on_insert(fields_to_set_on_insert=['state', 'start_time'])
+        # We're now racing with __call__, on_failure and on_success, any of which may
+        # have completed by now. To avoid overwriting TaskStatus updates from those callbacks,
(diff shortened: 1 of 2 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -483,10 +483,12 @@ def apply_async(self, *args, **kwargs):
...
             state=constants.CALL_WAITING_STATE, worker_name=routing_key, tags=tag_list,
             group_id=group_id)
-        # To avoid the race condition where __call__ method below is called before
-        # this change is propagated to all db nodes, using an 'upsert' here and setting
-        # the task state to 'waiting' only on an insert.
-        task_status.save_with_set_on_insert(fields_to_set_on_insert=['state', 'start_time'])
+        # We're now racing with __call__, on_failure and on_success, any of which may
+        # have completed by now. To avoid overwriting TaskStatus updates from those callbacks,
+        # we'll do an upsert and only touch the fields listed below if we've inserted the object.
+        task_status.save_with_set_on_insert(fields_to_set_on_insert=[
```
rohanpm: Brian suggested changing this to be more like the Pulp 3 implementation: try to create the object and catch/ignore exception if it already exists.  However this seemed like a riskier change, as I couldn't find where fields such as tags and task_type would end up being initialized, if not here.

bmbouter: Yeah this sounds fine. I mostly agree w/ the risk minimization approach. I agree the should also solve the bug.

rohanpm:
@bmbouter what should I do to progress this pull request?

bmbouter:
@rohanpm I was told from @dralley that he is testing it and will LGTM after testing.

dralley:
@rohanpm @bmbouter I'm currently running pulp-smash on it, if ends up clean, this is LGTM for me

PR #3160, opened 2017-09-20
Made some changes to the installation docs for clarity

dralley:
Changes:

- Made section header names more consistent
- Moved the SELinux permissive mode step, since it applies to all installation types, not just Ansible
- Added the command for setting SELinux to permissive mode
- Clarified that the python version (major *and* minor) found by virtualenv influences the path where the ```server.yaml``` file will be located.  It will be different on Fedora 25 and Fedora 26 because F26 uses Python 3.6, and the python version number becomes a component in the path name
- Added a command to exit the postgres user shell after setting up the database
- Added a reminder for the user to add the real location of ```server.yaml``` to the ```Environment``` variable in the systemd unit files

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -1,25 +1,30 @@
 Installation Instructions
 =========================
 
-Pulp 3 Ansible Galaxy Installation
-----------------------------------
+.. note::
+
+    As Pulp 3 currently does not have an SELinux Policy, it currently requires the target
+    machine to have SELinux set to permissive mode::
+
+    $ sudo setenforce 0
+
+Ansible Galaxy Installation
+---------------------------
 
 1. In order to be able to run playbook, ensure python3 and ansible( >2.2) are installed::
 
    $ sudo dnf install python3 ansible
 
-2. Target machine should have SELinux set to permissive mode.
-
-3. Target machine should have virtualenv installed::
+2. Target machine should have virtualenv installed::
 
    $ sudo pip3 install virtualenv
 
-4. Download ``requirements.yml`` and install Galaxy roles specified in the file::
+3. Download ``requirements.yml`` and install Galaxy roles specified in the file::
 
    $ wget https://raw.githubusercontent.com/pulp/devel/3.0-dev/ansible/requirements.yml
    $ ansible-galaxy install -r requirements.yml -p ./roles
 
-5. To install Pulp locally, download the ``deploy-pulp3.yml`` playbook and run::
+4. To install Pulp locally, download the ``deploy-pulp3.yml`` playbook and run::
 
    $ wget https://raw.githubusercontent.com/pulp/devel/3.0-dev/ansible/deploy-pulp3.yml
    $ ansible-playbook deploy-pulp3.yml -i localhost, -c local -K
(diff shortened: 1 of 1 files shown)


docs/installation/instructions.rst
```diff
@@ -1,25 +1,30 @@
...
+================================
 
-Pulp 3 Ansible Galaxy Installation
-----------------------------------
+.. note::
+
+    As Pulp 3 currently does not have an SELinux Policy, it currently requires the target
+    machine to have SELinux set to permissive mode::
+
+    $ sudo setenforce 0
```
bmbouter: This is more of an FYI right, because the galaxy playbook and the vagrant env both disable selinux. The way its written now makes it seem like the user needs to take this step but in practice I don't think they will. I recommend moving this note further down so users don't think this is the first thing they need to do.

bmbouter: Oh I see this is a step that users were being instructed to take. I guess it's good that people be required to turn it off manually. This is good as is, please disregard my comments.

docs/installation/instructions.rst
```diff
@@ -1,25 +1,30 @@
-Installation Instructions
-=========================
+Pulp 3 Installation Instructions
```
bmbouter: The entire docs site is for Pulp 3 so I think s/Pulp 3 // from this line.

docs/installation/instructions.rst
```diff
@@ -1,25 +1,30 @@
...
-Pulp 3 Ansible Galaxy Installation
-----------------------------------
+.. note::
+
+    As Pulp 3 currently does not have an SELinux Policy, it currently requires the target
+    machine to have SELinux set to permissive mode::
+
+    $ sudo setenforce 0
+
+Ansible Galaxy Installation
```
bmbouter: I like that this heading also does not contain the `Pulp 3`.

docs/installation/instructions.rst
```diff
@@ -38,15 +43,15 @@ Pulp 3 Ansible Galaxy Installation
...
+PyPI Installation
+-----------------
 
 .. tip::
 
     These are the manual steps to install Pulp. There are Ansible roles that will do all
     of the following for you.
 
-1. Install python3.5 and virtualenv::
+1. Install python3.x and virtualenv::
```
bmbouter: We require 3.5 specifically.

docs/installation/instructions.rst
```diff
@@ -61,8 +66,11 @@ PyPI
    $ pip3 install pulpcore
 
 4. If the the server.yaml file isn't in the default location of `/etc/pulp/server.yaml`, set the
-   PULP_SETTINGS environment variable to tell pulp where to find you server.yaml file.
-   ``export PULP_SETTINGS={virtualenv}/lib/python3.5/site-packages/pulpcore/etc/pulp/server.yaml``
+   PULP_SETTINGS environment variable to tell pulp where to find you server.yaml file (note that
```
bmbouter: s/pulp/Pulp/

bmbouter: The parentheses contain a complete sentence, can the parens be removed and that turned into its own sentence with `Note`... ?

docs/installation/instructions.rst
```diff
@@ -61,8 +66,11 @@ PyPI
...
+   the exact path will depend on the major *and* minor python version found by virtualenv e.g.
```
bmbouter: s/python/Python/

docs/installation/instructions.rst
dralley: My thinking was that since it was currently listed as a step in the Ansible Galaxy instructions, it's nbd to just move it out and have it apply to all of them.

bmbouter: Yeah I thought the galaxy playbook did this for you, but since it doesn't this is good.

PR #3162, opened 2017-09-21
Minor install docs tweak

dralley:


docs/installation/instructions.rst
```diff
@@ -229,7 +229,7 @@ pulp_resource_manager::
     WantedBy=multi-user.target
 
 
-pulp_worker@::
+```pulp_worker@```.service::
```
daviddavis: why is service outside the backticks here?

dralley: agh, yup, you're right!

PR #3163, opened 2017-09-21
Remove db-reset script

asmacdo:
Where db-reset is used in this repo, it is replaced with pulp-manager use
directly.
https://pulp.plan.io/issues/3012
To be merged ~along with~ after https://github.com/pulp/devel/pull/91

re: #3012


.travis.yml
```diff
@@ -25,7 +25,7 @@ before_script:
 script:
     # flake8
     - "flake8 --config flake8.cfg"
-    # tests can't run without migrations being made
-    - "python manage.py makemigrations pulp_app"
     # chain these so we don't try to run tests if the db reset fails
-    - "platform/pulpcore/app/db-reset.sh && python manage.py test"
+    - "pulp-manager makemigrations pulp_app --noinput"
+    - "pulp-manager migrate auth --noinput"
+    - "pulp-manager migrate --noinput && pulp-manager test"
```
asmacdo: @dkliban, do you think that chaining only the final migrate will give us the intended result?

dkliban: maybe you can put an after_success section after the migrate step. and run test in there.

dkliban: https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

asmacdo: After IRC discussion, we decided to try moving this around in a separate change.

PR #3168, opened 2017-09-22
UnicodeEncodeError in comments provided with SSL cert/key/CA

ammaritiz:
This is an additional fix needed in Pulp apart from
https://github.com/pulp/nectar/pull/63, since Pulp writes
out the SSL certs to a different path than nectar

closes #2960

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -359,7 +359,7 @@ class Importer(AutoRetryDocument):
                 misc.mkdir(os.path.dirname(self._pki_path))
                 os.mkdir(self._pki_path, 0700)
             with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0600), 'w') as pem_file:
-                pem_file.write(self.config[config_key])
+                pem_file.write(self.config[config_key].encode('utf-8'))
 
 
 signals.pre_delete.connect(Importer.pre_delete, sender=Importer)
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -359,7 +359,10 @@ def _write_pem_file(self, config_key, path):
                 misc.mkdir(os.path.dirname(self._pki_path))
                 os.mkdir(self._pki_path, 0700)
             with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0600), 'w') as pem_file:
-                pem_file.write(self.config[config_key])
+                try:
+                    pem_file.write(self.config[config_key])
+                except UnicodeEncodeError:
+                    pem_file.write(self.config[config_key].encode('ascii', 'replace'))
```
goosemania: Same as in pulp/nectar#63, I think you need `self.config[config_key].encode('utf-8')`.

goosemania: And I guess no `try/except` block is needed, you can always encode the string into UTF-8.

dralley:
This looks good to me.  One thing you might want to do is make the strings on the right hand side of the assertions unicode literals, e.g. ```u'CA Cert'```.  I'm like 99.9% certain it doesn't matter in this instance though.

  @goosemania, did you have any other reservations?

ammaritiz:
@goosemania I have fixed the pep8speaks errors. Thanks for reviewing the PR.

PR #3170, opened 2017-09-25
Fix syntax error in documentation

Ichimonji10:
The reST specification states that inline code must be identified by
double backticks. Triple backticks have no special meaning.

Triple backticks are used to denote `<pre>` blocks in various markdown
derivatives, such as CommonMark and GitHub-flavored Markdown.


dralley:
My bad!  Thanks for catching this @Ichimonji10

PR #3173, opened 2017-09-25
Fix the traceback during publisher deletion.

ipanova:
closes #3011
https://pulp.plan.io/issues/3011


dralley:
Looks good @ipanova !

PR #3174, opened 2017-09-25
Remote artifact created with correct digests.

jortel:
https://pulp.plan.io/issues/3017

Mainly separated the _pending_ and _stored_ models:  `PendingArtifact._model` and `PendingArtifact_stored_model`.

### DIFF
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -66,7 +66,7 @@ class Artifact(Model):
         Returns:
             bool: True when equal.
         """
-        for field in Artifact.RELIABLE_DIGEST_FIELDS[:-1]:
+        for field in Artifact.RELIABLE_DIGEST_FIELDS:
             digest = getattr(self, field)
             if not digest:
                 continue
--- a/plugin/pulpcore/plugin/changeset/model.py
+++ b/plugin/pulpcore/plugin/changeset/model.py
@@ -18,15 +18,15 @@ class Pending:
     Represents content related things contained within the remote repository.
 
     Attributes:
-        model (Model): A pending (wanted) model instance.
-        fetched (bool): model has been fetched from the DB.
+        _model (django.db.models.Model): A pending (wanted) model instance.
+        _stored_model (django.db.models.Model): The model stored in the database.
         _settled (bool): All matters are settled and the object is ready
             to be (optionally created) and added to the repository.
     """
 
     __slots__ = (
-        'model',
-        'fetched',
+        '_model',
+        '_stored_model',
         '_settled'
     )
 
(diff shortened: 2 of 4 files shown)


platform/pulpcore/app/models/content.py
```diff
@@ -66,7 +66,7 @@ def is_equal(self, other):
         Returns:
             bool: True when equal.
         """
-        for field in Artifact.RELIABLE_DIGEST_FIELDS[:-1]:
+        for field in Artifact.RELIABLE_DIGEST_FIELDS:
```
jortel: Noticed this and fixed it.

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -176,15 +184,18 @@ def save(self):
...
+        When this happens, the model is fetched and _stored_model is updated.
         """
         with transaction.atomic():
-            try:
-                with transaction.atomic():
-                    self.model.save()
-            except IntegrityError:
-                content = type(self.model).objects.get(**self.key)
-                self.update(content)
+            if not self._stored_model:
```
goosemania: Why is this check needed? To decrease reads from DB (`IntegrityError` case)?

platform/pulpcore/app/models/content.py
goosemania: Nice catch

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -306,57 +339,50 @@ def save(self):
...
-                content=self.content.model)
-            if self.fetched:
-                content_artifact.artifact = artifact
+                content=self.content.stored_model)
+            if self._stored_model:
+                content_artifact.artifact = self._stored_model
                 content_artifact.save()
 
-        digests = {f: getattr(self.model, f) for f in Artifact.DIGEST_FIELDS}
+        digests = {f: getattr(self._model, f) for f in Artifact.DIGEST_FIELDS}
```
goosemania: This fixes issue because you now take data from "pending" model here and not from the stored one. And before this PR there was no such separation and you updated the only model you had. Correct?

goosemania: Heh, now I see your comment to this PR about separation between pending and stored models :)

plugin/pulpcore/plugin/changeset/model.py
jortel: Pretty much.  Having the `_stored_model` set indicates that the content already existed in the DB.  Saving would certainly cause `IntegretyError` which would result in an unnecessary update.  Both the insert and update would be unnecessary.  The `IntegretyError` is there to catch race conditions whereby the content didn't exist when the changeset looked but exists when another process creates it.

plugin/pulpcore/plugin/changeset/model.py
jortel: Exactly.

PR #3177, opened 2017-09-27
	Reconnecting to mongodb after celery fork

daviddavis:


bmbouter:
This patch makes sense to me. When the forker is reforked this signal will cause the db to re-initialize. Can this be tested some with pulp-smash. I'd also like @dralley to lgtm it so he is aware of this change too. Thanks @daviddavis

dralley:
@bmbouter @daviddavis This makes sense to me also, although I would like to play around with it a bit.

server/pulp/server/async/app.py
```diff
@@ -179,6 +180,11 @@ def initialize_worker(sender, instance, **kwargs):
         get_resource_manager_lock(sender)
 
 
+@worker_process_init.connect
+def reconnect_services(sender, **kwargs):
```
dralley: Add a simple docstring here explaining the purpose of this, perhaps linking to the issue.  See ```initialize_workers()``` for reference but no need to be so comprehensive.

server/pulp/server/db/connection.py
```diff
@@ -335,3 +336,13 @@ def get_connection():
     :rtype:  pymongo.connection.Connection
     """
     return _CONNECTION
+
+
+def reconnect():
```
dralley: Add a docstring here too

dralley:
Looks fine to me apart from the two places I marked for adding docstrings.

daviddavis:
@dralley added docstrings to the functions. I am running this code tonight one final time to test for deadlocks. Hopefully if the docstrings look good and there are no deadlocks, we can merge. Thanks for the review.

PR #3178, opened 2017-09-28
Reimplements the custom FileStorage backend

dkliban:
The FileSystemStorage backend provided by Django has been extended to conform to Pulp needs. When saving a file, the storage backend does nothing if a file by that name already exists. All TemporaryDownloadedFile and TemporaryUploadedFile objects are moved using os.rename() when possible. 

The new ArtifactFileField allows plugin writers to instantiate an Artifact with a string path to a downloaded file. This helps the plugin writer avoid having to open and close a downloaded file. The pre_save()[0] method instantiates a TemporaryDownloadedFile object and passes it to the storage backend to save into place. The save()[1] method on the Artifact then closes the file opened by pre_save(). 

The get_artifact_path() and published_metadata_path() static methods are now helper
methods that can be used with any storage backend. In particular, I expect these to
work with S3 storage backend without any modification.

The signal handler for pre_delete of the Artifact model has also been removed. This handler relied on the custom storage backend to remove empty directories left behind after an Artifact has been removed. This is not necesary because there can only be 255 directories inside /var/lib/pulp/artifact. It is highly unlikely that a directory will remain empty for long.

[0] https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.Field.pre_save    
[1] https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.save

closes: #2950
https://pulp.plan.io/issues/2950


platform/pulpcore/app/models/storage.py
```diff
@@ -205,13 +117,12 @@ def _save(path, content):
...
         os.makedirs(os.path.dirname(path), exist_ok=True)
-
         try:
             try:
-                os.link(content.name, path)
+                os.link(content.file.name, path)
             except OSError as e:
                 if e.errno in (errno.EXDEV, errno.ENOENT):
-                    FileSystem.copy(path, content)
+                    super()._save(path, content)
```
jortel: Most (perhaps all) of the unwanted behavior is in FileSystemStorage._save().
- directory creation is not thread safe.
- new file-name is created when file already exists.
 - uses os.rename() instead of link.

It also tries to create the dir and rename the file which is 2 unnecessary system calls.

However in this case, it's only being called when the directory has already been created and the link() would raise `FileExistsError`.  So it works for now.  What if other undesirable behaviors get introduced in `FIleSystemStorage._save()`?  I don't think removing `FileSystem.copy()` and saving 11 lines of code is worth the risk of using `FIleSystemStorage._save()`.

jortel:
Thanks for the PR, @dkliban.

While you're at it, how about changing `TLSLocation` to module functions as well for consistency.  You'll need to update the `Importer` model.

platform/pulpcore/app/models/storage.py
```diff
@@ -225,176 +136,35 @@ def _save(path, content):
...
-            str: The absolute storage path.
-        """
-        return os.path.join(
-            settings.MEDIA_ROOT,
-            'published',
-            'metadata',
-            str(model.pk),
-            name)
-
-    def get_available_name(self, name, max_length=None):
```
jortel: This method is used to derive a new unique file-name for the path in cases where the file already exists.  We never want this to happen for artifacts.  To be safe, Let's override and raise an exception.

platform/pulpcore/app/models/storage.py
```diff
@@ -225,176 +136,35 @@ def _save(path, content):
...
-        files = []
-        directories = []
-        for entry in os.listdir(path):
-            if os.path.isdir(os.path.join(path, entry)):
-                directories.append(entry)
-            else:
-                files.append(entry)
-        return (directories, files)
-
-    def path(self, name):
```
jortel: The `FileSystemStorage.path` uses a `safe_path()` that does 2 things.

1. Ensure that absolute paths are within the MEDIA_ROOT.  This is good.

2. For relative paths, it joins the MEDIA_ROOT & path.  This is bad.  For artifacts, we never want the storage to change the storage location.

This should be overriden to raise an exception whenever called.

platform/pulpcore/app/models/storage.py
dkliban: Thank you for the feedback. I read through the _save() method in Django and  it looks like it already does what we want it to do for TemporaryUploadedFile objects. When Django sees that the file object has a temporary_file_path method, it attempts to move the file instead of copying it. 

Why is the directory creation not thread safe? It looks like it tries to create the repo and if it already exists it keeps going

Instead of implementing the _save() method we can implement save() and get_available_name(). The get_available_name() method should raise an Exception when the file already exists. save() can catch that exception and just keep going without doing any other operations.

I am not sure that I understand why os.rename is not desired for our use case. Could you please elaborate?

platform/pulpcore/app/models/storage.py
dkliban: I don't think we need to worry about enforcing the paths. I don't think plugin writers will be providing paths for where to save files. I am not 100% sure on this, so if you can provide an example of when we need to guard against this, that would help me understand this requirement better.

jortel: I don't think making assumptions about how the class is used is a good practice.  This storage class may be used by every model in pulp.  Both core and plugin models.  Plugin writers could add `FileField` to any of their models.  Passing a relative path could be unintentional by the plugin writer or a coding error.

The `path()` method goes beyond the `Storage` API and adds the behavior of joining relative paths to the MEDIA_ROOT.  There is no guarantee that any other storage back-end will do this so pulp should not rely on it.  Looking at S3 storage back-end, it uses the path *as-is* for the object key.

We should continue to override this method as suggested (or as it originally was) to ensure consistent, deterministic  behavior across storage back-ends.

platform/pulpcore/app/models/storage.py
```diff
@@ -101,300 +97,86 @@ def __call__(self, model, name):
...
-        if root == path:
-            return
-        try:
-            os.rmdir(path)
-        except OSError as e:
-            if e.errno in [errno.ENOENT, errno.ENOTEMPTY]:
-                return
-            else:
-                raise
+        if self.exists(name):
```
jortel: This is not thread safe.

platform/pulpcore/app/models/storage.py
```diff
@@ -101,300 +97,86 @@ def __call__(self, model, name):
...
-
-        Returns:
-            File: An open file object.
-        """
-        return File(open(path, mode))
+            return name
 
-    @staticmethod
-    def _save(path, content):
+    def save(self, name, content, max_length=None):
```
bmbouter: The `name` is called `path` in the docstring.

platform/pulpcore/app/models/storage.py
```diff
@@ -101,300 +97,86 @@ def __call__(self, model, name):
...
-            A tuple of two lists: (directories, files).
-        """
-        files = []
-        directories = []
-        for entry in os.listdir(path):
-            if os.path.isdir(os.path.join(path, entry)):
-                directories.append(entry)
+        try:
+            name = self.get_available_name(name, max_length=max_length)
+            return self._save(name, content)
```
bmbouter: Are you catching exceptions from this line? If not, it can move to an `else` block.

platform/pulpcore/app/models/storage.py
jortel: The `Storage` base class intends subclasses to implement `_save()`.

platform/pulpcore/app/models/storage.py
dkliban: i am catching exceptions in self._save(). self.get_available_name() can get called again during self._save() if the file already exists.

platform/pulpcore/app/models/storage.py
dkliban: Yes, but this method is called again later in _save() if a file exists exception is raised.

platform/pulpcore/app/models/storage.py
jortel: Delegating everything FileSystemStorage._save() has reintroduced a lot of bad and unwanted behavior.
- I thought there was a race condition flaw with directory creation logic but after looking again it does catch EEXIST.  So, although the logic is poor, it works.
- Problems with `file_move_safe()`
  - Logic related to *allow_overwrite* has race condition flaw.
  - Uses `os.reneme()` which removes the source (storage should not do this) and has the _nasty_ behavior of silently overwriting the target.
- Determines new filename when already exists.

What you had before with the _save() was better.  Just needed to deal with `path()` and `get_available_name()` overrides.

platform/pulpcore/app/models/storage.py
```diff
@@ -101,300 +97,86 @@ def __call__(self, model, name):
...
+        except Exception as e:
```
jortel: Should be catch IOError or OSError, not Exception.

plugin/pulpcore/plugin/download/file.py
```diff
@@ -0,0 +1,29 @@
+import errno
+
+from django.core.files import File
+
+
+class DownloadedFile(File):
```
jortel: This is not part of the `File` interface.  We should not be doing special things so that a specific `Storage` will work correctly.

platform/pulpcore/app/models/storage.py
jortel: The `Storage` interface does not document that this method should raise `FileExistsError`.  This behavior is convenient for changing the logic in _save() but inconsistent with the `Storage` interface and dangerous.  What if the `FileSystemStorage._save()` logic changes?  This could result in unpredictable behavior.

platform/pulpcore/app/models/storage.py
```diff
@@ -101,300 +97,86 @@ def __call__(self, model, name):
...
-            return
-        try:
-            os.rmdir(path)
-        except OSError as e:
-            if e.errno in [errno.ENOENT, errno.ENOTEMPTY]:
-                return
-            else:
-                raise
+        if self.exists(name):
+            raise OSError(errno.EEXIST, "File Exists")
```
jortel: Side note: better to use `FileExistsError` instead.

jortel:
@dkliban, I think your original PR was on the right track to making some good improvements.  Please consider going back to that approach.

plugin/pulpcore/plugin/download/file.py
dkliban: I borrowed this code from the TemporaryUploadedFile[0]. The FileSystemStorage is designed to work with 2 types of uploaded files: In memory or on disk. Our upload API always writes files to disk. So when we hand an uploaded file to the FileSystemStorage, the file system storage calls file_move_safe(). I would like us to handle temporary downloaded files in the same manner.



[0] https://github.com/django/django/blob/1.11/django/core/files/uploadedfile.py#L59

jortel: The original (code you removed) FileSystem._save() handled both in memory and on disk files with very little complexity and code.  It cleanly supported both linking and copying.  And, did not need the `content` to be a special type of `File` to do so.  The `FileSystemStorage._save()` seems over engineered and does not behave like pulp needs it to.  I don't see the point to writing code to coerce FileSystemStorage._save() to behave the way we want instead of just overriding _save(), path() and get_available_name().

After reviewing the `Storage` and `FileSystemStorage` code (again),  I still think the original FileSystem.path() and FileSystem.get_available_name() overrides that returned *name* was the correct thing to do.

I'm all for leveraging django code as much as possible so long as it's safe, practical and provides desirable functionality.

dkliban:
ok test

platform/pulpcore/app/models/storage.py
```diff
@@ -1,400 +1,110 @@
...
-            path(str): The storage path.
-            content (File): Source file object.
+            Name of the file.
 
         Raises:
-            FileExistsError: File already stored.
-            OSError: On error.
+            OSError if a file with requested name already exists.
         """
-        tmp_path = '.'.join((path, str(uuid4()), 'tmp'))
```
jortel: This tmp file logic ensures that a file being copied into storage is atomic.  This ensures that if the process is killed, it cannot leave a half-baked file in place.  If half-baked, the logic will think the file already exists in subsequent save requests.  Best I can tell, the FileSystemStorage copy is not atomic.  Am I missing something?

jortel:
There are a lot of good changes in the PR:
- Moving the class methods on the storage to module functions for determining path.
- Subclassing FileSystemStorage to reduce implemented methods.  Not a lot of value but good none the less.

Summary of concerns:
- Overriding get_available_name() to trick the logic in FileSystemStorage._save() to not store the file with a new name when it already exists instead of overriding FileSystemStorage._save()  is brittle.  It's based on knowledge of the superclass implementation of _save() which is never a good idea.  Especially when the logic is not specified in the Storage interface.  Any changes to the impl of FileSystemStorage._save() will have unpredictable results.
- The FileSystemStorage._save() uses os.rename() which removes the source file and will blindly overwrite the target file if given the chance.  The source file should not be removed.  There is nothing in the Storage interface that suggests this.  And, it's presumptuous not to mention inconsistent with copy. 
- The FileSystemStorage._save() has logic Pulp does not want.  Why work so hard to reuse it.
- The FileSystemStorage._save() copy logic is not atomic.  If interrupted the half-baked file will prevent the file from ever being stored in subsequent save attempts.
- I don't understand the value of the TemporaryDownloadFile.
- Using the File with FileField is standard django, I don't see the value in adding code to hide this from plugin writers.

daviddavis:
I think there are some tests in `test_content.py` that should be removed/updated.

platform/pulpcore/app/files.py
```diff
@@ -43,3 +43,18 @@ def receive_data_chunk(self, raw_data, start):
...
         for hasher in hashlib.algorithms_guaranteed:
             self.file.hashers[hasher].update(raw_data)
+
+
+class TemporaryDownloadedFile(TemporaryUploadedFile):
+    """
+    A temporary downloaded file.
+
+    The FileSystemStorage backend treats this object the same as a TemporaryUploadedFile. The
+    storage backend attempts to link the file to it's final location. If the final location is on a
```
daviddavis: s/it's/its/

same thing on the next line too.

platform/pulpcore/app/models/storage.py
```diff
@@ -1,400 +1,110 @@
...
-        >>> from django.db import models
-        >>>
-        >>> class Person(models.Model):
-        >>>     ssl_certificate = models.FileField()
-        >>>
-        >>> person = Person()
-        >>> certificate = 'PEM ENCODED CERTIFICATE HERE'
-        >>> person.ssl_certificate = FileContent(certificate)
-        >>>
+    Modifies how Django saves files
```
daviddavis: A better description of how this class differs from the Django one would be helpful.

platform/pulpcore/app/models/storage.py
```diff
@@ -1,400 +1,162 @@
...
+    TemporaryUploadedFile is the most efficient for Pulp's use case. However, the behavior from 2b
+    can result in duplicate files being created, but not associated with any Artifact. There is a
+    race condition if the same file is being uploaded in two parallel requests. One of the requests
+    will create the Artifact in /var/lib/pulp/artifact and then save it to the database. The second
+    request will also create a file in /var/lib/pulp/artifacts, but it will have a random name. The
+    database insert will fail due to uniqueness constraint and the user will receive a 400 error.
+    No cleanup happens when this occurs.
+
+    Here is how FileSystem saves a file to its destination for the two File types we are
+    interested in:
```
daviddavis: This seems like a repeat of the stuff above.

platform/pulpcore/app/models/content.py
```diff
@@ -75,6 +76,21 @@ def is_equal(self, other):
...
+        Saves Artifact model and closes the file associated with the Artifact
+
+        Args:
+            args (list): list of positional arguments for Model.save()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.save()
+        """
+        try:
+            super().save(*args, **kwargs)
+            self.file.close()
+        except:
```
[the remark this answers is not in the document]
dkliban: We don't care about what kind of exception occured here. We just want to make sure that we close the file before reraising.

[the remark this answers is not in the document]
daviddavis: I'm almost certain you mean `except Exception:` though. More info:

https://docs.python.org/2/howto/doanddont.html#except

PR #3180, opened 2017-10-04
fix task cancel

asmacdo:
https://pulp.plan.io/issues/3054

### DIFF
--- a/platform/pulpcore/app/viewsets/task.py
+++ b/platform/pulpcore/app/viewsets/task.py
@@ -3,12 +3,13 @@ from django_filters.rest_framework import filterset
 from pulpcore.app.models import Task, Worker
 from pulpcore.app.serializers import TaskSerializer, WorkerSerializer
 from pulpcore.app.viewsets import NamedModelViewSet
+from pulpcore.app.viewsets.base import GenericNamedModelViewSet
 from pulpcore.app.viewsets.custom_filters import CharInFilter
 from pulpcore.tasking.util import cancel as cancel_task
 
 from rest_framework.decorators import detail_route
 from rest_framework.response import Response
-from rest_framework import status
+from rest_framework import status, mixins
 
 
 class TaskFilter(filterset.FilterSet):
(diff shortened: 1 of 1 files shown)


platform/pulpcore/app/viewsets/task.py
```diff
@@ -19,18 +20,19 @@ class Meta:
         fields = ['state', 'worker__name', 'tags']
 
 
-class TaskViewSet(NamedModelViewSet):
+class TaskViewSet(mixins.RetrieveModelMixin,
+                  mixins.ListModelMixin,
+                  GenericNamedModelViewSet):
```
goosemania: Do you have any objections to create a base class `ReadNamedModelViewSet` like [this one](https://github.com/asmacdo/pulp/blob/3104d240fc7d4e2a6a882420b095a43edda09686/platform/pulpcore/app/viewsets/base.py#L149)? I don't think we would need all the combinations of different mixins, likely there will be several common combinations. What do you think?

platform/pulpcore/app/viewsets/task.py
```diff
@@ -19,18 +20,19 @@ class Meta:
...
     queryset = Task.objects.all()
     serializer_class = TaskSerializer
     endpoint_name = 'tasks'
     filter_class = TaskFilter
-    http_method_names = ['get', 'options']
 
-    @detail_route(methods=['post'])
+    @detail_route(methods=('post',))
```
goosemania: Why does it fix the issue? Is it because by not including "create" mixin there is no default handler for POST but POST itself is not explicitly disallowed and you can specify it in detail route?

asmacdo: This line actually doesn't change anything, I just switched to a tuple to be consistent with other parts of the code.

goosemania: Yeah, I just meant that you can specify POST while it seems to be disallowed for the viewset as a whole. But I think my understanding is correct that it's not actually disallowed but default "create" method is just not defined. Never mind :) Thanks

platform/pulpcore/app/viewsets/task.py
asmacdo: I think that those base classes only make sense in certain circumstances. For instance, Repositories, Importers, Publishers, Distributions, (and Publications?) will all share a set of behaviors. Synchronous create, list, and retrieve. Async update and delete. To me, it makes sense to have a base viewset for those because it is applicable to so many viewsets. I don't think we have that here, so I'd rather use mixins directly.

Mostly, I think that direct mixin use is more clear, less error prone, and simpler to user. You don't have to hunt for the correct base class, you don't have to guess what class names mean, and you don't have to leave viewsets.base.py open to understand the behaviors of your viewset.

goosemania: I'm not convinced but not against either :)

platform/pulpcore/app/viewsets/task.py
```diff
@@ -19,18 +20,19 @@ class Meta:
...
 
-class TaskViewSet(NamedModelViewSet):
+class TaskViewSet(mixins.RetrieveModelMixin,
+                  mixins.ListModelMixin,
+                  GenericNamedModelViewSet):
     queryset = Task.objects.all()
     serializer_class = TaskSerializer
     endpoint_name = 'tasks'
     filter_class = TaskFilter
-    http_method_names = ['get', 'options']
```
asmacdo: @goosemania, The `create` method was defined by the create mixin, which was inherited from the NamedModelViewSet, but POST itself was disallowed for every views that is a part of the TaskViewSet. I'm going to merge, but lets chat about this and make sure we are on the same page.

PR #3182, opened 2017-10-05
Removing natural key fields attribute

daviddavis:
refs #3025
https://pulp.plan.io/issues/3025

### DIFF
--- a/platform/pulpcore/app/models/content.py
+++ b/platform/pulpcore/app/models/content.py
@@ -4,6 +4,7 @@ Content related Django models.
 from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import models
+from itertools import chain
 
 
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation
@@ -79,10 +80,6 @@ class Content(MasterModel):
     """
     A piece of managed content.
 
-    Attributes:
-
-        natural_key_fields (tuple): Natural key fields.  Must be models.Field subclasses.
-
     Relations:
 
         notes (GenericKeyValueRelation): Arbitrary information stored with the content.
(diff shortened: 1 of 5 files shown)


platform/pulpcore/app/models/content.py
```diff
@@ -90,22 +90,18 @@ class Content(MasterModel):
...
-    def natural_key(self):
-        """
-        Get the model's natural key based on natural_key_fields.
-
-        :return: The natural key.
-        :rtype: tuple
-        """
-        return tuple(getattr(self, f.name) for f in self.natural_key_fields)
+    @classmethod
+    def natural_key_fields(cls):
```
dkliban: Want to add a docbloc here?

platform/pulpcore/app/models/content.py
```diff
@@ -79,24 +79,28 @@ class Content(MasterModel):
...
     artifacts = models.ManyToManyField(Artifact, through='ContentArtifact')
 
     class Meta:
         verbose_name_plural = 'content'
+        unique_together = ()
+
+    @classmethod
+    def natural_key_fields(cls):
+        """
+        Returns a list of the natural key fields which usually equates to unique_together fields
```
dkliban: This is a generator actually, so it yields field names as you iterate over the generator. If you want to convert it to return a list, I think that would be alright in this case. Either way, the docstring should match the behavior.

platform/pulpcore/app/models/content.py
```diff
@@ -79,24 +79,26 @@ class Content(MasterModel):
...
+        """
+        return [field for fields in cls._meta.unique_together for field in fields]
```
jortel: Did you consider using?: https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable
Like:
```
return list(chain.from_iterable(cls._meta.unique_together))
```

jortel:
👍 I really like just doing this once the base class.

daviddavis:
ok test

daviddavis:
@jortel @dkliban I think this is ready to be reviewed again. Thanks for the reviews.

PR #2994, opened 2017-04-12
Update systemd tasking service files

BrnoPCmaniak:
closes #2479
https://pulp.plan.io/issues/2479

PR #3183, opened 2017-10-06
Update timestamp of unit association for all repositories

goosemania:
To ensure that next publish of all the repo copies is operational.

closes #2994
https://pulp.plan.io/issues/2994

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -317,6 +317,10 @@ def associate_single_unit(repository, unit):
     """
     Associate a single unit to a repository.
 
+    Association can happen for various reasons, including the case when unit has been updated.
+    Updating timestamp for every repo when unit is associated ensures that the next publish
+    of all the repositories containing this unit will be operational.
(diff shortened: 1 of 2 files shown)


goosemania:
I'm not sure it's the best solution because there will be less no-op publishes. I'm open for other ideas.

goosemania:
F24 test failure is unrelated to this change.

goosemania:
ok test

goosemania:
One of the use cases which is affected by this fix is multiple copies of the same repo (like Katello content views). When yet another copy of the repo is created, all the previous copies will have operational publish (in 99% cases it's unnecessary).
It makes it even worse for the "smart proxy" use case, because more repositories will need to be synced to "smart proxy".

I'm closing this PR, because the proposed fix has substantial impact on sync/publish performance in certain cases. The issue itself (no-op publish in copies of the repo if an erratum update is the only change) will be solved automatically with the next changes to the affected repository copies, or it can be overcome using force_full publish.

PR #3195, opened 2017-10-17
Return the repository content href so they can be deleted

daviddavis:
fixes #3073
https://pulp.plan.io/issues/3073


daviddavis:
ok test

PR #3198, opened 2017-10-19
Add uniquenessand length validators to serializer fields.

werwty:
Move username regex validator from model to serializer, since the model validation
does not run automatically on model save:
https://docs.djangoproject.com/en/dev/ref/validators/#how-validators-are-run

closes #2984
https://pulp.plan.io/issues/2984

### DIFF
--- a/platform/pulpcore/app/serializers/repository.py
+++ b/platform/pulpcore/app/serializers/repository.py
@@ -1,6 +1,9 @@
 from gettext import gettext as _
 
+from django.core import validators
+
 from rest_framework import serializers
+from rest_framework.validators import UniqueValidator, UniqueTogetherValidator
 from rest_framework_nested.serializers import NestedHyperlinkedModelSerializer
 
 from pulpcore.app import models
@@ -25,7 +28,8 @@ class RepositorySerializer(ModelSerializer):
         lookup_field='name',
     )
     name = serializers.CharField(
-        help_text=_('A unique name for this repository.')
+        help_text=_('A unique name for this repository.'),
+        validators=[UniqueValidator(queryset=models.Repository.objects.all())]
     )
 
     description = serializers.CharField(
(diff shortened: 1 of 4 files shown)


asmacdo:
Thanks @werwty! There are at least a couple more. The "name" field on Importers and Publishers needs a UniqueValidator too.

platform/pulpcore/app/serializers/user.py
```diff
@@ -49,11 +63,17 @@ class UserSerializer(ModelSerializer):
     jwt_secret = serializers.CharField(
         help_text=_("User JWT authentication secret"),
         required=False,
-        write_only=not settings.DEBUG  # If pulp in DEBUG mode secret is visible
+        write_only=not settings.DEBUG,  # If pulp in DEBUG mode secret is visible
+        validators=[validators.MaxLengthValidator(
+            User._meta.get_field('jwt_secret').max_length,
+            message=_('The length of jwt_secret must be less than 150 characters'.format(
```
daviddavis: Should `150` here be `{}` instead?

platform/pulpcore/app/serializers/repository.py
```diff
@@ -202,9 +206,21 @@ class DistributionSerializer(ModelSerializer):
     )
     name = serializers.CharField(
         help_text=_('The name of the distribution. Ex, `rawhide` and `stable`.'),
+        validators=[validators.MaxLengthValidator(
+            models.Distribution._meta.get_field('name').max_length,
+            message=_('Distribution name length must be less than {} characters'.format(
```
daviddavis: I thought `format` needs to be called on what's returned from `_()`.

bmbouter: Yes the format needs to be called on the return string from gettext aka `_()`. Otherwise the string translation would include the specific filled variables which would require and endless amount of unpredictable string translations to exist since runtime can produce a huge number of those unique strings.

werwty:
@daviddavis updated, back to you. 

Can you also review my PR to the plugin_template to include the importer and publisher validators?
https://github.com/pulp/plugin_template/pull/2

PR #3069, opened 2017-07-04
Create Workflows and Use Cases section and stub out pages

goosemania:
closes #2854
https://pulp.plan.io/issues/2854

PR #3200, opened 2017-10-23
Fix exception when deleting consumer

dralley:
Fix NotFound exception if consumer is deleted when
its queue is gone.

closes #3069
https://pulp.plan.io/issues/3069


dralley:
@goosemania Is this the sort of thing you were looking for?

PR #3201, opened 2017-10-25
Create a new jwt reset endpoint instead of a user field

daviddavis:
Creating a new endpoint for resetting jwts instead of having a field on
user to reset them. This also fixes user creation which is currently
broken.

fixes #3075
https://pulp.plan.io/issues/3075

### DIFF
--- a/docs/integration_guide/rest_api/authentication.rst
+++ b/docs/integration_guide/rest_api/authentication.rst
@@ -86,7 +86,14 @@ To achieve ability to invalidate all user's tokens each user have their own secr
 used to sign their tokens. The change of the secret will lead to invalidation of all user's
 tokens. The change is independent on password.
 
-You can reset user secret to random by setting ``reset_jwt_secret`` on user API endpoint to True.
+You can reset user secret to random by calling the jwt reset endpoint:
+
+
+ * **method:** ``post``
+ * **path:** ``api/v3/users/<username>/jwt_reset/``
+ * response list:
+    * **code:** ``200`` - jwt reset
+    * **code:** ``400`` - invalid credentials
 
 If you have enabled ``JWT_AUTH.JWT_ALLOW_SETTING_USER_SECRET`` you can set the user's secret
 via user API endpoint.
--- a/platform/pulpcore/app/models/auth.py
+++ b/platform/pulpcore/app/models/auth.py
@@ -88,3 +88,7 @@ class User(AbstractBaseUser, PermissionsMixin):
         :rtype: str
         """
         return self.get_full_name()
+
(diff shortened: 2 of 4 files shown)


daviddavis:
Blocked by https://pulp.plan.io/issues/3093

daviddavis:
Also, we're discussing this on pulp-dev mailing list before proceeding.

platform/pulpcore/app/models/auth.py
```diff
@@ -88,3 +88,7 @@ def get_short_name(self):
         :rtype: str
         """
         return self.get_full_name()
+
+    def reset_jwt(self):
```
werwty: can this be renamed into jwt_reset to be consistent with the endpoint?

werwty:
Current the jwt_secret field on the serializer does not have read_only set. I think it would be good to set this to True since the user should not be able to set their own jwt_secret.

werwty:
Can the documentation for jwt_secret reset also be updated? https://github.com/pulp/pulp/blob/3.0-dev/docs/integration_guide/rest_api/authentication.rst#user-secret

platform/pulpcore/app/serializers/user.py
```diff
@@ -85,15 +79,7 @@ def __init__(self, *args, **kwargs):
         if not settings.JWT_AUTH.get("JWT_ALLOW_SETTING_USER_SECRET"):
```
werwty: This line is causing jwt_secret to not be displayed even when DEBUG mode is True. I would suggest moving this check to the jwt_secret serialzier field as:
`read_only=settings.JWT_AUTH.get("JWT_ALLOW_SETTING_USER_SECRET") `

werwty: After talking to @daviddavis I will file a new issue for this since we are not sure what the correct behavior is supposed to be.

daviddavis:
@werwty just pushed updates.

docs/integration_guide/rest_api/authentication.rst
```diff
@@ -86,7 +86,14 @@ To achieve ability to invalidate all user's tokens each user have their own secr
 used to sign their tokens. The change of the secret will lead to invalidation of all user's
 tokens. The change is independent on password.
 
-You can reset user secret to random by setting ``reset_jwt_secret`` on user API endpoint to True.
+You can reset user secret to random by calling the jwt reset endpoint:
+
+
+ * **method:** ``post``
+ * **path:** ``api/v3/users/<username>/jwtreset/``
```
werwty: s/jwtreset/jwt_reset/

PR #3202, opened 2017-10-26
Get unit tests running

daviddavis:
Note: The tests in test_urls.py used `router_for_nested_viewset()` which doesn't seem to exist anymore so they are failing. I am removing them since I don't see a quick and easy way to fix them. They probably need to be rewritten from scratch.

fixes #3064
https://pulp.plan.io/issues/3064


jortel:
The tests under tests/models related to concrete model classes were intended as just early smoke tests.  I think they should be removed now that smash tests are being written.

daviddavis:
I removed test_consumer.py and test_repositoy.py. I did not see any tests related to content though.

PR #3203, opened 2017-10-27
adds duration to the log message for completion of each migration

mhrivnak:
I threw this together to help someone who is testing how long the migrations take for various upgrades, and it appears useful enough to benefit others.

### DIFF
--- a/server/test/unit/server/db/test_manage.py
+++ b/server/test/unit/server/db/test_manage.py
@@ -166,48 +166,6 @@ class TestManageDB(MigrationTest):
                                                     "not allowed because the platform handlesit "
                                                     "for you.")
 
-    @patch.object(manage, 'ensure_database_indexes')
-    @patch('logging.config.fileConfig')
-    @patch('pkg_resources.iter_entry_points', iter_entry_points)
-    @patch('pulp.server.db.manage.connection.initialize')
-    @patch('pulp.server.db.manage.factory')
-    @patch('pulp.server.db.manage.logging.getLogger')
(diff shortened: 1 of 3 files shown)


server/test/unit/server/db/test_manage.py
```diff
@@ -166,48 +166,6 @@ def test_ensure_database_indexes_throws_exception(self, mock_model, mock_plugin_
...
-    @patch('pulp.server.db.manage.connection.initialize')
-    @patch('pulp.server.db.manage.factory')
-    @patch('pulp.server.db.manage.logging.getLogger')
-    @patch('pulp.server.db.manage.RoleManager.ensure_super_user_role')
-    @patch('pulp.server.db.manage.managers.UserManager.ensure_admin')
-    @patch('pulp.server.db.migrate.models.pulp.server.db.migrations',
-           migration_packages.platform)
-    @patch('sys.argv', ["pulp-manage-db"])
-    @patch.object(models.MigrationPackage, 'apply_migration')
-    def test_admin_is_ensured(self, apply_migration, ensure_admin, ensure_super_user_role,
```
mhrivnak: I spent 10 minutes trying to make this "unit" test work, but it's really an integration test that mocks the whole world. Its test case is covered by smash, and it's real hard to work with, so I don't think it's worth keeping.

daviddavis:
FYI, it looks like the new release of flake8 is causing PRs against master to fail. I opened an issue:

https://pulp.plan.io/issues/3106

PR #3210, opened 2017-10-31
Change the default header prefix for jwt requests

daviddavis:
fixes #3107
https://pulp.plan.io/issues/3107


daviddavis:
@bmbouter I updated the server.yaml file. We pass through the JWT config to the drf jwt code so I added a link to where users can view those variables. I did add a entry for JWT_AUTH_HEADER_PREFIX since it's a rather important setting. Let me know what you think.

PR #3212, opened 2017-11-01
Propagate coded exceptions raised during upload.

jortel:
https://pulp.plan.io/issues/3090

### DIFF
--- a/server/pulp/server/managers/content/upload.py
+++ b/server/pulp/server/managers/content/upload.py
@@ -226,7 +226,9 @@ class ContentUploadManager(object):
             repo_controller.rebuild_content_unit_counts(repo_obj)
             repo_controller.update_last_unit_added(repo_obj.repo_id)
             return result
-
+        except PulpCodedException:
+            # propagate coded exceptions.
+            raise
         except PulpException:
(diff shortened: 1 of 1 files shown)


server/pulp/server/managers/content/upload.py
```diff
@@ -226,7 +226,9 @@ def import_uploaded_unit(repo_id, unit_type_id, unit_key, unit_metadata, upload_
             repo_controller.rebuild_content_unit_counts(repo_obj)
             repo_controller.update_last_unit_added(repo_obj.repo_id)
             return result
-
+        except PulpCodedException:
+            # propagate coded exceptions.
+            raise
```
mhrivnak: How is this behaviorally different from the block below? I think all exceptions matching here would have otherwise matched the next block, and the only difference seems to be that the next block logs the error.

jortel: Yes, that's exactly the difference.  The `PulpCodedExection` is logged by our REST middlewhare.  This block eliminates double logging.

mhrivnak: Gotcha!

jortel:
ok test

jortel:
Unit test failure on f24 is unrelated.  Merging.

PR #3215, opened 2017-11-07
Removing celerybeat code

daviddavis:
fixes #3076
https://pulp.plan.io/issues/3076


platform/pulpcore/tasking/celery_app.py
```diff
@@ -91,18 +91,20 @@ def stop(self, consumer):
             self.timer_ref.cancel()
             self.timer_ref = None
 
-    def _record_heartbeat(self, consumer):
+    def _handle_heartbeat(self, consumer):
```
dralley: Perhaps this should be renamed to ```on_tick()``` or something more generic.  Ditto to the name of the class -- maybe name it something like ```PulpWorkerStep```

The docstring here should also be improved -- maybe "This method performs cleanup for dead workers, and records a heartbeat for the current worker"

platform/pulpcore/tasking/celery_app.py
```diff
@@ -126,9 +128,9 @@ def initialize_worker(sender, instance, **kwargs):
     removed along with the setting of online=False on the worker record in the database itself.
     The working directory specified in /etc/pulp/server.conf (/var/cache/pulp/<worker_name>)
     by default is removed and recreated. This is called early in the worker start process,
-    and later when it's fully online, pulp_celerybeat will discover the worker as usual to allow
-    new work to arrive at this worker. If there is no previous work to cleanup, this method
-    still runs, but has no effect on the database.
+    and later when it's fully online, workers will discover the worker as usual to allow new work to
```
dralley: Not strictly true, the worker will write to the database and the resource manager will "discover" it itself by checking the db.  There isn't really any discovery happening on the worker, just checks to make sure there aren't zero of them.

daviddavis:
ok test

PR #3216, opened 2017-11-08
Problem: async update and delete code is duplicated

dkliban:
Solution: add general update and delete tasks and viewsets

This patch creates two new tasks that are used by both the publisher viewset and the importer viewset.
This patch also introduces a new base viewset. Both the importer and publisher viewset inherit from it.

closes #3038
https://pulp.plan.io/issues/3038

### DIFF
--- a/pulpcore/pulpcore/app/tasks/base.py
+++ b/pulpcore/pulpcore/app/tasks/base.py
@@ -1,0 +1,56 @@
+from celery import shared_task
+from django.http import QueryDict
+
+from pulpcore.tasking.tasks import UserFacingTask
+from pulpcore.app.apps import get_plugin_config
+
+
+@shared_task(base=UserFacingTask)
+def general_update(instance_id, app_label, serializer_name, *args, **kwargs):
+    """
+    Update a model
+
+    The model instance is identified using the app_label, id, and serializer name. The serializer is
+    used to perform validation.
+
+    Args:
+        id (str): the id of the model
+        app_label (str): the Django app label of the plugin that provides the model
+        serializer_name (str): name of the serializer class for the model
+        data (dict): dictionary whose keys represent the fields of the model and their corresponding
+            values.
+        partial (bool): When true, only the fields specified in the data dictionary are updated.
+            When false, any fields missing from the data dictionary are assumed to be None and
+            their values are updated as such.
+
+    Raises:
+        :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
+            due to validation error. This theoretically should never occur since validation is
+            performed before the task is dispatched.
+    """
+    data = kwargs.pop('data', None)
+    partial = kwargs.pop('partial', False)
+    serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
+    instance = serializer_class.Meta.model.objects.get(id=instance_id).cast()
+    data_querydict = QueryDict('', mutable=True)
+    data_querydict.update(data)
+    serializer = serializer_class(instance, data=data_querydict, partial=partial)
+    serializer.is_valid(raise_exception=True)
+    serializer.save()
+
+
+@shared_task(base=UserFacingTask)
+def general_delete(instance_id, app_label, serializer_name):
+    """
+    Delete a model
+
+    The model instance is identified using the app_label, id, and serializer name.
+
+    Args:
+        id (str): the id of the model
+        app_label (str): the Django app label of the plugin that provides the model
+        serializer_name (str): name of the serializer class for the model
+    """
+    serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
+    instance = serializer_class.Meta.model.objects.get(id=instance_id).cast()
+    instance.delete()
--- a/pulpcore/pulpcore/app/viewsets/base.py
+++ b/pulpcore/pulpcore/app/viewsets/base.py
@@ -1,6 +1,9 @@
 import warnings
 
+from pulpcore.app import tasks
 from pulpcore.app.models import MasterModel
+from pulpcore.app.response import OperationPostponedResponse
+
 from rest_framework import viewsets, mixins
 from rest_framework.generics import get_object_or_404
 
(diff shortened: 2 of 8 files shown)


platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The model instance is identified using the app_label, id, and serializer name. The serializer is
+    used to perform validation.
+
+    Args:
+        id (str): the id of the model
+        app_label (str): the Django app label of the plugin that provides the importer
+        serializer_name (str): name of the serializer class for the importer
+        data (dict): dictionary whose keys represent the fields of the importer that need to be
+            updated with the corresponding values.
```
asmacdo: This is only perfectly true if partial=False (PATCH). 

Suggestion: dictionary whose keys represent the fields of the model and their corresponding values.

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+def general_update(instance_id, app_label, serializer_name, *args, **kwargs):
+    """
+    Update a model
+
+    The model instance is identified using the app_label, id, and serializer name. The serializer is
+    used to perform validation.
+
+    Args:
+        id (str): the id of the model
+        app_label (str): the Django app label of the plugin that provides the importer
```
asmacdo: s/importer/model/

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+        serializer_name (str): name of the serializer class for the importer
```
asmacdo: same as ^

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+        data (dict): dictionary whose keys represent the fields of the importer that need to be
+            updated with the corresponding values.
+        partial (bool): When true, only the fields specified in the data dictionary are updated.
+            When false, any fields missing from the data dictionary are assumed to be None and
+            their values are updated as such.
```
asmacdo: Might be inappropriate here, but I think it would be nice to mention PUT and PATCH.

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Raises:
+        :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
+            due to validation error. This theoretically should never occur since validation is
+            performed before the task is dispatched.
```
asmacdo: This could occur if the another task updates this unit before this task is executed.

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    data = kwargs.pop('data', None)
+    partial = kwargs.pop('partial', False)
+    serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
+    instance = serializer_class.Meta.model.objects.get(id=instance_id).cast()
+    data_querydict = QueryDict('', mutable=True)
+    data_querydict.update(data)
```
asmacdo: I don't think its super important, but it might be a good idea to set the QueryDict to immutable after the update.

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+@shared_task(base=UserFacingTask)
+def general_delete(instance_id, app_label, serializer_name):
+    """
+    Delete a model
+
+    The model instance is identified using the app_label, id, and serializer name.
+
+    Args:
+        id (str): the id of the model
+        app_label (str): the Django app label of the plugin that provides the importer
```
asmacdo: s/importer/model/

platform/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+        serializer_name (str): name of the serializer class for the importer
```
asmacdo: same^

pulpcore/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
+from celery import shared_task
```
asmacdo: Both pulpcore/pulpcore/app/tasks/base.py and platform/pulpcore/app/tasks/base.py?

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -158,3 +162,58 @@ class CreateDestroyReadNamedModelViewSet(mixins.CreateModelMixin,
...
+        serializer = self.get_serializer(instance)
+        app_label = instance._meta.app_label
+        async_result = tasks.base.general_delete.apply_async_with_reservation(
+            instance._meta.db_table, pk,
+            args=(pk, app_label, serializer.__class__.__name__)
+        )
+        return OperationPostponedResponse([async_result], request)
+
+
+class CreateReadAsyncUpdateDestroyNamedModelViewset(mixins.CreateModelMixin,
```
asmacdo: I strongly prefer (up to you though) that the models inherit from their mixins directly rather than creating these base ViewSets.

pulpcore/pulpcore/app/tasks/base.py
dkliban: No .... that was wrong on my part. This got put in as part of the rebase :(

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -158,3 +161,58 @@ class CreateDestroyReadNamedModelViewSet(mixins.CreateModelMixin,
...
+    Provides an update method that dispatches a task with reservation for a repository
+    """
+    def update(self, request, pk, **kwargs):
+        partial = kwargs.pop('partial', False)
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        app_label = instance._meta.app_label
+        async_result = tasks.base.general_update.apply_async_with_reservation(
+            instance._meta.db_table, pk,
```
werwty: Ah does this mean that importers will reserve the importer table, and publisher will reserve the publisher table? Is this the correct behavior (instead of both reserving the repository table)?

pulpcore/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+        data (dict): dictionary whose keys represent the fields of the model and their corresponding
+            values.
+        partial (bool): When true, only the fields specified in the data dictionary are updated.
+            When false, any fields missing from the data dictionary are assumed to be None and
+            their values are updated as such.
+
+    Raises:
+        :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
+            due to validation error. This theoretically should never occur since validation is
+            performed before the task is dispatched.
```
asmacdo: (non-blocking) I think we can remove this sentence, since it could happen.

pulpcore/pulpcore/app/tasks/base.py
```diff
@@ -0,0 +1,56 @@
...
+        app_label (str): the Django app label of the plugin that provides the model
+        serializer_name (str): name of the serializer class for the model
+        data (dict): dictionary whose keys represent the fields of the model and their corresponding
+            values.
+        partial (bool): When true, only the fields specified in the data dictionary are updated.
+            When false, any fields missing from the data dictionary are assumed to be None and
+            their values are updated as such.
+
+    Raises:
+        :class:`rest_framework.exceptions.ValidationError`: When serializer instance can't be saved
```
asmacdo: "When the serializer can't save the model instance" would be more accurate.

pulpcore/pulpcore/app/viewsets/base.py
dkliban: Yes, that is correct. With introduction of versioned repositories, it does not make sense to lock on a particular repository. Syncs will lock on an importer also. Publishes will lock on publisher, and repository version.

PR #3218, opened 2017-11-09
Problem: repository URI is mutable

dkliban:
Solution: use repository id in the URL instead of name

This patch switches repository URI from `/api/v3/repositories/<name>/` to
`/api/v3/repositories/<pk>/`.

closes #3101
https://pulp.plan.io/issues/3101

### DIFF
--- a/pulpcore/pulpcore/app/serializers/fields.py
+++ b/pulpcore/pulpcore/app/serializers/fields.py
@@ -23,7 +23,6 @@ class HrefWritableRepositoryRelatedField(serializers.HyperlinkedRelatedField):
     read_only should passed as a kwarg to this field.
     """
     view_name = 'repositories-detail'
-    lookup_field = 'name'
     href_writable = True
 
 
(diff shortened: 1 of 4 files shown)


platform/pulpcore/app/serializers/fields.py
```diff
@@ -23,7 +23,7 @@ class HrefWritableRepositoryRelatedField(serializers.HyperlinkedRelatedField):
     read_only should passed as a kwarg to this field.
     """
     view_name = 'repositories-detail'
-    lookup_field = 'name'
+    lookup_field = 'pk'
```
goosemania: Do you need to specify `pk` as a `lookup_field` explicitly? I think it's a default.

dkliban: I believe you are correct. I'll test without it and see if it still works.

dkliban: removing the lookup_field seems to work. however, i am not sure what other places in this PR could use a similar change.

PR #3221, opened 2017-11-10
Log scenarios where a full sync or publish is being forced

dralley:
closes #3119
https://pulp.plan.io/issues/3119

### DIFF
--- a/docs/dev-guide/integration/rest-api/consumer/applicability.rst
+++ b/docs/dev-guide/integration/rest-api/consumer/applicability.rst
@@ -95,11 +95,11 @@ URL and perform a GET request to retrieve the :ref:`task_group_summary`.
 
 :sample_request:`_` ::
 
- { 
+ {
   "repo_criteria": {
-   "filters": {"id": {"$in": ["test-repo", "test-errata"]}},
-   "parallel": true
-  }
+   "filters": {"id": {"$in": ["test-repo", "test-errata"]}}
+  },
+  "parallel": true
  }
 
 


server/pulp/server/controllers/repository.py
```diff
@@ -1163,6 +1170,13 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
             repo_id=repo_obj.repo_id,
             distributor_id=dist_id).update(set__last_override_config=config_override)
 
+    if force_full:
+        _logger.info(_("Fully republishing due to use of force_full in config"))
```
goosemania: I suggest to log only this case (when force_full is specified in config/override_config).
Other cases (e.g. when configs changed) don't enforce full sync/publish necessarily, plugin still can apply its optimizations and perform incremental publish or make sync almost no-op. When force_full is specified explicitly, plugin is expected to respect that.

server/pulp/server/controllers/repository.py
```diff
@@ -948,6 +948,13 @@ def check_perform_full_sync(repo_id, conduit, call_config):
     config_changed = check_config_updated_since_last_sync(conduit, repo_id)
     override_config_changed = check_override_config_change(repo_id, call_config)
 
+    if force_full:
+        _logger.info(_("Fully resyncing due to use of force_full in config"))
```
goosemania: I suggest to log only this case. See my reasoning in the other comment.

PR #3227, opened 2017-11-28
UnicodeEncodeError in comments provided with SSL cert/key/CA

ammaritiz:
As Python by default decodes objects as 'ascii', a decoding error
was being raised in Satellite. This is a hotfix to resolve that.

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -359,7 +359,10 @@ class Importer(AutoRetryDocument):
                 misc.mkdir(os.path.dirname(self._pki_path))
                 os.mkdir(self._pki_path, 0700)
             with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0600), 'w') as pem_file:
-                pem_file.write(self.config[config_key].encode('utf-8'))
+                if type(self.config[config_key]) is unicode:
+                    pem_file.write(self.config[config_key].encode('utf-8'))
+                else:
+                    pem_file.write(self.config[config_key])
 
 
 signals.pre_delete.connect(Importer.pre_delete, sender=Importer)


dralley:
I'm confused, issue 2960 is marked CLOSED-CURRENTRELEASE.  Did the issue crop back up?

ammaritiz:
Yes, unfortunately it did crop back up.

dralley:
Could you reopen the issue with an update about the new error?

ammaritiz:
Sure, will do that.

server/pulp/server/db/model/__init__.py
```diff
@@ -359,7 +359,7 @@ def _write_pem_file(self, config_key, path):
                 misc.mkdir(os.path.dirname(self._pki_path))
                 os.mkdir(self._pki_path, 0700)
             with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0600), 'w') as pem_file:
-                pem_file.write(self.config[config_key].encode('utf-8'))
+                pem_file.write(self.config[config_key].decode('utf-8').encode('utf-8'))
```
goosemania: Could you elaborate what is the goal to decode and encode it again with utf-8?
If it's expected in utf-8 you can just eliminate decode/encode parts. And if you're trying to validate somehow that it's utf-8 compatible as expected I think we should do it during import of the key (if it's not done already) and not when we write it out to the file.
What do you think?

ammaritiz: This file doesn't have a utf-8 encoding declaration at the top, hence, it was being decoded as `ascii` by default which caused an exception. We could actually instead of decoding it separately like this, just add the encoding declaration at the top, right?

goosemania: @ammaritiz , the encoding declaration at the top is about the encoding of the source file, so it won't help in this case. You can try it out if you read file with non-ascii characters and then try to encode it - you'll see exception in Python 2.

If `self.config[config_key]` was a unicode, no decoding would be needed and thus no exception would occur and encoding would work with no issue.

So `self.config[config_key]` is just of type `str` and you can write it as-is, decode-encode doesn't add any value and can/should be dropped.
If it's the only change needed, the question is, how it worked before? Why `self.config[config_key]` was a unicode before and now it is not?

dralley:
@ammaritiz What is the status on this?

ammaritiz:
@dralley - Currently, in discussion with @goosemania w.r.t the optimal solution to this issue.

server/pulp/server/db/model/__init__.py
```diff
@@ -359,7 +359,10 @@ def _write_pem_file(self, config_key, path):
                 misc.mkdir(os.path.dirname(self._pki_path))
                 os.mkdir(self._pki_path, 0700)
             with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0600), 'w') as pem_file:
-                pem_file.write(self.config[config_key].encode('utf-8'))
+                if self.config[config_key] is unicode:
```
dralley: I don't think this code would work.  ```isinstance(config_key, unicode)``` would be better, I think.

@goosemania do you agree?

goosemania:
@ammaritiz , have you tested it? 
What are the two cases when config can be a unicode string and a utf-8 encoded string?
Why do we have two different ways and not just unicode?

ammaritiz:
During the import it's a utf-8 encoded string. In Nectar, the config is getting updated and stored as unicode, hence, during sync when the importer config is updated, the config is in unicode. Therefore, it has been handled in this way.

dralley:
ok test

dralley:
ok test

dralley:
ok test

dralley:
@goosemania , does that reasoning make sense to you?

As long as it tests correctly, I think I am a +1.

PR #3231, opened 2017-11-30
Update selinux definitions for celery so it runs on f27

werwty:
closes #3159
https://pulp.plan.io/issues/3159

Compiled and installed without errors on f26, f27 and rhel7.
Resolves selinux denials on f27.

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -17,6 +17,7 @@ files_tmp_file(pulp_tmp_t)
 require {
         type celery_t;
         type httpd_sys_rw_content_t;
+        type pstore_t;
         type pulp_var_run_t;
         type pulp_tmp_t;
         type pulp_var_cache_t;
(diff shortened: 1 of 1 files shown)


server/selinux/server/pulp-celery.te
```diff
@@ -52,6 +57,12 @@ allow celery_t tmpfs_t:filesystem getattr;
 allow celery_t tmpfs_t:dir { write add_name remove_name };
 allow celery_t tmpfs_t:file { create open read write link getattr unlink };
 
+
+#============= qpidd_t ===============
+#
+
+allow qpidd_t qpidd_var_lib_t:file map;
```
bmbouter: Let's leave out the qpidd ones. I suspect this failure didn't cause Pulp failures, but it prevented Qpid code from doing something it tried to do. If Pulp still has an issue after these other changes are merged, then we can raise the issue with them. Sound ok?

PR #3233, opened 2017-12-01
Update selinux definitions for qpid so it runs on f27

werwty:
re #3159
https://pulp.plan.io/issues/3159

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -23,6 +23,8 @@ require {
         type pulp_var_cache_t;
         type pulp_cert_t;
         type puppet_etc_t;
+        type qpidd_t;
+        type qpidd_var_lib_t;
         type tmp_t;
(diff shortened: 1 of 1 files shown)


bmbouter:
This will fix the problem, but I think it will cause the Pulp SELinux policy to have the qpidd SELinux policy installed. Does it compile and install on a system that doesn't know about the qpidd_t label?

One other option is to not ship this with Pulp, but to add a note with a selinux command that allows this via the command line as a workaround. If the answer the the first section is that it's safe for us to ship this line then we should do that, otherwise I think the one-line workaround is our best option.

bmbouter:
@werwty pointed out on IRC that it's probably in the base policy and indeed it is! It's here https://github.com/fedora-selinux/selinux-policy-contrib/blob/rawhide/qpid.te

That means this is safe for Pulp to ship because it's already in the base policy of Fedora.

PR #3093, opened 2017-07-16
Refactor Plugins section in docs

goosemania:
closes #2852
https://pulp.plan.io/issues/2852

PR #3234, opened 2017-12-03
Include filter fields into schema for read actions only

goosemania:
closes #3093
https://pulp.plan.io/issues/3093

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/base.py
+++ b/pulpcore/pulpcore/app/viewsets/base.py
@@ -12,6 +12,33 @@ from django.core.exceptions import ValidationError
 
 from rest_framework import viewsets, mixins, serializers
 from rest_framework.generics import get_object_or_404
+from rest_framework.schemas import AutoSchema
+
+
+class DefaultSchema(AutoSchema):
+    """
+    Overrides _allows_filters method to include filter fields only for read actions.
+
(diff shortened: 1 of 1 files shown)


pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -33,11 +60,14 @@ class GenericNamedModelViewSet(viewsets.GenericViewSet):
             to django model filter expressions that can be used with the corresponding value from
             self.kwargs, used only by a nested ViewSet to filter based on the parent object's
             identity.
+        schema (rest_framework.schemas.AutoSchema): The schema class to use by default in
```
dparalen: qq: shouldn't this read `DefaultSchema`?

goosemania: By default it's a `DefaultSchema` but you can set this `schema` attribute to any subclass of `AutoSchema` in your subclass of `GenericNamedModelViewSet`.

dparalen: I'd say that it should document what it is rather than what it might be replaced with but this is a minor thing.

goosemania: @dparalen , fixed

PR #3235, opened 2017-12-03
Delete an Artifact only if it is not associated with any Content

goosemania:
closes #3016
https://pulp.plan.io/issues/3016

### DIFF
--- a/pulpcore/pulpcore/app/models/content.py
+++ b/pulpcore/pulpcore/app/models/content.py
@@ -132,8 +132,9 @@ class ContentArtifact(Model):
     A relationship between a Content and an Artifact.
 
     Serves as a through model for the 'artifacts' ManyToManyField in Content.
+    Artifact is protected from deletion if it's present in a ContentArtifact relationship.
     """
-    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE, null=True)
+    artifact = models.ForeignKey(Artifact, on_delete=models.PROTECT, null=True)
     content = models.ForeignKey(Content, on_delete=models.CASCADE)
     relative_path = models.CharField(max_length=64)
 
--- a/pulpcore/pulpcore/app/viewsets/content.py
+++ b/pulpcore/pulpcore/app/viewsets/content.py
@@ -1,4 +1,6 @@
-from django.db import transaction
+from gettext import gettext as _
+
+from django.db import models, transaction
 from django_filters.rest_framework import filterset
 from rest_framework import status
 from rest_framework.response import Response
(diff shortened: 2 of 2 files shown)


pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -79,3 +81,25 @@ class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):
     queryset = Artifact.objects.all()
     serializer_class = ArtifactSerializer
     filter_class = ArtifactFilter
+
+    def destroy(self, request, pk):
```
jortel: This kind of "look-and-see" logic cannot be used in concurrent applications without some kind of mutual exclusion.  It is subject to race conditions.

goosemania: True. Thanks!

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -79,3 +81,24 @@ class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):
...
+                with information about the artifact being deleted.
+            pk (:class:`uuid.UUID`): Unique id of the artifact being deleted.
+
+        Returns:
+            :class:`rest_framework.response.Response` with a 204 status code in case of
+                successful deletion, 409 status code if deletion is not allowed.
+        """
+        try:
+            return super(ArtifactViewSet, self).destroy(request, pk)
+        except models.ProtectedError:
```
goosemania: @jortel , does it make sense to return the error message that `ProtectedError` creates:  
`("Cannot delete some instances of model 'Artifact' because they are referenced through a protected foreign key: 'ContentArtifact.artifact'", <QuerySet [<ContentArtifact: pk=ebf4eea0-7942-4b22-8ad4-f63ba944899d>]>)` instead on the vague one I wrote?
But I wonder if queryset contains many entries... maybe at least log it at debug log level? What do you think?

jortel: I think catching the `Protected` exception and including a _user facing_ description as you did is better.  However, I would suggest a small change in wording.

_'The Artifact cannot be deleted because it is associated with Content.'_

pulpcore/pulpcore/app/models/content.py
```diff
@@ -132,8 +132,9 @@ class ContentArtifact(Model):
     A relationship between a Content and an Artifact.
 
     Serves as a through model for the 'artifacts' ManyToManyField in Content.
+    Artifact is protected from deletion if it's present in a ContentArtifact relationship.
     """
-    artifact = models.ForeignKey(Artifact, on_delete=models.CASCADE, null=True)
+    artifact = models.ForeignKey(Artifact, on_delete=models.PROTECT, null=True)
```
jortel: 👍

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -79,3 +81,24 @@ class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):
     queryset = Artifact.objects.all()
     serializer_class = ArtifactSerializer
     filter_class = ArtifactFilter
+
+    def destroy(self, request, pk):
+        """
+        Remove :class:`~pulpcore.app.models.Artifact` only if it doesn't belong to any
```
jortel: Since artifacts don't belong to any one Content, I would suggest:

_only if it is not associated with any ~pulpcore.app.models.Content_

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -79,3 +81,24 @@ class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):
...
+        Returns:
+            :class:`rest_framework.response.Response` with a 204 status code in case of
+                successful deletion, 409 status code if deletion is not allowed.
+        """
+        try:
+            return super(ArtifactViewSet, self).destroy(request, pk)
+        except models.ProtectedError:
+            msg = _('The Artifact cannot be deleted because it is a part of the Content.')
+            data = {'detail': msg}
+            return Response(data, status=status.HTTP_409_CONFLICT)
```
jortel: Suggest:

s/detail/reason/

goosemania: I called it `detail` becasue DRF uses this terminology in many cases: https://github.com/encode/django-rest-framework/blob/15024f3f07b9311887ad6f5afb366f6c032c0486/rest_framework/views.py#L83-L97

jortel: Ah, okay.

PR #3237, opened 2017-12-05
Move nested urls to flat structure

asmacdo:
closes https://pulp.plan.io/issues/3160

### DIFF
--- a/pulpcore/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/pulpcore/app/serializers/__init__.py
@@ -3,10 +3,8 @@
 # - all can import directly from base and fields if needed
 from pulpcore.app.serializers.base import (DetailRelatedField, GenericKeyValueRelatedField,  # noqa
     ModelSerializer, MasterModelSerializer, DetailIdentityField, DetailRelatedField,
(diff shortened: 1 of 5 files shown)


asmacdo:
@bmbouter, I'm not sure, but I am testing my guess by adding `six` to the pulpcore requirements in setup.py.

asmacdo:
@bmbouter btw, that commit is just to test. IMO, we should get those string types some other way. We probably shouldn't keep six as a dep since we only support Python 2.

asmacdo:
ok test

werwty:
ok test

werwty:
ok test

pulpcore/pulpcore/app/settings.py
```diff
@@ -114,7 +114,7 @@
     'DEFAULT_FILTER_BACKENDS': ('django_filters.rest_framework.DjangoFilterBackend',),
     'DEFAULT_PAGINATION_CLASS': 'pulpcore.app.pagination.UUIDPagination',
     'PAGE_SIZE': 100,
-    'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',),
```
werwty: Don't forget to uncomment :)

jortel:
Nice work, @asmacdo!  LGTM.

PR #3240, opened 2017-12-08
Remove unused consumer model

daviddavis:


daviddavis:
@dkliban I had to make some changes. Want to re-review? Thanks.

PR #3242, opened 2017-12-12
Update user model to have a UUID

daviddavis:
fixes #3174
https://pulp.plan.io/issues/3174


daviddavis:
@asmacdo I'm not sure I completely understand. Currently users have an id and their usernames are mutable. This changes id from an Integer to a UUID. It doesn't change the mutability of usernames. So not really sure how that would affect users that have changed their username?

asmacdo:
Ok, seems fine then. I assumed that username had been the pk.

PR #3244, opened 2017-12-14
First draft, definitions of Pulp's core concepts

asmacdo:
This document is intended to be a base language so we can be internally consistent with how we write about and discuss Pulp.


daviddavis:
I'm wondering if we should use the term "RepositoryVersion" given that we use "Repository" instead of "Repo". Thoughts?

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Each paragraph is preceded by a list of terms that are covered for the first time. In the paragraph
+body, terms that correspond to Python code are capitalized based on their type (see PEP 8) and are
+indicated by double backticks \`\`object\`\`. (``ClassName``, ``function_name``).  Terms that are
+not code objects in in bold. (**plugin**)
+
+Definitions
+-----------
+
+[``ContentUnit``, **plugin**, ``type``, ``Artifact``]
```
daviddavis: 👍 to ContentUnit

asmacdo:
@daviddavis +1 to RepositoryVersion.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    ``pulpcore`` is a generalized backend with a REST API and a plugin API. You will also need at
+    least one **plugin** to manage content.  Each **plugin** defines at least one ``type`` of
+    ``ContentUnit`` (like .rpm or .deb), which is the smallest set of data that can be managed by
+    Pulp. (The plural form of ``ContentUnit`` is a ``ContentUnits``, rather than Content or Units).
```
bmbouter: s/is a/is/
Also this sentence probably doesn't need parentheses.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    Files that belong to a ``ContentUnit`` are called ``Artifacts``, and each ``ContentUnit`` can
+    have 0 or many ``Artifacts``.  ``ContentUnits`` can share ``Artifacts``.
```
bmbouter: The last sentence is confusing to me. Maybe talking about it from the Artifact side would be an idea. Something like: `An Artifact can be shared by multiple ContentUnits.`

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+
+[``Repository``, **add**, **remove**, **RepoVersion**]
+    ``ContentUnits`` in Pulp are organized by their membership in ``Repositories``. You can **add**
```
bmbouter: The first two sentences need updating with repo versions.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    or **remove** ``ContentUnits`` to/from a ``Repository``. Pulp provides deduplication of
+    ``ContentUnits`` and ``Artifacts`` so you don’t need be concerned about **adding** a
```
bmbouter: I don't understand the sentence starting from "so you don't ..."

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    ``ContentUnit`` to too many ``Repositories``.  ``Repositories`` are versioned, and you can
+    **add** or **remove** ``ContentUnits`` to a ``Repository`` by creating a new ``RepoVersion``
```
bmbouter: They only associate with a RepoVersion, not a Repository anymore.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    and specifying the ``ContentUnits`` to **add** or **remove**.
+
+[**upload**]
+    ``ContentUnits`` can be created in Pulp manually. You will need to specify the ``Artifacts``
```
bmbouter: In places where it says "you" maybe that should be "the user" and then the possessives that refer to them like "your" be "the user's", etc. If this is for language clarity the reader could be a variety of roles.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,60 @@
...
+    that belong to the ``ContentUnit`` and the **plugin** that defines the ``ContentUnit``
+    ``type``.  ``Artifacts`` that are not already managed by Pulp should be **uploaded** to Pulp
+    prior to creating a new ``ContentUnit``. ``ContentUnits`` created manually must also be
+    **added** to a ``Repository`` manually by creating a new ``RepoVersion``.
+
+[**external source**, **sync**]
+    You can fetch ``ContentUnits`` and **add** them to your ``Repository`` by **syncing** with an
```
bmbouter: Similar with the "user" language here, and similar throughout the doc.

docs/contributing/language_guide.rst
asmacdo: I changed this a little bit, (added "over time"), but that change doesn't address your concerns. My thinking was that we will still talk about "adding ContentUnits to a "Repository", and that RepositoryVersions are /how/ that is done.

docs/contributing/language_guide.rst
asmacdo: Same as above.

asmacdo:
@bmbouter, let me know if these changes fit how you want to think/talk about Repositories and RepositoryVersions.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+
+[``Repository``, **add**, **remove**, **RepositoryVersion**]
+    ``ContentUnits`` in Pulp are organized by their membership in a ``Repository`` over time. Users
+    can **add** or **remove** ``ContentUnits`` to a ``Repository`` by creating a new
+    ``RepositoryVersion`` and specifying the ``ContentUnits`` to **add** or **remove**.
+
+[**upload**]
+    ``ContentUnits`` can be created in Pulp manually. Users specify the ``Artifacts`` that belong
+    to the ``ContentUnit`` and the **plugin** that defines the ``ContentUnit`` ``type``.
+    ``Artifacts`` that are not already managed by Pulp should be **uploaded** to Pulp prior to
```
bmbouter: What about    s/managed/known/

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+    Users can fetch ``ContentUnits`` and **add** them to their ``Repository`` by **syncing** with an
+    **external source**. The logic and configuration that specifies how Pulp should to interact
+    with an **external source** is provided by an ``Importer`` and is defined by the same
+    **plugin** that defines that ``type`` of ``ContentUnit`` that the **external source** contains.
+    ``pulpcore`` supports multiple ``sync_modes``, including ``additive`` (``ContentUnits`` are
+    only **added**) and ``mirror`` (``ContentUnits`` are **added** and **removed** to match the
+    **external source**.)
+
+[``hosted``, **metdata**, ``Publisher``, ``Publication``, ``Distribution``]
+    All ``ContentUnits`` that are managed by Pulp are **hosted** as part of the ``pulpcore`` REST
```
bmbouter: s/are/can be/

Also I think it's s/REST API/Content App/ because the content app is it's own distinct webservice from the REST API.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+    API.  A ``Publication`` referes to the **metadata** (flat files) for the set of
```
bmbouter: Somehow this needs to also include `Artifacts` because a Publication is made of both `PublishedMetadata` and `PublishedArtifact` objects. Also the metadata does not entirely have to describe content, it is content, just not the kind that has an Artifact backing it. I'm not sure how to best bring this in.

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+[``Repository``, **add**, **remove**, **RepositoryVersion**]
+    ``ContentUnits`` in Pulp are organized by their membership in a ``Repository`` over time. Users
+    can **add** or **remove** ``ContentUnits`` to a ``Repository`` by creating a new
+    ``RepositoryVersion`` and specifying the ``ContentUnits`` to **add** and **remove**.
+
+[**upload**]
+    ``ContentUnits`` can be created in Pulp manually. Users specify the ``Artifacts`` that belong
+    to the ``ContentUnit`` and the **plugin** that defines the ``ContentUnit`` ``type``.
+    ``Artifacts`` that are not already known by Pulp should be **uploaded** to Pulp prior to
+    creating a new ``ContentUnit``. ``ContentUnits`` are manually **added** to a
```
bmbouter: s/are/can be/

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+    ``Repository`` manually by creating a new ``RepositoryVersion``.
```
bmbouter: and then on the next line.   s/manually//

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+    Users can fetch ``ContentUnits`` and **add** them to their ``Repository`` by **syncing** with an
+    **external source**. The logic and configuration that specifies how Pulp should to interact
+    with an **external source** is provided by an ``Importer`` and is defined by the same
+    **plugin** that defines that ``type`` of ``ContentUnit`` that the **external source** contains.
+    ``pulpcore`` supports multiple ``sync_modes``, including ``additive`` (``ContentUnits`` are
+    only **added**) and ``mirror`` (``ContentUnits`` are **added** and **removed** to match the
+    **external source**.)
+
+[``hosted``, **metdata**, ``Publisher``, ``Publication``, ``Distribution``]
+    All ``ContentUnits`` that are managed by Pulp are **hosted** as part of the ``pulpcore``
```
bmbouter: s/are/can be/

docs/contributing/language_guide.rst
```diff
@@ -0,0 +1,57 @@
...
+    Content App. **plugin** defined ``Publishers`` generate ``Publications``, which
+    refer to the **metadata** and ``Artifacts`` of the ``ContentUnits`` in a ``RepositoryVersion``
```
bmbouter: How is `**metadata**` different from `PublishedMetadata` as a formal software object? I think introducing both `PublishedMetadata` and `PublishedArtifacts` may be good. What do you think?

asmacdo: I wanted to avoid defining any objects that aren't user facing, since they wont be semantically versioned.

bmbouter: What about plugin writers as a form of user? Those two objects are in the plugin API so they would be semver governed by the plugin API version. If it's just for users that is OK too, but I figured it would be in the user guide, not the contributor guide.

asmacdo: I'm thinking of this as a contributor guide for how to write about user facing concepts. I am +1 on creating a second document in this style in the plugin writers guide

bmbouter: Sounds good to me. I'm +1 to merge with the other small grammar fixes.

dralley:
ok test

daviddavis:
👍

PR #3245, opened 2017-12-15
Add warning and config variables for heartbeat timeout/interval

daviddavis:
refs #3135
https://pulp.plan.io/issues/3135

### DIFF
--- a/common/pulp/common/constants.py
+++ b/common/pulp/common/constants.py
@@ -69,13 +69,6 @@ SCHEDULER_WORKER_NAME = 'scheduler'
 # Resource manager worker name
 RESOURCE_MANAGER_WORKER_NAME = 'resource_manager'
 
-# The amount of time (in seconds) between process wakeups to "heartbeat" and perform
-# their tasks.
-PULP_PROCESS_HEARTBEAT_INTERVAL = 5
-
-# The amount of time (in seconds) after which a Celery process is considered missing.
-PULP_PROCESS_TIMEOUT_INTERVAL = 25
-
 # The amount of time the migration script will wait to confirm that no processes are running.
 # This is the 90s CELERY_TICK_DEFAULT_WAIT_TIME used in Pulp version < 2.12 and a 2s buffer.
 # This ensures that the process check feature works correctly even in cases where a user
--- a/docs/user-guide/release-notes/2.16.x.rst
+++ b/docs/user-guide/release-notes/2.16.x.rst
@@ -1,0 +1,12 @@
+=======================
+Pulp 2.16 Release Notes
+=======================
+
+Pulp 2.16.0
+===========
+
+New Features
+------------
+
+* Added a `worker_timeout` config setting for users to customize when Pulp should consider workers
+  are dead and attempt to kill them.
--- a/server/etc/pulp/server.conf
+++ b/server/etc/pulp/server.conf
@@ -318,6 +318,10 @@
 #
 # login_method: Select the SASL login method used to connect to the broker. This should be left
 #     unset except in special cases such as SSL client certificate authentication.
+#
+# worker_timeout: The amount of time (in seconds) before considering a worker as missing. If Pulp's
+#     mongo database has slow I/O, then setting a higher number may resolve issues where workers are
+#     going missing incorrectly. Defaults to 30.
 
 [tasks]
 # broker_url: qpid://localhost/
@@ -326,6 +330,7 @@
 # keyfile: /etc/pki/pulp/qpid/client.crt
 # certfile: /etc/pki/pulp/qpid/client.crt
 # login_method:
+# worker_timeout: 30
 
 
 # = Email =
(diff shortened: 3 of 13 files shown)


common/pulp/common/constants.py
```diff
@@ -69,13 +69,6 @@
 # Resource manager worker name
 RESOURCE_MANAGER_WORKER_NAME = 'resource_manager'
 
-# The amount of time (in seconds) between process wakeups to "heartbeat" and perform
-# their tasks.
-PULP_PROCESS_HEARTBEAT_INTERVAL = 5
-
-# The amount of time (in seconds) after which a Celery process is considered missing.
-PULP_PROCESS_TIMEOUT_INTERVAL = 25
```
daviddavis: I had to remove these constants from common and add them to pulp.server.constants. For more info see https://pulp.plan.io/issues/3135#note-22.

daviddavis:
@dralley @bmbouter can you review/re-review my changes?

server/etc/pulp/server.conf
```diff
@@ -318,6 +318,9 @@
 #
 # login_method: Select the SASL login method used to connect to the broker. This should be left
 #     unset except in special cases such as SSL client certificate authentication.
+#
+# worker_failover_time: The maximum time a worker will run for before being considered dead and
```
dralley: It should probably be named something other than ```worker_failover_time``` since it doesn't apply only when HA is being used.  

Maybe ```worker_expiration_time```.  I don't care too much about the name though so if you prefer it this way that's fine with me.

server/pulp/server/constants.py
```diff
@@ -1,6 +1,16 @@
...
 LOCAL_STORAGE = "/var/lib/pulp/"
 
 PULP_USER_METADATA_FIELDNAME = 'pulp_user_metadata'
 PULP_DJANGO_SETTINGS_MODULE = 'pulp.server.webservices.settings'
 PULP_STREAM_REQUEST_HEADER = 'Pulp-Stream-Request'
 SUPER_USER_ROLE = 'super-users'
+
+# The amount of time (in seconds) between process wakeups to "heartbeat" and perform
+# their tasks.
+PULP_PROCESS_HEARTBEAT_INTERVAL = int(config.getint('tasks', 'worker_failover_time') / 5)
```
dralley: This is probably ok as-is, but maybe it would be good to add a brief explanation behind the math and magic numbers going on, the distinction between worker_failover_time and the process_timeout_interval, etc.

I'm fine without it too, though.

server/etc/pulp/server.conf
bmbouter: What about a name like `worker_timeout` or `worker_missing_time` or `worker_missing_timeout` ?

bmbouter: The `killed in most cases` makes it sound like we're actively killing the process which we don't do. I see you meant though. I'm also wondering about a little bit of guidance here too. Here's an idea, feel free to edit wholly/rewrite:

`The amount of time (in seconds) before considering a worker as missing. If Pulp's mongo database has slow I/O, then setting a higher number may resolve issues where workers are going missing incorrectly. Defaults to 30.`

server/pulp/server/constants.py
bmbouter: I think a link to the redmine comments where this is discussed would be a good way to help us (and others) remember.

bmbouter:
We didn't talk about this before, but what about adding another another "timing check" [here too](https://github.com/pulp/pulp/blob/master/server/pulp/server/async/scheduler.py#L74)? I think it's possible that the read operations also be really slow and that if they are so slow that they are > the heartbeat interval than that is probably also bad. I was thinking of the same check in between the creation of `now` on L#72 and `worker_list` being created to in L#74.

server/pulp/server/constants.py
dralley: +1 on that

server/etc/pulp/server.conf
dralley: +1 for ```worker_timeout``` or ```worker_missing_timeout```

daviddavis:
@dralley @bmbouter thanks for reviews. think I addressed all the feedback.

bmbouter:
Thanks @daviddavis. The code you pushed all looks exactly correct. I forgot to ask on the first round of feedback, but can a release note also be added about the new setting. This would be going into 2.16 I think. That is the last thing I can think of. Thanks a lot for putting this together. It's really an important fix for our users.

docs/user-guide/release-notes/2.16.x.rst
```diff
@@ -0,0 +1,12 @@
+=======================
+Pulp 2.16 Release Notes
+=======================
+
+Pulp 2.16.0
+===========
+
+New Features
+---------
```
dralley: Docs builder warning: underline too short

PR #3250, opened 2018-01-02
Add importer filters

daviddavis:
fixes #3080
https://pulp.plan.io/issues/3080


daviddavis:
Note: the name param is added by ContentAdaptorFilter.

asmacdo:
I thought we had decided not to do id_in_list anymore since the users won't really see the id. They would probably want to do href_in_list.

asmacdo:
> Note: the name filter param is added by ContentAdaptorFilter.
repo_name?

daviddavis:
I removed the id_in_list filter and opened [a href_in_list issue](https://pulp.plan.io/issues/3240) per our discussion on IRC.

PR #3252, opened 2018-01-04
Fix broken link to protected-branches.py

dparalen:
closes #3252
https://pulp.plan.io/issues/3245

### DIFF
--- a/docs/dev-guide/contributing/branching.rst
+++ b/docs/dev-guide/contributing/branching.rst
@@ -43,7 +43,7 @@ on GitHub. The basic protection that disallows force-push and deletion is the
 only option that should be enabled. There should be no restrictions on required
 status checks or who can push. There is a script at
 `devel/scripts/protected-branches.py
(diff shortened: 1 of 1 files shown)


docs/dev-guide/contributing/branching.rst
```diff
@@ -43,7 +43,7 @@ on GitHub. The basic protection that disallows force-push and deletion is the
 only option that should be enabled. There should be no restrictions on required
 status checks or who can push. There is a script at
 `devel/scripts/protected-branches.py
-<https://github.com/pulp/devel/tree/master/scripts/protected-branches.py>`_
+<https://github.com/pulp/devel/blob/master/scripts/protect-branches.py>`_
```
dparalen: the tree--blob change might not be necessary, though https://github.com/tree/devel/blob/master/scripts/protect-branches.py auto-redirects to https://github.com/pulp/devel/blob/master/scripts/protect-branches.py

dparalen:
Thanks for the review, @goosemania! :)

PR #3246, opened 2017-12-18
Update upload documentation for large files.

droslean:
This information should be included in the documents for upload cases of large files without using multiple upload bits.
User should be informed that he can use this functionality.

PR #3253, opened 2018-01-04
Update branching.rst to match PUP-0003

dparalen:
The bugfix branches guide contradicted PUP-0003 by recommending to apply
bug fixes to ``x.y-dev`` branches.

The correct approach is to apply fix agains master and cherry-pick it to
any relevant ``x.y-dev`` branch later.

closes #3246
https://pulp.plan.io/issues/3246


daviddavis:
The docs are definitely out-of-date but the build team is working on supporting the new workflow in PUP-3 and they're supposedly going to modify it. So PUP-3 is not quite exactly right either. Might be worth pinging @pcreech before we update our docs.

PR #3185, opened 2017-10-06
Pins pulp to DRF 3.6.4

bmbouter:
Pulp needs updates to be compatible with 3.7.0. This should be undone
with issue: https://pulp.plan.io/issues/3057

re #3057

PR #3256, opened 2018-01-06
Problem: Distribution lacks a base URL

dkliban:
Solution: Add 'base_url' field to the Distribution resource

This patch adds a read only 'base_url' field to the Distribution resource. It's value is derived
from a new 'hostname' config in server.yaml and the 'base_path' of the Distribution.
for the base_url field of the Distribution resource.

closes #3185
https://pulp.plan.io/issues/3185

### DIFF
--- a/pulpcore/pulpcore/app/settings.py
+++ b/pulpcore/pulpcore/app/settings.py
@@ -206,6 +206,7 @@ _DEFAULT_PULP_SETTINGS = {
     },
     'content': {
         'web_server': 'django',
+        'host': None,
     },
     'broker': {
         'url': 'amqp://guest@localhost//',
--- a/pulpcore/pulpcore/etc/pulp/server.yaml
+++ b/pulpcore/pulpcore/etc/pulp/server.yaml
@@ -93,6 +93,11 @@
 #                 streaming the content to Apache.  Requires: mod_xsendfile to be installed.
 #                 When set to 'nginx', the X-Accel-Redirect header is injected which delegates
 #                 streaming the content to NGINX.
+#   `host`: The host name or IP and an optional port number for the Content App. (e.g.
+#           example.com:8000) This value should be specified only if the Content App is served on
+#           a host that's different from the REST API. The default value is the same as the host
+#           used to serve the REST API.
 #
 # content:
 #   web_server: django
+#   host: pulp.example.com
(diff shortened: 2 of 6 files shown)


pulpcore/pulpcore/app/settings.py
```diff
@@ -206,6 +206,7 @@
     },
     'content': {
         'web_server': 'django',
+        'hostname': 'localhost:8000',
```
asmacdo: I think you can access the original request in the serializer via `context`

dkliban: If the content app and the rest api are being served on different hosts, the context will contain the wrong hostname and port.

jortel: Perhaps this settings should default to `None` and have viewset code inject the _host_ from the request when not specified here.  Most users will run the API and content app on the same box and would only need to set this when the content app is running on another box.  Less to keep up with.

Also, based on URL specifications, the setting should be named _host_ instead of _hostname_.  It can be either an IP or hostname and can optionally include the port.

dkliban: +1 @jortel - i'll rename it and change the default

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -102,3 +104,12 @@ class Distribution(Model):
...
         default_related_name = 'distributions'
+
+    def get_base_url(self):
+        """
+        Helper method for constructing the base_url field
+
+        Returns:
+            URL that can be used for consuming content distributed by this distribution
+        """
+        return ''.join([settings.CONTENT['hostname'], reverse('content-app'), self.base_path])
```
dkliban: @jortel I don't have access to the request object here. So I am not sure how I can get the host without a setting. Do you have any suggestions?

asmacdo: This should be handled by a serializer, similar to what we did with the CreatedResourceSerializer.to_representation. Models shouldn't know about their urls.

pulpcore/pulpcore/etc/pulp/server.yaml
```diff
@@ -93,6 +93,12 @@
 #                 streaming the content to Apache.  Requires: mod_xsendfile to be installed.
 #                 When set to 'nginx', the X-Accel-Redirect header is injected which delegates
 #                 streaming the content to NGINX.
+#   `host`: The hostname and port number for the content app. This value should be specified
```
jortel: Suggest: `The host name or IP and optional port number`

pulpcore/pulpcore/etc/pulp/server.yaml
```diff
@@ -93,6 +93,12 @@
...
+#           only if the Content App is served on a host that's different from the REST API. The
+#           hostname portion needs to match at least one hostname in the `allowed_hosts` config.
+#           The default value is None. When a value is specified, Distributions use it to form
+#           the 'base_url'.
```
jortel: My concern about mentioning `Distributions use it to form the base_url` is that if it's used for something else in the future it would be easy to miss updating this.  Also, I don't think it's useful information.

PR #3260, opened 2018-01-10
Allow users to delete the latest repo version

daviddavis:
fixes #3219
https://pulp.plan.io/issues/3219

This goes with https://github.com/pulp/pulp_file/pull/25.

### DIFF
--- a/pulpcore/pulpcore/app/models/repository.py
+++ b/pulpcore/pulpcore/app/models/repository.py
@@ -18,6 +18,9 @@ class Repository(Model):
 
         name (models.TextField): The repository name.
         description (models.TextField): An optional description.
+        last_version (models.PositiveIntegerField): A record of the last created version number.
+            Used when a repository version is deleted so as not to create a new vesrion with the
+            same version number.
 
     Relations:
 
@@ -26,6 +29,7 @@ class Repository(Model):
     """
     name = models.TextField(db_index=True, unique=True)
     description = models.TextField(blank=True)
+    last_version = models.PositiveIntegerField(default=0)
 
     notes = GenericKeyValueRelation(Notes)
 
@@ -241,7 +245,7 @@ class RepositoryVersion(Model):
         repository (models.ForeignKey): The associated repository.
     """
     repository = models.ForeignKey(Repository)
-    number = models.PositiveIntegerField(db_index=True, default=1)
+    number = models.PositiveIntegerField(db_index=True)
     created = models.DateTimeField(auto_now_add=True)
 
     class Meta:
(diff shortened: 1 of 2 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -18,6 +18,7 @@ class Repository(Model):
 
         name (models.TextField): The repository name.
         description (models.TextField): An optional description.
+        last_version (models.PositiveIntegerField): A record of the last created version number
```
mhrivnak: I think it's worth adding a note that the corresponding version may no longer exist, and that's why the max number is tracked here.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -323,8 +325,7 @@ def save(self, *args, **kwargs):
         """
         Save the version while setting the number automatically.
         """
-        with suppress(self.DoesNotExist):
-            self.number = self.repository.versions.latest().number + 1
+        self.number = self.repository.last_version + 1
```
mhrivnak: It would be an improvement here to ensure that if ``save()`` was called in the future for some reason, it wouldn't overwrite the number.

mhrivnak: But that would be above and beyond the scope of this fix, so don't feel obligated. :)

pulpcore/pulpcore/app/tasks/importer.py
```diff
@@ -51,6 +51,5 @@ def sync(importer_pk):
             new_version.delete()
             raise
 
-    if new_version.added().count() == 0 and new_version.removed().count() == 0:
-        log.debug('no changes; deleting repository version')
-        new_version.delete()
+    new_version.repository.last_version += 1
```
mhrivnak: It may be worth directly setting it to the value ``new_version.number``. In case there were ever a bug around these numbers getting out of sync, that would bring them back together.

dkliban:
ok test

pulpcore/pulpcore/app/tasks/importer.py
```diff
@@ -51,6 +51,5 @@ def sync(importer_pk):
             new_version.delete()
             raise
 
-    if new_version.added().count() == 0 and new_version.removed().count() == 0:
```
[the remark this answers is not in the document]
daviddavis: We discussed this and decided that sync should always create a new version even if content doesn't change. Creating repo versions with this design is cheap/free so—as a form of record keeping—we should always create a version unless the sync errors or fails. 

I'm not committed to this decision though. Interested to get your thoughts.

PR #3262, opened 2018-01-11
Problem: Artifacts can't store git repos

dkliban:
Solution: Provide shared persistent storage for all plugins.

This patch updates the plugin api with an interface for finding the path to a plugin's
storage location.

closes #3182
https://pulp.plan.io/issues/3182


plugin/pulpcore/plugin/storage.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+from pulpcore.app.apps import get_plugin_config
+
+
+def get_plugin_storage_path(plugin_app_label):
+    """
+    An interface for finding the path to a plugin's persistent storage location. It is
```
bmbouter: Can a single 1-line description be added? I think that is what PEP 257 expects.

PR #3266, opened 2018-01-12
Adopt asyncio downloading.

jortel:
https://pulp.plan.io/issues/3271

The diff is huge on this but mainly because a lot of files moved and deleted.  Also, a lot of imports needed to change.  The changes to the ChangeSet are relatively small but seems larger because a few things needed to be moved.


jortel:
ok test

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -405,3 +407,66 @@ def save(self):
...
+        """
+        Args:
+            artifact (PendingArtifact):
+            downloader (BaseDownloader):
+        """
+        def done(task):
+            try:
+                task.exception()
+            except Exception:
+                pass
```
bmbouter: Should this log here or is the point to silence exceptions and let the plugin writer report them at the end?

plugin/pulpcore/plugin/changeset/model.py
```diff
@@ -405,3 +407,66 @@ def save(self):
...
+        return NopDownloader()
+
+    def settle(self):
+        pass
+
+    def save(self):
+        pass
+
+
+class DownloadTask(asyncio.Task):
```
bmbouter: I was just reading about inheriting from `asyncio.Task` and this does use Futures to run co-routines. What I'm gathering from this is that the changeset uses Futures on it's own, outside of the futures based downloaders. Or is the main benefit that this allows for the callback opportunity?

bmbouter: Also the docs say `Don’t directly create Task instances: use the ensure_future() function or the AbstractEventLoop.create_task() method`. It sounds like we shouldn't be subclassing this. Am I reading it right?

plugin/pulpcore/plugin/changeset/model.py
jortel: I chose not to log the exception here thinking it would be the responsibility of the code using this iterator to handle the exception as seen fit.

bmbouter: Sounds good.

plugin/pulpcore/plugin/download/file.py
```diff
@@ -35,11 +35,13 @@ def __init__(self, url, **kwargs):
         Read, validate, and compute digests on the `url`. This is a coroutine.
 
         This method provides the same return object type and documented in
-        :meth:`~pulpcore.plugin.download.asyncio.BaseDownloader.run`.
+        :meth:`~pulpcore.plugin.download.BaseDownloader.run`.
         """
-        with open(self._path, 'r') as f_handle:
+        async def read():
```
bmbouter: I see why this needed to be changed. Unlike the aiohttp objects which are _awaitable_ the `read()` from stdlib is not. I think we should consider using an aio based file like the one this library provides: https://pypi.python.org/pypi/aiofiles/0.3.2

bmbouter:
I tested both the `http://` downloading and `file://` downloading. Everything worked for me.

bmbouter:
The docs are erroring, but it doesn't look related. I'm going to rerun the docs builders now.

bmbouter:
ok test

werwty:
ok test

werwty:
ok test

plugin/pulpcore/plugin/download/file.py
jortel: A little concerned aiofile is only at version 0.3 but probably a good idea.

plugin/pulpcore/plugin/changeset/model.py
jortel: The _asyncio_ API is a bit schizophrenic with regards to `Future` and `Task`.  It cannot decide if it's procedural or object-oriented.  The `Task` class is exposed and the API is fairly `Future` oriented and exposing and documenting a class then saying don't use it does not make sense.  That said, the warning against instantiating a `Task` directly is clear (though I think unfounded) and does preclude us from extending it (which is unfortunate).

I pushed another approach which uses `ensure_future()` to create the task, then adds the callback.  The `DownloadTask` included a `artifact` attribute which provided linkage to the `PendingArtifact` related to the download.  This is used by the `ChangeSet`.  I moved this correlation to the `DownloadIterator` and got rid of the `DownloadTask`.

jortel:
I noticed that I omitted the read buffer size when rewriting the FileDownloader.  Fixed and pushed.

jortel:
@bmbouter, I squashed the commits and retested all use cases:
- file content using file://
- file content using http://
- file content with duplicate artifacts and artifacts already existing.
- content without artifacts.
- failed downloads.

Built and inspected docs.

Should be ready for retest.

bmbouter:
Thanks @jortel ! I will totally test this out asap. Can you push one small fix for this docs error. It's at the bottom of the 'Details' link. I'll start testing either way though.

bmbouter:
@jortel Thanks so much for the docs. I am getting an error though because of a change that was merged to master. Now the `RepositoryVersion` has a required field called `number`.

pulp_file (when tested with this PR) gives this error. I also reproduced it in several vagrant environments:
```
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/celery/app/trace.py", line 374, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/vagrant/devel/pulp/pulpcore/pulpcore/tasking/tasks.py", line 272, in __call__
    return super(UserFacingTask, self).__call__(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/celery/app/trace.py", line 629, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/vagrant/devel/pulp_file/pulp_file/app/tasks.py", line 53, in sync
    new_version.save()
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/base.py", line 808, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/base.py", line 838, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/base.py", line 924, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/base.py", line 963, in _do_insert
    using=using, raw=raw)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/query.py", line 1076, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/models/sql/compiler.py", line 1112, in execute_sql
    cursor.execute(sql, params)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/utils/six.py", line 685, in reraise
    raise value.with_traceback(tb)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: null value in column "number" violates not-null constraint
DETAIL:  Failing row contains (3a0e039f-38a1-4717-8947-0ccd14b75120, null, 2018-01-18 20:55:25.283175+00, f, 42bac590-6b9f-4ad6-9e18-57b54dd5a0ad).
```

PR #3268, opened 2018-01-16
Add complete field

asmacdo:
closes #3222
https://pulp.plan.io/issues/3222

### DIFF
--- a/pulpcore/pulpcore/app/models/repository.py
+++ b/pulpcore/pulpcore/app/models/repository.py
@@ -330,14 +330,6 @@ class RepositoryVersion(Model):
         )
         q_set.update(version_removed=self)
 
-    def save(self, *args, **kwargs):
-        """
-        Save the version while setting the number automatically.
-        """
-        if self.number is None:
-            self.number = self.repository.last_version + 1
-        super().save(*args, **kwargs)
-
     def next(self):
         """
         Returns:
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -239,6 +239,8 @@ class RepositoryVersion(Model):
             1 + the most recent version.
         created (models.DateTimeField): When the version was created.
         action  (models.TextField): The action that produced the version.
+        complete (models.BooleanField): False indicates that the Repository Version is still being
```
dkliban: I think this description needs to state that the field indicates whether or not the Repository Version is user visible. Then you can say what the False means

daviddavis:
LGTM (aside from @dkliban's suggestion)

daviddavis:
@asmacdo I thought you were looking at a way to exclude RepositoryVersions with complete=False by default when querying them. Did that not pan out or did you decide against it?

daviddavis:
What do you think about creating a Manager to encapsulate the complete/incomplete logic?

asmacdo:
@daviddavis +1, complete/incomplete should be handled by pulpcore. This is an intermediate step. After demonstrating what needs to be done (this PR), I created an issue to handle this logic in core. https://pulp.plan.io/issues/3285

PR #3270, opened 2018-01-16
Adds ability to POST to repository version API

dkliban:
This patchs adds ability to create new repository versions by specifying a list of
content to add to the previous version and remove from the previous version.


pulpcore/pulpcore/app/tasks/repository.py
```diff
@@ -109,3 +109,40 @@ def delete_version(pk):
...
+    try:
+        for pk in add_content_units:
+            content = models.Content.objects.get(pk=pk)
+            new_version.add_content(content)
+        for pk in remove_content_units:
+            content = models.Content.objects.get(pk=pk)
+            new_version.remove_content(content)
+    except Exception:
+        with transaction.atomic():
+            new_version.delete()
```
asmacdo: This cascade deletes all of the RepositoryContent associations too right?

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -178,6 +187,63 @@ def destroy(self, request, repository_pk, number):
...
+        Returns:
+            PK of the Content
+
+        Raises:
+            :class:`rest_framework.exceptions.ValidationError`: When the URL is not a valid
+                Content resource
+        """
+
+        try:
+            resolver = resolve(urlparse(url).path)
```
asmacdo: I'm surprised this is necessary. It seems like DRF should be able to get_or_404 from an href.

pulpcore/pulpcore/app/tasks/repository.py
```diff
@@ -109,3 +109,40 @@ def delete_version(pk):
 
         # With no more relationships remaining, delete the version.
         version.delete()
+
+
+@shared_task(base=UserFacingTask)
+def create_version(repository_pk, add_content_units, remove_content_units):
```
asmacdo: could we call this add_or_remove? This name could confuse plugin writers because this is not involved in other tasks that create a version

pulpcore/pulpcore/app/tasks/repository.py
```diff
@@ -109,3 +109,40 @@ def delete_version(pk):
...
+    with transaction.atomic():
+        repository = models.Repository.objects.get(pk=repository_pk)
+        new_version = models.RepositoryVersion(repository=repository)
+        new_version.save()
+        created = models.CreatedResource(content_object=new_version)
+        created.save()
+        repository.last_version = new_version.number
+        repository.save()
+
+    try:
```
asmacdo: should this whole block be a transaction?

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -178,6 +187,63 @@ def destroy(self, request, repository_pk, number):
...
+            e.detail[0] = "%s %s" % (e.detail[0], url)
+            raise e
+        return resolver.kwargs['pk']
+
+    def create(self, request, repository_pk):
+        add_content_units = []
+        remove_content_units = []
+        if 'add_content_units' in request.data:
+            for url in request.data['add_content_units']:
+                add_content_units.append(self.validate_content_url(url))
```
asmacdo: A problem with this pattern is that the content_urls are validated synchronously. It is conceivable that a user issues a sync task and then a manual remove. They might know the sync will add some packages that they want to remove, so they might do another task to create a new repository version that removes them. This will fail synchronously because the units don't exist yet. 

I wonder if it would be better to just validate the form of the hrefs, to make sure that they could be valid. Then in the task, we just try it and inform the users if any hrefs were not added because they didn't exist.

pulpcore/pulpcore/app/tasks/repository.py
dkliban: I don't think it does. I'll file an issue for this.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -178,6 +187,63 @@ def destroy(self, request, repository_pk, number):
...
+
+        if 'remove_content_units' in request.data:
+            for url in request.data['remove_content_units']:
+                remove_content_units.append(self.validate_content_url(url))
```
daviddavis: It seems weird that you're appending the result of a validation function to a list of content units. Probably should call the function something like `parse_content_url` instead since it parses out the pk and validating the url is a secondary purpose.

dkliban: You are right. Originally I was just passing the whole URL to the task and then I switched to using just the PKs. I'll rename the method.

pulpcore/pulpcore/app/tasks/repository.py
daviddavis: I think this issue already kind of exists here - https://pulp.plan.io/issues/3226

dkliban: This would be for a non-crash scenario. That issue does not specify that the RepositoryContent table needs to be cleaned up also.

daviddavis: Right, I was imagining though that the task would create some sort of cleanup code we could call into. Sorry for the confusion.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -330,6 +331,24 @@ def remove_content(self, content):
         )
         q_set.update(version_removed=self)
 
+    def delete_incomplete_version(self):
+        """
+        Delete's an incomplete Repository Version
```
daviddavis: s/Delete's/Delete/ or Deletes

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -330,6 +331,24 @@ def remove_content(self, content):
...
+
+        This method delete's a RepositoryVersion only if its 'complete' property is False. All
```
daviddavis: here too.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +186,63 @@ def destroy(self, request, repository_pk, number):
         )
         return OperationPostponedResponse([async_result], request)
 
+    def get_content_pk(self, url):
+        """
+        Validates that the path belongs to a Content resource
```
daviddavis: nitpick - this needs to be updated.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +186,63 @@ def destroy(self, request, repository_pk, number):
...
+
+        Args:
+            url (str): href that needs validation
```
daviddavis: this too.

pulpcore/pulpcore/app/tasks/repository.py
```diff
@@ -109,3 +109,42 @@ def delete_version(pk):
 
         # With no more relationships remaining, delete the version.
         version.delete()
+
+
+@shared_task(base=UserFacingTask)
+def add_or_remove(repository_pk, add_content_units, remove_content_units):
```
asmacdo: add_and_remove

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +186,65 @@ def destroy(self, request, repository_pk, number):
...
+                                                            source='*', initial=url)
+        try:
+            content_field.run_validation(data=url)
+        except ValidationError as e:
+            # Append the URL of missing Content to the error message
+            e.detail[0] = "%s %s" % (e.detail[0], url)
+            raise e
+        return resolver.kwargs['pk']
+
+    def create(self, request, repository_pk):
```
asmacdo: Docstring please

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +186,65 @@ def destroy(self, request, repository_pk, number):
         )
         return OperationPostponedResponse([async_result], request)
 
+    def get_content_pk(self, url):
```
asmacdo: get_content_unit_pk please.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +186,65 @@ def destroy(self, request, repository_pk, number):
...
+
+        if 'remove_content_units' in request.data:
+            for url in request.data['remove_content_units']:
+                remove_content_units.append(self.get_content_pk(url))
+
+        result = tasks.repository.add_or_remove.apply_async_with_reservation(
+            tags.RESOURCE_REPOSITORY_TYPE, repository_pk,
+            kwargs={
+                'repository_pk': repository_pk,
+                'add_content_units': add_content_units,
```
asmacdo: Could we just pass request.data.get('add_content_units')? That way all the url wrangling can happen in task land

pulpcore/pulpcore/app/viewsets/repository.py
asmacdo: I also don't think this really belongs on the RepositoryVersion model.

pulpcore/pulpcore/app/viewsets/repository.py
dkliban: I was originally passing the URLs to the task, and I discovered that I was doing the same thing in the viewset and the task. Since I already validated that the URLs map to content units, I think it's ok to just let the database handle anything that goes missing between the REST API call and when the task executes.

pulpcore/pulpcore/app/viewsets/repository.py
dkliban: I'll add one

pulpcore/pulpcore/app/viewsets/repository.py
dkliban: I'll update the name, but I want to leave it in here for now. We should discuss moving all these plugin API methods somewhere else.

pulpcore/pulpcore/app/tasks/repository.py
dkliban: thanks. will update it.

asmacdo:
My main concern is that we are doing some url-wrangling that may not be necessary. For instance, when we create an importer, we pass an argument `repository=repository_href`. I think this works because the Importer is created by the ImporterSerializer, which is aware of hyperlinked fields.

Expanding that concept to this work, I think we could have a RepositoryContent serializer, without needing a RepositoryContentViewset. The RepositoryContent serializer could have Hyperlinked related fields, and the urls could be used directly to make associations.

daviddavis:
@asmacdo I agree. I tried digging into DRF to see how to avoid having to parse PKs out of URLs but gave up after a while. I like your idea though. Since @dkliban has already merged this, should we open a redmine issue?

PR #3273, opened 2018-01-18
Update views docstring to not expose internal details

daviddavis:
The docstring in our views and viewsets is exposed to users through our
REST API schema documentation. This updates our docstring to be more
user friendly. For more info see:

http://www.django-rest-framework.org/api-guide/schemas/#schemas-as-documentation

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/content.py
+++ b/pulpcore/pulpcore/app/viewsets/content.py
@@ -49,20 +49,6 @@ class ContentViewSet(CreateDestroyReadNamedModelViewSet):
 
     @transaction.atomic
     def create(self, request):
-        """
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -83,19 +69,6 @@ class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):
     filter_class = ArtifactFilter
 
     def destroy(self, request, pk):
-        """
-        Remove :class:`~pulpcore.app.models.Artifact` only if it is not associated with any
```
werwty: Maybe we should keep this help text?

PR #3275, opened 2018-01-18
Consolidated tasking modules in pluign API.

jortel:
https://pulp.plan.io/issues/3280

Also, refactored the services.storage functions into a WorkingDirectory class (context manager).

File plugin: https://github.com/pulp/pulp_file/pull/30

### DIFF
--- a/pulpcore/pulpcore/tasking/services/storage.py
+++ b/pulpcore/pulpcore/tasking/services/storage.py
@@ -1,152 +1,128 @@
-import errno
 import os
 import shutil
-import stat
-from contextlib import contextmanager, suppress
+
+from gettext import gettext as _
 
 from celery import task
-from django.conf import settings as pulp_settings
+from django.conf import settings
 
 
-def _working_dir_root(worker_name):
-    """Return the path to the working directory of a worker.
-
-    :param worker_name:     Name of worker for which path is requested
-    :type  worker_name:     basestring
-
-    :returns: working_directory setting from server config
-    :rtype:   str
+class WorkingDirectory:
     """
-    return os.path.join(pulp_settings.SERVER['working_directory'], worker_name)
+    A context manager used to manage a Celery task's working directory.
 
-
-def create_worker_working_directory(worker_name):
-    """Create a working directory for a worker.
-
-    Create a working directory inside the worker working_directory as specified in the setting file.
-
-    :param worker_name:     Name of worker that uses the working directory created
-    :type  worker_name:     basestring
-
-    :returns: Path to the working directory for a worker
-    :rtype:   str
+    Examples:
+        >>>
+        >>> with WorkingDirectory.create() as working_dir:
+        >>>     ...
+        >>>
     """
-    working_dir_root = _working_dir_root(worker_name)
-    os.mkdir(working_dir_root)
-    return working_dir_root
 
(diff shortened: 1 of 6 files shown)


pulpcore/pulpcore/app/tasks/publisher.py
```diff
@@ -6,7 +6,7 @@
 from django.db import transaction
```
asmacdo: I just deleted this module, probably the source of the conflict.

pulpcore/pulpcore/tasking/services/storage.py
```diff
@@ -1,152 +1,102 @@
...
+            WorkingDirectory: self
+
+        Raises:
+            OSError: On failure.
+        """
+        self.create()
+        os.chdir(self.path)
+        return self
+
+    def __exit__(self, *ignored):
```
asmacdo: IIRC, this will cause any exceptions to be swallowed. I think we want to reraise.

jortel: Exception raised within the content do not need to be raised in the __exit__() to prevent swallowing.  The arguments passed to __exit__() are so that the code can know an exception was raised within the content so it can make clean up decisions.  No to swallow them.

jortel:
ok test

pulpcore/pulpcore/tasking/services/storage.py
```diff
@@ -1,152 +1,104 @@
...
+        try:
+            os.makedirs(self.path)
+        except FileExistsError:
+            self.delete()
+            os.makedirs(self.path)
+
+    def delete(self):
+        """
+        Delete the directory (tree).
+        On permission denied - an attempt is made to recursively fix the
```
asmacdo: new line before "On permission". The first line is supposed to be a summary and fit on 1 line for autodocs reasons.

pulpcore/pulpcore/tasking/services/storage.py
```diff
@@ -1,152 +1,104 @@
...
+
+        Returns:
+            str: The absolute directory path.
+        """
+        return self._path
+
+    def create(self):
+        """
+        Create the directory.
+        Deleted and recreated when already exist.
```
asmacdo: same

pulpcore/pulpcore/tasking/services/storage.py
```diff
@@ -1,152 +1,104 @@
...
+        """
+        self.create()
+        os.chdir(self.path)
+        return self
+
+    def __exit__(self, *unused):
+        """
+        Delete the directory (tree).
+        """
+        self.delete()
```
asmacdo: Do we need to os.chdir back?

pulpcore/pulpcore/tasking/services/storage.py
```diff
@@ -1,152 +1,104 @@
...
-    working_dir_root = _working_dir_root(worker_name)
-    os.mkdir(working_dir_root)
-    return working_dir_root
-
-
-def delete_worker_working_directory(worker_name):
-    """Delete worker's working directory.
-
-    Delete a working directory inside the working_directory as specified in the setting file.
+    Task working directory.
```
asmacdo: Context manager for creating and using a working directory in a celery task.

PR #3279, opened 2018-01-22
Removes 'create' from Publications viewset

asmacdo:
This patch also removes publish task from pulpcore


pulpcore/pulpcore/app/models/publication.py
```diff
@@ -18,7 +18,7 @@ class Publication(Model):
 
     publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
 
-    repo_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
+    repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
```
werwty: s/repo_version/repository_version/ in docstring

werwty:
Please also fix PEP8 errors before merge. Everything else looks good

PR #3283, opened 2018-01-24
Fixes workers crashing on restart edge case

dralley:
Fixes an issue where workers would crash when
attempting to revoke() a previously running task
on startup. If the user sends SIGKILL to the
worker while running a task, the worker will
attempt to mark the task as stale when restarted.
This causes a hang for yet-unknown reasons. Calling
revoke() is not necessary when a dead worker is being
cleaned up because all work goes through dedicated
queues which disappear when the worker does. We
leave revoke() active only when a task is being
cancelled outside of worker cleanup.

closes #2835
https://pulp.plan.io/issues/2835

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -270,7 +270,7 @@ def _delete_worker(name, normal_shutdown=False):
     # Cancel all of the tasks that were assigned to this worker's queue
     for task_status in TaskStatus.objects(worker_name=name,
                                           state__in=constants.CALL_INCOMPLETE_STATES):
-        cancel(task_status['task_id'])
+        cancel(task_status['task_id'], revoke_task=False)
 
 
 @task(base=PulpTask)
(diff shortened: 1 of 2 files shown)


bmbouter:
I just hand tested this. Against the current head of master f223542d7dfe08e9ad4e63add83120b73e162596 I was able to reproduce the deadlocking when killing a worker with sigkill, restarting it, and then sending it new work. With this patch I confirmed that I could not reproduce the issue, and that new work started as expected.

dralley:
@bmbouter Any objections to the logic of the change?

server/pulp/server/async/tasks.py
```diff
@@ -631,14 +631,18 @@ def _handle_cProfile(self, task_id):
             self.pr.dump_stats("%s/%s" % (profile_directory, task_id))
 
 
-def cancel(task_id):
+def cancel(task_id, worker_cleanup=False):
```
bmbouter: What about making this `worker_cleanup=True` and making the if statement instead be `if worker_cleanup`.

I think that would be clearer because currently if worker_cleanup is False, it actually does cleanup.

daviddavis:
@dralley if you're going to go back and make changes, can I request the changes to server/pulp/server/async/app.py be removed from this PR and be opened in a separate PR?

server/pulp/server/async/tasks.py
dralley: I see your point about this being unclear but I disagree that the logic should be changed (the naming, perhaps. I think you parsed it in the opposite way from what I intended).  

This is the exception, not the rule - only once place in the codebase attempts to cancel tasks during worker cleanup, but lots of places in the codebase attempt to cancel tasks during normal operation.  If we were to flip this default, we would then need to specify ```worker_cleanup=False``` on every call to ```cancel()``` but this one, and remember to do so in the future.

bmbouter: The changes I'm suggestion would keep it as the exception not the rule because in addition to changing the default value from False to True you would also change the s/if not worker_cleanup/if worker_cleanup/

Since the default would be to perform cleanup, we would only need to specify `worker_cleanup=False` on that one call that wants to modify the "cleanup by default" behavior.

dralley: Ah, so this is basically switching the semantic meaning of ```worker_cleanup```.  That is reasonable, I'll change it.

dralley:
@daviddavis sure

server/pulp/server/async/tasks.py
bmbouter: Exactly! Thank you.

dralley:
@bmbouter @daviddavis changes made

PR #3285, opened 2018-01-25
add repoversion context_manager

werwty:


plugin/pulpcore/plugin/repository.py
```diff
@@ -0,0 +1,171 @@
+from contextlib import suppress
+from gettext import gettext as _
+from django.db import transaction
+from django.db.utils import IntegrityError
+
+
+from pulpcore.app import models
+from pulpcore.tasking.util import get_current_task_id
```
jortel: I think using `from pulpcore.plugin.tasking import Task` is cleaner.

jortel:
I'm a concerned that the use of _assertions_ here instead of raising exceptions isn't appropriate.  Assertions should be used to ensure the integrity of programming and not for reporting error conditions as explained here[1].  Assertions are more commonly enable and used for testing.  Not production code  Further, as with many programming languages, assertions can an often are turned off in production.  IIRC, for python it's the `-O` option.  This seems dangerous because with exceptions disabled, this code could be used to do bad things.  For example, add content to a complete repository version.  Also, any raised exceptions should be documented in the docstring:

```
Raises:
    AssertionError: When adding content to a complete version.
```

Requires the call to potentially be:

```
Try:
    version.add_content()
except AssertionError:
    ...
```

which seems undesirable.


[1] http://www.drdobbs.com/jvm/assertations-vs-exceptions/228701655

plugin/pulpcore/plugin/repository.py
```diff
@@ -0,0 +1,171 @@
...
+        assert not self._model.complete, _("Version {version} for repository {repo} is complete.").\
+            format(version=self._model.number, repo=self._model.repository.name)
+        with transaction.atomic():
+            models.RepositoryContent.objects.filter(version_added=self._model).delete()
+            models.RepositoryContent.objects.filter(version_removed=self._model)\
+                .update(version_removed=None)
+            self._model.repository.last_version = self._model.number - 1
+            self._model.repository.save()
+            self._model.delete()
+            if self._created_resource is not None:
```
jortel: Suggest `CreatedResource.objects.delete(content_object=self._model)` so you don't need to store `self._created_resource` as an attribute.  It would be cleaner.

plugin/pulpcore/plugin/repository.py
```diff
@@ -0,0 +1,171 @@
...
+            repository=self._model.repository,
+            content=content,
+            version_removed=None)
+        q_set.update(version_removed=self._model)
+
+    def complete(self):
+        """
+        Save the repository version when all operations are done on it
+        """
+        with transaction.atomic():
```
jortel: Why is this in a transaction?

werwty: Whoops, this does not need to be, will remove

plugin/pulpcore/plugin/repository.py
werwty: I am not sure if this is possible without updating CreatedResource.content_object to be a GenericRelation [0] 
Running `` models.CreatedResource.objects.filter(content_object=self._model).delete()`` generates the error message
``django.core.exceptions.FieldError: Field 'content_object' does not generate an automatic reverse relation and therefore cannot be used for reverse querying. If it is a GenericForeignKey, consider adding a GenericRelation.``

[0] https://stackoverflow.com/questions/36164654/how-to-traverse-a-genericforeignkey-in-django/36166644#36166644

werwty:
@jortel back to you

plugin/pulpcore/plugin/repository.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns:
+            pulpcore.plugin.repository.RepositoryVersion: The Created RepositoryVersion
+
+        Raises:
+            pulpcore.exception.UnsafeOperationError: if RepositoryVersion create() is called
+                outside of a task.
+        """
+
+        if get_current_task_id() is None:
+            raise UnsafeOperationError(_('RepositoryVersion creation must be ran inside a task'))
```
asmacdo: s/ran/run/

plugin/pulpcore/plugin/repository.py
```diff
@@ -0,0 +1,191 @@
...
+
+        Args:
+           content (pulpcore.plugin.models.Content): a content model to add
+
+        Raise:
+            pulpcore.exception.RepositoryVersionExistsError: if add_content is called on a
+                complete RepositoryVersion
+        """
+        if self._model.complete:
+            raise RepostoryVersionExistsError(self._model)
```
asmacdo: I dont like RepositoryVersionExists for this. The RepositoryVersion exists either way, it is just complete or not. UnsafeOperationError seems better.

pulpcore/pulpcore/exceptions/repository.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Args:
+            repository_version (pulpcore.app.models.RepositoryVersion):
+                The complete repositoryversion that the user is trying to update
+        """
+        super(UnsafeOperationError, self).__init__("PLP0004")
+        self.msg = msg
+
+    def __str__(self):
+        return self.msg.encode('utf-8')
```
jortel: Let's consider: `ResourceImmutableError`

pulpcore/pulpcore/exceptions/repository.py
```diff
@@ -0,0 +1,45 @@
+from gettext import gettext as _
+
+from .base import PulpException
+
+
+class RepostoryVersionExistsError(PulpException):
```
jortel: Let's consider renaming as: `ResourceImmutableError`.  I think it more accurately and generically represents the error condition.

pulpcore/pulpcore/exceptions/base.py
```diff
@@ -45,3 +45,24 @@ def exception_to_dict(exc, traceback=None):
...
         result['code'] = exc.error_code
     return result
+
+
+class ResourceImmutableError(PulpException):
+    """
+    Exceptions that are raised due to trying to update an immutable resource
+    """
+
+    def __init__(self, resource_name, resource_type):
```
jortel: Consider passing resource that _IsA_ `Model`.  This seems more user friendly.

PR #3288, opened 2018-01-25
Updated certificates in /oid_validation and /repodata

dralley:
Updated certificates using oid_validation/test/data/regenerate.sh

closes #3319
https://pulp.plan.io/issues/3319


dralley:
@daviddavis This seems relevant to https://pulp.plan.io/issues/2733

It seems like the regenerate.sh script used to generate these certs might need to be updated?  It looks like my PR is changing some of the things that your PR originally changed - that the script is generating the certificates incorrectly.

daviddavis:
@dralley I suspect that the change in formatting you're seeing is due to different openssl versions. In fact, when I checkout our your PR and run regenerate.sh, I get the format that's currently on master. I am using `openssl-1.0.1e-42.el7.9.x86_64`.

Looks like we need to update repoauth/test/data/chain/certs/test_cert.pem and it doesn't look like regnerate.sh handles that currently.

PR #3289, opened 2018-01-25
Clean up incomplete created-resources.

jortel:
https://pulp.plan.io/issues/3226

Also fixed a few issues with `RepositoryVersion`.


jortel:
@daviddavis, yes, I missed that one.  good catch.

PR #3292, opened 2018-01-27
Adds lots of plugin writer content

bmbouter:


docs/plugins/plugin-writer/planning-guide.rst
```diff
@@ -71,14 +71,12 @@ command line tool.
...
-
-It is important to understand how a content type needs to be made available for a client to consume
-it. Typically *http://* is used which allows clients to fetch content. the plugin API is well suited
-to the serving of files via *http(s)://*.
+How is Content Discovered?
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If the client does not already know the url to download content with, there are two typical designs
+It is important to understand how a content type is discovered by a client. In the simplest case,
+the client is explicitly told a url to fetch the client via, for example via *http://*. If the
```
dkliban: s/client/content unit/

bmbouter: fixed in next push.

docs/plugins/plugin-writer/planning-guide.rst
```diff
@@ -0,0 +1,168 @@
...
+package tarball or egg via *http://* also.
+
+In the simplest case, *pip* is told the url and can download it without "discovering" it. For
+example with a command like:
+
+``pip3 install -e "git+https://github.com/pulp/pulp.git@3.0-dev#egg=pulpcore&subdirectory=common"``
+
+The *Remote Content Unit* can be downloaded using the url.
+
+Here is an example of how *pip* downloads a specific *Remote Content Unit* that is needs to
```
dkliban: s/that is/that it/

bmbouter: fixed in next push

PR #3294, opened 2018-01-29
Cleaning up some docstring

daviddavis:


jortel:
LGTM but need to fix the docbuilder failure.

PR #3299, opened 2018-01-30
Add complete field to Publication

dparalen:
Only complete Publications can be distributed.

closes: #3184
https://pulp.plan.io/issues/3184

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -6,6 +6,7 @@ from pulpcore.app.models import Model, storage
 class Publication(Model):
     """
     Fields:
+        complete (models.BooleanField): State tracking; for internal use. Indexed.
         created (models.DatetimeField): When the publication was created UTC.
 
     Relations:
@@ -14,11 +15,17 @@ class Publication(Model):
             create this Publication.
     """
 
+    complete = models.BooleanField(default=False)
     created = models.DateTimeField(auto_now_add=True)
 
     publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
 
     repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
+
+    class Meta:
+        indexes = [
+            models.Index(fields=['complete']),
+        ]
 
 
 class PublishedFile(Model):
--- a/pulpcore/pulpcore/app/serializers/repository.py
+++ b/pulpcore/pulpcore/app/serializers/repository.py
@@ -226,7 +226,7 @@ class DistributionSerializer(ModelSerializer):
     publication = serializers.HyperlinkedRelatedField(
         allow_null=True,
         help_text=_('The publication being served as defined by this distribution'),
-        queryset=models.Publication.objects.all(),
+        queryset=models.Publication.objects.exclude(complete=False),
         view_name='publications-detail'
     )
     base_url = BaseURLField(
--- a/pulpcore/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/pulpcore/app/viewsets/repository.py
@@ -244,7 +244,7 @@ class PublicationViewSet(GenericNamedModelViewSet,
                          mixins.ListModelMixin,
                          mixins.DestroyModelMixin):
     endpoint_name = 'publications'
-    queryset = Publication.objects.all()
+    queryset = Publication.objects.exclude(complete=False)
     serializer_class = PublicationSerializer
 
 


pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -239,11 +239,27 @@ class PublisherViewSet(CreateReadAsyncUpdateDestroyNamedModelViewset):
...
 
 
+class PublicationFilter(filterset.FilterSet):
+    """
+    A base Publication filter that always excludes incomplete Publications.
+    """
+
+    class Meta:
+        model = Publication
+        fields = ('created', 'publisher', 'repository_version', 'complete')
```
dparalen: I'm unsure whether the `complete` field should be listed here; unsure what the purpose of the `Meta.fields` actually is. More docs digging I shall....

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -239,11 +239,28 @@ class PublisherViewSet(CreateReadAsyncUpdateDestroyNamedModelViewset):
     filter_class = PublisherFilter
 
 
+class PublicationFilter(filterset.FilterSet):
```
asmacdo: You don't need to add a filterset as part of this work. FilterSets are required for filtering over the REST API. The model layer queryset.filter() is unrelated.

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -190,6 +190,28 @@ class Meta:
         ]
 
 
+class HyperlinkedRelatedPublicationField(serializers.HyperlinkedRelatedField):
+    """
+    Avoids the API consumer passing an incomplete Publication to a Serializer.
+    """
+    queryset = models.Publication.objects.all()
```
asmacdo: I think a similar effect can be reached by `queryset = models.Publication.objects.exclude(complete=False)` without overriding `get_object`

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -20,6 +20,8 @@ class Publication(Model):
 
     repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
 
+    complete = models.BooleanField(default=False)
```
jortel: An unwritten convention has been to group the fields together followed by the relations (grouped).  For consistency, please move the `complete` field to be with `created` and add to `Fields` in the docstring.

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -223,11 +245,9 @@ class DistributionSerializer(ModelSerializer):
                     'served as defined by this distribution'),
         queryset=models.Publisher.objects.all(),
     )
-    publication = serializers.HyperlinkedRelatedField(
+    publication = HyperlinkedRelatedPublicationField(
```
asmacdo: We may not need to add a new publication field. Instead, we can just set the queryset to exclude complete=false here and use the vanilla HyperlinkedRelatedField

asmacdo:
I made a similar change recently to RepositoryVersions:
https://github.com/pulp/pulp/pull/3268/

pulpcore/pulpcore/app/serializers/repository.py
dparalen: ack.

pulpcore/pulpcore/app/models/publication.py
dparalen: will do.

pulpcore/pulpcore/app/serializers/repository.py
dparalen: ack; I'm all in for reducing the commit size :D

pulpcore/pulpcore/app/viewsets/repository.py
dparalen: I guess we better have an explicit filter to only allow API filtering on the `Meta.fields`, without the `complete` field, so I'd rather keep this, probably changing the `parent.filter(complete=True)` to `parent.exclude(complete=True)`

pulpcore/pulpcore/app/serializers/repository.py
dparalen: Actually I think a specific hyperlink here serves both reusability and documentation of intention. But I'm also cool with the vanilla hyperlink; just posted a revision to see how it feels the other way.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -14,6 +15,7 @@ class Publication(Model):
             create this Publication.
     """
 
+    complete = models.BooleanField(default=False)
```
jortel: The complete field needs an index.

dparalen: I guess that's to make the `.exclude()` effective? Will update..

jortel:
LGTM but defer to @asmacdo for approval on the django-ness.

pulpcore/pulpcore/app/serializers/repository.py
asmacdo: I can see the benefit of creating a custom field, but I think its helpful to keep these explicit rather than create classes that hide the details. If we follow this pattern, we would have a custom field class for every relationship, which might be easier to use, but makes the serializers more opaque. I strongly prefer vanilla, especially in this case, since there is only one Publication relation.

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -266,6 +272,7 @@ class PublicationSerializer(ModelSerializer):
         parent_lookup_kwargs={'repository_pk': 'repository__pk'},
         read_only=True,
     )
+    # NOTE: models.Publication.complete field is being left out on purpose
```
asmacdo: We don't specify other fields that are left out on other serializers, so I don't think we need to here, but its just my opinion. If you want to keep this, please explain why it is left out.

pulpcore/pulpcore/app/viewsets/repository.py
asmacdo: We can set the queryset to exclude(complete=False) on the Viewset.queryset instead, which will exclude for Read, List, and filter. As it is, v3/publications/ without filters will still include complete=False. I don't see a story to add filters to Publications but we should write one, and the filters we offer should be discussed. For instance "created" should probably be a date range filter.

dparalen:
@asmacdo @jortel  thanks for the reviews!
I hope I've addressed all the concerns in this revision.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -6,6 +6,7 @@
 class Publication(Model):
     """
     Fields:
+        complete (models.BooleanField): State tracking; for internal use. Indexed.
```
asmacdo: Great comment

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -14,12 +15,18 @@ class Publication(Model):
...
 
+    complete = models.BooleanField(default=False)
     created = models.DateTimeField(auto_now_add=True)
 
     publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
 
     repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
 
+    class Meta:
+        indexes = [
```
asmacdo: I think models.BooleanField(default=False, db_index=True) will do the same thing.

dparalen:
@jortel or @asmacdo guys, please merge as I don't have the bit ;)

Thanks!
milan

pulpcore/pulpcore/app/models/publication.py
jortel: Agreed with @asmacdo and would be more consistent with the other models.

PR #3301, opened 2018-01-31
Move RepositoryVersion context manger to the model

werwty:


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -1,10 +1,17 @@
...
 Repository related Django models.
 """
+from contextlib import suppress
 from django.db import models
+from django.db import transaction
+from django.db.utils import IntegrityError
+
 
 from pulpcore.app.models import Model, Notes, MasterModel, GenericKeyValueRelation
+from .task import CreatedResource
```
werwty: This import was done like this to avoid circular imports, would love suggestions on how to better do this.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -1,10 +1,18 @@
 """
 Repository related Django models.
 """
+from contextlib import suppress
 from django.db import models
+from django.db import transaction
+from django.db.utils import IntegrityError
+
+
```
jortel: Extra blank.

PR #3302, opened 2018-02-01
Remove relationship between publishers/importers and repos

daviddavis:
closes #3341
https://pulp.plan.io/issues/3341

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -73,6 +73,16 @@ class Publication(Model):
             resource.save()
             return publication
 
+    @property
+    def repository(self):
+        """
+        Return the associated repository
+
+        Returns:
+            pulpcore.app.models.Repository: The repository associated to this publication
+        """
+        self.repository_version.repository
+
     def delete(self, **kwargs):
         """
         Delete the publication.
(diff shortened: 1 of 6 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -158,9 +119,18 @@ def tls_storage_path(self, name):
     sync_mode = models.TextField(choices=SYNC_MODES)
     last_synced = models.DateTimeField(blank=True, null=True)
 
-    class Meta(ContentAdaptor.Meta):
+    class Meta:
         default_related_name = 'importers'
 
+    def natural_key(self):
```
asmacdo: I think we can get rid of the natural key since it isn't used in the url.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -85,7 +85,9 @@ class Distribution(Model):
 
     Relations:
         publisher (models.ForeignKey): The associated publisher.
-            All publications created by the specified publisher will be automatically associated.
+            All publications created by the specified publisher/repo combo will be automatically
+            associated.
```
jortel: Consider:
"All publications of the repository that are created by the publisher will be automatically associated."

jortel:
LGTM

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -180,6 +195,7 @@ class Distribution(Model):
 
     publication = models.ForeignKey(Publication, null=True, on_delete=models.SET_NULL)
     publisher = models.ForeignKey('Publisher', null=True, on_delete=models.SET_NULL)
+    repository = models.ForeignKey('Repository', null=True, on_delete=models.SET_NULL)
 
     class Meta:
         unique_together = ('publisher', 'name')
```
jortel: Should the distribution name needs to be unique my itself?

PR #3306, opened 2018-02-01
Publishing context manager.

jortel:
https://pulp.plan.io/issues/3295

I originally added `add_artifact()` and `add_metadata()` as defined by the issue but after writing them they did not seem to provide much value.  Perhaps more discussion is needed.

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -1,31 +1,106 @@
-from django.db import models
+from django.db import models, transaction
 
-from pulpcore.app.models import Model, storage
+from . import storage
+from .base import Model
+from .task import CreatedResource
 
 
 class Publication(Model):
     """
+    A publication contains metadata and artifacts associated with content
+    contained within a RepositoryVersion.
+
+    Using as a context manager is highly encouraged.  On context exit, the complete attribute
+    is set True provided that an exception has not been raised.  In the event and exception
+    has been raised, the publication is deleted.
+
     Fields:
         complete (models.BooleanField): State tracking; for internal use. Indexed.
         created (models.DatetimeField): When the publication was created UTC.
 
     Relations:
         publisher (models.ForeignKey): The publisher that created the publication.
-        repository_version (models.ForeignKey): The RepositoryVersion whose content set was used to
+        repository_version (models.ForeignKey): The RepositoryVersion used to
             create this Publication.
+
+    Examples:
+        >>> publisher = ...
+        >>> repository_version = ...
+        >>>
+        >>> with Publication.create(repository_version, publisher) as publication:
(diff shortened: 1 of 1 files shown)


pulpcore/pulpcore/app/models/publication.py
```diff
@@ -20,6 +42,68 @@ class Publication(Model):
 
     repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
 
+    complete = False  # placeholder
```
asmacdo: dont forget to make this a model.Boolean

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -20,6 +42,68 @@ class Publication(Model):
...
+
+        Args:
+            **kwargs (dict): Delete options.
+
+        Notes:
+            Deletes the Task.created_resource when incomplete.
+        """
+        with transaction.atomic():
+            if not self.complete:
+                CreatedResource.objects.delete(object_id=self.pk)
```
asmacdo: I think we should delete the CreatedResource even if the Publication was never marked complete.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -1,17 +1,39 @@
...
 
-from pulpcore.app.models import Model, storage
+from . import storage
+from .base import Model
+from .task import CreatedResource
 
 
 class Publication(Model):
     """
+    A publication.
```
asmacdo: From the language guide: 
Publications refer to the metadata and arifacts of the content units in a repository version.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -20,6 +42,68 @@ class Publication(Model):
...
+            repository_version (pulpcore.app.models.RepositoryVersion): The repository
+                version to be published.
+            publisher (pulpcore.app.models.Publisher): The publisher used
+                to create the publication.
+
+        Returns:
+            pulpcore.app.models.Publication: self
+
+        Notes:
+            The created publication with complete=False.
```
asmacdo: Would you move this note to the returns line? Something like
Returns:
     A created Publication in an incomplete state.

jortel: 👍

pulpcore/pulpcore/app/models/publication.py
jortel: The complete field will be added by #3299.  This just a placeholder to get it working that will be removed before merge.

pulpcore/pulpcore/app/models/publication.py
jortel: 👍

PR #3307, opened 2018-02-06
Make config.get_boolean return given default value

mdellweg:
Greetings from CfgMgmtCamp!


dkliban:
ok test

mdellweg:
@dkliban Thanks for the advise.

PR #3308, opened 2018-02-06
Update versions for 2.16.0 alpha

pcreech:
Update versions for 2.16.0 alpha


pcreech:
I'm proposing to ignore the pep8 issues here, as this is automated code changes from long standing releng tooling

PR #3309, opened 2018-02-06
Add RepositoryVersion Filter

werwty:
fixes #3238
https://pulp.plan.io/issues/3238

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/pulpcore/app/viewsets/custom_filters.py
@@ -9,3 +9,4 @@ class CharInFilter(filters.BaseInFilter, filters.CharFilter):
     Enables the user to filter a field by comma separated strings, allowing them to retrieve more
     than one object in a single query.
     """
+    pass
--- a/pulpcore/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/pulpcore/app/viewsets/repository.py
@@ -1,7 +1,8 @@
+import itertools
+
 from django_filters.rest_framework import filters, filterset
 from django_filters import CharFilter
 from rest_framework import decorators, mixins
-
 
 from pulpcore.app import tasks
 from pulpcore.app.models import (Content,
@@ -10,6 +11,7 @@ from pulpcore.app.models import (Content,
                                  Publication,
                                  Publisher,
                                  Repository,
+                                 RepositoryContent,
                                  RepositoryVersion)
 from pulpcore.app.pagination import UUIDPagination, NamePagination
 from pulpcore.app.response import OperationPostponedResponse
@@ -29,11 +31,10 @@ from pulpcore.common import tags
 
 class RepositoryFilter(filterset.FilterSet):
     name_in_list = CharInFilter(name='name', lookup_expr='in')
-    content_added_since = filters.Filter(name='last_content_added', lookup_expr='gt')
 
     class Meta:
         model = Repository
-        fields = ['name', 'name_in_list', 'content_added_since']
+        fields = ['name', 'name_in_list']
 
 
 class RepositoryViewSet(NamedModelViewSet):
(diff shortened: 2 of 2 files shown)


werwty:
The filter can be tested with the following query: http://pulp3.dev:8000/api/v3/repositories/f6bf16e4-e419-482b-8032-942e3fa5b93b/versions/?version_range=3,4&created_0=2018-02-06 19:27&created_1=&content_in=1ac31f86-04ff-47cf-a3ff-2578dcaebeec,00ab453f-dbb0-481e-9934-cf68784989dd

daviddavis:
I was thinking that the ranges would be like two fields like `created_after` and `created_before`. Not sure I understand the field names `created_0` and `created_1`.

Also, I think that the content filter should accept hrefs like we do when you create a repo version with a list of content units:

https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/viewsets/repository.py#L227

daviddavis:
It's late so apologies if I am not thinking clearly but it seems like your content filter algorithm is not removing one version number when it generates it range. In other words, if I search for a content unit that was added in version 4 and removed in version 6, will your code with search versions 4-6 or will it only search versions 4 and 5?

daviddavis:
Nevermind, I see that `range(4, 6)` will return `[4, 5]`. Ignore my previous comment 😄

daviddavis:
Is this `version_range` really a `range` filter or is it a `in` filter? I would expect `version_range=3,6` to return all versions from 3-6 whereas with `version_in=3,6` I would expect versions 3 and 6 to be returned.

werwty:
Thanks for the review! I updated so that instead of trying to use a fancy DateTimeFromToRangeFilter (which generated the `created_0` and `created_1` queries, that needed a custom widget to override https://django-filter.readthedocs.io/en/master/ref/widgets.html#suffixedmultiwidget ) I just separated it into two filters.

I did the same for the VersionRange filter, since a `NumberRangeFilter(BaseRangeFilter, NumberFilter)` expected both a lower and upper bound, and doing it through two filters does not.

`RepositoryVersionContentInFilter` was also updated to take content hrefs.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -128,6 +129,86 @@ class Meta:
...
+        2. Get a list of version_added and version_removed where the content was
+           changed on the repository
+        3. Calculate and return the versions that the contents can be found on
+
+    """
+
+    def filter(self, qs, value):
+        """
+        Args:
+            qs (pulpcore.app.viewsets.repository.RepositoryVersionViewSet.queryset):
```
asmacdo: This type is very informative, but I think it violates the spirit of the docstring conventions. It probably ought to be:

qs (django.db.models.query.QuerySet)

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -128,6 +129,86 @@ class Meta:
...
+class RepositoryVersionContentInFilter(CharInFilter):
+    """
+    Filter used to get the repository versions where some given content can be found.
+
+    Given a list of content_id, this filter will:
+        1. Get the RepositoryContent that the id can be found in
+        2. Get a list of version_added and version_removed where the content was
+           changed on the repository
+        3. Calculate and return the versions that the contents can be found on
+
```
asmacdo: No space here

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -30,17 +32,16 @@
...
                                    GenericNamedModelViewSet,
                                    CreateReadAsyncUpdateDestroyNamedModelViewset)
-from pulpcore.app.viewsets.custom_filters import CharInFilter
+from pulpcore.app.viewsets.custom_filters import CharInFilter, NumberRangeFilter
 from pulpcore.common import tags
 
 
 class RepositoryFilter(filterset.FilterSet):
     name_in_list = CharInFilter(name='name', lookup_expr='in')
-    content_added_since = filters.Filter(name='last_content_added', lookup_expr='gt')
```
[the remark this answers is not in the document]
werwty: the last_content_added field was removed from Repository a while back

daviddavis:
The code looks good. I just realized though: if I filter on content_in=A,B,C and I have a version that contains only A, will it be returned?

werwty:
@daviddavis it would be returned. Currently this is an 'OR' filter, and would return all repositories containing (A | B | C)

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -128,6 +129,87 @@ class Meta:
...
+
+        Returns:
+            Queryset of the RepositoryVersions containing specified content
+        """
+
+        # If no value is passed, return queryset as-is
+        if not value:
+            return qs
+
+        contend_pks = [RepositoryVersionViewSet.get_content_unit_pk(url) for url in value]
```
daviddavis: [A PR just got merged that affects this](https://github.com/pulp/pulp/pull/3311). I think you just need this now:

```python
contend_pks = [GenericNamedModelViewSet.get_resource(url, Content).pk for url in value]
```

daviddavis:
One comment. Otherwise LGTM.

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -9,3 +9,8 @@ class CharInFilter(filters.BaseInFilter, filters.CharFilter):
     Enables the user to filter a field by comma separated strings, allowing them to retrieve more
     than one object in a single query.
     """
+    pass
+
+
+class NumberRangeFilter(filters.BaseInFilter, filters.NumberFilter):
```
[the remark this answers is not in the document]
werwty: I updated this to use 2 NumberFilters:
```
version_min = filters.NumberFilter(name='number', lookup_expr='gte')
version_max = filters.NumberFilter(name='number', lookup_expr='lte')```

Mainly so we don't get `version_0=&version_1=` in the field name

PR #3312, opened 2018-02-07
Importer and publisher names must be unique.

jortel:
For importer and publisher, the `name` is the natural key and must be unique.


pulpcore/pulpcore/app/models/publication.py
```diff
@@ -81,7 +81,7 @@ def repository(self):
         Returns:
             pulpcore.app.models.Repository: The repository associated to this publication
         """
-        self.repository_version.repository
+        return self.repository_version.repository
```
jortel: Noticed this while editing.

PR #3315, opened 2018-02-08
Support Exporters.

jortel:
https://pulp.plan.io/issues/2894


pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -102,6 +110,15 @@ class Meta:
         fields = ['name', 'last_updated']
 
 
+class ExporterFilter(filterset.FilterSet):
+    class Meta:
+        model = Exporter
+        fields = [
+            'name',
+            'last_updated'
```
asmacdo: last_export seems more useful as a filter.

jortel: agreed

PR #3318, opened 2018-02-09
Prevent duplicate content

daviddavis:
Prevent users/plugins from adding content to a repo version if that content would otherwise already exist for that repo version.

Added validation to the viewset so we don't kick off a task and a nice user error gets return if the content addition/removal isn't allowed. Also added validation to the model mostly for the benefit of plugins who might call these methods directly.

closes #3304
https://pulp.plan.io/issues/3304

### DIFF
--- a/pulpcore/pulpcore/app/models/repository.py
+++ b/pulpcore/pulpcore/app/models/repository.py
@@ -4,7 +4,6 @@ Repository related Django models.
 from contextlib import suppress
 from django.db import models
 from django.db import transaction
-from django.db.utils import IntegrityError
 
 from .base import Model, MasterModel
 from .content import Content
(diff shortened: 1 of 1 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -373,12 +382,15 @@ def add_content(self, content):
...
             raise ResourceImmutableError(self)
-        with suppress(IntegrityError):
-            association = RepositoryContent(
-                repository=self.repository,
-                content=content,
-                version_added=self)
-            association.save()
+
+        if self.content_exists(content):
+            raise DuplicateContentException(content, self.repository)
```
jortel: Do we really want to raise an exception here?  Perhaps add_content() should be idempotent.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -395,6 +407,9 @@ def remove_content(self, content):
         if self.complete:
             raise ResourceImmutableError(self)
 
+        if not self.content_exists(content):
```
jortel: Do we really want to raise an exception here?  Perhaps remove_content() should be idempotent.

daviddavis:
@jortel updated the PR based on your feedback.

daviddavis:
ok test

PR #3321, opened 2018-02-09
 Don't define 'queue' in the task detail view if the worker is None 

dralley:
Queue should not be defined in the task detail view when the worker is none.

### DIFF
--- a/server/pulp/server/webservices/views/tasks.py
+++ b/server/pulp/server/webservices/views/tasks.py
@@ -125,7 +125,7 @@ class TaskResourceView(View):
             raise MissingResource(task_id)
 
         task_dict = task_serializer(task)
-        if 'worker_name' in task_dict:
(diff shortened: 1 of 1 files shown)


dralley:
Would it?  The problem is that currently it is returning something entirely invalid, a queue called "None.dq".  An alternate fix would be to make Worker.queue_name property return an empty string in that situation instead.

bmbouter:
Semver allows for backwards incompatible changes if its for a bugfix. Given that it would make a backwards incompatible change filing a formal bug to associate with this commit would be good.

I also think that having it set to an empty string would be better than omitting the key. What do you think about that?

dralley:
I wouldn't have a problem doing it that way.

dralley:
I'll close this PR and make a new one with that fix.

bmbouter:
@dralley Thanks!

PR #3322, opened 2018-02-12
Fix case where queue_name property returns None.dq

dralley:
The queue_name property does not handle the case where the worker name
is None.


dralley:
I will file an issue

bmbouter:
This looks great. Thanks @dralley!

PR #3325, opened 2018-02-12
Fields: created & last_updated moved to Model.

jortel:
Related to https://pulp.plan.io/issues/3337, the FileContent needed to be ordered by a `created` field.  This seemed generally useful for other plugins for the same reason.  Then, considered that a `created` timestamp would be useful on most models.  So, why not add it to `Model`.  Along the same line of thinking, why not add `last_updated` while we're at it.  Should be generally useful for filtering/search and troubleshooting.

There are some concerns within the django community regarding *auto_now* and *auto_now_add* but I suggest we make this change independently of that.  We should be revisited and fix on all models as needed.

### DIFF
--- a/pulpcore/pulpcore/app/models/base.py
+++ b/pulpcore/pulpcore/app/models/base.py
@@ -11,7 +11,9 @@ class Model(models.Model):
     behavior.
 
     Fields:
-        id: UUID ID Primary Key Field
+        id (models.UUIDField): UUID ID Primary Key Field
+        created (models.DateTimeField): Created timestamp UTC.
+        last_updated (models.DateTimeField): Last updated timestamp UTC.
 
     References:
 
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/models/base.py
```diff
@@ -11,7 +11,9 @@ class Model(models.Model):
     behavior.
 
     Fields:
-        id: UUID ID Primary Key Field
+        id (models.UUIDField): UUID ID Primary Key Field
+        create (models.DateTimeField): Created timestamp UTC.
```
dparalen: s/create/created/

pulpcore/pulpcore/app/models/base.py
```diff
@@ -11,7 +11,9 @@ class Model(models.Model):
     behavior.
 
     Fields:
-        id: UUID ID Primary Key Field
+        id (models.UUIDField): UUID ID Primary Key Field
```
dparalen: Nit: unrelated but OK

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -271,7 +266,7 @@ def content(self):
             >>> repository_version = ...
             >>>
             >>> for content in repository_version.content:
-            >>>     content = content.case()  # optional downcast.
+            >>>     content = content.cast()  # optional downcast.
```
dparalen: Nit: unrelated, but OK

pulpcore/pulpcore/app/models/base.py
```diff
@@ -20,10 +22,9 @@ class Model(models.Model):
         * https://www.postgresql.org/docs/current/static/datatype-uuid.html
 
     """
-    # ...we have zero interest in using a mongo-specific datatype (ObjectId) as
```
dparalen: nit: unrelated but OK

PR #3327, opened 2018-02-14
Refer to RabbitMQ as the default broker

dparalen:
Since Pulp still won't run with vanilla qpid broker[1] it seems reasonable
to refer to the RabbitMQ as the default broker instead.

Fixes #3366
https://pulp.plan.io/issues/3366

[1] https://github.com/celery/kombu/pull/810


dkliban:
ok test

dkliban:
ok test

dparalen:
Hey Brian!

Thanks for the review! I've just updated this PR...

--
milan

PR #3340, opened 2018-02-26
Updating distribution validation

daviddavis:
fixes #3394
https://pulp.plan.io/issues/3394


pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -240,7 +241,8 @@ class DistributionSerializer(ModelSerializer):
         help_text=_('Publications created by this repository and publisher are automatically'
                     'served as defined by this distribution'),
         queryset=models.Repository.objects.all(),
-        view_name='repositories-detail'
+        view_name='repositories-detail',
+        allow_null=True
```
dralley: What is the use case for a distribution which does not have an associated publication?

PR #3341, opened 2018-02-26
Point users to pulp_file instead of pulp_example

daviddavis:
fixes #3364
https://pulp.plan.io/issues/3364


daviddavis:
@ammaritiz done.

PR #3343, opened 2018-02-26
Namespace DRF's ValidationError under serializers

daviddavis:
For more info see https://git.io/vAKBA

### DIFF
--- a/pulpcore/pulpcore/app/serializers/fields.py
+++ b/pulpcore/pulpcore/app/serializers/fields.py
@@ -3,7 +3,6 @@ import os
 
 from django.conf import settings
 from rest_framework import serializers
-from rest_framework.exceptions import ValidationError
 from rest_framework.reverse import reverse
 from rest_framework_nested.relations import NestedHyperlinkedRelatedField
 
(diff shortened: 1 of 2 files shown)


daviddavis:
@ammaritiz good catch. updated.

ammaritiz:
Please make the changes in this [file](https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/serializers/fields.py) too.

daviddavis:
Done.

ammaritiz:
Please update the docstring [here.](https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/serializers/fields.py#L54) Otherwise, LGTM 👍

daviddavis:
I think the docstring is correct: [ValidationError is defined in the exceptions module](https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L138).

ammaritiz:
Agreed. But we need to be consistent. You have updated the docstring in the [base module.](https://github.com/pulp/pulp/pull/3343/files#diff-ae7f375b1b9eaee42aecdf250679cd6eR58)

daviddavis:
Ah good catch. Do you think the docstring should point to `serializers` or `exceptions`? I've updated the PR to use the latter but let me know if you disagree.

PR #3345, opened 2018-02-27
Add section about DNF to consumer docs

daviddavis:
refs #1875
https://pulp.plan.io/issues/1875

### DIFF
--- a/docs/user-guide/consumer-client/introduction.rst
+++ b/docs/user-guide/consumer-client/introduction.rst
@@ -44,3 +44,12 @@ Troubleshooting with verbose flag
 You can run pulp-consumer commands in verbose mode to get additional information
 in case of a failure. Look at the example usage of verbose flag in the
 :ref:`admin client troubleshooting section <client-verbose-flag>`.
+
+
+DNF Support
+-----------
+
+It should be noted that as of Pulp 3, consumers will no longer be supported in
+Pulp. As such, we have no plans to add support for DNF in Pulp 2. For more
+information, see our `blog post on this
+<https://pulpproject.org/2018/03/01/deprecating-consumers/>`_.


bmbouter:
Does it make any sense to make this a 2.16 release note?

docs/user-guide/consumer-client/introduction.rst
```diff
@@ -44,3 +44,12 @@ Troubleshooting with verbose flag
 You can run pulp-consumer commands in verbose mode to get additional information
 in case of a failure. Look at the example usage of verbose flag in the
 :ref:`admin client troubleshooting section <client-verbose-flag>`.
+
+
+DNF Support
+-----------
+
+It should be noted that as of Pulp 3, consumers will no longer be supported in
+Pulp. As such, we have no plans to add support for DNF. For more information,
```
goosemania: Does it make sense, just for clarity, to add `in Pulp 2.`?

`As such, we have no plans to add support for DNF in Pulp 2.`

When I quickly read this note, I thought it's only about Pulp 3, but maybe it's just me :)

goosemania:
@bmbouter , I'm not sure what exactly to add to the release notes. We've never supported DNF and now we just decided that we won't add it. This decision is also not strictly attached to the 2.16 release, imo.

daviddavis:
I updated the docs. I neglected to add a release note because I agree with @goosemania but it's not a strong opinion I have.

bmbouter:
+1 to not add a release note. It's not specific to this release. Thank you for pointing that out. Sorry for the delay in my comment.

goosemania:
@daviddavis , I'm not sure `refs` in the commit message is handled correctly by redmine. Maybe `closes` is better?

daviddavis:
I've updated the PR but I am not sure why this is failing.

bmbouter:
ok test

bmbouter:
I'm also not sure why the one test is failing, but it could be a timing problem with the test. I pasted the failure below. I'm rerunning to see if another run makes it green.

```
======================================================================
FAIL: test_increments_last_run (unit.server.db.model.test_dispatch.TestScheduleEntryNextInstance)
----------------------------------------------------------------------
  File "/home/jenkins/.virtualenvs/test/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/jenkins/workspace/unittest-pulp-pr/node-type/f26-np/pulp/server/test/unit/server/db/model/test_dispatch.py", line 906, in test_increments_last_run
    self.assertTrue(now - next_entry.last_run_at < timedelta(seconds=1))
AssertionError: False is not true
```

daviddavis:
Yea, I saw the timing error. I was thinking that maybe the test ran slowly for whatever reason and maybe we should bump up the duration maybe?

bmbouter:
Yeah bumping up the timing to reduce the false positive failure rate would be good. I'm not sure what @Ichimonji10 would suggest.

daviddavis:
I'm tempted to see if it happens again before trying to remedy it.

bmbouter:
That sounds good.

daviddavis:
@bmbouter @goosemania thanks for the reviews

PR #3352, opened 2018-03-02
Fix referencing attribute that does not exist

dralley:
Task models do not have a ```self.request``` attribute

WIP, trying to reproduce anpther issue I had.

### DIFF
--- a/pulpcore/pulpcore/app/models/task.py
+++ b/pulpcore/pulpcore/app/models/task.py
@@ -246,7 +246,7 @@ class Task(Model):
         """
         if self.state != self.WAITING:
             msg = _('Task __call__() occurred but Task %s is not at WAITING')
-            _logger.warning(msg % self.request.id)
+            _logger.warning(msg % self.id)
         self.state = Task.RUNNING
         self.started_at = timezone.now()
         self.save()
@@ -269,7 +269,7 @@ class Task(Model):
             self.state = Task.COMPLETED
         else:
             msg = _('Task set_completed() occurred but Task %s is already in final state')
-            _logger.warning(msg % self.pk)
+            _logger.warning(msg % self.id)
 
         self.save()
 


pulpcore/pulpcore/app/models/task.py
```diff
@@ -213,7 +213,7 @@ class Task(Model):
     )
 
     group = models.UUIDField(null=True)
-    state = models.TextField(choices=STATES)
+    state = models.TextField(choices=STATES, null=False)
```
dralley: Should we default to ```waiting``` state like Pulp 2?

daviddavis: According to the [django documentation](https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.Field.null) `null=False` is the default.

daviddavis: I'd probably lean towards having the code explicitly set the state rather than having a default. A default could disguise a bug.

dralley: Ok, that is good.  There's still a weird bug somewhere then, because when I ran a task it gave me the warning on line 247 that the __call__() had occurred but that the ```self.state``` was not "waiting".  I thought maybe it was null (and should have checked before destroying my env).

But I will try to figure out what is going on.

dralley:
Reproduced that issue from earlier.

```
Mar 02 16:09:09 pulp3.dev celery[9259]: [2018-03-02 16:09:09,925: WARNING/ForkPoolWorker-1] Task __call__() occurred but Task 79e234d6-29e2-47c4-9b86-e4f167c345b7 is not at WAITING                               
Mar 02 16:09:11 pulp3.dev celery[9259]: [2018-03-02 16:09:11,187: WARNING/ForkPoolWorker-1] Task __call__() occurred but Task e664110d-ee07-4235-b412-25dd359f5739 is not at WAITING
```

daviddavis:
Interesting. Any idea on what the state was?

dralley:
At the time I checked it, it was ```completed```, but it may have been something different at the time.

I added a line to log that in-place and I'll try to repro again.

dralley:
I haven't seen it again, strangely.

dralley:
ok test

PR #3360, opened 2018-03-06
Use server.yaml files to customize Travis env

daviddavis:


.travis/server.postgres.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+# `SECRET_KEY`: A secret key for a particular Django installation. This is used to provide
+# cryptographic signing, and should be set to a unique, unpredictable value.
+# Pulp does not provide a default secret key. This must be user provided in order for the
+# Pulp Django Application to run.
+# SECURITY WARNING: keep the secret key used in production secret!
+#
+# SECRET_KEY:
+
+SECRET_KEY: '%25(i14$828q74)@k3d7=rn2hjg1jd%^zwa92qc6nxsto2x@(#'
```
daviddavis: This seemed like the quick and easy way to deal with SECRET_KEY and I'm not sure if it matters if we hardcode it in Travis?

PR #3361, opened 2018-03-06
Remove Ansible docs from installer

bmbouter:
The Ansible installer docs are being temporarily removed so that a new,
unified Ansible installer can be re-introduced later.

This commit also makes the install docs slightly for cross platform.

https://pulp.plan.io/issues/3437
closes #3437

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -8,56 +8,10 @@ Installation Instructions
 
     $ sudo setenforce 0
 
-Ansible Galaxy Installation
----------------------------
-
-1. In order to be able to run playbook, ensure python3 and ansible (2.3.2+) are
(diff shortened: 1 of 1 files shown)


bmbouter:
I pretty much agree since pip can take source installs and we could provide maybe a `note` section that also has some of the source install commands formed. I'm +1 to that change. I can easily do that, but maybe it needs to go on the issue as a comment?

asmacdo:
Good plan, https://pulp.plan.io/issues/3437#note-3

bmbouter:
@Ichimonji10 ty for the review. I pushed another change which moves them up closer to the `pip3` command step and also fixes a formatting error. I have not compiled the docs locally to make sure they look good.

PR #3367, opened 2018-03-11
Make compatible with Django 2.0

dralley:
Fix a few issues which prevented Pulp from working with Django 2.0

* Changed requirement string
* Added ```on_delete``` behavior on all ForeignKey field types, required by Django 2 (needs to be double checked)
* Removed reference to ```six```

closes #3453
https://pulp.plan.io/issues/3453

### DIFF
--- a/pulpcore/pulpcore/app/models/repository.py
+++ b/pulpcore/pulpcore/app/models/repository.py
@@ -205,9 +205,11 @@ class RepositoryContent(Model):
     content = models.ForeignKey('Content', on_delete=models.CASCADE,
                                 related_name='version_memberships')
     repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
-    version_added = models.ForeignKey('RepositoryVersion', related_name='added_memberships')
+    version_added = models.ForeignKey('RepositoryVersion', related_name='added_memberships',
+                                      on_delete=models.CASCADE)
     version_removed = models.ForeignKey('RepositoryVersion', null=True,
-                                        related_name='removed_memberships')
+                                        related_name='removed_memberships',
+                                        on_delete=models.CASCADE)
 
     class Meta:
         unique_together = (('repository', 'content', 'version_added'),
@@ -243,7 +245,7 @@ class RepositoryVersion(Model):
 
         repository (models.ForeignKey): The associated repository.
     """
-    repository = models.ForeignKey(Repository)
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
     number = models.PositiveIntegerField(db_index=True)
     complete = models.BooleanField(db_index=True, default=False)
 
--- a/pulpcore/pulpcore/app/models/task.py
+++ b/pulpcore/pulpcore/app/models/task.py
@@ -36,7 +36,7 @@ class ReservedResource(Model):
     resource = models.TextField(unique=True, blank=False)
 
     tasks = models.ManyToManyField("Task", related_name="reserved_resources")
-    worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
+    worker = models.ForeignKey("Worker", related_name="reservations", on_delete=models.CASCADE)
 
 
 class WorkerManager(models.Manager):
@@ -260,8 +260,10 @@ class Task(Model):
     non_fatal_errors = JSONField(default=list)
     error = JSONField(null=True)
 
-    parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks")
-    worker = models.ForeignKey("Worker", null=True, related_name="tasks")
+    parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks",
+                               on_delete=models.SET_NULL)
+    worker = models.ForeignKey("Worker", null=True, related_name="tasks",
+                               on_delete=models.SET_NULL)
 
     @staticmethod
     def current():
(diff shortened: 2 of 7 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -205,9 +205,11 @@ class RepositoryContent(Model):
     content = models.ForeignKey('Content', on_delete=models.CASCADE,
                                 related_name='version_memberships')
     repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
-    version_added = models.ForeignKey('RepositoryVersion', related_name='added_memberships')
+    version_added = models.ForeignKey('RepositoryVersion', related_name='added_memberships',
+                                      on_delete=models.CASCADE)  # TODO: correct behavior?
```
daviddavis: I think it needs to be `PROTECT` as the repo version delete code cleans up these relationships.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -205,9 +205,11 @@ class RepositoryContent(Model):
...
     version_removed = models.ForeignKey('RepositoryVersion', null=True,
-                                        related_name='removed_memberships')
+                                        related_name='removed_memberships',
+                                        on_delete=models.CASCADE)  # TODO: correct behavior?
```
daviddavis: Same here.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -243,7 +245,7 @@ class RepositoryVersion(Model):
 
         repository (models.ForeignKey): The associated repository.
     """
-    repository = models.ForeignKey(Repository)
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)  # TODO: correct behavior?
```
daviddavis: 👍

pulpcore/pulpcore/app/models/repository.py
daviddavis: Actually, thinking this through some more, I think `CASCADE` is correct in order to support the case where a repository is deleted.

pulpcore/pulpcore/app/models/task.py
```diff
@@ -221,8 +221,10 @@ class Task(Model):
     non_fatal_errors = JSONField(default=list)
     error = JSONField(null=True)
 
-    parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks")
-    worker = models.ForeignKey("Worker", null=True, related_name="tasks")
+    parent = models.ForeignKey("Task", null=True, related_name="spawned_tasks",
+                               on_delete=models.SET_NULL)  # TODO: Correct behavior?
```
daviddavis: Is parent being used anywhere? If not, maybe delete this FK?

daviddavis:
Just a thought if we're going to support django 2 and django 1: have travis test both?

pulpcore/pulpcore/app/models/task.py
dralley: AFAIK "parent" is used to keep track of tasks that were spawned by other tasks, e.g. sync with auto-publish.  The publish task would set the sync task as its parent, and the sync task would then show the publish task as a child when you GET the record.

daviddavis: Yea, I know what the field does but AFAICT, we don't have any tasks in our code yet that use it. If that's the case, I'd probably remove this FK until we actually need it and then add it back.

dralley: Nothing is setting it, but since other code is looking at it, better just to leave it there IMO.

dralley:
Rebased after the Task tag removal, tests should pass now.

dralley:
ok test

daviddavis:
Since your changes failed on 1.11, I would either change $DJANGO_MIN to 1.11.1 or `Django>=$DJANGO_MIN` to `Django>$DJANGO_MIN`.

Also, remove the TODOs and I think this PR is good to go. 👍

dralley:
I'm fine with that but it isn't strictly necessary since Pip will use the newest version that meets the constraints, which in this case is the latest z-release of 1.11

daviddavis:
So there are certain cases where pip won't like if 1.11 is already installed. Or if another dependency somehow requires 1.11.

pulpcore/pulpcore/app/models/task.py
```diff
@@ -36,7 +36,7 @@ class ReservedResource(Model):
     resource = models.TextField(unique=True, blank=False)
 
     tasks = models.ManyToManyField("Task", related_name="reserved_resources")
-    worker = models.ForeignKey("Worker", on_delete=models.CASCADE, related_name="reservations")
+    worker = models.ForeignKey("Worker", related_name="reservations", on_delete=models.CASCADE)
```
bmbouter: Doesn't the worker crash recovery rely on the presence of old `ReservedResource` records in order to know which tasks to clean up `created_resources` with `complete=False`? I recognize you are only moving this, but I'm wondering if its correct to begin with.

dralley: I don't know... "practically" speaking this is fine since we don't delete worker models any more.. but I think we should probably audit the models before the beta to make sure nothing like that is slipping through the cracks.

dralley: I appreciate Django 2's decision to make ```on_delete``` explicit for that reason

bmbouter: That sounds fine. I agree we need to look over all the `on_delete()` relationships soon.

PR #3368, opened 2018-03-12
Remove http and https options from distributions

daviddavis:
fixes #3451
https://pulp.plan.io/issues/3451


daviddavis:
Pulp smash PR: https://github.com/PulpQE/pulp-smash/pull/903/files

dralley:
This all looks good to me but I would prefer to wait for a clean smash run before approving - as soon as that gets merged we can test again.

PR #3369, opened 2018-03-12
Simplify worker 'online' state tracking

dralley:
closes #3121
https://pulp.plan.io/issues/3121

### DIFF
--- a/pulpcore/pulpcore/app/models/task.py
+++ b/pulpcore/pulpcore/app/models/task.py
@@ -69,9 +69,9 @@ class WorkerManager(models.Manager):
         """
         Returns a queryset of workers meeting the criteria to be considered 'online'
 
-        To be considered 'online', a worker should both have its 'online' field set to True
-        and have a recent heartbeat. "Recent" is defined here as "within the pulp process timeout
-        interval".
+        To be considered 'online', a worker must have a recent heartbeat timestamp and must not
+        have the 'gracefully_stopped' flag set to True. "Recent" is defined here as "within
+        the pulp process timeout interval".
 
         Returns:
             :class:`django.db.models.query.QuerySet`:  A query set of the Worker objects which
@@ -79,9 +79,25 @@ class WorkerManager(models.Manager):
         """
         now = timezone.now()
         age_threshold = now - timedelta(seconds=TASKING_CONSTANTS.PULP_PROCESS_TIMEOUT_INTERVAL)
-        return self.filter(name__startswith='reserved',
-                           last_heartbeat__gte=age_threshold,
-                           online=True)
+
+        return self.filter(last_heartbeat__gte=age_threshold, gracefully_stopped=False)
+
+    def missing_workers(self):
+        """
+        Returns a queryset of workers meeting the criteria to be considered 'missing'
+
+        To be considered missing, a worker must have a stale timestamp while also having
+        cleaned_up=False, meaning that it has not been cleaned up after an improper shutdown.
+        Stale is defined here as "beyond the pulp process timeout interval".
+
+        Returns:
+            :class:`django.db.models.query.QuerySet`:  A query set of the Worker objects which
+                are considered by Pulp to be 'missing'.
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=TASKING_CONSTANTS.PULP_PROCESS_TIMEOUT_INTERVAL)
+
+        return self.filter(last_heartbeat__lt=age_threshold, cleaned_up=False)
 
     def with_reservations(self, resources):
         """
@@ -117,16 +133,31 @@ class Worker(Model):
 
     name = models.TextField(db_index=True, unique=True)
     last_heartbeat = models.DateTimeField(auto_now=True)
-    online = models.BooleanField(default=True)
+    gracefully_stopped = models.BooleanField(default=True)
+    cleaned_up = models.BooleanField(default=False)
+
+    @property
+    def online(self):
+        """
+        Whether a worker can be considered 'online'
+
+        To be considered 'online', a worker must have a recent heartbeat timestamp and must not
+        have the 'gracefully_stopped' flag set to True. "Recent" is defined here as "within
+        the pulp process timeout interval".
+
+        Returns:
+            bool: True if the worker is considered online, otherwise False
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=TASKING_CONSTANTS.PULP_PROCESS_TIMEOUT_INTERVAL)
+
+        return not self.gracefully_stopped and self.last_heartbeat >= age_threshold
 
     def save_heartbeat(self):
-        """Save a worker heartbeat
-
+        """
         Update the last_heartbeat field to now and save it.
 
-        Warnings:
-
-            1) Only the last_heartbeat field will be saved. No other changes will be saved.
+        Only the last_heartbeat field will be saved. No other changes will be saved.
 
         Raises:
             ValueError: When the model instance has never been saved before. This method can
--- a/pulpcore/pulpcore/app/models/task.py
+++ b/pulpcore/pulpcore/app/models/task.py
@@ -192,20 +192,30 @@ class TaskLock(Model):
     Fields:
 
         name (models.TextField): The name of the item that has the lock
-        timestamp (models.DateTimeField): The time the lock was acquired
+        timestamp (models.DateTimeField): The time the lock was last updated
         lock (models.TextField): The name of the lock acquired
 
     """
-    CELERY_BEAT = 'CeleryBeat'
     RESOURCE_MANAGER = 'ResourceManager'
     LOCK_STRINGS = (
-        (CELERY_BEAT, 'Celery Beat Lock'),
-        (RESOURCE_MANAGER, 'Resource Manager Lock')
+        (RESOURCE_MANAGER, 'Resource Manager Lock'),
     )
 
     name = models.TextField(db_index=True, unique=True)
-    timestamp = models.DateTimeField(auto_now_add=True)
+    timestamp = models.DateTimeField(auto_now=True)
     lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
+
+    def update_timestamp(self):
+        """
+        Update the timestamp field to now and save it.
+
+        Only the timestamp field will be saved. No other changes will be saved.
+
+        Raises:
+            ValueError: When the model instance has never been saved before. This method can
+                only update an existing database record.
+        """
+        self.save(update_fields=['timestamp'])
 
 
 class Task(Model):
--- a/pulpcore/pulpcore/app/serializers/task.py
+++ b/pulpcore/pulpcore/app/serializers/task.py
@@ -115,10 +115,17 @@ class WorkerSerializer(ModelSerializer):
     )
 
     online = serializers.BooleanField(
-        help_text='Whether the worker is online or not. Defaults to True.',
+        help_text=_('True if the worker is considered online, otherwise False'),
+        read_only=True
+    )
+
+    gracefully_stopped = serializers.BooleanField(
+        help_text=_('True when the worker was shut down properly, False when the worker is \
+            online, or if it crashed (determined by timestamp).'),
         read_only=True
     )
 
     class Meta:
         model = models.Worker
-        fields = ModelSerializer.Meta.fields + ('name', 'last_heartbeat', 'online')
+        fields = ModelSerializer.Meta.fields + ('name', 'last_heartbeat',
+                                                'online', 'gracefully_stopped')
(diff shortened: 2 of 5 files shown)


pulpcore/pulpcore/app/models/task.py
```diff
@@ -69,9 +69,9 @@ def online_workers(self):
         """
         Returns a queryset of workers meeting the criteria to be considered 'online'
 
-        To be considered 'online', a worker should both have its 'online' field set to True
-        and have a recent heartbeat. "Recent" is defined here as "within the pulp process timeout
-        interval".
+        To be considered 'online', a worker must have a recent heartbeat timestamp and must not
+        have the 'gracefully_stopped' flag set to True. "Recent" is defined here as "within
```
bmbouter: These docs duplicate the docs in the `online()` method which makes me concerned about maintaining them. Can these be removed and point at the `online()` via a Sphinx link instead?

pulpcore/pulpcore/app/models/task.py
```diff
@@ -165,17 +182,29 @@ class TaskLock(Model):
         lock (models.TextField): The name of the lock acquired
 
     """
-    CELERY_BEAT = 'CeleryBeat'
```
bmbouter: niiiice

pulpcore/pulpcore/app/models/task.py
```diff
@@ -165,17 +182,29 @@ class TaskLock(Model):
...
     RESOURCE_MANAGER = 'ResourceManager'
     LOCK_STRINGS = (
-        (CELERY_BEAT, 'Celery Beat Lock'),
```
bmbouter: :+1:

pulpcore/pulpcore/tasking/services/worker_watcher.py
```diff
@@ -76,17 +77,20 @@ def check_celery_processes():
     now = timezone.now()
     oldest_heartbeat_time = now - timedelta(seconds=TASKING_CONSTANTS.PULP_PROCESS_TIMEOUT_INTERVAL)
 
-    for worker in Worker.objects.filter(last_heartbeat__lt=oldest_heartbeat_time, online=True):
+    for worker in Worker.objects.filter(last_heartbeat__lt=oldest_heartbeat_time,
```
bmbouter: I think this should switch to using the `workers.online_workers()` manager queryset method. It seems to do the same thing. Is that right?

pulpcore/pulpcore/tasking/services/worker_watcher.py
```diff
@@ -76,17 +77,20 @@ def check_celery_processes():
...
+                                        gracefully_stopped=False):
         msg = _("Worker '%s' has gone missing, removing from list of workers") % worker.name
         _logger.error(msg)
 
         mark_worker_offline(worker.name)
 
     worker_count = Worker.objects.exclude(
-        name__startswith=TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME).filter(online=True).count()
+        name__startswith=TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME)\
```
bmbouter: Similar for this, can this be refactored to use `Worker.objects.online_workers()`. I think the current approach would include non-gracefully stopped workers incorrectly without a change.

pulpcore/pulpcore/tasking/services/worker_watcher.py
```diff
@@ -76,17 +77,20 @@ def check_celery_processes():
...
+        .filter(gracefully_stopped=False).count()
 
     resource_manager_count = Worker.objects.filter(
-        name__startswith=TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME).filter(online=True).count()
+        name__startswith=TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME)\
```
bmbouter: Similar for this, can this be refactored to use `Worker.objects.online_workers()`. I think the current approach would include non-gracefully stopped workers incorrectly without a change.

pulpcore/pulpcore/tasking/services/worker_watcher.py
```diff
@@ -147,14 +151,15 @@ def mark_worker_offline(name, normal_shutdown=False):
         _logger.info(msg)
 
     try:
-        worker = Worker.objects.get(name=name, online=True)
+        worker = Worker.objects.get(name=name, gracefully_stopped=False)
```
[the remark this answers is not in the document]
bmbouter: Oh this is in `mark_worker_offline` so maybe the query you have is right.

bmbouter:
+1 to serializing both `gracefully_shutdown` and `online`.

bmbouter:
In terms of removing the timestamp, I think we should do that as a separate piece of work. I think the design you propose is OK I'm just hesitant to jump into it while this work is in-progress. How does that sound?

pulpcore/pulpcore/app/models/task.py
dralley: I'd prefer not to because, since the code also has to be mostly duplicated but isn't exactly the same, both will need to be refactored together in sync anyway, and fixing the docstrings would be part of that refactoring.

pulpcore/pulpcore/tasking/services/worker_watcher.py
dralley: It's actually doing precisely the opposite.  It's getting workers that have timed out but haven't had ```gracefully_stopped``` set to ```True```, e.g. crashed/missing workers.

pulpcore/pulpcore/tasking/services/worker_watcher.py
dralley: Nope, because this code only wants "workers" and doesn't care about resource managers, and ```Worker.objects.online_workers()``` returns both.

I think it would be entirely reasonable to rewrite ```Worker.objects.online_workers()``` to exclude the resource manager but then the status API couldn't use it... so it would just be shifting the problem somewhere else.

Alternatively, we could have a kwarg to toggle whether or not to exclude the resource manager from the query.

pulpcore/pulpcore/tasking/services/worker_watcher.py
dralley: Again nope, for the same reason.  It's only counting resource managers.

dralley:
I'm fine with leaving the timestamp in for now.

Something else I would like to do is have some way to know whether a resource manager is actually doing anything or if it's just HA sitting around waiting for the lock to be freed.

pulpcore/pulpcore/tasking/services/worker_watcher.py
bmbouter: Querysets are chainable so you can Worker.objects.online_workers().exclude(name__startswith=TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME)` without having to "build" those capabilities into the `online_workers` manager method itself.

pulpcore/pulpcore/app/models/task.py
bmbouter: I agree since the code is also duplicated and deduplicating that in this PR isn't the goal.

pulpcore/pulpcore/app/models/task.py
```diff
@@ -117,11 +133,27 @@ class Worker(Model):
...
+    def online(self):
+        """
+        Whether a worker can be considered 'online'
+
+        To be considered 'online', a worker must have a recent heartbeat timestamp and must not
+        have the 'gracefully_stopped' flag set to True. "Recent" is defined here as "within
+        the pulp process timeout interval".
+
+        Returns:
+            bool: Whether the worker is online or not.
```
bmbouter: Would this be clearer if it was written `True if the worker is online, False otherwise`? Just an idea.

pulpcore/pulpcore/app/models/task.py
```diff
@@ -117,11 +133,27 @@ class Worker(Model):
...
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=TASKING_CONSTANTS.PULP_PROCESS_TIMEOUT_INTERVAL)
+
+        return not self.gracefully_stopped and self.last_heartbeat > age_threshold
```
bmbouter: Probably not an important aspect, but to be consistent w/ the inequality used in the manager method I think this should this be `>=`

pulpcore/pulpcore/app/models/task.py
```diff
@@ -161,21 +194,33 @@ class TaskLock(Model):
...
     name = models.TextField(db_index=True, unique=True)
-    timestamp = models.DateTimeField(auto_now_add=True)
+    timestamp = models.DateTimeField(auto_now=True)
     lock = models.TextField(unique=True, null=False, choices=LOCK_STRINGS)
 
+    def update_timestamp(self):
+        """
+        Update the timestamp field to now and save it.
+
+        Warnings:
```
bmbouter: This looks like a Napolean style doc aspect but it's not. Can `Warnings:` not be used and instead move that text to be just another sentence on L#214?

pulpcore/pulpcore/app/serializers/task.py
```diff
@@ -115,10 +115,17 @@ class WorkerSerializer(ModelSerializer):
     )
 
     online = serializers.BooleanField(
-        help_text='Whether the worker is online or not. Defaults to True.',
+        help_text=_('Whether the worker is online'),
```
bmbouter: Similar here, I'm hoping the values True and False will be explicitly called out with its case for any bool

pulpcore/pulpcore/app/serializers/task.py
```diff
@@ -115,10 +115,17 @@ class WorkerSerializer(ModelSerializer):
...
+        read_only=True
+    )
+
+    gracefully_stopped = serializers.BooleanField(
+        help_text=_('If this worker is currently offline, this represents whether the worker \
```
bmbouter: Similar here w/ the True and False values themselves called out in the docstring help.

pulpcore/pulpcore/app/serializers/task.py
```diff
@@ -115,10 +115,17 @@ class WorkerSerializer(ModelSerializer):
...
+            was shut down properly, as opposed to e.g. crashed'),
         read_only=True
     )
 
     class Meta:
         model = models.Worker
-        fields = ModelSerializer.Meta.fields + ('name', 'last_heartbeat', 'online')
+        fields = ModelSerializer.Meta.fields + ('name', 'last_heartbeat',
```
bmbouter: What does this line do? It looks right but I want to learn more.

dralley: Here's the docs page: http://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include

It looks like it's ensuring that all of the fields specified to include on the base ModelSerializer are included on the final Model, but I'm not sure if there is a better way to accomplish this.  Maybe worth looking into.

bmbouter:
I found one issue:
1. Have workers started normally
2. Look at worker2 via the browsable API and see that online=True and gracefully_shutdown=False
3. pkill -9 worker 2 non gracefully:  `sudo pkill -9 -f _2`
4. Wait enough time to see log statements like this appear:

```
Mar 15 19:30:21 pulp3.dev celery[21727]: [2018-03-15 19:30:21,158: ERROR/MainProcess] Worker 'reserved_resource_worker_2@pulp3.dev' has gone missing, removing from list of workers
Mar 15 19:30:21 pulp3.dev celery[21727]: [2018-03-15 19:30:21,160: ERROR/MainProcess] The worker named reserved_resource_worker_2@pulp3.dev is missing. Canceling the tasks in its queue.
```

Reload the browsable API for worker 2 that was killed. Observe that online = False which is correct, but gracefully_shutdown=True which is not correct because it was not gracefully shutdown.

pulpcore/pulpcore/app/serializers/task.py
bmbouter: Thank you!

pulpcore/pulpcore/tasking/celery_app.py
```diff
@@ -232,6 +212,7 @@ def get_resource_manager_lock(name):
         worker.save_heartbeat()
 
         try:
+            # this whole block is gross and we shouldn't do it this way. TODO
```
bmbouter: I agree but this comment probably should be removed since

dralley:
Here is why: https://github.com/pulp/pulp/pull/3369/files#diff-b9365647194214f918f5930ca9f8eb93L153

I suppose ```gracefully_shutdown``` actually, semantically means something more along the lines of "this worker has been cleaned up after".  What do you think we should do about this?

dralley:
I see utility in having a value that tells you "this worker crashed" but we also do need to keep track of workers that have already been cleaned up so we're not wasting cycles.

bmbouter:
I like the user value and utility of having the data that "this worker crashed". It's useful in a variety of ways.

We could add another field called cleaned_up and never expose that to the user. Could the logic then be built on that?I would like to keep that out of the API some because I don't think we are going to use it long term with the switch to Celery4 there won't be a cleanup.

What do you think about ^?

dralley:
@bmbouter updated.  I recommend looking at the commits separately to make things easier.

bmbouter:
@dralley I manually reviewed the second commit. It all looks good. I just retested it now, and it tests exactly as I expected it to for both the graceful and non-graceful cases. To merge can you either squash it into one commit or add the redmine association keywords to the second commit? If we merge as-is Redmine won't have all the commits.

PR #3370, opened 2018-03-13
Using /var/lib/pulp/tmp instead of /var/cache/pulp

daviddavis:
fixes #3406
https://pulp.plan.io/issues/3406


ammaritiz:
An update [here](https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/tasking/celery_app.py#L129), otherwise LGTM

daviddavis:
@ammaritiz updated.

PR #3371, opened 2018-03-13
Validate distribution base paths

daviddavis:
fixes #3412
https://pulp.plan.io/issues/3412

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -1,4 +1,8 @@
 from collections import OrderedDict
+from gettext import gettext as _
+from urllib.parse import urljoin
+
+from django.core.validators import URLValidator
 
 from rest_framework import serializers
 from rest_framework.fields import SkipField
@@ -149,6 +153,32 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
         for field_name, mapping in field_mappings.items():
             field = getattr(instance, field_name)
             field.mapping.replace(mapping)
+
+    def _validate_relative_path(self, path):
+        """
+        Validate a relative path (eg from a url) to ensure it forms a valid url and does not begin
+        or end with slashes
+
+        Args:
+            path (str): A relative path to validate
+
+        Returns:
+            str: the validated path
+
+        Raises:
+            django.core.exceptions.ValidationError: if the relative path is invalid
+
+        """
+        base = "{}://{}".format(self.context['request'].scheme,
+                                self.context['request'].get_host())
+        validate = URLValidator()
+        validate(urljoin(base, path))
+
+        if path != path.strip("/"):
+            raise serializers.ValidationError(detail=_("Relative path cannot begin or end with "
+                                                       "slashes."))
+
+        return path
 
 
 class MasterModelSerializer(ModelSerializer):
--- a/pulpcore/pulpcore/app/serializers/repository.py
+++ b/pulpcore/pulpcore/app/serializers/repository.py
@@ -249,6 +249,9 @@ class DistributionSerializer(ModelSerializer):
             'name', 'base_path', 'publisher', 'publication', 'base_url', 'repository',
         )
 
+    def validate_base_path(self, path):
+        return self._validate_relative_path(path)
+
     def validate(self, data):
         if 'publisher' in data:
             publisher = data['publisher']
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ cover
 .DS_Store
 *,cover
 tags
+.pytest_cache
 
 # KDevelop
 .kdev4/


pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -249,6 +249,9 @@ class Meta:
             'name', 'base_path', 'publisher', 'publication', 'base_url', 'repository',
```
dralley: It's a bit strange that we have ```base_path``` mapped to ```base_url``` instead of using one name consistently (not part of this PR specifically)

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -249,6 +249,9 @@ class Meta:
...
         )
 
+    def validate_base_path(self, path):
+        return self._validate_relative_path(path)
+
```
dralley: Why do this separately from the ```validate()``` we use for everything else?

pulpcore/pulpcore/app/serializers/repository.py
daviddavis: I'm not sure I understand. They're two separate things: base_path is like `foo/bar` and base_url is the full url like `localhost:8000/foo/bar`.

pulpcore/pulpcore/app/serializers/repository.py
daviddavis: Yes, DRF autodetects these methods:

http://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

The validate method is not field specific. In this case, it's checking two fields: publisher and repository. It's better to use field-specific validation as it sets the error on the particular failing field in the response. Also, you're suppose to raise an exception which halts execution so adding this to validate would run the other validation on publisher and repository.

pulpcore/pulpcore/app/serializers/repository.py
dralley: I was referring to:

```
 base_url = BaseURLField(
        source='base_path', read_only=True,
        help_text=_('The URL for accessing the publication as defined by this distribution.')
)
```

But I think maybe I'm just not understanding the code semantics.

pulpcore/pulpcore/app/serializers/repository.py
dralley: Ah, I was looking at the *Django* docs, not DRF.  My bad.  And that makes sense.

Ichimonji10:
Does this PR mean that if I send an invalid base_path to Pulp with a trailing slash, that Pulp will silently change what I've send and save a base path that doesn't have a trailing slash? If so, I dislike this solution. When I send a request to Pulp, I expect that request to be interpreted as is. Magical changes being made without my knowledge are extremely frustrating.

In other words, loudly complain when my code does something wrong, so that I can fix it. Don't assume that I *really* meant for my code to do something else.

daviddavis:
The problem I ran into is that 'my-base-path/' is actually a valid base path:

```
In [1]: from django.core.validators import URLValidator
In [2]: validate = URLValidator()
In [3]: validate("http://localhost/content/my-base-path//test.iso")
(no exception)
```

I could add some validation specifically for Pulp (which I'm a bit apprehensive to do) or fix the distribution code so that it properly handles requests to `http://localhost/content/my-base-path//test.iso`. I'm a bit apprehensive about the latter as well though as I think most users will assume they need to add a slash since they are specifying a path (and paths end in slashes).

Thoughts @dralley @Ichimonji10 ?

Ichimonji10:
I think letting base paths have trailing slashes - or even requiring that they have slashes - would lead to a more usable application. Consider these code snippets:

```python
from urllib.parse import urljoin

# In a real client-side application, `distribution` comes from Pulp's API.
distribution = {'base_url': 'pulp-3.example.com:8000/content/my-base-path'}
base_url = 'http://' + distribution['base_url']
basename = '1.iso'

print(base_url + basename)
print(urljoin(base_url, basename))
```

The results are not as intended:

```
http://pulp-3.example.com:8000/content/my-base-path1.iso
http://pulp-3.example.com:8000/content/1.iso
```

On the other hand, if base paths have trailing slashes, results are sane:

```python
from urllib.parse import urljoin

# In a real client-side application, `distribution` comes from Pulp's API.
distribution = {'base_url': 'pulp-3.example.com:8000/content/my-base-path/'}
base_url = 'http://' + distribution['base_url']
basename = '1.iso'

print(base_url + basename)
print(urljoin(base_url, basename))
```

Output:

```
http://pulp-3.example.com:8000/content/my-base-path/1.iso
http://pulp-3.example.com:8000/content/my-base-path/1.iso
```

I discuss this in [Pulp #3412](https://pulp.plan.io/issues/3412)`.

daviddavis:
Relying on the precise science of [internet polling](https://etherpad.net/p/D2UjRxz01J), I'm going to update this PR to check for slashes at the beginning/end of paths.

Ichimonji10:
+0

I still think there's a usability issue here. But at least we're telling users how they should use using Pulp.

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -150,6 +154,31 @@ def _populate_generic_fields(self, instance, field_mappings):
...
 
+    def _validate_relative_path(self, path):
+        """
+        Validate a relative path (eg from a url) and strip out beginning/end slashes
+
+        Args:
+            path (str): A relative path to validate
+
+        Returns:
+            str: the validated path stripped of beginning and end slashes
```
Ichimonji10: This needs to be updated, correct?

PR #3373, opened 2018-03-13
Allow users to filter by tasks

daviddavis:
fixes #3144
https://pulp.plan.io/issues/3144

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/pulpcore/app/viewsets/custom_filters.py
@@ -1,7 +1,14 @@
 """
 This module contains custom filters that might be used by more than one ViewSet.
 """
+from gettext import gettext as _
+from urllib.parse import urlparse
+from uuid import UUID
+
+from django.urls import resolve, Resolver404
 from django_filters import filters
+
+from rest_framework import serializers
 
 
 class CharInFilter(filters.BaseInFilter, filters.CharFilter):
@@ -10,3 +17,37 @@ class CharInFilter(filters.BaseInFilter, filters.CharFilter):
     than one object in a single query.
     """
     pass
+
+
+class HyperlinkRelatedFilter(filters.Filter):
+    """
+    Enables a user to filter by a foreign key using that FK's href
+    """
+
+    def filter(self, qs, value):
+        """
+        Args:
+            qs (django.db.models.query.QuerySet): The Queryset to filter
+            value (string): href containing pk for the foreign key instance
+
+        Returns:
+            django.db.models.query.QuerySet: Queryset filtered by the foreign key pk
+        """
+
+        if not value:
+            raise serializers.ValidationError(detail=_('No value supplied for {name} filter.').
+                                              format(name=self.name))
+
+        try:
+            match = resolve(urlparse(value).path)
+        except Resolver404:
+            raise serializers.ValidationError(detail=_('URI not valid: {u}').format(u=value))
+
+        pk = match.kwargs['pk']
+        try:
+            UUID(pk, version=4)
+        except ValueError:
+            raise serializers.ValidationError(detail=_('UUID invalid: {u}').format(u=pk))
+
+        key = "{}__pk".format(self.name)
+        return qs.filter(**{key: pk})
--- a/pulpcore/pulpcore/app/viewsets/task.py
+++ b/pulpcore/pulpcore/app/viewsets/task.py
@@ -1,9 +1,10 @@
-from django_filters.rest_framework import filterset
+from django_filters.rest_framework import filters, filterset
 
 from pulpcore.app.models import Task, Worker
 from pulpcore.app.serializers import TaskSerializer, WorkerSerializer
 from pulpcore.app.viewsets import NamedModelViewSet
 from pulpcore.app.viewsets.base import GenericNamedModelViewSet
+from pulpcore.app.viewsets.custom_filters import CharInFilter, HyperlinkRelatedFilter
 from pulpcore.tasking.util import cancel as cancel_task
 
 from rest_framework.decorators import detail_route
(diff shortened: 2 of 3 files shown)


pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -3,10 +3,37 @@
...
+        """
+
+        if not value:
+            return qs
+
+        model_class = self.model._meta.get_field(self.name).rel.to
+        instance = GenericNamedModelViewSet.get_resource(value, model_class)
+
+        key = "{}__pk".format(self.name)
+        return qs.filter(**{key: instance.pk})
```
dparalen: I think getting the instance from the DB just to fetch its pk to construct a query is an overkill and might introduce performance issues; better might be to duplicate the URL resolving code from `GenericNamedModelViewSet.get_resource()`:
```python
match = resolve(urlparse(value).path)
key = "{}__pk".format(self.name)
return qs.filter(**{key: match.kwargs['pk']})
```

Even better might be to draw that code from `GenericNamedModelViewSet.get_resource()` into a new static method e.g `GenericNamedModelViewSet.get_view_by_url()` and use that instead

daviddavis:
@dparalen thanks for the suggestion. I've updated the code.

I also had to catch `Exception` instead of an actual exception like ValidationError because it looks like if the UUID is malformed, the django exception handling code hits an exception:

https://gist.github.com/daviddavis/86ac89c3189839d0719f2f9865e258ab

Seems like a bug in django.

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -10,3 +13,31 @@ class CharInFilter(filters.BaseInFilter, filters.CharFilter):
...
+            Queryset filtered by the foreign key instance
+        """
+
+        if not value:
+            return qs
+
+        try:
+            match = resolve(urlparse(value).path)
+        except Exception:
+            return qs
```
dparalen: I'm a bit worried this might, in case one (by mistake) specifies a malformed URL, not limit the returned list of objects (tasks): `GET@/v3/tasks/?worker=this_aint_url&state__in=waiting` might return more tasks than expected. Shouldn't in such a case a 400 response be generated instead? Or a 404?

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -10,3 +13,31 @@ class CharInFilter(filters.BaseInFilter, filters.CharFilter):
...
+        Args:
+            qs (django.db.models.query.QuerySet): The Queryset to filter
+            value (string): href representing an instance of foreign key
+
+        Returns:
+            Queryset filtered by the foreign key instance
+        """
+
+        if not value:
+            return qs
```
dparalen: this seems now obsoleted by line 39 but I'm wondering the same: shouldn't a 400/404 be raised in case empty value in here?

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -1,12 +1,53 @@
...
+                                              format(name=self.name))
+
+        try:
+            match = resolve(urlparse(value).path)
+        except Resolver404:
+            raise serializers.ValidationError(detail=_('URI not valid: {u}').format(u=value))
+
+        pk = match.kwargs['pk']
+        try:
+            UUID(pk, version=4)
```
daviddavis: I am manually validating UUID here otherwise it gets validated much later on when the query gets executed.

Not sure if this is the best way to validate a UUID though.

daviddavis:
@dparalen updated.

pulpcore/pulpcore/app/viewsets/custom_filters.py
dparalen: I think this is the best we can do; another PR might follow up by making this value-url-pk workflow some sort of an utility function (to reuse e.g in the `GenericNamedModelViewSet.get_resource()`) but this is cool too.

Btw I'd say this may be worth unit testing for edge conditions.

dparalen:
Just by looking this seems LGTM; I'd like to play with it a bit in the console; will do a bit later.
If anyone else can do it faster, feel free to approve!

dralley:
I played around with this locally and everything seems functional.  Could you file an issue for pulp-smash to test this?

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -1,12 +1,53 @@
...
+    """
+
+    def filter(self, qs, value):
+        """
+        Args:
+            qs (django.db.models.query.QuerySet): The Queryset to filter
+            value (string): href representing an instance of foreign key
+
+        Returns:
+            Queryset filtered by the foreign key instance
```
dralley: Paste the type down here for the return as well.

dralley: Could you paste the queryset type down here as well?

daviddavis:
@dralley updated the pulp-smash issue and docstring 👍

dralley:
@daviddavis , approved :+1: 

I like @dparalen 's suggestion of having a reusable utility function, but instead of doing that now let's file a refactor issue.  Adding unit testing can be done then.

daviddavis:
So I opened this issue to refactor: https://pulp.plan.io/issues/3471

And I outlined some edge cases in the pulp-smash test.

pulpcore/pulpcore/app/viewsets/task.py
```diff
@@ -12,10 +13,19 @@
 
 
 class TaskFilter(filterset.FilterSet):
+    state_in_list = CharInFilter(name='state', lookup_expr='in')
+    worker = HyperlinkRelatedFilter(name='worker')
```
dparalen: I checked docs generated for `GET@/api/v3/tasks/` but I can't see neither type nor description of that field. I'm not able to figure out how to get that though (`label='text'` doesn't work) so I'm cool ;)

daviddavis: I looked into this but the solutions seem painful. It would be nice to have an easy way to add descriptions for filters though and maybe something we could look into in the future.

PR #3375, opened 2018-03-13
Document worker commands in install doc

dralley:
closes #3430
https://pulp.plan.io/issues/3430

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -25,9 +25,9 @@ PyPI Installation
 
 .. note::
 
-   To install from source, replace the pip3 install commands to specify a source install such as:
+   To install from source, replace the pip3 install commands to specify a source install such as::
 
-   ``pip3 install -e "git+https://github.com/pulp/pulp.git@3.0-dev#egg=pulpcore&subdirectory=pulpcore"``
+   $ pip3 install -e "git+https://github.com/pulp/pulp.git@3.0-dev#egg=pulpcore&subdirectory=pulpcore"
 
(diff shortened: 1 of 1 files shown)


docs/installation/instructions.rst
```diff
@@ -50,6 +50,15 @@ PyPI Installation
 
 7. Go through the :ref:`database-install`, :ref:`broker-install`, and `systemd-setup` sections
 
+.. note::
+
+    In place of using the systemd unit files provided in the `systemd-setup` section, you can run
+    the commands yourself inside of a shell. This is fine for development but not recommended in production.
+
+    $ /path/to/python/bin/celery worker -A pulpcore.tasking.celery_app:celery -n resource_manager@%%h -Q resource_manager -c 1 --events --umask 18
```
daviddavis: Is there a way to escape this section?

dralley: what do you mean exactly?

daviddavis: screen shot 2018-03-14 at 2 56 40 pm

dralley: Updated, we'll see if this fixes it.

dralley:
I don't think github renders the rst properly.  It doesn't render the notes, whereas docs.pulproject.org does.

daviddavis:
Yea, I just use Github as quick way to view the text.

I think if you replace `production.` with `production.::` and remove the backticks, you'll be good.

Also, mind throwing some `$` at the beginning of each of lines. Seems like other places are doing that.

PR #3376, opened 2018-03-13
Set a 20 second timeout for sqlite

dralley:
https://pulp.plan.io/issues/3456

We can't specify it in settings.py because it is a database-specific option.  Therefore it needs to be specified in the server.yaml file.  Unfortunately this means that server.yaml will be providing some settings even on a default installation.

### DIFF
--- a/pulpcore/pulpcore/app/settings.py
+++ b/pulpcore/pulpcore/app/settings.py
@@ -5,7 +5,7 @@ Never import this module directly, instead `from django.conf import settings`, s
 https://docs.djangoproject.com/en/1.8/topics/settings/#using-settings-in-python-code
 
 For the full list of settings and their values, see
-https://docs.djangoproject.com/en/1.8/ref/settings/
+https://docs.djangoproject.com/en/1.11/ref/settings/
 """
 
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/settings.py
```diff
@@ -161,23 +161,26 @@
...
-    # https://docs.djangoproject.com/en/1.8/ref/settings/#databases
+    # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
     'databases': {
         'default': {
             'ENGINE': 'django.db.backends.sqlite3',
             'NAME': '/var/lib/pulp/sqlite3.db',
             'USER': '',
             'CONN_MAX_AGE': 0,
+            'OPTIONS': {
+                'timeout': 20,
```
asmacdo: any reason you chose 20?

dralley: no reason other than that it was the value Django used as an example of raising the sqlite timeout here: https://docs.djangoproject.com/en/2.0/ref/databases/#database-is-locked-errors

PR #3378, opened 2018-03-14
Adds release note on celery

bmbouter:
https://pulp.plan.io/issues/3407
re #3407


bmbouter:
The dev freeze for 2.16 already happened and this won't be included so I'm closing.

PR #2987, opened 2017-04-06
Add a DownloadCatalog ViewSet

BrnoPCmaniak:
closes #2303
https://pulp.plan.io/issues/2303

PR #3380, opened 2018-03-14
Validate overlapping distribution paths

daviddavis:
fixes #2987
https://pulp.plan.io/issues/2987

### DIFF
--- a/pulpcore/pulpcore/app/serializers/repository.py
+++ b/pulpcore/pulpcore/app/serializers/repository.py
@@ -1,6 +1,7 @@
 from gettext import gettext as _
 
 from django.core import validators
+from django.db.models import Q
 
 from rest_framework import serializers
 from rest_framework.validators import UniqueValidator
@@ -203,7 +204,7 @@ class DistributionSerializer(ModelSerializer):
     )
     base_path = serializers.CharField(
         help_text=_('The base (relative) path component of the published url. Avoid paths that \
-                    overlap with other distributor base paths (e.g. "foo" and "foo/bar")'),
+                    overlap with other distribution base paths (e.g. "foo" and "foo/bar")'),
         validators=[validators.MaxLengthValidator(
             models.Distribution._meta.get_field('base_path').max_length,
             message=_('Distribution base_path length must be less than {} characters').format(
@@ -244,8 +245,31 @@ class DistributionSerializer(ModelSerializer):
             'name', 'base_path', 'publisher', 'publication', 'base_url', 'repository',
         )
 
+    def _validate_path_overlap(self, path):
+        # look for any base paths nested in path
+        search = path.split("/")[0]
+        q = Q(base_path=search)
+        for subdir in path.split("/")[1:]:
+            search = "/".join((search, subdir))
+            q |= Q(base_path=search)
+
+        # look for any base paths that nest path
+        q |= Q(base_path__startswith='{}/'.format(path))
+        qs = models.Distribution.objects.filter(q)
+
+        if self.instance is not None:
+            qs = qs.exclude(pk=self.instance.pk)
+
+        match = qs.first()
+        if match:
+            raise serializers.ValidationError(detail=_("Overlaps with existing distribution '"
+                                                       "{}'").format(match.name))
+
+        return path
+
     def validate_base_path(self, path):
-        return self._validate_relative_path(path)
+        self._validate_relative_path(path)
+        return self._validate_path_overlap(path)
 
     def validate(self, data):
         if 'publisher' in data:
--- a/pulpcore/tests/serializers/test_repository.py
+++ b/pulpcore/tests/serializers/test_repository.py
@@ -1,0 +1,59 @@
+from unittest import TestCase
+
+from pulpcore.app.models import Distribution
+from pulpcore.app.serializers import DistributionSerializer
+
+
+class TestDistributionPath(TestCase):
+    def test_overlap(self):
+        Distribution.objects.create(base_path="foo/bar", name="foobar")
+        overlap_errors = {'base_path': ["Overlaps with existing distribution 'foobar'"]}
+
+        # test that the new distribution cannot be nested in an existing path
+        data = {"name": "foobarbaz", "base_path": "foo/bar/baz"}
+        serializer = DistributionSerializer(data=data)
+        self.assertFalse(serializer.is_valid())
+        self.assertDictEqual(overlap_errors, serializer.errors)
+
+        # test that the new distribution cannot nest an existing path
+        data = {"name": "foo", "base_path": "foo"}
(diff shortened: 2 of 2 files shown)


daviddavis:
Note that there's a separate issue for preventing overlap in the DB layer: https://pulp.plan.io/issues/3051

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -249,6 +250,28 @@ class Meta:
...
+        q |= Q(base_path__startswith='{}/'.format(path))
+        qs = models.Distribution.objects.filter(q)
+
+        if self.instance is not None:
+            qs = qs.exclude(pk=self.instance.pk)
+
+        match = qs.first()
+        if match:
+            raise serializers.ValidationError(detail=_("overlaps with existing distributor "
+                                                       "{}").format(match.name))
```
dralley: Please capitalize "overlaps" and put single quotes around the name of the distributor to distinquish it e.g.

```Overlaps with existing distrubutor 'bar'```

**Comment by dralley** [comment]
I tested this from both directions (creating ```foo``` and then ```foo/test```, and also ```foo/test``` and then ```foo```) and it was functional, however I don't feel qualified to review the logic itself.

Also, this should get a Pulp smash test.  I think it should piggyback on the previous test issue for validating the distribution paths instead of making a separate issue for this.

**Comment by bmbouter** [comment]
Nice work on having it go to the db once. This is a simple implementation. :+1: 

The only question I have is about a race condition. Do you think write-write race conditions of two distributions be validated concurrently that conflict with each other is a problem?

If ^ is an issue what should/could be done about it in this PR?

One possible solution would be to use a transactional level lock on the entire http call for this one specific endpoint and HTTP method type? I'm not exactly sure how to tell Django to do that, but if it could be done I think that would resolve it because as soon as one writes the other's transaction will fail.

What do you think?

**Comment by daviddavis** [comment]
Yea, the race condition still exists with this PR. It wasn't something I was trying to solve. I think we might be able to solve it easily with transactions. I can look into it.

So the one problem is that there isn't one specific endpoint. You can set base_path through distributor create AND update. Maybe we can set a transaction on both though?

**Comment by bmbouter** [comment]
Doing it on both would be a good solution. Could this as simple as adding [`@transaction.atomic`](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#controlling-transactions-explicitly) to decorate those two view calls?

I think since we know it has a race condition we need to take some sort of action pre-merge since we may defer the DB solution out to maybe never.

**Comment by asmacdo** [comment]
This seems like a good candidate for unit tests.

**Comment by jortel** [comment]
I do not think using transactions will address the race condition problem.  DB transactions provide atomic update of multiple tables and are optimized to minimize table/row read locks.

**Comment by daviddavis** [comment]
@jortel is correct. DB transactions don't lock. My testing confirms this so https://pulp.plan.io/issues/3051 will need to be addressed separately.

**Comment by bmbouter** [comment]
Doesn't the "isolation" property of ACID provide what we need here? While concurrent writes can occur around existing transactions if they touch the same data they are serialized. Maybe if I understood how it was being tested I would be able to see how I'm misunderstanding things.

**Comment by daviddavis** [comment]
For testing, I defined `@transaction.atomic` on the `create()` action in the distribution viewset and then added a sleep statement:

https://gist.github.com/daviddavis/7bdea3cc6022f50a4e1dff50e3f395af

When I submitted two distributions simultaneously, they both created records:

```
$ http POST http://localhost:8000/api/v3/distributions/ name=$RANDOM base_path='foo'
$ http POST http://localhost:8000/api/v3/distributions/ name=$RANDOM base_path='foo/bar'
```

**Comment by daviddavis** [comment]
It looks like django does provide a [`select_for_update()`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#select-for-update) function that will lock until the end of a db transaction but I don't see anything for inserting.

**Comment by daviddavis** [comment]
@bmbouter and I did some research into transaction isolation and the level we need is 'serializable' but postgresql and django use 'read commit' by default. Changing the transaction level looks quite challenging and can't really be done on a per-transaction basis so I think we agreed to handle the db level later (in issue 3051).

I'll work on adding some unit tests.

**Comment by daviddavis** [comment]
Added unit tests.

File: pulpcore/tests/serializers/test_repository.py [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+from pulpcore.app.models import Distribution
+from pulpcore.app.serializers import DistributionSerializer
+
+
+class TestDistributinPath(TestCase):
+    def test_overlap(self):
+        Distribution.objects.create(base_path="foo/bar", name="foobar")
+        overlap_errors = {'base_path': ["Overlaps with existing distribution 'foobar'"]}
+
+        # test that the new distribution is nested in an existing
```
asmacdo: s/is/cannot be/

pulpcore/tests/serializers/test_repository.py
```diff
@@ -0,0 +1,31 @@
...
+        data = {"name": "foobarbaz", "base_path": "foo/bar/baz"}
+        serializer = DistributionSerializer(data=data)
+        self.assertFalse(serializer.is_valid())
+        self.assertDictEqual(overlap_errors, serializer.errors)
+
+        # test that the new distribution nests an existing
```
asmacdo: s/nests/cannot nest/

pulpcore/tests/serializers/test_repository.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertDictEqual(overlap_errors, serializer.errors)
+
+    def test_no_overlap(self):
+        Distribution.objects.create(base_path="fu/bar", name="fubar")
+        Distribution.objects.create(base_path="fu", name="fu")
+
+        data = {"name": "fufu", "base_path": "fubar"}
+        serializer = DistributionSerializer(data=data)
+        self.assertTrue(serializer.is_valid())
+        self.assertDictEqual({}, serializer.errors)
```
asmacdo: There are a few other cases I'd like to see here

1) Identical base path "foo/bar"
2) Trailing slash difference "foo/bar/"
3) Common base, different path "foo/legal"

daviddavis: I think this is outside the scope of this PR. This PR only adds validation for overlapping paths.

asmacdo: Feel free to merge without, bug IMO those would count as "overlapping"

daviddavis: So 1 would be a violation of our uniq constraint and I'm not sure it would hit this overlapping validation. The same thing goes for two only that we validate slashes. 

Test 3 is actually an overlapping case. Will add a unit test for it.

PR #3382, opened 2018-03-15
Fixing problems with bad hostnames

daviddavis:
Looks like django requires certain features of hostnames like a "." so
using `get_host()` was not a good idea. Using a known valid hostname
instead.

fixes #3476
https://pulp.plan.io/issues/3476

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -169,8 +169,8 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
             django.core.exceptions.ValidationError: if the relative path is invalid
 
         """
-        base = "{}://{}".format(self.context['request'].scheme,
-                                self.context['request'].get_host())
+        # in order to use django's URLValidator we need to construct a full url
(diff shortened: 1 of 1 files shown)


daviddavis:
Django permits `localhost` here:

https://github.com/django/django/blob/master/django/core/validators.py#L93

dralley:
So, I tested this and it looks good.  The code block confuses me a little bit though.  Is the distinction between "base path" and "relative path" that base path starts from the content endpoint and relative path starts from the hostname?

If the variable ```path``` is the relative path, then how does ```path.strip()``` properly validate the base path?  And if it's the base path, then how is the full and proper URL being validated?

daviddavis:
> So, I tested this and it looks good. The code block confuses me a little bit though. Is the distinction between "base path" and "relative path" that base path starts from the content endpoint and relative path starts from the hostname?

I didn't really have a distinction in mind. I just wanted to write a validator that could validate paths in a URL so if relative path is not the correct term, I can rename it.

> If the variable `path` is the relative path, then how does `path.strip()` properly validate the base path? And if it's the base path, then how is the full and proper URL being validated?

I made some changes around this to make it clearer. Let me know if you don't know what's going on.

PR #3385, opened 2018-03-16
Remove ability to delete a content unit used by a repo version

CodeHeeler:
Remove ability to delete a content unit used by a
repo version

You should not be able to delete a unit of content that is used by a
repo version so that ability has been removed.

fixes #3418
https://pulp.plan.io/issues/3418

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/content.py
+++ b/pulpcore/pulpcore/app/viewsets/content.py
@@ -6,7 +6,7 @@ from rest_framework import status
 from rest_framework.response import Response
 
 
-from pulpcore.app.models import Artifact, Content, ContentArtifact
+from pulpcore.app.models import Artifact, Content, ContentArtifact, RepositoryContent
 from pulpcore.app.serializers import ArtifactSerializer, ContentSerializer
 from pulpcore.app.viewsets import CreateDestroyReadNamedModelViewSet
 
@@ -60,6 +60,23 @@ class ContentViewSet(CreateDestroyReadNamedModelViewSet):
 
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def destroy(self, request, pk):
+        """
+        Remove Content unit only if it is not associated with a repository version
+        """
+        repository_content_set = RepositoryContent.objects.filter(content__pk=pk)
+
+        if repository_content_set:
+            msg = (_('The Content unit cannot be deleted because it is associated with a repository'
+                   ' version, which is immutable. Repository: ') +
+                   repository_content_set[0].repository.name)
+            data = {'detail': msg}
+            return Response(data, status=status.HTTP_409_CONFLICT)
+        else:
+            instance = self.get_object()
+            self.perform_destroy(instance)
+            return Response(status=status.HTTP_204_NO_CONTENT)
 
 
 class ArtifactViewSet(CreateDestroyReadNamedModelViewSet):


dralley:
Looks good, works great!  :+1: 

Could you move the status of the pulp.plan.io issue to POST, and leave a comment linking to the PR?

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -61,6 +61,22 @@ def create(self, request):
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def destroy(self, request, pk):
+        """
+        Remove Content unit only if it is not associated with a repository version
+        """
+        repository_content_set = RepositoryContent.objects.filter(content__pk=pk,
+                                                                  version_removed=None)
```
dparalen: I think `version_removed` might not have to be `None` and the content still thought of as being associated with a repository; see e.g the [RepositoryVersion.content()](https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/models/repository.py#L275,#L279)

dparalen:
fwiw there's the [F-expression](https://docs.djangoproject.com/en/2.0/topics/db/queries/#filters-can-reference-fields-on-the-model) serving exactly this purpose

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -61,6 +61,22 @@ def create(self, request):
...
+        if repository_content_set:
+            msg = _('The Content unit cannot be deleted because it is associated with a repository '
+                    'version, which is immutable.')
+            data = {'detail': msg}
+            return Response(data, status=status.HTTP_409_CONFLICT)
+        else:
```
dparalen: `else` might not really be needed here

dparalen: we probably should, it just needn't be under the `else` statement as there's the `return` on previous line but matter of taste I guess...

pulpcore/pulpcore/app/viewsets/content.py
CodeHeeler: Ok.  If content may be added, removed, then re-added in another version, do we wish to track all prior version_added/version_removed data?  If no, we can have version_removed changed back to None when re-adding content and this would still be fine.  If yes, we may want this stored in a different data structure or handled differently.

Either way, both the code you linked to earlier regarding relationships:  https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/models/repository.py#L275,#L279
and the code in remove_content:  https://github.com/pulp/pulp/blob/c36ab038100ebdc23d23654901dde5f7425fb242/pulpcore/pulpcore/app/models/repository.py#L412-L416
make assumptions that the content is not re-added by filtering out those with version_removed dates without checking how they relate to the current version_added date (the former) and by assuming version_removed to be None (the latter).

daviddavis: The short answer is yes—we want to track all the information about content being added/removed/re-added/etc. If a RepositoryContent record exists then the content is in some RepositoryVersion (regardless of whether version_removed is None or not).

We store the changes between RepositoryVersions instead of storing a direct relationship between RepositoryVersions and content. If you're curious about why this is, here's some info:

https://www.redhat.com/archives/pulp-dev/2017-December/msg00023.html

CodeHeeler: Thanks!

dralley:
The PEP8 issues mentioned by pep8speaks bot and travis should be resolved but otherwise I am OK with this.

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -61,6 +61,23 @@ def create(self, request):
...
+    def destroy(self, request, pk):
+        """
+        Remove Content unit only if it is not associated with a repository version
+        """
+        repository_content_set = RepositoryContent.objects.filter(content__pk=pk)
+        
+        if repository_content_set:
+            msg = _('The Content unit cannot be deleted because it is associated with a repository '
+                    'version, which is immutable. Repository: '
+                    + repository_content_set[0].repository.name)
```
daviddavis: I'm not super familiar with gettext but I think the variable needs to be outside the `_(...)`

daviddavis:
I tested this and it works great 👍

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -61,6 +61,23 @@ def create(self, request):
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def destroy(self, request, pk):
+        """
+        Remove Content unit only if it is not associated with a repository version
+        """
+        repository_content_set = RepositoryContent.objects.filter(content__pk=pk)
+
```
dparalen: I'm wondering whether it's OK for repository content, that was removed in particular version, should be prevented from being destroyed. In case it shouldn't, the filter can be expressed like this:
```python
query = models.Q(content__pk=pk) & (models.Q(version_removed=None) |
                                    models.Q(version_removed__lt=models.F('version_added')))
if RepositoryContent.objects.filter(query):
    ...
```
Ideally, an explanation why this kind of filtering isn't necessary would be provided (as a comment in the code) for further reference.

pulpcore/pulpcore/app/viewsets/content.py
dparalen: +1 for suggested change

pulpcore/pulpcore/app/viewsets/content.py
daviddavis: `version_removed` is never going to be less than `version_added`. If a content unit gets re-added to a repository, a new RepositoryContent record gets created.

pulpcore/pulpcore/app/viewsets/content.py
dralley: yes, use ```Repository: {repo}".format(repo=repository_content_set[0].repository.name)```

dralley:
Please update the issue on pulp.plan.io to be "POST" state and leave a comment with a link to this PR

PR #3141, opened 2017-09-13
Squash nest classes

asmacdo:
Add nesting functionality to base classes to simplify inheritance hierarchy.

PR #3387, opened 2018-03-17
Remove the 'is_superuser' field from user-facing API

dralley:
It's not currently being used for anything, so we shouldn't expose it.

closes #3141
https://pulp.plan.io/issues/3141

### DIFF
--- a/pulpcore/pulpcore/app/serializers/user.py
+++ b/pulpcore/pulpcore/app/serializers/user.py
@@ -44,12 +44,6 @@ class UserSerializer(ModelSerializer):
                     ],
     )
 
-    is_superuser = serializers.BooleanField(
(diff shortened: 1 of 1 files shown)


dparalen:
Just checked this works indeed; [un-patched](https://paste.fedoraproject.org/paste/LIO4NKDzRvsj2InWewx8zg) getting the admin user and [patched](https://paste.fedoraproject.org/paste/sl-yEEImVkCmtKWqQRPd9Q); patched creating an user with `is_superuser=true`, getting it back un-patched, then creating un-patched with the same and getting it back patched again: [edgy](https://paste.fedoraproject.org/paste/LoIXlynokLe94EVEwGcGFQ)

PR #3388, opened 2018-03-20
Set batch size to 800 to resolve sqlite issues

dralley:
closes #3499
https://pulp.plan.io/issues/3499

### DIFF
--- a/plugin/pulpcore/plugin/changeset/iterator.py
+++ b/plugin/pulpcore/plugin/changeset/iterator.py
@@ -93,7 +93,8 @@ class ContentIterator(Iterable):
             dict: A dictionary of {model_class: [content,]}
                 Each content is: PendingContent.
         """
-        for batch in BatchIterator(self.content, 1024):
+        # should be < 1000 as higher values may hit the sqlite expression tree size limit
+        for batch in BatchIterator(self.content, 800):
             collated = {}
             for content in batch:
                 _list = collated.setdefault(type(content.model), list())


plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -32,7 +32,7 @@ class BatchIterator(Iterable):
     """
 
     # Default batch size.
-    BATCH = 1000
+    BATCH = 800
```
jortel: The BatchIterator is a general purpose object for which the default of 1024 seems reasonable.  The usage specific (more appropriate) size should be passed in.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -93,7 +93,7 @@ def _collated_content(self):
             dict: A dictionary of {model_class: [content,]}
                 Each content is: PendingContent.
         """
-        for batch in BatchIterator(self.content, 1024):
```
jortel: Let's make this 800 here.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -93,7 +93,8 @@ def _collated_content(self):
...
+        # higher values may hit the sqlite expression tree size limit
```
jortel: Suggest: `should be < 1000 due to sqlite limitation. ` That way if someone changes the value later the comment is still accurate and relevant.

dralley: +1

PR #3391, opened 2018-03-21
Clean up file on Artifact deletion

daviddavis:
fixes #3508
https://pulp.plan.io/issues/3508

### DIFF
--- a/pulpcore/pulpcore/app/models/content.py
+++ b/pulpcore/pulpcore/app/models/content.py
@@ -91,6 +91,17 @@ class Artifact(Model):
             self.file.close()
             raise
 
+    def delete(self, *args, **kwargs):
+        """
+        Deletes Artifact model and the file associated with the Artifact
+
+        Args:
+            args (list): list of positional arguments for Model.delete()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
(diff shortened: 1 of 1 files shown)


pulpcore/pulpcore/app/models/content.py
```diff
@@ -91,6 +91,17 @@ def save(self, *args, **kwargs):
...
+    def delete(self, *args, **kwargs):
+        """
+        Deletes Artifact model and the file associated with the Artifact
+
+        Args:
+            args (list): list of positional arguments for Model.delete()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
+        """
+        super().delete(*args, **kwargs)
+        self.file.delete(save=False)
```
dralley: Why ```save=False```?  Because the model has already been deleted?

Do you think maybe these lines should go in opposite order?

daviddavis: With `save=False`, the model was being saved as a new record in the database. I'm assuming that we want to delete the file after we delete the model in case the model delete fails which would leave us with a record that has no file.

dralley: That makes sense.  Approved!

PR #3393, opened 2018-03-23
Remove AUTHORS

alexxa:
Please consider removing AUTHORS file.

- One can find all contributors in [Contributors](https://github.com/pulp/pulp/graphs/contributors).
- There are Git commands to grep a list of all contributors, e.g. `git shortlog -sne`. It displays a sum of all commits+names+emails, the same as AUTHORS currently. 

^^ Therefore, reasoning "That way you can prove to all your friends that you contributed to Pulp!" doesn't make much sense. 

- Hardcoding AUTHORS requires an extra effort and maintenance. Currently, AUTHORS file is outdated. Not all contributors are listed there. 
- Some people don't like to share their personal and working emails that way and somebody can use it. Recently, I have received several emails from HRs saying they found my name and contact details in that file.


dralley:
@alexxa 

I don't expect we will remove the entire file, but if you update this PR to remove your name, we would gladly do that.

bmbouter:
If you can see any ways the AUTHORS can be improved (without deleting it) that would be good. Maybe auditing that the git log and the AUTHORS file are in sync? Just an idea I wanted to share.

alexxa:
Hello,

@dralley,

True, there are several entries per one name. True, it can be because one uses several email addresses or it is misconfigured . However,

- In our world, names are not unique. There can be at least two persons sharing the same name and contributing to Pulp. But those two will never have the same email address.

- People can change their emails and even names. For example, one leaves RH. Do you suggest to update the AUTHORS file every time it happens? I doubt one did/will do it.

Therefore `git shortlog -sne` showing contributors by email entry and not squashing them by name is completely logical.

@bmbouter, Still, I don't consider it's reasonable to keep the AUTHORS file, especially by adding some scripts, which will be triggered by every new PR.

I know Katello guys had/have it, since their [CONTRIBUTORS](https://github.com/Katello/katello/blob/master/CONTRIBUTORS) file has 'duplicate' entries [1]. But if one looks at the date it was updated the last time (July'15) and names in commits history, it seems they don't use it anymore.

BTW, not every repo in Pulp GItOrg has AUTHORS file. And those, which do, are outdated too.

Guys, ... AUTHORS file is kind of old-fashion thing, when Git didn't have means to list contributors.  It was essential when the project was tend to be migrated from other versioning system... and if there are any contributors sending their patches by e-mail. If the project is intended to be kept in git and all contributors will use git to contribute, it's just redundant relict of the past era of early open source communities.

Let me know if it's still not convincing, I'm cool to update this PR and remove my name only.

--alexxa

daviddavis:
👍 from me.

daviddavis:
@alexxa would you mind starting a discussion on our mailing list? 

https://www.redhat.com/mailman/listinfo/pulp-dev

Just want to discuss this with the entire project. If you cannot, I'll do it. Thanks!

alexxa:
Thanks for the support, @daviddavis. Please feel free to move it to pulp-dev list. I wrote all my reasoning here, I have nth to add.

And please pass my Hi to Pulp Team, some old-timers may remember me as igulina, QE, which was often confused with @ipanova  =)

bmbouter:
@alexxa Can the docs be updated for this change? Like references about AUTHORS on [this page](https://docs.pulpproject.org/dev-guide/contributing/index.html#contribution-checklist). Also can the AUTHORS file be kept but have it's contents replaced with some simple text recommending the user to use `git shortlog -sne` command.

What do you think about ^ ideas?
Also note that the source tree of Pulp3 is fully branched and will never merge again so if possible could the exact same thing be also done in a separate PR against pulp/pulp:3.0-dev?

Thank you for this idea, communication, and code!

dralley:
I would rather get rid of the AUTHORS entirely and put the documentation for ```git shortlog -sne``` elsewhere

alexxa:
@bmbouter I have just opened a similar PR against 3.0-dev, #3410 . Let me know if I should do anything else, please.

bmbouter:
Thanks @alexxa I'm going to go ahead and merge.

PR #3394, opened 2018-03-23
Master/Detail tasks

asmacdo:


pulpcore/pulpcore/app/response.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, task_results, request):
         """
         tasks = []
         for result in task_results:
-            task = {"_href": reverse('tasks-detail', args=[result.task_id], request=request),
+            task = {"_href": reverse('tasks-core-updates-detail', args=[result.task_id],
+                                     request=request),
```
asmacdo: TODO: Can this be generalized? 

something like:

    task = CoreTask.objects.get(id=result.task_id).cast()
    
Except I don't think the model object knows about its serializer. Could we find some way to use a serializer to have DetailRelatedField(many=True) that would cast() to render with the detail serializer? Would replace "OperationPostponedResponse"

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -110,8 +110,8 @@ def endpoint_pieces(cls):
                 except AttributeError:
                     # no endpoint_name defined, need to get more specific in the MRO
                     continue
-
-            pieces = (master_endpoint_name, cls.endpoint_name)
+            pieces = tuple(cls.endpoint_name.split('/'))
```
asmacdo: allows plugins to namespace tasks.

v3/tasks/core-update/ becomes tasks-core-update-list and tasks-core-update-detail

pulpcore/pulpcore/app/viewsets/task.py
```diff
@@ -43,6 +45,14 @@ def cancel(self, request, pk=None):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
+class CoreUpdateTaskViewSet(TaskViewSet):
+
+    endpoint_name = 'core/updates'
```
asmacdo: includes namespace which allows plugins to use the same name.

pulpcore/pulpcore/tasking/tasks.py
```diff
@@ -176,7 +177,8 @@ class UserFacingTask(PulpTask):
     # this tells celery to not automatically log tracebacks for these exceptions
     throws = (PulpException,)
 
-    def apply_async_with_reservation(self, resources, args=None, kwargs=None, **options):
+    def apply_async_with_reservation(self, reservations, task_status=None, args=None, kwargs=None,
```
asmacdo: @bmbouter, didn't want to slip this in without you noticing

pulpcore/pulpcore/tasking/tasks.py
```diff
@@ -200,20 +202,27 @@ def apply_async_with_reservation(self, resources, args=None, kwargs=None, **opti
...
-        resources = {util.get_url(resource) for resource in resources}
-        inner_task_id = str(uuid.uuid4())
+        resources = {util.get_url(resource) for resource in reservations}
         task_name = self.name
 
         # Set the parent attribute if being dispatched inside of a Task
         parent_arg = self._get_parent_arg()
 
-        TaskStatus.objects.create(pk=inner_task_id, state=TaskStatus.WAITING, **parent_arg)
+        if task_status is None:
```
asmacdo: TODO
Task (task_status) should be created and passed in here everytime. Creation of CoreUpdateTask is temporary.

pulpcore/pulpcore/tasking/tasks.py
```diff
@@ -237,7 +246,7 @@ def apply_async(self, args=None, kwargs=None, **options):
         parent_arg = self._get_parent_arg()
 
         try:
-            TaskStatus.objects.create(pk=async_result.id, state=TaskStatus.WAITING, **parent_arg)
+            CoreUpdateTask.objects.create(pk=async_result.id, state=TaskStatus.WAITING, **parent_arg)
```
asmacdo: TODO Should this be a new Task type?

pulpcore/pulpcore/app/response.py
asmacdo: ODO: Can this be generalized? 

something like:

    task = CoreTask.objects.get(id=result.task_id).cast()
    
Except I don't think the model object knows about its serializer. Could we find some way to use a serializer to have DetailRelatedField(many=True) that would cast() to render with the detail serializer? Would replace "OperationPostponedResponse"

Maybe here? a new serializer that is used here, and in the AsyncRemoveMixin https://github.com/pulp/pulp/pull/3394/files#diff-ae7f375b1b9eaee42aecdf250679cd6eR232

dparalen: yeah, esp. in case this was an async result of some plug-in specific thing, returning this hard-coded `tasks-core-updates-detail` reverse might not be cool. 

I think part of the issue is that we change this logic to be an object-like endpoint instead a verb; posting to a serializer to create an object, we might want to avoid returning a list of them, it feels odd.

If instead there was a `WaitingTask` that had a custom, read-only serializer, a read only `parent` field  and a `RunningTask` had a read-only `children` field in addition, we might  avoid having to have a custom response handling logic (here) and make the tasks appear what they actually are, a lazily-unspooling, directed acyclic graph.

See also my off-topic comment below

pulpcore/pulpcore/app/serializers/task.py
```diff
@@ -93,14 +94,56 @@ class TaskSerializer(ModelSerializer):
...
+        task_kwargs = {}
+        for (key, value) in self.task_kwarg_structure.items():
+            task_kwargs[key] = self._str_to_nested_value(value)
+        return task_kwargs
+
+    def _str_to_nested_value(self, nested_string):
+        nested_layers = nested_string.split('.')
+        value = self.task
+        for layer in nested_layers:
+            value = getattr(value, layer)
```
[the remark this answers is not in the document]
asmacdo: this allows us to create a string structure like "importer.pk"
first loop does task.importer, second loop does task.importer.pk

pulpcore/pulpcore/app/viewsets/task.py
asmacdo: The idea is that all tasks that are deployed by pulpcore will be listed at v3/tasks/core/, but CoreTask is just abstract.
Updates to Repositories, Importers, Publishers will deploy CoreUpdateTask, listed at v3/tasks/core/updates/
Deletes of the same will deploy CoreDeleteTask, listed at v3/tasks/core/deletes/.

pulpcore/pulpcore/app/models/task.py
```diff
@@ -298,6 +299,14 @@ def release_resources(self):
             if not reservation.tasks.exists():
                 reservation.delete()
 
+class CoreTask(Task):
+
+    TYPE = "core"
```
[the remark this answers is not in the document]
asmacdo: Either Repository, Importer, or Publisher

pulpcore/pulpcore/app/response.py
asmacdo: The OperationPostponed will only ever be used for asynchronous updates/deletes. The reason for this is that these tasks are created by the ViewSets for the objects that are being updated/deleted (repos, importers, publishers).

Other tasks are created/read/listed at their own endpoints and handled by their own viewsets.

pulpcore/pulpcore/app/serializers/task.py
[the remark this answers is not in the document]
dparalen: aha, it's `getattr` thru each nested layer; probably a sample would help

pulpcore/pulpcore/app/viewsets/task.py
dparalen: Ah, I see, cool

asmacdo:
Closing for now, I'll rebase and reopen if it is helpful for discussion.

PR #3401, opened 2018-04-02
As a user, I can cleanup orphan content and artifacts

daviddavis:
fixes #3442
https://pulp.plan.io/issues/3442

### DIFF
--- a/pulpcore/pulpcore/tasking/tasks.py
+++ b/pulpcore/pulpcore/tasking/tasks.py
@@ -13,7 +13,7 @@ from celery.result import AsyncResult
 from django.conf import settings as pulp_settings
 from django.db import IntegrityError
 
-from pulpcore.app.models import Task as TaskStatus
+from pulpcore.app.models import ReservedResource, Task as TaskStatus
 from pulpcore.app.models import Worker
 from pulpcore.common import TASK_STATES
 from pulpcore.exceptions import PulpException
@@ -99,6 +99,16 @@ def _queue_reserved_task(name, inner_task_id, resources, inner_args, inner_kwarg
     """
     task_status = TaskStatus.objects.get(pk=inner_task_id)
     while True:
+        if name == "pulpcore.app.tasks.orphan.orphan_cleanup":
+            if ReservedResource.objects.exists():
+                # wait until there are no reservations
+                time.sleep(0.25)
+                continue
+            else:
+                celery.tasks[name].apply(args=inner_args, task_id=inner_task_id,
+                                         kwargs=inner_kwargs, **options)
+                return
+
         try:
             worker = _acquire_worker(resources)
         except Worker.DoesNotExist:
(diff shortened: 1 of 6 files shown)


pulpcore/pulpcore/tasking/tasks.py
```diff
@@ -99,6 +99,16 @@ def _queue_reserved_task(name, inner_task_id, resources, inner_args, inner_kwarg
     """
     task_status = TaskStatus.objects.get(pk=inner_task_id)
     while True:
+        if name == "pulpcore.app.tasks.orphan.orphan_cleanup":
+            if ReservedResource.objects.exists():
+                # wait until there are no reservations
+                time.sleep(0.25)
+                continue
+            else:
+                celery.tasks[name].apply(args=inner_args, task_id=inner_task_id,
```
bmbouter: I had imagined this orphan cleanup code would run synchronously here instead of being routed to a worker. With this design, another piece of work could be routed quickly to another worker and then it's possible the orphan cleanup code isn't running as a singleton. Can this call the tasking code synchronously somehow?

Also assuming the task code run synchronously, for cancellation purposes I believe we would have this take _queue_release_reasource use the user facing task ID so that if a cancel happens it's actually this task (in the resource manager) that will get killed.

bmbouter: Oh you are right `apply()` does run locally.

daviddavis: Updated `apply_async_with_reservation` to send the inner_task_id as task_id to `_queue_reserved_task`. That fixed task cancel.

daviddavis:
I think this is ready for review/testing again. I also added progress reporting to the task.

bmbouter:
The code review looks all correct. It's time for some hand testing!

bmbouter:
Hand testing of the latest code works great. I verified the only orphaned artifact and the only orphaned unit both were deleted. Also the progress reporting is all correct. The task shows me:

```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: Accept

{
    "_href": "http://localhost:8000/api/v3/tasks/dc6f3f7c-5610-401c-93d4-fe150b724bc2/",
    "state": "completed",
    "started_at": "2018-04-05T12:57:45.412145Z",
    "finished_at": "2018-04-05T12:57:45.552251Z",
    "non_fatal_errors": [],
    "error": null,
    "worker": null,
    "parent": null,
    "spawned_tasks": [],
    "progress_reports": [
        {
            "message": "Clean up orphan Content",
            "state": "completed",
            "total": 1,
            "done": 1,
            "suffix": "",
            "task": "http://localhost:8000/api/v3/tasks/dc6f3f7c-5610-401c-93d4-fe150b724bc2/"
        },
        {
            "message": "Clean up orphan Artifacts",
            "state": "completed",
            "total": 1,
            "done": 1,
            "suffix": "",
            "task": "http://localhost:8000/api/v3/tasks/dc6f3f7c-5610-401c-93d4-fe150b724bc2/"
        }
    ],
    "created_resources": []
}
```

bmbouter:
I just tested task cancellation and it works exactly as expected with rabbitMQ!

bmbouter:
I just tested fatal exception recording from inside `orphan_cleanup()` and it works great!

PR #3403, opened 2018-04-02
Change how workers are represented

dralley:
* Don't serialize the 'gracefully_shutdown' field
* Create a new 'missing' property and serialize it
* In the status API, list both online and missing workers

### DIFF
--- a/pulpcore/pulpcore/app/serializers/status.py
+++ b/pulpcore/pulpcore/app/serializers/status.py
@@ -48,8 +48,16 @@ class StatusSerializer(serializers.Serializer):
         many=True
     )
 
-    known_workers = WorkerSerializer(
-        help_text=_("List of celery workers known to the application"),
+    online_workers = WorkerSerializer(
+        help_text=_("List of online celery workers known to the application. An online worker is "
+                    "actively heartbeating and can respond to new work"),
+        many=True
+    )
+
+    missing_workers = WorkerSerializer(
+        help_text=_("List of missing celery workers known to the application. A missing worker is "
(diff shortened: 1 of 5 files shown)


dralley:
The only thing I'm concerned about is confusion about the intricacies of "online" and "missing".  Just because a worker isn't online doesn't mean it's missing.  If it was shutdown properly, it isn't missing, but it's still not online.

pulpcore/pulpcore/app/serializers/status.py
```diff
@@ -48,8 +48,13 @@ class StatusSerializer(serializers.Serializer):
         many=True
     )
 
-    known_workers = WorkerSerializer(
-        help_text=_("List of celery workers known to the application"),
+    online_workers = WorkerSerializer(
+        help_text=_("List of online celery workers known to the application"),
```
daviddavis: What about expanding this out a bit more and let users know what 'online' means?

pulpcore/pulpcore/app/serializers/status.py
```diff
@@ -48,8 +48,13 @@ class StatusSerializer(serializers.Serializer):
...
+        many=True
+    )
+
+    missing_workers = WorkerSerializer(
+        help_text=_("List of missing celery workers known to the application"),
```
daviddavis: Same comment here.

dralley:
This PR will need a change made to pulp-smash but I'll wait until approval on these changes before making it.

dralley:
@daviddavis updated

dralley:
Pulp Smash PR: https://github.com/PulpQE/pulp-smash/pull/921

PR #3404, opened 2018-04-02
Renames Importer to Remote

dkliban:
closes #3488
https://pulp.plan.io/issues/3488


alexxa:
@dkliban There are 5 occurrences of `an remote`.

PR #3406, opened 2018-04-03
Fix date filtering on tasks and repository versions

dralley:
closes #3536
https://pulp.plan.io/issues/3536


daviddavis:
What about these filters:

https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/viewsets/repository.py#L190-L191

dralley:
@daviddavis pffft, I forgot the original reason for the issue.  lol.

Thanks.  I'll fix it.

PR #3408, opened 2018-04-04
Simplify travis config as it relates to Django versions

dralley:


.travis/install.sh
```diff
@@ -2,7 +2,7 @@
 set -v
 
 # dev_requirements should not be needed for testing; don't install them to make sure
-pip install "Django>=$DJANGO_MIN,<$DJANGO_MAX"
+pip install "Django<=DJANGO_MAX"
```
daviddavis: Why not just use `<` here and `2.3` and `1.12` as values?

PR #3409, opened 2018-04-05
Prevent query matching all content/artifact when not criteria included.

jortel:
https://pulp.plan.io/issues/3524

An empty `Q` object passed to `Model.objects.filter()` results in matching everything.  The PK will never be NULL so initializing the `Q(pk=None)` assures that when nothing else is added to the query, it will match nothing.  For the `Artifact` query, it means that when none of the digests are added to the query, it won't match everything.  Same for `Content` to be safe.

### DIFF
--- a/plugin/pulpcore/plugin/changeset/iterator.py
+++ b/plugin/pulpcore/plugin/changeset/iterator.py
@@ -73,7 +73,7 @@ class ContentIterator(Iterable):
         Returns:
             Q: The built query.
         """
-        q = Q()
+        q = Q(pk=None)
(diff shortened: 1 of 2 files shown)


bmbouter:
I'm hitting an error and this patch would resolve it. I'm going to go ahead and merge it since it's ready to go and we are experiencing the errors without it.

daviddavis:
Sorry, I probably should have merged this when I approved it.

bmbouter:
Naw it's all good. It didn't cause me any issues.

PR #3419, opened 2018-04-06
Added back two mixins removed earlier

dralley:
Create provides more than just the ```create()``` method, too.


dralley:
@asmacdo It wasn't that, actually

https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L28

PR #3421, opened 2018-04-06
Swap drf_openapi in favor of drf-yasg

dralley:
closes: #3552
https://pulp.plan.io/issues/3552


pulpcore/pulpcore/app/urls.py
```diff
@@ -117,12 +118,29 @@ def __repr__(self):
     url(r'^api/v3/orphans/', OrphansView.as_view()),
 ]
 
-# if drf_openapi is installed add live docs route
+# if drf_yasg is installed add live docs route
 with suppress(ImportError):
```
werwty: Since you added 'drf-yasg' to the requirements in setup.py we don't need to treat it as an optional dependency anymore, and can remove the suppress and move the import statements to the top.

pulpcore/pulpcore/app/settings.py
```diff
@@ -66,7 +66,7 @@
 OPTIONAL_APPS = [
     'crispy_forms',
     'django_extensions',
-    'drf_openapi'
+    'drf_yasg'
```
werwty: This should be moved to "INSTALLED_APPS"

docs/integration_guide/rest_api/index.rst
```diff
@@ -11,14 +11,7 @@ pulpcore REST API
 -----------------
 
 Each instance of Pulp hosts dynamically generated API documentation located at `http://pulpserver/api/v3/docs/`
-if `drf_openapi` is installed. This documentation includes the installed pulp plugin endpoints and is a more
+if `drf_yasg` is installed. This documentation includes the installed pulp plugin endpoints and is a more
```
werwty: s/if `drf_yasg` is installed.//

pulpcore/doc_requirements.txt
```diff
@@ -1 +1 @@
--e git+https://github.com/limdauto/drf_openapi.git@54d24fb#egg=drf_openapi
```
werwty: I don't think we need this file anymore, since this was only separate so the user can install it separately from a pinned git commit

These docs should be removed: https://github.com/pulp/pulp/blame/3.0-dev/docs/contributing/3.0-development/rest-api.rst#L473

An associated devel change should be made to remove doc_requirements here: https://github.com/pulp/devel/blob/3.0-dev/ansible/roles/dev/tasks/main.yml#L97

There also needs to be a jjb change- I've opened a PR of it here: https://github.com/pulp/pulp-ci/pull/505

daviddavis:
What is this `docs/api.yaml` file and should it be checked into git? Asking because it seems to contain some plugin specific routes like `/remotes/file/` but not others like `/remotes/python/`.

dralley:
@daviddavis Good question.

That was due to a snafu we found where ```sphinxcontrib_openapi``` couldn't parse the ```drf_yasg``` output even though the ```drf_yasg``` output conformed perfectly to the OpenAPI schema (while it worked fine with output from ```drf_openapi```, whose output *failed* to validate against the OpenAPI schema).  Due to that, we switched from using ```sphinxcontrib_swagger``` to ```sphinxcontrib_openapi```, which works great.  Except ```sphinxcontrib_openapi``` doesn't support grabbing the schema from a live query, it needs it from a local file, and if the schema isn't present it fails hard instead of failing silently like ```sphinxcontrib_swagger``` did. 

Bihan and I hated the idea of committing the schema file, but it was the best of a bunch of crappy options.  She was planning to change the actual docs hosting job to replace the schema so that the one in-tree wouldn't need to be kept up to date in order for the hosted docs on the website to be correct.  Really, the only purpose of its presence is to keep ```sphinxcontrib_openapi``` from puking.

daviddavis:
@dalley Sounds good. Thanks for answering my question.

werwty:
@dalley I think we should commit the `docs/api.yaml` without plugin specific endpoints.

@daviddavis after thinking about it, I don't hate the idea of commiting it: it would be a good sanity check when we release a new version and look at the diff of the schema, to make sure we're not breaking semver, and pcreech can use it to generate docs for the rpm build.

dralley:
@werwty I updated the schema (manually, but it should work)

PR #3424, opened 2018-04-08
Remote auto_publish field from Publisher

dralley:
closes: #3545
https://pulp.plan.io/issues/3545

### DIFF
--- a/pulpcore/pulpcore/app/models/repository.py
+++ b/pulpcore/pulpcore/app/models/repository.py
@@ -145,8 +145,6 @@ class Publisher(MasterModel):
 
     Fields:
 
-        auto_publish (models.BooleanField): Indicates that the adaptor may publish automatically
-            when the associated repository's content has changed.
         last_published (models.DatetimeField): When the last successful publish occurred.
 
     Relations:
@@ -155,8 +153,6 @@ class Publisher(MasterModel):
     TYPE = 'publisher'
 
     name = models.TextField(db_index=True, unique=True)
-
-    auto_publish = models.BooleanField(default=True)
     last_published = models.DateTimeField(blank=True, null=True)
 
     class Meta:
@@ -169,8 +165,7 @@ class Exporter(MasterModel):
 
     Fields:
 
-        name (models.CharField): The exporter unique name.
-        last_updated (models.DatetimeField): Timestamp of the last update.
+        name (models.TextField): The exporter unique name.
         last_export (models.DatetimeField): When the last successful export occurred.
 
     Relations:
@@ -178,7 +173,7 @@ class Exporter(MasterModel):
     """
     TYPE = 'exporter'
 
-    name = models.CharField(max_length=255, db_index=True, unique=True)
+    name = models.TextField(db_index=True, unique=True)
     last_export = models.DateTimeField(blank=True, null=True)
 
     class Meta:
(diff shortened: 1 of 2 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -170,7 +166,6 @@ class Exporter(MasterModel):
     Fields:
 
         name (models.CharField): The exporter unique name.
-        last_updated (models.DatetimeField): Timestamp of the last update.
```
dralley: Documented but nonexistent

pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -165,7 +160,7 @@ class Meta:
         abstract = True
         model = models.Publisher
         fields = MasterModelSerializer.Meta.fields + (
-            'name', 'last_updated', 'auto_publish', 'last_published', 'distributions',
```
dralley: ```last_updated``` is documented but isn't on the model.

dralley: We really need to comb back through the models for mistakes like this.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -155,8 +154,6 @@ class Publisher(MasterModel):
     TYPE = 'publisher'
 
     name = models.TextField(db_index=True, unique=True)
```
dralley: Was missing docstring

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -170,7 +167,6 @@ class Exporter(MasterModel):
     Fields:
 
         name (models.CharField): The exporter unique name.
```
dralley: Why is ```Exporter.name``` a CharField while ```Publisher.name``` is a TextField?

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -170,15 +166,14 @@ class Exporter(MasterModel):
...
-        last_updated (models.DatetimeField): Timestamp of the last update.
         last_export (models.DatetimeField): When the last successful export occurred.
 
     Relations:
 
     """
     TYPE = 'exporter'
 
-    name = models.CharField(max_length=255, db_index=True, unique=True)
```
dralley: This should probably be a TextField like all the others.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -169,16 +165,15 @@ class Exporter(MasterModel):
 
     Fields:
 
-        name (models.CharField): The exporter unique name.
-        last_updated (models.DatetimeField): Timestamp of the last update.
```
dralley: Defined on the Master model, not here.

pulpcore/pulpcore/app/models/repository.py
dralley: ...or maybe **all** the names should be CharField, instead.  TextField is really meant for long blobs of text, CharField for short ones that need to be size-bounded. 

We just need to be consistent.

asmacdo: I don't think we know the max length of all strings, so that change would be pretty big. https://docs.pulpproject.org/en/3.0/nightly/contributing/3.0-development/data-modeling.html?highlight=textfield#introduction

dralley: +1.

PR #3425, opened 2018-04-09
Adds 429 Retry Support

bmbouter:
All retry support is provided by the 'backoff' library which also
provides an asyncio compatible interface.

This updates the docs and docstrings to call out the new retry
functionality.

https://pulp.plan.io/issues/3421
closes #3421

### DIFF
--- a/plugin/pulpcore/plugin/download/http.py
+++ b/plugin/pulpcore/plugin/download/http.py
@@ -1,6 +1,32 @@
+import logging
+
 import aiohttp
+import backoff
 
 from .base import attach_url_to_exception, BaseDownloader, DownloadResult
+
+
+log = logging.getLogger(__name__)
+
+
+logging.getLogger('backoff').addHandler(logging.StreamHandler())
+
+
+def giveup(exc):
+    """
+    Inspect a raised exception and determine if we should give up.
+
+    Do not give up when the status code is one of the following:
+
+        429 - Too Many Requests
+
+    Args:
+        exc (aiohttp.ClientResponseException): The exception to inspect
+
+    Returns:
+        True if the download should give up, False otherwise
+    """
+    return not exc.code == 429
 
 
 class HttpDownloader(BaseDownloader):
@@ -49,6 +75,10 @@ class HttpDownloader(BaseDownloader):
         >>>         task.result()  # This is a DownloadResult
         >>>     except Exception as error:
         >>>         pass  # fatal exceptions are raised by result()
+
+    The HTTPDownloaders contain automatic retry logic if the server responds with HTTP 429 response.
+    The coroutine will automatically retry 10 times with exponential backoff before allowing a
+    final exception to be raised.
 
     Attributes:
         session (aiohttp.ClientSession): The session to be used by the downloader.
@@ -118,9 +148,13 @@ class HttpDownloader(BaseDownloader):
                               url=self.url, exception=None)
 
     @attach_url_to_exception
+    @backoff.on_exception(backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=giveup)
     async def run(self):
         """
         Download, validate, and compute digests on the `url`. This is a coroutine.
+
+        This method is decorated with a backoff-and-retry behavior to retry HTTP 429 errors. It
+        retries with exponential backoff 10 times before allowing a final exception to be raised.
 
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader.run`.
--- a/docs/plugins/plugin-api/download.rst
+++ b/docs/plugins/plugin-api/download.rst
@@ -61,10 +61,10 @@ the downloader's `run()` method when the download is complete.
 .. autoclass:: pulpcore.plugin.download.DownloadResult
     :no-members:
 
-.. _configuring-from-an-remote:
+.. _configuring-from-a-remote:
 
-Configuring from an Remote
-----------------------------
+Configuring from a Remote
+-------------------------
 
 When fetching content during a sync, the remote has settings like SSL certs, SSL validation, basic
 auth credentials, and proxy settings. Downloaders commonly want to use these settings while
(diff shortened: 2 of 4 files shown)


plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
+
+
+logging.getLogger('backoff').addHandler(logging.StreamHandler())
+
+
+def inspect_to_retry(exc):
+    """
+    Inspect a raised method to determine if a retry is needed
+
+    Retries on when status code is one of the following:
```
dralley: type "Retries on"

plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
 
+log = logging.getLogger(__name__)
+
+
+logging.getLogger('backoff').addHandler(logging.StreamHandler())
+
+
+def inspect_to_retry(exc):
+    """
+    Inspect a raised method to determine if a retry is needed
```
dralley: s/method/exception ?

plugin/pulpcore/plugin/download/http.py
bmbouter: fixed in next push. Thanks.

plugin/pulpcore/plugin/download/http.py
bmbouter: fixed in next push thanks.

plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
 from .base import attach_url_to_exception, BaseDownloader, DownloadResult
 
 
+log = logging.getLogger(__name__)
+
+
+logging.getLogger('backoff').addHandler(logging.StreamHandler())
+
+
+def inspect_to_retry(exc):
```
dralley: I feel like this name is misleading.  The function returns true when a download should **not** be retried, and False if it should.  

A more correct name would be ```is_fatal_exception(exc)``` or something along those lines

plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
+    """
+    Inspect a raised exception to determine if a retry is needed
+
+    Retries when the status code is one of the following:
+
+        429 - Too Many Requests
+
+    Args:
+        exc (Exception): The exception to inspect
```
dralley: Presumably, if there is a special "status" field on the exception, there is some more specific Exception base class being expected here?

plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
 
+log = logging.getLogger(__name__)
+
+
+logging.getLogger('backoff').addHandler(logging.StreamHandler())
+
+
+def inspect_to_retry(exc):
+    """
+    Inspect a raised exception to determine if a retry is needed
```
dralley: For the same reason as above, probably best to be clear about what the function is doing in the description.

```Inspect a raised exception to determine if the download should not be retried```

It's kind of a double negative and not very ergonomic but, eh, that's what the library does.

File: plugin/pulpcore/plugin/download/http.py [hunk 6]
```diff
@@ -118,10 +148,14 @@ def __init__(self, url, session=None, auth=None, proxy=None, proxy_auth=None,
                               url=self.url, exception=None)
 
     @attach_url_to_exception
+    @backoff.on_exception(backoff.expo, aiohttp.ClientError, max_tries=10, giveup=inspect_to_retry)
```
dralley: Since not all ```aiohttp.ClientError``` have a status code, either this code should be made more specific by specifying ```aiohttp.ClientResponseError``` or the function above should account for the fact that not all subclasses of ```aiohttp.ClientError``` have status codes since they aren't all http-related.  Potentially some of those would still warrant a retry but I haven't really looked thoroughly (```ServerTimeoutError```? ```ServerDisconnectedError```?  ```ClientConnectionError```?)

plugin/pulpcore/plugin/download/http.py
dkliban: +1 to changing the name to be more appropriate

bmbouter: I agree completely. Let me fix this up.

bmbouter: What is the right name for this? Here are some ideas:

* giveup
* maybe_giveup
* inspect_exc_and_giveup
* inspect_exc_and_bail
* inspect_exc_and_fail

dralley: I'm fine with ```maybe_giveup```

dkliban: 'giveup' by itself is the best. @dralley  what do yo uthink?

dralley: I prefer ```maybe_giveup``` because it's more descriptive.  Giving up isn't a guarantee.  I don't care that much about this one point though.

plugin/pulpcore/plugin/download/http.py
bmbouter: I've updated this language some in my next push to match the renaming of the function. I think it's an improvement, but we can see what you think.

plugin/pulpcore/plugin/download/http.py
```diff
@@ -1,8 +1,34 @@
...
+
+        429 - Too Many Requests
+
+    Args:
+        exc (Exception): The exception to inspect
+
+    Returns:
+        False if the download should be retried, True otherwise
+    """
+    return not exc.status == 429
```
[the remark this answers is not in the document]
bmbouter: I agree completely, this needs to be switched to `exc.code`. Thanks for pointing this out. This is what I get for developing via rpdb. I'll test and push the change.

plugin/pulpcore/plugin/download/http.py
bmbouter: I see exactly what you are saying, and I agree.

I only intend to capture HTTP 429's for now. That's the only error code we have a concrete use case for. That exception type will only ever come as a the `aiohttp.ClientResponseError` so I'm updating the type to be that. This also matches the `exc.code` change from the above comment.

bmbouter:
I just tested it again to make sure it works as well. I will change the docstring, I want to make it the best.

PR #3430, opened 2018-04-12
closes #3563

dkliban:
https://pulp.plan.io/issues/3563


Ichimonji10:
Do this instead:

```python
"shell": {"transport": "local"},
```

PR #3434, opened 2018-04-13
DRY up task states

dralley:


pulpcore/pulpcore/app/models/task.py
```diff
@@ -274,23 +274,7 @@ class Task(Model):
...
-        (WAITING, 'Waiting'),
-        (SKIPPED, 'Skipped'),
-        (RUNNING, 'Running'),
-        (COMPLETED, 'Completed'),
-        (FAILED, 'Failed'),
-        (CANCELED, 'Canceled')
-    )
-
-    state = models.TextField(choices=STATES)
+    state = models.TextField(choices=TASK_CHOICES, default=TASK_STATES.WAITING)
```
dralley: This is the only change I am a little unsure about.  ProgressReport does set the default to "waiting", and it is a reasonable default IMO.  Better than setting no state at all, personally.

asmacdo: This might be reasonable, but lets restrict this change to DRYing up only.

PR #3435, opened 2018-04-13
[WIP] Require PRs from pulp_file and pulp-smash

dkliban:


.travis/install.sh
```diff
@@ -8,26 +8,34 @@ pushd common/ && pip install -e . && popd
 pushd pulpcore/ && pip install -e . && popd
 pushd plugin/ && pip install -e .  && popd
 
-export COMMIT_MSG=$(git log --format=%B -n 1 HEAD)
+export COMMIT_MSG=$(git show HEAD^2 -s)
```
daviddavis: FYI, there's a variable for this I believe. Something like $TRAVIS_COMMIT_MESSAGE.

dkliban: There is, but it's for the merge commit of the PR into the branch it's going to. So it only says "Merge of blah into bleh". I have to find the parent commit of that.

dralley: Git has an option for ignoring merge commits ```--no-merges```.  You could try that.

.travis/install.sh
```diff
@@ -8,26 +8,31 @@ pushd common/ && pip install -e . && popd
...
-  PULP_SMASH_SHA='master';
+  pip install git+https://github.com/PulpQE/pulp-smash.git#egg=pulp-smash
 else
   export PULP_SMASH_SHA=$(http https://api.github.com/repos/PulpQE/pulp-smash/pulls/$PULP_SMASH_PR_NUMBER | jq -r '.merge_commit_sha')
+  cd ../
+  git clone https://github.com/PulpQE/pulp-smash.git
+  cd pulp-smash
+  git fetch origin +refs/pull/$PULP_SMASH_PR_NUMBER/merge
+  git checkout $PULP_SMASH_SHA
+  pip install -e .
```
daviddavis: Wondering if we should DRY this up with a bash function?

docs/contributing/continuous_integration.rst
```diff
@@ -0,0 +1,26 @@
+.. _tests: https://github.com/pulp/pulp/blob/3.0-dev/tests/
+.. _pulpcore: https://github.com/PulpQE/pulp-smash/tree/master/pulp_smash/tests/pulp3/pulpcore
+.. _file: https://github.com/PulpQE/pulp-smash/tree/master/pulp_smash/tests/pulp3/file
+.. _pulp-smash: https://github.com/PulpQE/pulp-smash/
+
+Continuous Integration
```
bmbouter: I think these docs are great, thanks!

.travis/install.sh
dkliban: let's DRY up in a separate PR. please?

daviddavis:
Should the commit be updated to remove the lines `Required PR: ...`?

dkliban:
The commit message says that those lines are in the commit message for testing purposes only. Should I remove them and test that the PR still passes tests?

PR #3436, opened 2018-04-13
General cleanup

dralley:
* Added some comments
* Converted some comments to google-style
* Moved some filters around for clarity
* Deleted an old reference to Scheduler
* Added Exporter to the ViewSets __init__.py
* Removed filters from "retrieve" action in the DefaultSchema
  (this actually does not change the generated schema at all, I checked)

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/__init__.py
+++ b/pulpcore/pulpcore/app/viewsets/__init__.py
@@ -2,6 +2,7 @@ from pulpcore.app.viewsets.base import (AsyncRemoveMixin, AsyncUpdateMixin,  # n
                                         NamedModelViewSet)
 from pulpcore.app.viewsets.content import ArtifactViewSet, ContentViewSet  # noqa
 from pulpcore.app.viewsets.repository import (DistributionViewSet,  # noqa
+                                              ExporterViewSet,
                                               RemoteViewSet,
                                               PublicationViewSet,
                                               PublisherViewSet,
--- a/pulpcore/pulpcore/app/viewsets/base.py
+++ b/pulpcore/pulpcore/app/viewsets/base.py
@@ -48,7 +48,7 @@ class DefaultSchema(AutoSchema):
             return False
 
         if hasattr(self.view, 'action'):
-            return self.view.action in ["list", "retrieve"]
+            return self.view.action in ["list"]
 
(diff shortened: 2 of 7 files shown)


pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -10,34 +10,54 @@
 from pulpcore.app.viewsets import NamedModelViewSet
 
 
-class ContentFilter(filterset.FilterSet):
```
dralley: Ignore this, github is confused.  I just moved the filters around to sit next to their viewsets, and changed the docstrings to be consistent with the changes Austin wanted in the filter overhaul PR

pulpcore/pulpcore/app/viewsets/__init__.py
```diff
@@ -2,6 +2,7 @@
                                         NamedModelViewSet)
 from pulpcore.app.viewsets.content import ArtifactViewSet, ContentViewSet  # noqa
 from pulpcore.app.viewsets.repository import (DistributionViewSet,  # noqa
+                                              ExporterViewSet,
```
dralley: This should probably be here

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -169,7 +169,7 @@ def urlpattern(cls):
     def endpoint_pieces(cls):
         # This is a core ViewSet, not Master/Detail. We can use the endpoint as is.
         if cls.queryset.model._meta.master_model is None:
-            return (cls.endpoint_name,)
+            return [cls.endpoint_name]
```
dralley: Function should preferably return only one type (could return either tuple or list).  Since the length of the return value is variable, we should probably use a list (it doesn't really matter but it's more intuitive).

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -47,7 +47,7 @@ def _allows_filters(self, path, method):
             return False
 
         if hasattr(self.view, 'action'):
-            return self.view.action in ["list", "retrieve"]
+            return self.view.action in ["list"]
```
dralley: I double checked and this does not, actually, change the output schema at all.  But we shouldn't have it there anyway.

PR #3437, opened 2018-04-15
get_resource() shouldn't accept a non-specific URI

dralley:
closes #3567
https://pulp.plan.io/issues/3567


daviddavis:
I'm a bit hesitant to change `serializers.ValidationError` to `DRFValidationError` since elsewhere in the code we call `serializers.ValidationError`. Thoughts?

dralley:
I feel you, but here at least, using the full name is a massive waste of space and it was difficult/unweildy to get the message to wrap in a PEP8 compliant way without doing this.

PR #3438, opened 2018-04-15
Add more unit tests around NamedModelViewSet

dralley:
closes #3567
https://pulp.plan.io/issues/3568

### DIFF
--- a/pulpcore/tests/viewsets/test_base.py
+++ b/pulpcore/tests/viewsets/test_base.py
@@ -1,34 +1,224 @@
 from uuid import uuid4
+import unittest
 
+from django.http import Http404
 from django.test import TestCase
+from rest_framework.serializers import ValidationError as DRFValidationError
 
-from pulpcore.app import models, viewsets
+from pulpcore.app import models, viewsets, serializers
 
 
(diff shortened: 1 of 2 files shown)


pulpcore/tests/viewsets/test_base.py
```diff
@@ -1,34 +1,222 @@
...
+from django.http import Http404
 from django.test import TestCase
+from rest_framework.serializers import ValidationError as DRFValidationError
 
-from pulpcore.app import models, viewsets
+from pulpcore.app import models, viewsets, serializers
 
 
 class TestGetQuerySet(TestCase):
     def test_adds_filters(self):
```
dralley: This test currently fails and I'm not sure why.  ```queryset``` is empty for some reason.

dralley:
@daviddavis should I disable the test for now?  As I mentioned, it was not *actually* testing anything previously.

daviddavis:
It would be a nice test to have but if you don't have the time or can't figure it out, then +1.

dralley:
I'll disable it for now.  We can figure it out when we do some unit testing passes later on.

PR #3448, opened 2018-04-19
Order list views where applicable

dralley:
closes: 3576
https://pulp.plan.io/issues/3576

### DIFF
--- a/pulpcore/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/pulpcore/app/viewsets/repository.py
@@ -4,6 +4,7 @@ import itertools
 from django_filters.rest_framework import filters, filterset
 from django_filters import Filter
 from rest_framework import decorators, mixins, serializers
+from rest_framework.filters import OrderingFilter
 
 from pulpcore.app import tasks
 from pulpcore.app.models import (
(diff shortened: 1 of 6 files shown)


daviddavis:
Cool, I tried to add the OrderingFilter backend to our entire app last night (like what we do for [the Filter backend](https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/settings.py#L111)) but I couldn't get it to work with the paginated endpoints. I think we may want that eventually but this works for now.

pulpcore/pulpcore/app/viewsets/repository.py
```diff
@@ -177,6 +178,8 @@ class RepositoryVersionViewSet(NamedModelViewSet,
     serializer_class = RepositoryVersionSerializer
     queryset = RepositoryVersion.objects.exclude(complete=False)
     filter_class = RepositoryVersionFilter
+    filter_backends = (OrderingFilter,)
+    ordering = ('-number',)
```
daviddavis: I feel like this should count up and not down. But then again I guess if this is eventually paginated, you'd want to see the latest ones first?

PR #3449, opened 2018-04-19
Prevent task deletion with locked resources

dralley:
closes #3509
https://pulp.plan.io/issues/3509


dralley:
This needs a close review.  It worked fine in the testing I did, but I'm not confident that I did everything properly, just because my database knowledge is not great when it comes to stuff like Many-to-Many relationships.

PR #3451, opened 2018-04-19
Provide more informative invalid base path error

CodeHeeler:
Default message provided by Django's URLValidator is simply:
"Enter a valid URL"

fixes #3475
https://pulp.plan.io/issues/3475

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -162,7 +162,7 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
     def _validate_relative_path(self, path):
         """
         Validate a relative path (eg from a url) to ensure it forms a valid url and does not begin
-        or end with slashes
+        or end with slashes nor contain spaces
 
         Args:
             path (str): A relative path to validate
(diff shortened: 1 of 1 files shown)


CodeHeeler:
There is some larger conversation going on today around whether URLValidator is too strict and perhaps our validation will change, but if we go ahead and put additional messaging in place, it will serve as a reminder that we need more informative errors on future work as well.

We already have an error message covering leading and trailing slashes, this one covers spaces.  Is there any other cases that should be covered for now?  It's overkill to link to W3's ideas of what constitutes a valid URL.

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -171,7 +171,7 @@ def _validate_relative_path(self, path):
         """
         # in order to use django's URLValidator we need to construct a full url
         base = "http://localhost"  # use a scheme/hostname we know are valid
-        validate = URLValidator()
+        validate = URLValidator(message=_("Relative path cannot contain spaces."))
```
dralley: Suggestion:

Instead, move lines 177 and 178 up here, and add another case ```if ' '  in path: raise ...```

dralley: Suggestion:

Instead of doing this, you could move lines 177-178 up here, and raise a validation error with a specific error  message in the condition  ```if ' ' in path```

CodeHeeler: Ok, good point!

PR #3457, opened 2018-04-21
Switch to codecov

daviddavis:
VVV This is expected since there are no prior codecov data. VVV


daviddavis:
^^^ This is expected since there are no prior codecov data. ^^^

This PR is ready for review/merging.

PR #3458, opened 2018-04-21
Speed up travis by using containers

daviddavis:
More info: https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/


dkliban:
Image of Awesome!

daviddavis:
Unfortunately, I'm not seeing a huge improvement using containers. Without this change, builds take about 9-10 min total and with this, it's still in the same 9-10 min range. ☹️ 

Closing out.

PR #3461, opened 2018-04-23
Adding pulp_ansible to list

bmbouter:


docs/plugins/index.rst
```diff
@@ -35,6 +35,13 @@ Plugin List
      - Yes
      - No
 
+   * - Ansible Roles
```
daviddavis: I don't think we should limit this plugin to ansible roles since there's already talk about supporting other content types. Thoughts?

bmbouter: I agree, but can we update the table again later as type support gets added?

bmbouter: Actually let's go with `Ansible plug-in`. I'll push a change.

PR #3464, opened 2018-04-23
ChangeSet Improvements

jortel:
https://pulp.plan.io/issues/3582

Improvements:
- The `ChangeSet.apply_and_drain()` removed.
- The `SizedIterator` no longer used.
- ChangeSet.apply() signature updated.
- Progress reported for artifact downloads.
- Exposed `ContentIterator` as it will be useful for other abstractions.

### DIFF
--- a/plugin/pulpcore/plugin/changeset/iterator.py
+++ b/plugin/pulpcore/plugin/changeset/iterator.py
@@ -1,7 +1,7 @@
 import asyncio
 import itertools
 
-from collections.abc import Iterable
+from collections import Iterable, Sized
 from logging import getLogger
 
 from django.db.models import Q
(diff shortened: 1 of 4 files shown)


plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -284,10 +315,16 @@ def _iter(self):
...
             correlation = {d: a for a, d in batch}
             pending = correlation.keys()
+            if self.bar:
+                self.bar.total += len(batch)
+                self.bar.save()
             while pending:
                 future = asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
                 completed, pending = loop.run_until_complete(future)
                 for task in completed:
+                    if self.bar:
```
gmbnomis: I find the use of the BatchIterator unusual for downloading.

Say for example, the we have to download 20 artifacts overall. The size of the artifacts are 1* 1GB + 19 * 1MB (they are yielded in this order by the artifact iterator). 
With a batch size of 10, the first 10 downloads will be started in parallel. However, all of them need to terminate (including the 1 GB one) before the second batch is started. In the first batch, there will be only one concurrent download for most of the time with 10 queued downloads that do not start.

I would expect `pending` to be filled with up to 10 downloads initially. Then, it is backfilled by one download for each completed task in order to always have 10 concurrent downloads. But perhaps I am missing something here...

jortel: Good catch @gmbnomis.  You have identified a flaw where. Using the `BatchIterator` here to support the progressive _progress_ reporting will still be useful but the batch needs to be bigger (like the default: 1024).  Then within each batch the `self.concurrent` needs to be used differently to constrain the number of concurrent downloads.  Will fix.

plugin/pulpcore/plugin/changeset/iterator.py
```diff
@@ -48,11 +48,38 @@ def __iter__(self):
         while True:
             batch = tuple(itertools.islice(generator, 0, self.batch))
             if batch:
-                yield batch
+                yield Batch(batch, len(batch))
             else:
                 return
 
 
+class Batch(Iterable, Sized):
```
jortel: Removing the `Batch` because it adds no value.  Not sure what I was thinking :-/

dralley:
Closing this since ChangeSets are being removed in favor of stages @bmbouter @jortel

PR #3468, opened 2018-04-23
Have get_resource() use pk and only pk if pk is provided

daviddavis:
We should ignore other kwargs since pk should uniquely identify the
resource. The problem we were experiencing in pulp_ansible is that we
had a pk and a role_pk. The conversion from role_pk to role_id only
happens inside the else block and so we're getting a FieldError because
role_pk doesn't exist on the model. Instead, I think we should just use
the pk kwarg and ignore the rest.


daviddavis:
Another alternative fix would be to move this logic outside of the else:

https://github.com/pulp/pulp/pull/3468/files#diff-ae7f375b1b9eaee42aecdf250679cd6eL137

PR #3475, opened 2018-04-25
Switch pulp's default db back to postgres

daviddavis:
fixes #3606
https://pulp.plan.io/issues/3606


.travis/before_script.sh
```diff
@@ -14,11 +14,6 @@ sudo mkdir /var/lib/pulp/tmp
 sudo mkdir /etc/pulp/
 sudo chown -R travis:travis /var/lib/pulp
 
-if [ "$DB" = 'postgres' ]; then
-  sudo cp .travis/server.postgres.yaml /etc/pulp/server.yaml
-else
-  # docs job also requires server.yaml
-  sudo cp .travis/server.sqlite.yaml /etc/pulp/server.yaml
-fi
+sudo cp .travis/server.postgres.yaml /etc/pulp/server.yaml
```
daviddavis: I decided to leave `.travis/server.sqlite.yaml` in case we support it again one day but if anyone disagrees, I'm happy to remove it.

dkliban: I would remove the file. We can use git to bring it back.

daviddavis:
@dkliban removed the sqlite server.yaml.

PR #3476, opened 2018-04-26
Move single-artifact related code into core

daviddavis:
fixes #3607
fixes https://pulp.plan.io/issues/3607

### DIFF
--- a/pulpcore/pulpcore/app/models/content.py
+++ b/pulpcore/pulpcore/app/models/content.py
@@ -138,6 +138,27 @@ class Content(MasterModel):
         return tuple(getattr(self, f) for f in self.natural_key_fields())
 
 
+class SingleArtifactContent(Content):
+    """
+    Class for plugin content that only have a single artifact.
+    """
+
+    def _calc_relative_path(self):
+        pass
+
+    @property
+    def artifact(self):
+        return self.artifacts.first().pk
+
+    @artifact.setter
+    def artifact(self, artifact):
+        if self.pk:
+            ca = ContentArtifact(artifact=artifact,
+                                 content=self,
+                                 relative_path=self._calc_relative_path())
+            ca.save()
+
+
 class ContentArtifact(Model):
     """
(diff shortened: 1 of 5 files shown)


pulpcore/pulpcore/app/models/content.py
```diff
@@ -138,6 +138,27 @@ def natural_key(self):
         return tuple(getattr(self, f) for f in self.natural_key_fields())
 
 
+class SingleArtifactContent:
```
jortel: Why is this a _mixin_ instead of a subclass of `Content`?  Generally, I'm opposed to _mixins_.

Why not add these convenience properties to `Content`?

daviddavis: Will make SingleArtifactContent a subclass 👍 

The reason I didn't want to add these to `Content` is that I didn't want plugin writers to accidentally call them if they're dealing with Content that may have multiple artifacts.

daviddavis: I'm cool with adding them to `Content` though and maybe being explicit in the docstring that they're not for use with Content that can have multiple artifacts?

pulpcore/pulpcore/app/viewsets/content.py
```diff
@@ -74,14 +74,4 @@ def create(self, request):
...
-        artifacts = serializer.validated_data.pop('artifacts')
-        content = serializer.save()
-
-        for relative_path, artifact in artifacts.items():
-            ca = ContentArtifact(artifact=artifact, content=content, relative_path=relative_path)
-            ca.save()
-
-        headers = self.get_success_headers(serializer.data)
-        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+        return super().create(request)
```
dralley: Does this need to be overridden at all if we're just calling ```super()```?

daviddavis:
Abandoning this work. May create a serializer in the future to handle this.

PR #3479, opened 2018-04-27
Fixes using multiple serializers on derived viewsets

dralley:
closes #3587
https://pulp.plan.io/issues/3587

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -189,6 +189,34 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
 
         return path
 
+    def to_representation(self, instance):
+        """
+        Object instance -> Dict of primitive datatypes.
+
+        This is very similar to DRF's default to_representation implementation in
+        ModelSerializer, but checks whether a 'minimal_fields' attribute is defined,
+        and uses it when many objects are being serialized at once (i.e. many=True,
+        when the viewset is performing a 'list' action).
+        """
+        ret = OrderedDict()
+        fields = self._readable_fields
+        if self.parent and self.parent.many and hasattr(self.Meta, 'minimal_fields'):
+            fields = [field for field in fields if field.field_name in self.Meta.minimal_fields]
+
+        for field in fields:
+            try:
+                attribute = field.get_attribute(instance)
+            except SkipField:
+                continue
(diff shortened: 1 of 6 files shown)


pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -232,8 +232,7 @@ def to_representation(self, instance):
         ret = OrderedDict()
 
         instance = instance.cast()
-        viewset = viewset_for_model(instance)
-        fields = viewset.serializer_class(context=self._context)._readable_fields
```
dralley: I need some special attention paid to this line.  I think it should be fine but just want to be sure.

The reason this line is problematic is because we want to be able to redefine which serializer class is being used at runtime based on the action being exercised (e.g. use a different, stripped-down serializer for listing objects).  DRF is fine with that, which is why internally they *never* use ```.serializer_class``` and instead use ```get_serializer_class()``` everywhere - explicitly so that people can redefine that behavior.

dralley: And FWIW, I don't understand what this method is doing.  I commented the whole thing out and as far as I could tell everything worked fine and exactly the same, and I tested a number of things.

dralley: viewset is just a class, not an instance

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -189,6 +189,40 @@ def _validate_relative_path(self, path):
...
+        # ModelSerializer, but checks whether a 'minimal_fields' attribute is defined,
+        # and uses it when many objects are being serialized at once (i.e. many=True,
+        # when the viewset is performing a 'list' action).
+        ret = OrderedDict()
+        fields = self._readable_fields
+        if self.parent and self.parent.many and hasattr(self.Meta, 'minimal_fields'):
+            fields = [field for field in fields if field.field_name in self.Meta.minimal_fields]
+
+        for field in fields:
+            try:
```
bmbouter: This looks like a good job for [suppress](https://docs.python.org/3.5/library/contextlib.html#contextlib.suppress)

bmbouter:
I think this looks like a sane approach.

I'm just curious. Does this kind of feature make any sense contributed upstream in a future version of DRF provided serializers?

PR #3484, opened 2018-05-02
Beta2 release notes and version bump

werwty:


werwty:
I think we should start committing beta version bumps to github- when travis starts releasing we would create a commit for this anyways so it know what to release. 

pulpcore-plugin and pulpcore-common were not bumped since there were no changes to release there.

PR #3482, opened 2018-05-01
Update migration doc.

jortel:
https://pulp.plan.io/issues/3612

PR #3487, opened 2018-05-09
Update installation instructions

dralley:


docs/installation/instructions.rst
```diff
@@ -18,16 +26,25 @@ PyPI Installation
...
 3. Install Pulp::
 
    $ pip3 install pulpcore
 
-
 .. note::
 
-   To install from source, replace the pip3 install commands to specify a source install such as::
+   To install from source, clone the git repository, checkout the 3.0-dev branch, navigate to the
+   directory containing the ``setup.py`` file, and do a local, editable pip installation::
```
werwty: It might be good to call out the 3 packages needed: pulpcore, plugin, and common

docs/installation/instructions.rst
```diff
@@ -36,8 +53,7 @@ PyPI Installation
 
    .. note::
 
-       The exact path will depend on the major *and* minor Python version found by venv e.g.
-       /lib/python3.5/, /lib/python3.6/
+       The exact path will depend on the operating system, Python version, and other factors.
```
werwty: Maybe link to https://raw.githubusercontent.com/pulp/pulp/3.0-dev/pulpcore/pulpcore/etc/pulp/server.yaml in case they want to put it in /etc/pulp/server.yaml and not want to go look for where venv installed pulp

docs/installation/instructions.rst
```diff
@@ -1,10 +1,18 @@
...
 
+Supported Platforms
+-------------------
+
+Pulp has been tested to work on Ubuntu, Fedora, and CentOS. It should work on other operating systems
+such as Debian or OSX, but no guarentees.
+
 .. note::
 
     As Pulp 3 currently does not have an SELinux Policy, it currently requires the target
```
bmbouter: What about mentioning AppArmor here too? It's the SELinux for Ubuntu.

docs/installation/instructions.rst
dralley: I thought that too but since ```pulpcore``` depends on ```pulpcore-common``` and any plugin should (and currently does) depend on ```pulpcore-plugin```, I think it's fine to leave it out.

docs/installation/instructions.rst
dralley: +1

docs/installation/instructions.rst
dralley: +1

docs/installation/instructions.rst
dralley: I know very little about AppArmor but I think it works differently?  I'm not sure there is anything to disable by default.

docs/installation/instructions.rst
```diff
@@ -1,10 +1,18 @@
 Installation Instructions
 =========================
 
+Supported Platforms
+-------------------
+
+Pulp has been tested to work on Ubuntu, Fedora, and CentOS. It should work on other operating systems
```
bmbouter: I don't think of this as "Supported Platforms". I hope we don't get into "cerifying" distributions. What about this as an idea:

Pulp can run on any distribution that provides a the Python 3.5+ runtime environment, the necessary dependencies, and supporting services, e.g. a database. Pulp has been demonstrated to run on Ubuntu, Fedora, CentOS, Debian, and OS X, and likely runs in many more places.

docs/installation/instructions.rst
```diff
@@ -76,9 +92,29 @@ PyPI Installation
 Database Setup
 --------------
 
-Databases can be configed in the `databases` section of your server.yaml. See the `Django database
-settings documentation <https://docs.djangoproject.com/en/1.11/ref/settings/#databases>`_ for more
-information on setting the `databases` values in settings.yaml.
+You must provide a compatible SQL database for Pulp to use. At this time Pulp 3.0 is only known to work
+properly with PostgreSQL. It may work with other databases, but no guarantees.
```
bmbouter: s/databases/databases Django supports/

dralley: +1

dralley:
This PR should also be taken with https://github.com/pulp/pulp/pull/3482 if tested

docs/installation/instructions.rst
```diff
@@ -1,10 +1,19 @@
 Installation Instructions
 =========================
 
+Supported Platforms
+-------------------
+
+Pulp has been demonstrated to work on Ubuntu, Debian, Fedora, CentOS, and Mac OSX, but it should
+work on any operating system that can provide a Python 3.5+ runtime environment and the supporting
```
bmbouter: Can we reverse the order of these statements. The primary claim is the runtime environment so I'm hoping that will be first.

docs/installation/instructions.rst
```diff
@@ -1,10 +1,19 @@
...
+dependencies e.g. a database.
+
 .. note::
 
     As Pulp 3 currently does not have an SELinux Policy, it currently requires the target
-    machine to have SELinux set to permissive mode::
+    machine to have SELinux set to permissive mode.
```
bmbouter: s/mode/mode or disabled/

docs/installation/instructions.rst
```diff
@@ -1,10 +1,19 @@
 Installation Instructions
 =========================
 
+Supported Platforms
+-------------------
+
+Pulp should work on any operating system that can provide a Python 3.5+ runtime environment and
+the supporting dependencies e.g. a database, and has been demonstrated to work on Ubuntu, Debian,
```
bmbouter: Can this become two sentences? The second sentence would start with "Pulp has been ...".

PR #3496, opened 2018-05-16
Moar coverage reports

bmbouter:
Adds a coverage report for:
* pulp-smash-workers
* pulp-smash-webservers

Also it renames the existing unittest coverage report to 'unittests'.


bmbouter:
Closing because it's not as great as the one from @mdellweg here:  https://github.com/pulp/pulp/pull/3814

PR #3497, opened 2018-05-18
FIPS: Setting the usedforsecurity flag to False if available

daviddavis:
In some builds of Python, a usedforsecurity flag is available to declare that md5 is not being used for security purposes. We're using md5 for checksums so set this flag if it's available. This is required for FIPS environments.

### DIFF
--- a/server/pulp/server/util.py
+++ b/server/pulp/server/util.py
@@ -4,6 +4,7 @@ This module contains utility code to be used by pulp.server.
 from contextlib import contextmanager
 from gettext import gettext as _
 import hashlib
+import inspect
 import logging
 import os
 from shutil import copy, Error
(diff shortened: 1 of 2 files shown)


server/pulp/server/util.py
```diff
@@ -16,19 +17,31 @@
...
+TYPE_MD5 = md5()().name
 TYPE_SHA = 'sha'
 TYPE_SHA1 = hashlib.sha1().name
 TYPE_SHA256 = hashlib.sha256().name
 
 HASHLIB_ALGORITHMS = (TYPE_MD5, TYPE_SHA, TYPE_SHA1, TYPE_SHA256)
 
 CHECKSUM_FUNCTIONS = {
-    TYPE_MD5: hashlib.md5,
+    TYPE_MD5: md5,
```
goosemania: I think either this should be `md5()` or `def md5` should return the result when called and not a function. WDYT?

daviddavis: Fixed.

daviddavis:
ok test

daviddavis:
ok test

daviddavis:
ok test

PR #3498, opened 2018-05-18
Consistent names in server.yaml

CodeHeeler:
The naming convention for yaml is lowercase snake.
We had some all-uppercase names mixed in, now made consistent.

fixes #3589
https://pulp.plan.io/issues/3589

### DIFF
--- a/pulpcore/pulpcore/app/settings.py
+++ b/pulpcore/pulpcore/app/settings.py
@@ -162,7 +162,7 @@ STATIC_URL = '/static/'
 # /etc/pulp/ is missing or if it does not have values for every setting
 _DEFAULT_PULP_SETTINGS = {
     # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
-    'databases': {
+    'DATABASES': {
         'default': {
             'ENGINE': 'django.db.backends.postgresql_psycopg2',
             'NAME': 'pulp',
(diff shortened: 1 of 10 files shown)


pulpcore/pulpcore/app/settings.py
```diff
@@ -197,21 +197,21 @@
...
-        'working_directory': '/var/lib/pulp/tmp',
+    'SERVER': {
+        'WORKING_DIRECTORY': '/var/lib/pulp/tmp',
     },
-    'content': {
-        'web_server': 'django',
-        'host': None,
+    'CONTENT': {
+        'WEB_SERVER': 'django',
+        'HOST': None,
```
dralley: I think this usage needs to be changed

https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/serializers/fields.py#L175

pulpcore/pulpcore/etc/pulp/server.yaml
```diff
@@ -73,38 +73,38 @@ databases:
...
 # of Redis traffic over untrusted networks Redis recommends spiped.
 #
-# redis:
-#   host: 127.0.0.1
-#   port: 6379
-#   password:
+# REDIS:
+#   HOST: 127.0.0.1
+#   PORT: 6379
+#   PASSWORD:
```
dralley: Same here: https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/tasking/connection.py#L11

PR #3508, opened 2018-05-30
Fix dependencies for source installation

dralley:
pulpcore is currently at 3.0.0b4, pulpcore-plugin requires 3.0.0b3,
so the editable installation fails and it falls back to installing from
PyPI. So the developer installation will not actually be a developer
installation.

### DIFF
--- a/plugin/setup.py
+++ b/plugin/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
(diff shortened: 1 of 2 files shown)


werwty:
Can you also make the change in pulpcore for the pulpcore-common dependency?
https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/setup.py#L17

PR #3513, opened 2018-06-01
Redact last_override_config

bmbouter:
This commit resolvs CVE-2018-1090 by redacting all data from
last_override_config. It does not modify what is saved in the database.
It also adds a release note for the 2.16.2 release.

https://pulp.plan.io/issues/3521
closes #3521


dkliban:
ok test

PR #3514, opened 2018-06-02
Fixes Pulp w/ Mongoengine v0.14.1+

dralley:
closes #3687
https://pulp.plan.io/issues/3687

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -553,7 +553,13 @@ class TaskStatus(AutoRetryDocument, ReaperMixin):
         set_on_insert = {}
         for field in fields_to_set_on_insert:
             set_on_insert[field] = stuff_to_update.pop(field)
+
         task_id = stuff_to_update.pop('task_id')
+
+        # Don't try to set Mongoengine internal values.
+        # Can cause silent failures if you accidentally overwrite 'id'.
+        for internal_field in ('id', '_ns'):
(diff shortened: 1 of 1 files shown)


server/pulp/server/db/model/__init__.py
```diff
@@ -553,8 +553,14 @@ def save_with_set_on_insert(self, fields_to_set_on_insert):
         set_on_insert = {}
         for field in fields_to_set_on_insert:
             set_on_insert[field] = stuff_to_update.pop(field)
+
         task_id = stuff_to_update.pop('task_id')
 
+        # Don't try to set Mongoengine internal values.
+        # Can cause silent failures if you accidentally overwrite 'id'.
+        for internal_field in ('id', '_ns'):
```
goosemania: How did you choose which fields are internal? E.g. `_id` is not in the tuple.

asmacdo: @goosemania Mongoengine uses `id` as an alias for `_id`

goosemania: @asmacdo , thank you

goosemania:
@dralley , thanks for the fix! :tada: 
It works for me both with the mongoengine prior to 0.14.1 and with 0.15.

server/pulp/server/db/model/__init__.py
dralley: @goosemania exactly, and as far as methods go, I interactively debugged the _data object to determine what was being added to it.

PR #3517, opened 2018-06-06
Remove installation step for installing RQ by git commit

dralley:
RQ v0.11.0 contains the changes we need.

### DIFF
--- a/pulpcore/setup.py
+++ b/pulpcore/setup.py
@@ -12,7 +12,7 @@ requirements = [
     'drf-yasg',
     'psycopg2-binary',
     'PyYAML',
-    'rq',
+    'rq>=0.11.0',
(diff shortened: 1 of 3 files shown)


pulpcore/setup.py
```diff
@@ -12,7 +12,7 @@
     'drf-yasg',
     'psycopg2-binary',
     'PyYAML',
-    'rq',
+    'rq>0.10.0',
```
bmbouter: What about having:    `'rq>=0.111'` ?

dralley: I don't really mind either way

asmacdo: We wouldn't want `0.10.1` (I assume)

PR #3520, opened 2018-06-11
Ignore missing consumer profile during applicability regeneration

goosemania:
closes #3745
https://pulp.plan.io/issues/3745


goosemania:
ok test

PR #3521, opened 2018-06-11
Use a shorter CA key in FIPS environments

dralley:
FIPS does not support 4096 bit length keys
https://pulp.plan.io/issues/3646


dralley:
ok test

PR #3524, opened 2018-06-14
Add Sync and Publish serializers

asmacdo:
https://pulp.plan.io/issues/3617
fixes #3617


daviddavis:
This looks great 👍 

What do you think of adding a couple unit tests?

PR #3531, opened 2018-06-19
Raise a ValidationError when undefined attributes are passed to an endpoint.

muattiyah:
refs: https://pulp.plan.io/issues/2970

# TODO
- [x] https://github.com/pulp/pulp_file/pull/96/
- [x] Tests, yo

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -76,6 +76,16 @@ def view_name_for_model(model_obj, view_action):
     raise LookupError('view not found')
 
 
+def validate_unknown_fields(initial_data, defined_fields):
+    """
+    This will raise a `ValidationError` if a serializer is passed fields that are unknown.
+    """
+    unknown_fields = set(initial_data) - set(defined_fields)
+    if unknown_fields:
+        unknown_fields = {field: _('Unexpected field') for field in unknown_fields}
+        raise serializers.ValidationError(unknown_fields)
+
+
 # Defined here instead of generic.py to avoid potential circular imports issues,
 # since this is used by ModelSerializer
 class GenericKeyValueRelatedField(serializers.DictField):
@@ -216,6 +226,11 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
                 ret[field.field_name] = field.to_representation(attribute)
 
         return ret
+
+    def validate(self, data):
+        if hasattr(self, 'initial_data'):
+            validate_unknown_fields(self.initial_data, self.fields)
+        return data
 
 
 class MasterModelSerializer(ModelSerializer):
--- a/pulpcore/tests/serializers/test_base.py
+++ b/pulpcore/tests/serializers/test_base.py
@@ -1,7 +1,9 @@
 from unittest import TestCase, mock
 
+from rest_framework import serializers
+
 from pulpcore.app import models
-from pulpcore.app.serializers import base
+from pulpcore.app.serializers import base, validate_unknown_fields
 
 
 class TestViewNameForModel(TestCase):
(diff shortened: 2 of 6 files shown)


muattiyah:
Ok, so the following code imitates the relevant parts of the test failure, running it against a local pulp instance indeed caused an HTTP400 error.

```python
from pulp_smash import api, config
from pulp_smash.constants import FILE_URL
from pulp_smash.pulp3.constants import ARTIFACTS_PATH, FILE_CONTENT_PATH


cfg = config.get_config()
content_unit = {}
client = api.Client(cfg, api.json_handler)
client.request_kwargs['auth'] = get_auth()
files = {'file': utils.http_get(FILE_URL)}
artifact = client.post(ARTIFACTS_PATH, files=files)

attrs = _gen_content_unit_attrs(self.artifact)
content_unit.update(client.post(FILE_CONTENT_PATH, attrs))


def _gen_content_unit_attrs(artifact):
    """Generate a dict with content unit attributes.

    :param: artifact: A dict of info about the artifact.
    :returns: A semi-random dict for use in creating a content unit.
    """
    return {'artifact': artifact['_href'], 'relative_path': utils.uuid4()}
```

After creating the artifact and before doing a POST on `FILE_CONTENT_PATH`, I sent the request via `httpie` like so `http http://127.0.0.1:8000/pulp/api/v3/content/file/files/ artifact=http://localhost:8000/pulp/api/v3/artifacts/a043e993-27c4-4a8a-aa7a-cd0d2035f321/ relative_path=1a5c9718-6cfa-45c2-a1cd-01f6541c5d76` and what I got in return was:

```
HTTP/1.1 400 Bad Request
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 53
Date: Fri, 22 Jun 2018 08:28:24 GMT
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "non_field_errors": [
        "Undefined Fields: {'digest'}"
    ]
}
```

So, this is indeed an error that was caused by the `validate()` function that I added on the `ModelSerializer`. I still haven't figured out where this is happening.

muattiyah:
We probably need to add a `digest` field here https://github.com/pulp/pulp_file/blob/master/pulp_file/app/serializers.py#L20

muattiyah:
Blocked by https://github.com/pulp/pulp_file/pull/96

daviddavis:
Just a heads up: some serializers are overriding the `validate` method of ModelSerializer and therefore are still allowing bad fields:

```
http POST http://localhost:8000/pulp/api/v3/distributions/ name='baz' base_path='foo' badfield='wat'
```

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -217,6 +217,12 @@ def to_representation(self, instance):
 
         return ret
 
+    def validate(self, data):
+        unknown_fields = set(self.initial_data.keys()) - set(self.fields.keys())
+        if unknown_fields:
+            raise serializers.ValidationError("Undefined Fields: {}".format(unknown_fields))
```
daviddavis: I'd recommend:

```python
raise serializers.ValidationError("Unknown fields: {}".format(", ".join(unknown_fields)))
```

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -217,6 +217,12 @@ def to_representation(self, instance):
 
         return ret
 
+    def validate(self, data):
+        unknown_fields = set(self.initial_data.keys()) - set(self.fields.keys())
+        if unknown_fields:
+            raise serializers.ValidationError("Unknown fields: {}".format(", ".join(unknown_fields)))
```
daviddavis: I forgot to mention that this needs i18n. And maybe use a mapping:

```python
raise serializers.ValidationError({unknown_fields.pop(): _('Unexpected field')})
```

muattiyah:
I see, I have updated them with the same validation logic, I am thinking about extracting the relevant  bits into a function `validate_unknown_fields` inside `serializers/base.py`, what do you think?

daviddavis:
Would calling `super` be an option?

muattiyah:
It would be for the serializers I've changed with the exception of `RepositoryPublishURLSerializer` because it inherits from DRF's `serializers.Serializer`.

muattiyah:
So we could either:

1- Replicate the logic inside `RepositoryPublishURLSerializer` `validate()` and call super on the rest of the serializers.

2- Factor it out into a function, call super on the rest of the serializers and call the function in `RepositoryPublishURLSerializer`.

pulpcore/pulpcore/app/serializers/base.py
muattiyah: Will do, won't `pop()` extract a single field only though?

daviddavis: Ah yea. Here we go:

```
    def validate(self, data):
        unknown_fields = set(self.initial_data.keys()) - set(self.fields.keys())
        if unknown_fields:
            unknown_fields = {field: _('Unexpected field') for field in unknown_fields}
            raise serializers.ValidationError(unknown_fields)
        return data
```

daviddavis:
@muattiyah I lean towards option 2 but either works. 👍

pulpcore/pulpcore/app/serializers/base.py
muattiyah: Perfect.

muattiyah:
@daviddavis Same, will go with 2.

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -76,6 +76,16 @@ def view_name_for_model(model_obj, view_action):
     raise LookupError('view not found')
 
 
+def validate_unknown_fields(initial_data, defined_fields):
+    """
+    This will raise a `ValidationError` if a serializer is passed fields that are unknown.
+    """
+    unknown_fields = set(initial_data.keys()) - set(defined_fields.keys())
```
asmacdo: Prefer: set(initial_data) and set(defined_fields)

muattiyah: Good catch, thanks.

daviddavis:
One thing: your commit message should conform to https://docs.pulpproject.org/dev-guide/contributing/branching.html#commit-messages.

Here's what that would look like:

```
Raise error on undefined parameters

Raise a ValidationError when undefined parameters are passed to an endpoint.

fixes #2970
https://pulp.plan.io/issues/2970
```

muattiyah:
@daviddavis oh, forgot about that. @dralley told me before. Will amend with the tests 👍

pulpcore/tests/serializers/test_base.py
```diff
@@ -1,7 +1,9 @@
 from unittest import TestCase, mock
 
 from pulpcore.app import models
-from pulpcore.app.serializers import base
+from pulpcore.app.serializers import base, validate_unknown_fields
+
+from rest_framework import serializers
```
muattiyah: I should probably place this before the `pulpcore` imports.

daviddavis:
For the validate_unknown_fields, I would probably test:

1. There's an unexpected field and a validation error is raised
2. There's not an unexpected field and no validation error is raised
3. Optional: Test that the initial data does NOT contain a field in defined fields but no validation error is raised

Testing things like empty set for defined fields seem less useful IMO.

muattiyah:
1. Cool, this one is covered.
2. I thought about this while writing the tests, given that the function has no side effects and doesn't return a value, I think I can call the function in a `try` and `except serializers.ValidationError` otherwise. I can also `assert` that the function didn't mutate the dicts. Do you have a better idea?
3. I don't think it's harmful to keep it, no?

daviddavis:
> I thought about this while writing the tests, given that the function has no side effects and doesn't return a value, I think I can call the function in a try and except serializers.ValidationError otherwise. I can also assert that the function didn't mutate the dicts. Do you have a better idea?

Yea something like:

```python
def run_test(self):
    try:
        validate_unknown_fields()
    except serializers.ValidationError:
        self.fail("...")
```

> I don't think it's harmful to keep it, no?

👍

dralley:
It occurs to me that this would potentially be a really cool feature to add to upstream Django Rest Framework, maybe as a ```strict=True``` param on the Serializer.

If you think it would be a cool thing to do, feel free to write an issue, or a PR against DRF.  (Not that you have to though, and short term, definitely +1 on solving it here).

muattiyah:
@dralley good idea. Will jump on their issue tracker tomorrow.

PR #3533, opened 2018-06-22
Adhering to Github Community Standards

CodeHeeler:
Github community standards expect a Code of Conduct, Contributing
guidelines, and a pull request template which have now been added.

fixes #3599
https://pulp.plan.io/issues/3599

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,6 @@
+## Code of Conduct
+
+Please be sure you have read our Code of Conduct here:
+https://pulpproject.org/conduct/
+
+Thank you for keeping our community a welcoming one!
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+## Contributing
+
+We have provided detailed documentation for ways in which you can
+contribute to Pulp here:
+https://docs.pulpproject.org/en/3.0/nightly/contributing/index.html
+
+This documentation includes:
+
+* Suggestions of how to contribute
+* How we track bugs
+* Ways to get in touch with other contributors who can advise you
+* A contribution checklist
+* A developer guide
+
+Join us!  We look forward to hearing from you.
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,23 @@
+### Pull Requests
+
+Please be sure you have read our documentation regarding contributing
+here:
+https://docs.pulpproject.org/en/3.0/nightly/contributing/index.html
+
+There you can read a Contribution Checklist and learn more about
+appropriate branches and rebasing among other things.
+
+Once you are ready to create a pull request:
+
+- [ ] For properly formatting commit messages, please see the
+documentation here:
+https://docs.pulpproject.org/en/3.0/nightly/contributing/git.html#commit-message
+- [ ] Once you have created your pull request, please add a link to
+your pull request into the notes/comments for the relevant bug in the
+issue tracker and change the status to POST.
+- [ ] Continuous integration tests will be automatically triggered
+by your pull request.  You can see the status of these tests on the
+page for your pull request.  If there are failures, please try to
+fix them and update your pull request.
+
+Thank you so much for contributing to Pulp!


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,73 @@
+## Contributor Covenant Code of Conduct
```
daviddavis: Is this the same as the one in PUP-4? If so, thoughts on linking to it instead of having to maintaining two copies?

CONTRIBUTING.md
```diff
@@ -0,0 +1,14 @@
+## Contributing
+
+We have provided detailed documentation for ways in which you can
+contribute to Pulp here: https://docs.pulpproject.org/
```
daviddavis: What about linking directly to our "Contributing" page:

Also, I think we want to link to our 3.0 docs on master (vs our 2.y docs).

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+Please be sure you have read our documentation regarding contributing
+here: https://docs.pulpproject.org/index.html
+
+There you can read a Contribution Checklist and learn more about
+appropriate branches and rebasing among other things.
+
+Once you are ready to create a pull request:
+
+- [ ] If there is a bug for this issue, please title the pull request
+"<bug_id> - Short Message"
```
daviddavis: Hmm we typically don't put the bug id in our PR title.

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] In the commit message or comment section for your pull request,
+please include a link to the issue.
```
daviddavis: For commit message, you might want to link to our guide on how to format commit messages:

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] If there is a Redmine issue associated with the commit, please
+add "closes #<issue number>" in your commit as this will set the
+issue to MODIFIED in the tracker upon merging.
```
daviddavis: Not sure I would mention setting the issue to MODIFIED since upon merging it should automatically happen.

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+### Pull Requests
+
+Please be sure you have read our documentation regarding contributing
+here: https://docs.pulpproject.org/index.html
```
daviddavis: Same issue here as in CONTRIBUTING - it links to Pulp 2 docs.

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] If there is a bug for this issue, please title the pull request
+"<bug_id> - Short Message"
+- [ ] In the commit message or comment section for your pull request,
+please include a link to the issue.
+- [ ] If there is a Redmine issue associated with the commit, please
+add "closes #<issue number>" in your commit as this will set the
+issue to MODIFIED in the tracker upon merging.
+- [ ] Once you have created your pull request, please add a link to
+your pull request into the notes/comments for the relevant bug in the
+issue tracker.
```
daviddavis: Also, the issue should be set to POST status.

CODE_OF_CONDUCT.md
CodeHeeler: This is the one from the PUP, but we should link to a standalone Code of Conduct and not link to the PUP as the PUP is worded as a proposal, and though it has been accepted, it is not clear that it was accepted in this form or that this is final and binding.  Is it anywhere else?  I didn't see it in our docs at all.

CONTRIBUTING.md
CodeHeeler: I really wanted to do this, but I don't want this to constantly end up out of date and we currently have our stable current docs linking to the 2.y docs, and 3.0 is nightly only so I think we need to address that before I link to it here.

pull_request_template.md
CodeHeeler: Agreed, but that is what our docs says to do.  We need to pick one and be consistent.

pull_request_template.md
CodeHeeler: Good idea.

pull_request_template.md
asmacdo: I think thats what our Pulp 2 docs say to do, not Pulp 3.

pull_request_template.md
CodeHeeler: It says that it will be done automatically, it's not an instruction, just information as to why the "closes #" is important.

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+### Pull Requests
+
+Please be sure you have read our documentation regarding contributing
+here: https://docs.pulpproject.org/index.html
+
+There you can read a Contribution Checklist and learn more about
+appropriate branches and rebasing among other things.
+
+Once you are ready to create a pull request:
```
asmacdo: This information is very similar to https://docs.pulpproject.org/en/3.0/nightly/contributing/pull-request-walkthrough.html

It might be best to cut this down a bit, and link to that. Some of the information here is not included in the docs, and could be added there.

pull_request_template.md
CodeHeeler: Good catch.  Happy to update this.  I was using the stable docs to write this, which is evidently out of date.

pull_request_template.md
CodeHeeler: I think we need to decide on a definitive version and then link to it so that we are not having to maintain things in multiple places.  As someone coming to the pulpproject.org website, there is nowhere that it is clear that these instructions differ, and no reason to suspect that you should use anything other than the docs listed as "current and stable".  If we are to use the nightly, then we need to update the current stable information to reflect those choices.

CONTRIBUTING.md
daviddavis: I'm not sure what you mean by "constantly out of date". Can you explain? Regardless though I see less harm in linking to docs that might be a little out of date than linking to our pulp 2 docs.

pull_request_template.md
daviddavis: Ah yes, I misread it. 👍

pull_request_template.md
daviddavis: Where are the other versions?

CODE_OF_CONDUCT.md
daviddavis: We have considered having the PUPs on our website. They are proposals but also living documents and are meant to be the canonical sources of information. What about linking to the PUP with some verbiage to explain that the proposal and how it contains our official CoC?

asmacdo: I think it makes sense to have the CoC as a separate file outside of the PUP. I don't think people reading our code of conduct need to see the proposal, unless they want to.

dralley:
@daviddavis @CodeHeeler what is the status of this PR?

daviddavis:
I believe @CodeHeeler is working on this issue: https://github.com/pulp/pulpproject.org/issues/134. After that is complete, per https://github.com/pulp/pups/blob/master/pup-0004.md#displaying-the-coc, we can update this PR and merge it.

PR #3534, opened 2018-06-25
Support ca_path passed in configuration for http notifier.

jortel:
https://pulp.plan.io/issues/3780


jortel:
> Please update the docs for the REST API: https://docs.pulpproject.org/dev-guide/integration/rest-api/event/crud.html#create-an-event-listener

@dkliban, I deliberately did not add the `ca_path` to the example configuration passed in the create call for these reasons.
- The documentation has a precedent of not including optional configuration fields.  The `username` and `password` are not included.
- The optional `ca_path` is an edge use case and concerned it may confuse users that don't dig deeper.

Thoughts?

jortel:
Added to the 2.17.x release notes.

PR #3535, opened 2018-06-26
Allow users to override page size

daviddavis:
fixes #3790
https://pulp.plan.io/issues/3790

Note: This is up for discussion on pulp-dev. Recommend waiting a bit before merging.

### DIFF
--- a/pulpcore/pulpcore/app/pagination.py
+++ b/pulpcore/pulpcore/app/pagination.py
@@ -13,6 +13,8 @@ class UUIDPagination(pagination.CursorPagination):
 
     """
     ordering = 'id'
+    page_size_query_param = 'page_size'
+    max_page_size = 5000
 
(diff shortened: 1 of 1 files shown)


dralley:
Looks good.  Should we set a max page size?

daviddavis:
Sure, any thoughts on a good number?

dralley:
5000 maybe?  I don't know what the guidelines are for determining that sort of thing.  And maybe we don't need one.

daviddavis:
5000 sounds good to me. @jlsherrill does that meet your requirements?

jlsherrill:
Yeah 5,000 is fine, we usually do it much lower in pulp 2, in the hundreds.  For pulp 3 i imagine we'll do some tests to try to optimize it to some degree.

PR #3536, opened 2018-06-26
 Use a query param approach for multiple serializers

dralley:
Certain models, such as {RPM|Python|Ansible}Content or Task, have a lot of very verbose information that the user doesn't care much about.  We would like to provide some way for a user to filter out the less-useful information so that responses can be easily human-readable.

e.g.

```
[vagrant@pulp3 ~]$ http GET http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/?pretty=True
HTTP/1.1 200 OK
...

{
    "_href": "http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/",
    "created": "2018-06-28T16:13:54.241649Z",
    "finished_at": "2018-06-28T16:13:55.467281Z",
    "id": "edbe9ad0-9787-4d76-97df-1411ff097fa6",
    "parent": null,
    "started_at": "2018-06-28T16:13:54.312332Z",
    "state": "completed",
    "worker": "http://localhost:8000/pulp/api/v3/workers/e340248c-59b1-40ce-9638-d902eaaf6e79/"
}
```

vs.

```
[vagrant@pulp3 ~]$ http GET http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/
HTTP/1.1 200 OK
...
{
    "_href": "http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/",
    "created": "2018-06-28T16:13:54.241649Z",
    "created_resources": [
        "http://localhost:8000/pulp/api/v3/repositories/6a53ecd4-f65d-48d4-89e3-4d83ad1142c9/versions/1/"
    ],
    "error": null,
    "finished_at": "2018-06-28T16:13:55.467281Z",
    "id": "edbe9ad0-9787-4d76-97df-1411ff097fa6",
    "non_fatal_errors": [],
    "parent": null,
    "progress_reports": [
        {
            "done": 3,
            "message": "Add Content",
            "state": "completed",
            "suffix": "",
            "task": "http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/",
            "total": 3
        },
        {
            "done": 0,
            "message": "Remove Content",
            "state": "completed",
            "suffix": "",
            "task": "http://localhost:8000/pulp/api/v3/tasks/edbe9ad0-9787-4d76-97df-1411ff097fa6/",
            "total": 0
        }
    ],
    "spawned_tasks": [],
    "started_at": "2018-06-28T16:13:54.312332Z",
    "state": "completed",
    "worker": "http://localhost:8000/pulp/api/v3/workers/e340248c-59b1-40ce-9638-d902eaaf6e79/"
}
```

The approach currently used is to do this only for viewset "list" actions (multiple records at once), and not for viewset "retrieve" actions (a single specific record).  This is problematic for several reasons:

* API users, such as Katello or the bindings, cannot fetch detailed information en-masse.  Non-human consumers don't care about the verboseness and may need all the information.
* A human user user might want to see a minimal representation of just one model, such as a very large content model, e.g. Python Content [0]
* It's inconsistent, as a human you can't tell whether you're getting all the fields from the list view unless you compare it against a detail view.  Most will provide everything but a few won't.

With this new approach, a human user can provide the ```?pretty=True``` query parameter when making your GET request in order to get the restricted set of information, if the viewset supports it.

For viewsets which don't support minimal serializers, providing this param will do nothing. This means that we can inject it into every request the CLI makes without consequence.  If a minimal serializer is supported, then it will use it, if not, then it will just use the normal serializer.

It works on both "list" actions and "retrieve" actions and is more consistent because you know that unless you explicitly ask for less, you're getting everything.

[0] Another example usage: https://github.com/pulp/pulp_python/pull/156

### DIFF
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -199,34 +199,6 @@ class ModelSerializer(serializers.HyperlinkedModelSerializer):
 
         return path
 
-    def to_representation(self, instance):
-        """
-        Object instance -> Dict of primitive datatypes.
-
-        This is very similar to DRF's default to_representation implementation in
-        ModelSerializer, but checks whether a 'minimal_fields' attribute is defined,
-        and uses it when many objects are being serialized at once (i.e. many=True,
-        when the viewset is performing a 'list' action).
-        """
-        ret = OrderedDict()
-        fields = self._readable_fields
-        if self.parent and self.parent.many and hasattr(self.Meta, 'minimal_fields'):
-            fields = [field for field in fields if field.field_name in self.Meta.minimal_fields]
-
-        for field in fields:
-            try:
-                attribute = field.get_attribute(instance)
-            except SkipField:
-                continue
-
-            check_for_none = attribute.pk if isinstance(attribute, PKOnlyObject) else attribute
-            if check_for_none is None:
-                ret[field.field_name] = None
-            else:
-                ret[field.field_name] = field.to_representation(attribute)
-
-        return ret
-
     def validate(self, data):
         if hasattr(self, 'initial_data'):
             validate_unknown_fields(self.initial_data, self.fields)
@@ -267,24 +239,20 @@ class MasterModelSerializer(ModelSerializer):
     def to_representation(self, instance):
         """
         Represent a cast Detail instance as a dict of primitive datatypes
-
-        This is very similar to DRF's default to_representation implementation in
-        ModelSerializer, but makes sure to cast Detail instances and use the correct
-        serializer for rendering so that all detail fields are included. If a serializer
-        defines a 'minimal_fields' attribute, it will use these fields instead of the full set
-        of detail fields when many objects are being serialized at once (i.e. many=True,
-        when the viewset is performing a 'list' action).
         """
+
+        # This is very similar to DRF's default to_representation implementation in
+        # ModelSerializer, but makes sure to cast Detail instances and use the correct
+        # serializer for rendering so that all detail fields are included.
         ret = OrderedDict()
 
         instance = instance.cast()
         viewset = viewset_for_model(instance)()
-        serializer_class = viewset.get_serializer_class()(context=self._context)
-
-        fields = serializer_class._readable_fields
-        if self.parent and self.parent.many and hasattr(serializer_class.Meta, 'minimal_fields'):
-            fields = [field for field in fields
-                      if field.field_name in serializer_class.Meta.minimal_fields]
+        if self.parent and self.parent.many:
+            viewset.action = 'list'
+        else:
+            viewset.action = 'get'
+        fields = viewset.get_serializer_class()(context=self._context)._readable_fields
 
         for field in fields:
             try:
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -248,10 +248,7 @@ class MasterModelSerializer(ModelSerializer):
 
         instance = instance.cast()
         viewset = viewset_for_model(instance)()
-        if self.parent and self.parent.many:
-            viewset.action = 'list'
-        else:
-            viewset.action = 'get'
+        viewset.request = self._context['request']
         fields = viewset.get_serializer_class()(context=self._context)._readable_fields
 
         for field in fields:
--- a/pulpcore/pulpcore/app/serializers/base.py
+++ b/pulpcore/pulpcore/app/serializers/base.py
@@ -3,6 +3,7 @@ from gettext import gettext as _
 from urllib.parse import urljoin
 
 from django.core.validators import URLValidator
+from drf_queryfields.mixins import QueryFieldsMixin
 
 from rest_framework import serializers
 from rest_framework.fields import SkipField
@@ -105,7 +106,8 @@ class GenericKeyValueRelatedField(serializers.DictField):
         return super().to_representation(value.mapping)
 
 
-class ModelSerializer(serializers.HyperlinkedModelSerializer):
+# Inheritance order matters, don't flip these
+class ModelSerializer(QueryFieldsMixin, serializers.HyperlinkedModelSerializer):
     """Base serializer for use with :class:`pulpcore.app.models.Model`
 
     This ensures that all Serializers provide values for the '_href` field, and
--- a/pulpcore/pulpcore/app/viewsets/base.py
+++ b/pulpcore/pulpcore/app/viewsets/base.py
@@ -84,6 +84,35 @@ class NamedModelViewSet(viewsets.GenericViewSet):
     parent_lookup_kwargs = {}
     schema = DefaultSchema()
 
+    def get_serializer_class(self):
+        """
+        Provides a way to customize the serializer based on the action being taken
+        (e.g. to provide a different serialization for a list view vs. a detail view)
+
+        If you do not specify a 'serializer_class', you must specify a 'serializers' dict which
+        defines a 'default' serializer, and can override that serializer on a per-action basis.
+
+        e.g. serializers = {'default': TaskSerializer, 'list': MinimalTaskSerializer}
+
+        If you define both, the 'serializer_class' attribute will be ignored and the 'serializers'
+        attribute will be used. This is because plugin viewsets may want to use multiple
+        serializers, but they inherit from basic viewsets such as Content which will have a
+        'serializer_class' attribute defined.
+        """
+        serializer_class = getattr(self, 'serializer_class', None)
+        serializers = getattr(self, 'serializers', None)
+
+        msg = _("{} must either have a 'serializer_class' attribute, or it must have a "
+                "'serializers' attribute with a 'default' key set").format(self.__class__.__name__)
+        assert serializer_class or serializers, msg
+
+        if serializers:
+            valid = isinstance(serializers, dict) and 'default' in serializers.keys()
+            assert valid, msg
+            return self.serializers.get(self.action, serializers['default'])
+
+        return serializer_class
+
     @staticmethod
     def get_resource(uri, model):
         """
--- a/pulpcore/pulpcore/app/viewsets/base.py
+++ b/pulpcore/pulpcore/app/viewsets/base.py
@@ -86,32 +86,31 @@ class NamedModelViewSet(viewsets.GenericViewSet):
 
     def get_serializer_class(self):
         """
-        Provides a way to customize the serializer based on the action being taken
-        (e.g. to provide a different serialization for a list view vs. a detail view)
+        Fetch the serializer class to use for the request.
 
-        If you do not specify a 'serializer_class', you must specify a 'serializers' dict which
-        defines a 'default' serializer, and can override that serializer on a per-action basis.
+        The default behavior is to use the "serializer_class" attribute on the viewset.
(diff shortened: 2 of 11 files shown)


pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -81,6 +81,32 @@ class NamedModelViewSet(viewsets.GenericViewSet):
...
+        the most important fields.
+        """
+        assert self.serializer_class is not None, (_(
+            "'{}' should either include a `serializer_class` attribute, or override the "
+            "`get_serializer_class()` method.".format(self.__class__.__name__)
+        ))
+        minimal_serializer_class = getattr(self, 'minimal_serializer_class', None)
+
+        if minimal_serializer_class:
+            if 'pretty' in self.request.query_params:
```
dralley: I like the name "pretty" but apparently it's often used to mean "JSON pretty print" with readable nesting.  Should it be changed?

dralley: The best alternate is 'minimal' I think.

dparalen: how about `core`?

dralley: ```core``` would also be a good choice

dralley:
This doesn't work on any of the content endpoints because they're implemented without viewsets, meaning there's no request passed in, meaning it doesn't know what query parameters were used.  Same applies to the other filters AFAIK.  This is a really annoying limitation.

pulpcore/pulpcore/app/serializers/base.py
```diff
@@ -267,24 +239,17 @@ class Meta:
...
 
         instance = instance.cast()
         viewset = viewset_for_model(instance)()
-        serializer_class = viewset.get_serializer_class()(context=self._context)
-
-        fields = serializer_class._readable_fields
-        if self.parent and self.parent.many and hasattr(serializer_class.Meta, 'minimal_fields'):
-            fields = [field for field in fields
-                      if field.field_name in serializer_class.Meta.minimal_fields]
+        viewset.request = self._context['request']
```
dralley: ^ aforementioned hack.  not that big a deal though, and kind of unavoidable

dparalen: meh not the worst ;)

dparalen:
Briefly tested; seems good:

```
[vagrant@pulp3 pulp]$ http http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/?pretty=True
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Content-Length: 385
Date: Wed, 04 Jul 2018 07:46:24 GMT
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_href": "http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/",
    "created": "2018-07-04T07:46:01.485486Z",
    "finished_at": "2018-07-04T07:46:06.210607Z",
    "id": "f0078367-3689-4d04-80b1-16343211fad7",
    "parent": null,
    "started_at": "2018-07-04T07:46:01.771666Z",
    "state": "completed",
    "worker": "http://localhost:8000/pulp/api/v3/workers/0b004368-6af5-4ad2-8660-03843ada95b0/"
}

[vagrant@pulp3 pulp]$ http http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Content-Length: 911
Date: Wed, 04 Jul 2018 07:46:28 GMT
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_href": "http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/",
    "created": "2018-07-04T07:46:01.485486Z",
    "created_resources": [
        "http://localhost:8000/pulp/api/v3/repositories/e4da1ddd-488c-4e64-a092-d2e1109cc5be/versions/1/"
    ],
    "error": null,
    "finished_at": "2018-07-04T07:46:06.210607Z",
    "id": "f0078367-3689-4d04-80b1-16343211fad7",
    "non_fatal_errors": [],
    "parent": null,
    "progress_reports": [
        {
            "done": 3,
            "message": "Add Content",
            "state": "completed",
            "suffix": "",
            "task": "http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/",
            "total": 3
        },
        {
            "done": 0,
            "message": "Remove Content",
            "state": "completed",
            "suffix": "",
            "task": "http://localhost:8000/pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/",
            "total": 0
        }
    ],
    "spawned_tasks": [],
    "started_at": "2018-07-04T07:46:01.771666Z",
    "state": "completed",
    "worker": "http://localhost:8000/pulp/api/v3/workers/0b004368-6af5-4ad2-8660-03843ada95b0/"
}

[vagrant@pulp3 pulp]$

## Django:
[04/Jul/2018 07:46:01] "POST /pulp/api/v3/remotes/file/9d1ebaa6-5a67-477c-8c57-15d03db4d0ba/sync/ HTTP/1.1" 202 138
[04/Jul/2018 07:46:24] "GET /pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/?pretty=True HTTP/1.1" 200 385
[04/Jul/2018 07:46:28] "GET /pulp/api/v3/tasks/f0078367-3689-4d04-80b1-16343211fad7/ HTTP/1.1" 200 911
```

dparalen:
@dralley,

I've done a brief research and what seems folks would do is to define a `?fields=` query parameter; this IMO makes sense as [what `pretty` is for one might not be so pretty for other](https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#limiting-fields)

The suggested [Stack Overflow approach]( https://stackoverflow.com/questions/23643204/django-rest-framework-dynamically-return-subset-of-fields) probably led someone to implement a [3rd party DRF plugin](https://github.com/wimglenn/djangorestframework-queryfields), one that in addition to specifying a list of wanted field names supports also enumerating not wanted fields: `?fields=id,!_href`

What do you think?

Thanks,
milan

dralley:
@dparalen I think there's still a usecase for ```?pretty=True``` just for user-simplicity's sake, but this implementation might be a better way of accomplishing it (or even both).

dralley:
@dparalen The mixin you linked to seems to work pretty well, so I've added that.  I'll see if the implementation for ```pretty``` can also be simplified.

dralley:
@daviddavis I've changed "pretty" to "minimal"

dparalen:
I see the motivation behind but I'm a bit worried that `minimal` could mean different things in different API endpoints (contexts) and might therefore break the least-surprise principle. Conversely the smallest set of fields is probably just `_id` and `_href` and I'm wondering whether calling it `core` wouldn't make more sense but I'm unsure about the value added.

An interesting filter conflict situation might arise if a plug-in has a `minimal` (or `core`) model field.
A field called `fields` is IMO less likely to cause a conflict.

dralley:
@dparalen I think it's unlikely that a content type would define "minimal" as a field, and probably less likely than "core".  I hear your concerns about introducing another potential name conflict though.  I think there are other options we could use like "human" or "format_min" or "human_summary" or "reduced_fields", etc.

I don't think we'll be able to avoid doing it in general though, because we need some way for the CLI to avoid flooding people with useless information, but I also don't think we want to have that enumeration logic inside the CLI.  Doing it this way makes the most sense I think.  I'm also not sure that manually specifying fields in order not to get overly verbose output is a reasonable thing to ask of a human using the raw API.

> I'm a bit worried that minimal could mean different things in different API endpoints (contexts) and might therefore break the least-surprise principle.

I'm not sure I understand this - it would mean the same thing in every context semantically speaking, except that not every endpoint would implement minimal fields.  But as far as meaning goes, it's always the same - "give me the minimum set of useful fields".  For endpoints where it isn't used like Remote or Distribution, it isn't really needed and also, all of that information is likely useful to the API user.

dparalen:
@dralley I don't disagree ;) I'm +0 on this, `minimal` query parameter.
FWIW `GET@repositories/<repo_uuid>/core/` would be my favorite esp. if we used composition of models rather than inheritance

dralley:
@dparalen would you mind LGTM'ing?  @bmbouter also

dparalen:
@dralley I shall re-test tomorrow and approve

pulpcore/pulpcore/app/viewsets/base.py
```diff
@@ -84,6 +84,35 @@ class NamedModelViewSet(viewsets.GenericViewSet):
     parent_lookup_kwargs = {}
     schema = DefaultSchema()
 
+    def get_serializer_class(self):
+        """
+        Provides a way to customize the serializer based on the action being taken
+        (e.g. to provide a different serialization for a list view vs. a detail view)
```
dparalen: nit: imo no need to put it in ( )

dparalen:
So the `?minimal=True` works as expected, enumerating fields does as well `?fields=id,finished_at` but what doesn't seem to work is excluding some field(s):
```
(pulp) [vagrant@pulp3 pulp]$ http http://localhost:8000/pulp/api/v3/tasks/c1f9eecb-2957-40f1-a17c-431cc5fbbfc2/?fields=\!id
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Content-Length: 2
Date: Tue, 17 Jul 2018 17:27:26 GMT
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{}
(pulp) [vagrant@pulp3 pulp]$
```
But maybe I misunderstand how to use those or it might be a bug in the `djangorestframework-queryfields` package.

dparalen:
OK LGTM; my bad, this works: `?fields\!=id,finished_at`:
```
pulp) [vagrant@pulp3 pulp]$ http http://localhost:8000/pulp/api/v3/tasks/c1f9eecb-2957-40f1-a17c-431cc5fbbfc2/?fields\!=id,finished_at
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Content-Length: 292
Date: Tue, 17 Jul 2018 17:43:20 GMT
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_href": "http://localhost:8000/pulp/api/v3/tasks/c1f9eecb-2957-40f1-a17c-431cc5fbbfc2/",
    "created": "2018-07-17T17:23:48.317889Z",
    "created_resources": [],
    "error": null,
    "non_fatal_errors": [],
    "parent": null,
    "progress_reports": [],
    "spawned_tasks": [],
    "started_at": null,
    "state": "waiting",
    "worker": null
}
```

PR #3540, opened 2018-07-05
Fix task assigned to the resource manager not cancel when it stops.

hao-yu:
When stopping resource manager worker, tasks that assigned to the
resource manager will not be cancelled and leaving the tasks in
"Waiting" state forever. This is because the "worker_name" of
the TaskStatus is not set during the creation and Pulp is unable
to find the tasks for the resource manager during the worker
cleanup.

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -8,6 +8,7 @@ import signal
 import time
 import traceback
 import uuid
+import platform
 
 from bson.json_util import dumps as bson_dumps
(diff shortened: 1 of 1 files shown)


dralley:
Could you try to reproduce this issue with a newer Pulp and Celery version as mentioned [here](https://pulp.plan.io/issues/3817#note-2)?  If it still occurs, we can definitely re-open this, but since @bmbouter was unable to reproduce it with newer code, best holding back until it can be re-confirmed.

PR #3542, opened 2018-07-06
Handle systems without fips_enabled in /proc

ekohl:
When /proc/sys/crypto/fips_enabled doesn't exist, the pulp-gen-ca-certificate script fails and exits with code 1 because of set -e. By first checking if it exists we avoid this.


ekohl:
FYI: our integration tests in puppet-pulp are failing because of this (https://travis-ci.org/theforeman/puppet-pulp/jobs/401000310).

PR #3543, opened 2018-07-06
Handle systems without fips_enabled in /proc

ekohl:
When /proc/sys/crypto/fips_enabled doesn't exist, the pulp-gen-ca-certificate script fails and exits with code 1 because of set -e. By first checking if it exists we avoid this.

Fixes #3828

This is the same patch as https://github.com/pulp/pulp/pull/3542 but against 2-master.


daviddavis:
ok test

PR #3546, opened 2018-07-09
Adds nightly doc building to travis

dkliban:
The 2-master branch will be setup to build using cron on Travis. When this build is trigerred, the
docs for branches defined in .travis/2-master.yaml will be built and published.

re: #3722
https://pulp.plan.io/issues/3722


.travis.yml
```diff
@@ -1,19 +1,27 @@
 language: python
 python:
-    - "2.7"
+- '3.6'
```
daviddavis: Minor grip but typically arrays in yaml are indented.

.travis/docs/all_content_index.rst
```diff
@@ -0,0 +1,13 @@
+Documentation Index
```
daviddavis: I feel like these files aren't necessarily specific to Travis? Thoughts on maybe moving these to the top level docs directory?

.travis.yml
daviddavis: In the files below (e.g. `.travis/2-master.yaml`) you're indenting array entries.

.travis/docs/all_content_index.rst
dkliban: They are specific to the way we build docs for docs.pulpproject.org. You can build the docs in this repo by themselves and they will not include plugin information.

PR #3547, opened 2018-07-09
Update installation docs

dralley:
closes #3786
https://pulp.plan.io/issues/3786

I went ahead and removed Fedora 23/24 references because I don't know that python2-qpid is a correct package name on Fedora 23, and in any case those are long EOL'd


dralley:
ok test

PR #3548, opened 2018-07-09
Using querysets for add/remove_content methods

daviddavis:
fixes #3814
https://pulp.plan.io/issues/3814

### DIFF
--- a/plugin/pulpcore/plugin/changeset/main.py
+++ b/plugin/pulpcore/plugin/changeset/main.py
@@ -6,7 +6,7 @@ from logging import getLogger
 
 from django.db import transaction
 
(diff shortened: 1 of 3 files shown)


dralley:
@bmbouter, @daviddavis I know this is something we don't really have bandwidth for, but it would be great to have a set of performance tracking tests to see the impact of changes like this, and future ones.  I'm not sure how difficult that would be to do.

bmbouter:
@dralley  I'm a +1 on tracking that how our code changes affect the code performance over time. We have a [performance tracking plan](https://pulp.plan.io/projects/pulp/wiki/Performance_Test_Plan) written up with some ideas. Overall I think we know what to track and how to do it, so we're blocked on time currently.

PR #3555, opened 2018-07-19
Makes the Redis string connection in server.yaml prevails for resource_manager and workers

tchellomello:
Make sure the REDIS connection string present in server.yaml is inherited by workers and resource_manager.

fixes #3861
https://pulp.plan.io/issues/3861


tchellomello:
**cc** @dkliban please let me know your thoughts about this one as we talked about this issue earlier.

omaciel:
@tchellomello how about adding some unittests? :)

tchellomello:
@omaciel let's do it :+1:

PR #3556, opened 2018-07-19
Fix a bunch of bad references in ReST docs

dralley:
Also did a little bit of formatting - because it didn't conform to what the docs said it should be

### DIFF
--- a/pulpcore/pulpcore/app/models/__init__.py
+++ b/pulpcore/pulpcore/app/models/__init__.py
@@ -2,8 +2,14 @@
 
 from .auth import User  # noqa
 from .base import Model, MasterModel  # noqa
-from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyValueRelation,  # noqa
-                      GenericKeyValueModel, Config, Notes)  # noqa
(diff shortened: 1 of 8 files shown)


pulpcore/pulpcore/app/models/__init__.py
```diff
@@ -2,8 +2,14 @@
 
 from .auth import User  # noqa
 from .base import Model, MasterModel  # noqa
-from .generic import (GenericRelationModel, GenericKeyValueManager, GenericKeyValueRelation,  # noqa
-                      GenericKeyValueModel, Config, Notes)  # noqa
+from .generic import (  # noqa
+    GenericRelationModel,
```
asmacdo: Since you are reformatting this, mind doing alphabetical order?

dralley: I think it's this way due to circular imports

asmacdo: Cool. If you are sure that this would cause circular imports, please add a comment. Don't worry about it if you aren't, didn't mean to scope creep you.

PR #3559, opened 2018-07-23
Introducing asyncio stages

bmbouter:
This PR pairs with pulp_file changes here: https://github.com/pulp/pulp_file/pull/102

I use the test script below. It resets my Pulp3 installation, sync pulp_file, upload another file, associate with the repo, and then resync again with mirror='true'.

It's expected to make 3 versions, and Pulp should have 4 content units in it.
v1 - 3 content units matching the remote pulp_file repo
v2 - 4 content units, included the newly uploaded one
v3 - 3 content units, after the sync with mirror='true'. This verifies that mirror is working.

```
set -v 

# Reset everything

read -p "Stop the webserver, then press enter"

sudo systemctl stop pulp_resource_manager
sudo systemctl stop pulp_worker@1
sudo systemctl stop pulp_worker@2

pulp-manager reset_db --noinput
pulp-manager migrate
pulp-manager reset-admin-password --password admin
rm -rf /var/lib/pulp/artifact/*

sudo systemctl start pulp_resource_manager
sudo systemctl start pulp_worker@1
sudo systemctl start pulp_worker@2

read -p "Start the webserver, then press enter"

# Create a repository, 'foo'

http POST http://localhost:8000/pulp/api/v3/repositories/ name=foo

export REPO_HREF=$(http :8000/pulp/api/v3/repositories/ | jq -r '.results[] | select(.name == "foo") | ._href')

# Create a remote, 'bar'

http POST http://localhost:8000/pulp/api/v3/remotes/file/ name='bar' url='https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/PULP_MANIFEST'

export REMOTE_HREF=$(http :8000/pulp/api/v3/remotes/file/ | jq -r '.results[] | select(.name == "bar") | ._href')

# Sync 'foo' using 'bar'

http POST $REMOTE_HREF'sync/' repository=$REPO_HREF

# Upload an artifact

http --form POST http://localhost:8000/pulp/api/v3/artifacts/ file@/home/vagrant/devel/foo.tar.gz

export ARTIFACT_HREF=$(http http://localhost:8000/pulp/api/v3/artifacts/?sha256=2a123bec2a2e1df9cd8705861f61cb6df5453f7eaab424f1984658152211e4a5 | jq -r '.results[] | ._href')

http POST http://localhost:8000/pulp/api/v3/content/file/files/ relative_path=foo.tar.gz artifact="$ARTIFACT_HREF"

export CONTENT_HREF=$(http :8000/pulp/api/v3/content/file/files/ | jq -r '.results[] | select(.relative_path == "foo.tar.gz") | ._href')

http POST $REPO_HREF'versions/' add_content_units:="[\"$CONTENT_HREF\"]"

# Sync 'foo' using 'bar'

http POST $REMOTE_HREF'sync/' repository=$REPO_HREF
```


pulpcore/pulpcore/app/models/content.py
```diff
@@ -138,6 +138,15 @@ def natural_key(self):
         """
         return tuple(getattr(self, f) for f in self.natural_key_fields())
 
+    def natural_key_dict(self):
```
jortel: This should be simplified to:
```
return {f: getattr(self, f) for f in self.natural_key_fields()}
```
This is simple enough for the caller to just do.  I don't think this (1-line) method adds value and should not be added.

bmbouter: The use case I think about with this is plugin writers wanting to query for a unit they have in memory using `filter()` or `Q()`. There are two places I needed to do this for in DeclarativeVersion. I wanted to state it's purpose and see what you thought.

plugin/pulpcore/plugin/stages/models.py
```diff
@@ -0,0 +1,66 @@
...
+class DeclarativeArtifact:
+    """
+    Relates an Artifact, how to download it, and its relative_path for publishing.
+
+    This is used by the Stages API stages to determine if an Artifact is already present and ensure
+    Pulp can download it in the future. The `artifact` can be either saved or unsaved. If unsaved,
+    the `artifact` attributes may be incomplete because not all digest information can be computed
+    until the Artifact is downloaded.
+
+    Attributes:
```
dkliban: This is causing the generated docs to not call these parameters, but variables instead. It renders weird: 
declarativeartifactdocs

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,252 @@
...
+
+class artifact_downloader:
+    """
+    An object containing a Stages API stage to download missing Artifacts, but don't call save()
+
+    The actual stage is the `stage` attribute which can be used as follows:
+
+    >>> artifact_downloader(max_concurrent_downloads=42).stage  # This is the real stage
+
+    in_q data type: A `~pulpcore.plugin.stages.DeclarativeContent` with potentially files missing
```
dkliban: The references to other docs are not resolving correctly in the built docs. 
artifact_downloader_docs

bmbouter: This problem is all over the docs. I'm going through and fixing all of them with the next push.

plugin/pulpcore/plugin/stages/models.py
bmbouter: I'm fixing this in the next push by using Sphinx's `:no_members:`

pulpcore/pulpcore/app/models/content.py
jortel: That is the use case I'd imaged as well so my opinion hasn't changed.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+            if content is None:
+                continue
+            await out_q.put(content)
+        declarative_content = []
+        if shutdown:
+            break
+    await out_q.put(None)
+
+
+class artifact_downloader:
```
jortel: Class name should be CapWords - PEP8.

[1] https://www.python.org/dev/peps/pep-0008/#class-names

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+        try:
+            content = in_q.get_nowait()
+        except asyncio.QueueEmpty:
+            if not declarative_content:
+                content = await in_q.get()
+                declarative_content.append(content)
+                continue
+        else:
+            declarative_content.append(content)
+            continue
```
jortel: This *batching* logic seems to rely on a momentary gap in the `in_q` to determine the batch size.  If the `in_q` is being fed faster than consumed here, won't that result in all of the content being in memory at once.  Conversely, what if the `in_q` is being fed in slow, small bursts.  Won't that result in really small batch sizes and poor performance?

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+    Args:
+        in_q (:class:`asyncio.Queue`): The queue to receive
+            :class:`~pulpcore.plugin.stages.DeclarativeContent` objects from.
+        out_q (:class:`asyncio.Queue`): The queue to put
+            :class:`~pulpcore.plugin.stages.DeclarativeContent` into.
+
+    Returns:
+        The query_existing_artifacts stage as a coroutine to be included in a pipeline.
+    """
+    declarative_content = []
```
jortel: This name of `declarative_content` is a bit confusing.  I seems to be a *batch* of `DeclarativeContent`.  Having the *type* in the name implies that other variables like `content` are not `DeclarativeContent` and less obvious as to what the list is for.

Recommend renaming to: `batch` or something.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+                shutdown = True
+                continue
+
+            for declarative_artifact in content.d_artifacts:
+                one_artifact_q = Q()
+                for digest_name in declarative_artifact.artifact.DIGEST_FIELDS:
+                    digest_value = getattr(declarative_artifact.artifact, digest_name)
+                    if digest_value:
+                        key = {digest_name: digest_value}
+                        one_artifact_q &= Q(**key)
```
jortel: The artifact is queried using AND of all known digests.  Why not query on the first (non-None)?  This implies you don't trust any single digest, right?  What if *some* of them match?  I know this would indicate something bad but wondering how this would play out.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+            continue
+
+        all_artifacts_q = Q(pk=None)
+        for content in declarative_content:
+            if content is None:
+                shutdown = True
+                continue
+
+            for declarative_artifact in content.d_artifacts:
+                one_artifact_q = Q()
```
jortel: This should be:
```
Q(pk=None)
```
to ensure you don't match everything in the case where a `DeclarativeArtifact` is queued with none of the digests set.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
+import asyncio
+
+from django.core.files import File
+from django.core.files.storage import default_storage
+from django.db.models import Q
+
+from pulpcore.plugin.models import Artifact, ProgressBar
+
+
+async def query_existing_artifacts(in_q, out_q):
```
jortel: This function contains multiple code blocks each doing something complex.  Recommend decomposing into smaller functions each doing *one* thing.  Better functional decomposition will be more readable, testable and maintainable.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+            run. Default is 100.
+
+    Returns:
+        An object containing the artifact_downloader stage to be included in a pipeline.
+    """
+
+    def __init__(self, max_concurrent_downloads=100):
+        self.max_concurrent_downloads = max_concurrent_downloads
+
+    async def stage(self, in_q, out_q):
```
jortel: Since stages can be any *callable* that is a *cooroutine*, perhaps stages that are classes should just implement `__call__()`?  The *Command* design pattern seems to be a good match.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+                        incoming_content.append(content)
+                        continue
+
+                for i, content in enumerate(incoming_content):
+                    if content is None:
+                        shutdown = True
+                        continue
+                    downloaders_for_content = []
+                    for declarative_artifact in content.d_artifacts:
+                        if declarative_artifact.artifact._state.adding:
```
jortel: Should not be accessing protected `_state` member.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+                                declarative_artifact.url
+                            )
+                            next_future = asyncio.ensure_future(downloader.run())
+                            downloaders_for_content.append(next_future)
+                    if not downloaders_for_content:
+                        await out_q.put(content)
+                        continue
+
+                    async def return_content_for_downloader(c):
+                        return c
```
jortel: I don't see the point of using this closure since it will immediately return `c`.  Can you help me understand this?

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+
+        """
+        pending = set()
+        incoming_content = []
+        outstanding_downloads = 0
+        shutdown = False
+        saturated = False
+        with ProgressBar(message='Downloading Artifacts') as pb:
+            while True:
+                if not saturated:
```
jortel: This *batching* logic seems to rely on a momentary gap in the `in_q` to determine the batch size.  If the `in_q` is being fed faster than consumed here, won't that result in all of the content being in memory at once.  Conversely, what if the `in_q` is being fed in slow, small bursts.  Won't that result in really small batch sizes and poor performance?

The `saturated` flag is set true later in the loop so don't think it will prevent this.

plugin/pulpcore/plugin/stages/artifact_stages.py
jortel: This method contains multiple code blocks each doing something complex.  Recommend decomposing into smaller methods each doing *one* thing.  Better functional decomposition will be more readable, testable and maintainable.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+                            content = results[-1]
+                            for declarative_artifact in content.d_artifacts:
+                                if declarative_artifact.artifact._state.adding:
+                                    new_artifact = Artifact(
+                                        **download_result.artifact_attributes,
+                                        file=download_result.path
+                                    )
+                                    declarative_artifact.artifact = new_artifact
+                            pb.done = pb.done + len(content.d_artifacts)
+                            outstanding_downloads = outstanding_downloads - len(content.d_artifacts)
```
jortel: Re:
```
 outstanding_downloads = outstanding_downloads - len(content.d_artifacts)
```
This seems to assume that all artifacts needed to be downloaded and had a downloader in `downloaders_for_content` which seems incorrect.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+        try:
+            content = in_q.get_nowait()
+        except asyncio.QueueEmpty:
+            if not declarative_content_list:
+                content = await in_q.get()
+                declarative_content_list.append(content)
+                continue
+        else:
+            declarative_content_list.append(content)
+            continue
```
jortel: This *batching* logic seems to rely on a momentary gap in the `in_q` to determine the batch size.  If the `in_q` is being fed faster than consumed here, won't that result in all of the content being in memory at once.  Conversely, what if the `in_q` is being fed in slow, small bursts.  Won't that result in really small batch sizes and poor performance?

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: This is going to make some stages lowercase function names and some CapWords. That would be like:

```
stages = [
    self.first_stage.gen_declarative_content,
    query_existing_artifacts, ArtifactDownloader().stage, artifact_saver, 
    query_existing_content_units, content_unit_saver,
    ContentUnitAssociation(new_version).stage
]
```

Do you think we should stick to PEP8 in this case still?

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+        in_q (:class:`asyncio.Queue`): The queue to receive
+            :class:`~pulpcore.plugin.stages.DeclarativeContent` objects from.
+        out_q (:class:`asyncio.Queue`): The queue to put
+            :class:`~pulpcore.plugin.stages.DeclarativeContent` into.
+
+    Returns:
+        The artifact_saver stage as a coroutine to be included in a pipeline.
+    """
+    shutdown = False
+    declarative_content_list = []
```
jortel: This name of `declarative_content` is a bit confusing.  I seems to be a *batch* of `DeclarativeContent`.  Having the *type* in the name implies that other variables like `content` are not `DeclarativeContent` and less obvious as to what the list is for.

Recommend renaming to: `batch` or something.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: It's possible that it could have small batch sizes, but it's also possible that adding delays could produce poor performance. I think we just don't know without data. Overall I think we need some performance tuning, but we need some fixture data to benchmark it against so we can see how our changes are affecting it. I think we should do these kinds of things later. The great thing is that each stage can be tuned and benchmarked individually.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: batch sounds like a fine name. I will look into updating this name throughout the code.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,274 @@
...
+            if declarative_content is None:
+                shutdown = True
+                break
+            for declarative_artifact in declarative_content.d_artifacts:
+                if declarative_artifact.artifact.pk is None:
+                    src_path = str(declarative_artifact.artifact.file)
+                    dst_path = declarative_artifact.artifact.storage_path(None)
+                    with open(src_path, mode='rb') as input_file:
+                        django_file_obj = File(input_file)
+                        default_storage.save(dst_path, django_file_obj)
```
jortel: I don't think there is anything in the django `Storage` interface that requires the implementation be stateless and thus no guarantee.  Using the global singleton `default_storage` could be dangerous.  Recommend using (instantiating) `DefaultStorage` instead.

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,158 @@
+import asyncio
+from collections import defaultdict
+
+from django.db.models import Q
+
+from pulpcore.plugin.models import ProgressBar
+
+
+class content_unit_association:
```
jortel: Class name should be CapWords - PEP8.

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,158 @@
...
+                try:
+                    declarative_content = in_q.get_nowait()
+                except asyncio.QueueEmpty:
+                    if not declarative_content_list and not shutdown:
+                        declarative_content = await in_q.get()
+                        declarative_content_list.append(declarative_content)
+                        continue
+                else:
+                    declarative_content_list.append(declarative_content)
+                    continue
```
jortel: Same concern here ^ about batching algorithm.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: good call. will fix in next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I overall agree, but I don't want to do big code rearranging now for a few reasons. I do agree though that the internals of the stages could be improved.

1) the code has undergone a lot of hand testing. If the stages get a decomposing refactor then it all has to be re hand tested. That is a lot of extra effort that I think we would do better to spend in other areas of Pulp3.

2) We should do refactoring after we have tests to ensure we don't regress and we don't spend a lot of time hand testing code over and over. I want to help write those tests even, but I can't write them right now.

3) Other PRs with significant diffs on top of this PR are already underway, especially for https://pulp.plan.io/issues/3812 so if this code starts changing a lot there will be a lot of conflicts to resolve.

Can we make these improvements (that I agree with) later?

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,158 @@
...
+                    for unit_key in self.unit_keys_by_type[unit_type]:
+                        query_dict = {}
+                        for i, key_name in enumerate(unit_type.natural_key_fields()):
+                            query_dict[key_name] = unit_key[i]
+                        units_to_unassociate |= Q(**query_dict)
+                    await out_q.put(unit_type.objects.filter(units_to_unassociate))
+            await out_q.put(None)
+
+
+class content_unit_unassociation:
```
jortel: Class name should be CapWords - PEP8.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I had also thought about that, but I decided it was kind of confusing. Unlike most objects the whole point of this object is to get the coroutine so you probably will chain together the call. When you do that w/ `__call__()` to get the coroutine you call: `artifact_downloader()()`. I think the `artifact_downloader().stage()` provides a more explicit experience for the plugin writer.

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Returns:
+        An object containing the configured content_unit_unassociation stage to be included in a
+        pipeline.
+    """
+
+    def __init__(self, new_version):
+        self.new_version = new_version
+
+    async def stage(self, in_q, out_q):
```
jortel: Same recommendation here for __call__()

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: yup this isn't right. will fix in next push. Thanks!

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,176 @@
...
+        try:
+            content = in_q.get_nowait()
+        except asyncio.QueueEmpty:
+            if not declarative_content_list:
+                content = await in_q.get()
+                declarative_content_list.append(content)
+                continue
+        else:
+            declarative_content_list.append(content)
+            continue
```
jortel: Same question/concern about batching algorithm.

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,176 @@
...
+        try:
+            declarative_content = in_q.get_nowait()
+        except asyncio.QueueEmpty:
+            if not declarative_content_list and not shutdown:
+                declarative_content = await in_q.get()
+                declarative_content_list.append(declarative_content)
+                continue
+        else:
+            declarative_content_list.append(declarative_content)
+            continue
```
jortel: Same question/concern about batching algorithm.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: Yes, it looks a little strange but it makes the code elagent. This closure has no purpose except to maintain a reference to c and then return it so that the closure completes. That reference is then returned in the list of returned results as part of `gather(...)`. This allows the content unit and specifically its artifacts to all come back together after all the artifacts are downloaded. That reference to the content unit is stripped out of the results [here](https://github.com/pulp/pulp/pull/3559/files#diff-026c51b6860b1ee72f80e464a00a2228R190)

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: re batching logic: we should tune the stages later with data. I have a similar answer [here](https://github.com/pulp/pulp/pull/3559#discussion_r206942577).

The saturated feature is for limiting the number of downloaders running. It's possible that a greater number of things to download could be drained from the queue already at which point on the next loop run nothing is read from the queue because it's saturated.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I agree, but I can't do that now. See reasons [here](https://github.com/pulp/pulp/pull/3559#discussion_r206947935).

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,176 @@
...
+        remote_artifact_bulk = []
+        remote_artifact_map = {}
+
+        with transaction.atomic():
+            for declarative_content in declarative_content_list:
+                if declarative_content is None:
+                    shutdown = True
+                    continue
+                if declarative_content.content.pk is None:
+                        declarative_content.content.save()
```
jortel: So, `Content` is not saved in bulk?  Seems like an opportunity for a big performance improvement here.

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,176 @@
...
+                                'md5': declarative_artifact.artifact.md5,
+                                'sha1': declarative_artifact.artifact.sha1,
+                                'sha224': declarative_artifact.artifact.sha224,
+                                'sha256': declarative_artifact.artifact.sha256,
+                                'sha384': declarative_artifact.artifact.sha384,
+                                'sha512': declarative_artifact.artifact.sha512,
+                                'remote': declarative_artifact.remote,
+                            }
+                            rel_path = content_artifact.relative_path
+                            remote_artifact_map[rel_path] = remote_artifact_data
```
jortel: The `remote_artifact_map` *dict* seems to be shared across `Content` but *keyed* by artifact relative path.  What if two different artifacts in two different content has the same relative path?

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,176 @@
...
+                    declarative_content.content = result
+        for declarative_content in declarative_content_list:
+            await out_q.put(declarative_content)
+        declarative_content_list = []
+        if shutdown:
+            break
+    await out_q.put(None)
+
+
+async def content_unit_saver(in_q, out_q):
```
jortel: This function contains multiple code blocks each doing something complex.  Recommend decomposing into smaller functions each doing *one* thing.  Better functional decomposition will be more readable, testable and maintainable.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -0,0 +1,146 @@
...
+from pulpcore.plugin.models import RepositoryVersion
+from pulpcore.plugin.tasking import WorkingDirectory
+
+from .api import create_pipeline, end_stage
+from .artifact_stages import artifact_downloader, artifact_saver, query_existing_artifacts
+from .association_stages import content_unit_association, content_unit_unassociation
+from .content_unit_stages import content_unit_saver, query_existing_content_units
+
+
+class FirstStage:
```
jortel: Can you help me understand the value of this interface?

By adopting the *Command* design pattern recommendation and requiring each stage to simply be a *callable* by implementing `__call__()` this interface could no longer be necessary.  Stages being an object is useful in that it supports separate stage construction/initialization and usage.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -0,0 +1,146 @@
...
+        Returns:
+            A Stages API compatible coroutine for
+            :class:`~pulpcore.plugin.stages.DeclarativeVersion` to use as the first stage.
+        """
+        raise NotImplementedError('A plugin writer needs to implement this')
+
+
+class DeclarativeVersion:
+
+    def __init__(self, first_stage, repository, sync_mode='mirror'):
```
jortel: How would I use this with a customized stage pipeline?  Would you consider supporting the plugin writer passing in the pipeline?

Instead of interpreting *string* mode names, I think a (bool) mirror=True argument would be better.  This object can only handle two choices.

plugin/pulpcore/plugin/stages/models.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, artifact=None, url=None, relative_path=None, remote=None):
+        if not url:
+            raise ValueError(_("DeclarativeArtifact must have a 'url'"))
+        if not relative_path:
+            raise ValueError(_("DeclarativeArtifact must have a 'relative_path'"))
+        if not remote:
+            raise ValueError(_("DeclarativeArtifact must have a 'remote'"))
+        if not artifact:
+            raise ValueError(_("DeclarativeArtifact must have a 'artifact'"))
```
jortel: would `assert` be more appropriate?

plugin/pulpcore/plugin/stages/models.py
```diff
@@ -0,0 +1,75 @@
...
+    :class:`~pulpcore.plugin.models.Artifact` is downloaded.
+
+    Attributes:
+        artifact (:class:`~pulpcore.plugin.models.Artifact`): An
+            :class:`~pulpcore.plugin.models.Artifact` either saved or unsaved. If unsaved, it
+            may have partial digest information attached to it.
+        url (str): the url to fetch the :class:`~pulpcore.plugin.models.Artifact` from.
+        relative_path (str): the relative_path this :class:`~pulpcore.plugin.models.Artifact`
+            should be published at for any Publication.
+        remote (:class:`~pulpcore.plugin.models.Remote`): The remote used to fetch this
```
jortel: Just curious: is the `:class:` necessary?  The docs seems to generate fine without it.

plugin/pulpcore/plugin/stages/models.py
```diff
@@ -0,0 +1,75 @@
...
+            :class:`~pulpcore.plugin.models.Artifact`.
+
+    Raises:
+        ValueError: If `artifact`, `url`, `relative_path`, or `remote` are not specified.
+    """
+
+    __slots__ = ('artifact', 'url', 'relative_path', 'remote')
```
jortel: +1 using `__slots__`

plugin/pulpcore/plugin/stages/models.py
```diff
@@ -0,0 +1,75 @@
...
+        ValueError: If `content` is not specified.
+    """
+
+    __slots__ = ('content', 'd_artifacts')
+
+    def __init__(self, content=None, d_artifacts=None):
+        if not content:
+            raise ValueError(_("DeclarativeContent must have a 'content'"))
+        if d_artifacts:
+            self.d_artifacts = d_artifacts
```
jortel: Suggest: 
```
self.d_artifacts = d_artifacts or []
```
would be simpler.

plugin/pulpcore/plugin/stages/api.py
```diff
@@ -0,0 +1,48 @@
...
+    >>>         await out_q.put(item)
+    >>>     await out_q.put(None)  # this stage is shutdown so send 'None'
+
+    Args:
+        stages (list of coroutines): A list of Stages API compatible coroutines.
+        maxsize (int): The maximum amount of items a queue between two stages should hold. Optional
+            and defaults to 100.
+
+    Returns:
+        A single coroutine that can be used to run, wait, or cancel the entire pipeline with.
```
jortel: This probably should document it returns an `asyncio.Future`.  (I think)

jortel:
@bmbouter, Review finished.  A few of the comments ended up being repetitive, sorry.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: Oh yeah I see what you are saying. I will fix this in the next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I agree but we should tune the stages later with data. I have a similar answer [here](https://github.com/pulp/pulp/pull/3559#discussion_r206942577).

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I can rename to batch in the next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I think this is a supported approach. I got it from the django docs here. https://docs.djangoproject.com/en/2.0/topics/files/#storage-objects

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: Same question as [here](https://github.com/pulp/pulp/pull/3559#discussion_r206939231).

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: I should outline the performance measurement approach I thought up. We really need data to know if we need to make changes.

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    def __init__(self, new_version):
+        self.new_version = new_version
+        self.unit_keys_by_type = defaultdict(set)
+        for unit in self.new_version.content.all():
+            unit = unit.cast()
+            self.unit_keys_by_type[type(unit)].add(unit.natural_key())
+
+    async def stage(self, in_q, out_q):
```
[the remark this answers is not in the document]
bmbouter: Same reasoning as [above](https://github.com/pulp/pulp/pull/3559#discussion_r206948991) to leave it as is.

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: Same question as [here](https://github.com/pulp/pulp/pull/3559#discussion_r206939231).

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: same answer as [above](https://github.com/pulp/pulp/pull/3559#discussion_r206948991) to leave it as is.

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: Same answer as [above](https://github.com/pulp/pulp/pull/3559#discussion_r206990521).

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: Same answer as [above](https://github.com/pulp/pulp/pull/3559#discussion_r206990521).

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: Yeah `Content` currently is not compatible with bulk save :( It throws an error because bulk_save() and multi-table inheritance are incompatible. This is actually the main thing I'm working on now: https://pulp.plan.io/issues/3812

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: I see what you're saying for Pulp generally, but for the pipelines, I don't think that is going to happen because it's an invalid repo. For that to happen in the stream, it would effectively be wanting to include two Artifacts with the same relative path in one repo which would fail in various places anyway.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: re: the string mode names. I see what you are saying, but what if we have more modes in the future. What if we had a retention behavior for example. If we adopt booleans here then we have to add more parameters later for additional modes. That's easy enough, but then we have to make choices about how those various options work together. If we leave them as strings like this we can avoid that problem which I think is good. It will be simpler because we can add more modes without changing the interface and having to worry about feature interactions.

plugin/pulpcore/plugin/stages/models.py
bmbouter: I got this from your code!

plugin/pulpcore/plugin/stages/models.py
bmbouter: It won't be clickable without this label. All of these class renders are clickable.

plugin/pulpcore/plugin/stages/models.py
bmbouter: I could see using `assert`, but when I was reading through the standard library docs the [ValueError description](https://docs.python.org/3/library/exceptions.html#ValueError) jumped out at me as describing this error situation perfectly.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: This interface came out of [Simon's feedback](https://github.com/pulp/pulp_file/commit/c23bec069e12a485000f048e486e1d89cce65beb#r29701942). For all the reasons he says, the plugin writer will need to make an object like this one. I think the value of providing it here is that we can clearly prompt the plugin writer to build the right thing because we can say they need to subclass this.

The only other option would be for them to make the object just like this on its own, and pass the function that returns the co-routine. This is tricky because if you call `MyPluginFirstStage(myarg=True).my_coroutine()` it won't work with `create_pipeline()` but if you call `MyPluginFirstStage(myarg=True).my_coroutine` it will. I don't think this is very intuative to people.

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: I agree, but I hope we can do this later. See similar answer [here](https://github.com/pulp/pulp/pull/3559#discussion_r206947935).

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: re: customizing the stage pipeline, The only way I can even think it would work is having a hook called `customize_stages` that is called that a subclassed version of DeclarativeVersion could implement. It would be called just before `pipeline = create_pipeline(stages)` [here](https://github.com/pulp/pulp/pull/3559/files/05bb5f86e1119f017025dfc00e8cb57b96d77565#diff-74a14c704caf7207fa01aaca7419b5feR145).

This means that the order of the stages as provided by `DeclarativeVersion` is definitely part of the plugin API's semantic versioning claim. Plugin code is probably going to add/remove stages based on it being the second or 4th stage so we can't be adding/removing/changing the stages. I'm ok w/ that and it's probably what we should be doing anyway, but I wanted to state it.

I think I'm going to implement this design with the next push.

plugin/pulpcore/plugin/stages/models.py
bmbouter: cool will include in next push.

plugin/pulpcore/plugin/stages/api.py
bmbouter: At this point it's only returning a `coroutine`. I looked into sphinx's support for indicating something is a coroutine, but they aren't there yet unfortunately. https://github.com/sphinx-doc/sphinx/issues/4777

plugin/pulpcore/plugin/stages/artifact_stages.py
jortel: Yes.

bmbouter: ok I'll port this w/ my next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
jortel: Yes.

plugin/pulpcore/plugin/stages/artifact_stages.py
jortel: Both approaches are supported but I believe constructing the `DefaultStorage` is safer for the reason given.

bmbouter: I don't think I understand how it's safer or what danger we are preventing with that safety. Can you explain it to me some more?

plugin/pulpcore/plugin/stages/content_unit_stages.py
jortel: Perhaps but the result here is silently creating a remote artifact with the (wrong) information from another (remote) artifact.  That seems unacceptable.

plugin/pulpcore/plugin/stages/declarative_version.py
jortel: Actually, I think it makes it more consistent and intuitive for plugin writers.  They would pass a list any *callable* as a stage to create a pipeline and in some cases it's a *function* and others it's a *callable* object.  Looking at `create_pipeline()` it calls each stage passing the queues.

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: Wait I'm confused. It's not happening because in practice you can't create a stream that creates two Artifacts at one relative_path. So there won't be duplicates like that.

Let me ask you this, what do you think should be done instead that is safer?

plugin/pulpcore/plugin/stages/declarative_version.py
jortel: Re: sync mode -  what if a plugin supports a sync-mode named `fast` and it is passed here?
Re: optionally passing the entire list of stages - By *hook* you mean a method they can override?  How would you know where in the stage ordering they should be integrated?

I think more thought needs to be put into this.  Especially since one of the main value propositions is customizable stages.

plugin/pulpcore/plugin/stages/models.py
jortel: Ah.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: After thinking about this more I agree w/ you. I'm going to switch the `stage` to be `__call__`. I'll do the same w/ `FirstStage` as well.

plugin/pulpcore/plugin/stages/artifact_stages.py
dralley: @bmbouter you could, in theory, make your classes Callable, so that they can be used as functions

no comment on whether you *should* :)

plugin/pulpcore/plugin/stages/artifact_stages.py
jortel: The `Storage` interface does not explicitly stipulate that implementations need to be safe to used as a global singleton. A stateful implementation *could* have unpredictable side effects when reused.  I'm mainly suggesting that since you have a choice, use the safer of the two.

IMHO, providing `default_storage` was a poor decision by the django project.

plugin/pulpcore/plugin/stages/content_unit_stages.py
jortel: How creating two artifacts (on different content) prevented?

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I updated with [this commit](https://github.com/pulp/pulp/pull/3559/commits/082e1e4888b65900e8fe6f1b20f252491ddc7deb). I used `batch` in almost all places except one where I thought `unhandled_content` was better since it's not guaranteed to handle the entire batch as a batch.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I can't just use `Q(pk=None)` because this is being combined with an AND query. I just tested and Q() == False and Q() | Q(anything=anything) is True, so I'm going with this patch. Let me know if I should do something different.

```
diff --git a/plugin/pulpcore/plugin/stages/artifact_stages.py b/plugin/pulpcore/plugin/stages/artifact_stages.py
index 64e78609b..aedce7808 100644
--- a/plugin/pulpcore/plugin/stages/artifact_stages.py
+++ b/plugin/pulpcore/plugin/stages/artifact_stages.py
@@ -65,7 +65,8 @@ async def query_existing_artifacts(in_q, out_q):
                     if digest_value:
                         key = {digest_name: digest_value}
                         one_artifact_q &= Q(**key)
-                all_artifacts_q |= one_artifact_q
+                if one_artifact_q:
+                    all_artifacts_q |= one_artifact_q
 
         for artifact in Artifact.objects.filter(all_artifacts_q):
             for content in batch:
```

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: I'm going to switch this to use `DefaultStorage()` instead which will ensure each instance is an unrelated copy. https://docs.djangoproject.com/en/2.0/ref/files/storage/#django.core.files.storage.DefaultStorage

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: I implemented the customizable hook, but then I pulled it out. Below is the diff that was there. For this hook to be successful, users have to rely on positional changes to the pipeline, either adding in a stage as stage 3, or removing stage 6, etc. They can only know the order, they can't get more info from them. This means that any customization is 100% dependant on DeclarativeVersion never changing the number, ordering, or types of the stages which means that object is basically fixed forever unelss we want to bump the major version of the plugin API with each change. I don't think we want to do that.

Instead let's offer `DeclarativeVersion` for usage only (not cusomtization) and its behavior can improve overtime (more/better stages, more features, etc). If plugin writers needs a custom set of stages they need to assemble it themselves.

If we need to provide additional machinery to help make that easier we could do that too, but I think that is still different from `DeclarativeVersion`.

```
diff --git a/plugin/pulpcore/plugin/stages/declarative_version.py b/plugin/pulpcore/plugin/stages/declarative_version.py
index b84cc0d62..ae5f340f9 100644
--- a/plugin/pulpcore/plugin/stages/declarative_version.py
+++ b/plugin/pulpcore/plugin/stages/declarative_version.py
@@ -125,6 +125,18 @@ class DeclarativeVersion:
         self.repository = repository
         self.sync_mode = sync_mode
 
+    def customize_stages(self, stages):
+        """
+        A hook that subclasses can override to modify the pipeline just before its created.
+
+        Args:
+            stages: The list of stages produced by the logic in :meth:`create`.
+
+        Returns:
+            The list of stages to be handed to `create_pipeline()`.
+        """
+        return stages
+
     def create(self):
         """
         Perform the work. This is the long-blocking call where all syncing occurs.
@@ -142,5 +154,5 @@ class DeclarativeVersion:
                     stages.append(end_stage)
                 elif self.sync_mode is 'mirror':
                     stages.extend([content_unit_unassociation(new_version).stage, end_stage])
-                pipeline = create_pipeline(stages)
+                pipeline = create_pipeline(self.customize_stages(stages))
                 loop.run_until_complete(pipeline)
```

jortel: I agree with the decision/reasons not include this patch.  We can let this evolve organically and I suspect we can find a better solution.

bmbouter:
@dralley @jortel Now that all stages inherit from `BaseStage` I think `FirstStage` should be deleted, and I'll refactor the docs to have plugin writer's make their first stage directly from `BaseStage`. Sound good?

bmbouter:
@jortel This PR still needs 3 things I'll do tomorrow morning:
1. remove `FirstStage` and use `BaseStage` instead
2. Handle the artifact collision concern
3. Switch default_storage() to DefaultStorage()

I'll ping you when it's ready for final review, Friday AM.

plugin/pulpcore/plugin/stages/api.py
```diff
@@ -0,0 +1,73 @@
...
+    in_q = None
+    futures = []
+    for stage in stages:
+        out_q = asyncio.Queue(maxsize=maxsize)
+        futures.append(stage(in_q, out_q))
+        in_q = out_q
+    await asyncio.gather(*futures)
+
+
+class EndStage(BaseStage):
```
jortel: Big :+1:  for making the stages objects!  This will unlock a lot of potential.

At the risk of being pedantic :(

Why not just `Stage`?

The rare cases I see the *Base* prefix, it strikes me as odd. The class name is a *classification* of objects and should read naturally using the **IsA** lexicon.

For example, the natural way to describe an `QueryExistingArtifacts` is to say:

 *"A QueryExistingArtifacts` IsA Stage"*.

not

 *"A QueryExistingArtifacts` IsA BaseStage"*.

I don't think designating this as a *base* class though naming adds value.  This isn't a blocker but, IMHO, it would be more correct in the classic object modeling sense.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -0,0 +1,146 @@
...
+from pulpcore.plugin.models import RepositoryVersion
+from pulpcore.plugin.tasking import WorkingDirectory
+
+from .api import create_pipeline, EndStage
+from .artifact_stages import ArtifactDownloader, ArtifactSaver, QueryExistingArtifacts
+from .association_stages import ContentUnitAssociation, ContentUnitUnassociation
+from .content_unit_stages import ContentUnitSaver, QueryExistingContentUnits
+
+
+class FirstStage:
```
jortel: Why is `FirstStage` still needed?

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,284 @@
...
+    downloads completed. Since it's a stream the total count isn't known until it's finished.
+
+    This stage drains all available items from `in_q` and starts as many downloaders as possible.
+
+    Args:
+        max_concurrent_downloads (int): The maximum number of concurrent downloads this stage will
+            run. Default is 100.
+    """
+
+    def __init__(self, max_concurrent_downloads=100, *args, **kwargs):
```
jortel: What are the `*args, **kwargs` for?

If needed, please docstring what can be passed.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,284 @@
...
+        self.max_concurrent_downloads = max_concurrent_downloads
+        super().__init__(*args, **kwargs)
```
jortel: It's better practice to initialize objects in heirarchical order.  Please call `super().__init__()` (first) before setting extended attributes.

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,150 @@
...
+            stage associates content with.
+    """
+
+    def __init__(self, new_version, *args, **kwargs):
+        self.new_version = new_version
+        self.unit_keys_by_type = defaultdict(set)
+        for unit in self.new_version.content.all():
+            unit = unit.cast()
+            self.unit_keys_by_type[type(unit)].add(unit.natural_key())
+        super().__init__(*args, **kwargs)
```
jortel: Same re: calling *super* as: [here](https://github.com/pulp/pulp/pull/3559/files#r207544584)
Same re: *args, kwargs* as: [here](https://github.com/pulp/pulp/pull/3559/files#r207543875)

plugin/pulpcore/plugin/stages/association_stages.py
```diff
@@ -0,0 +1,150 @@
...
+    un-associated. Since it's a stream the total count isn't known until it's finished.
+
+    Args:
+        new_version (:class:`~pulpcore.plugin.models.RepositoryVersion`): The repo version this
+            stage unassociates content from.
+    """
+
+    def __init__(self, new_version, *args, **kwargs):
+        self.new_version = new_version
+        super().__init__(*args, **kwargs)
```
jortel: Same re: calling *super* as: [here](https://github.com/pulp/pulp/pull/3559/files#r207544584)
Same re: *args, kwargs* as: [here](https://github.com/pulp/pulp/pull/3559/files#r207543875)

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,189 @@
...
+                                    'md5': declarative_artifact.artifact.md5,
+                                    'sha1': declarative_artifact.artifact.sha1,
+                                    'sha224': declarative_artifact.artifact.sha224,
+                                    'sha256': declarative_artifact.artifact.sha256,
+                                    'sha384': declarative_artifact.artifact.sha384,
+                                    'sha512': declarative_artifact.artifact.sha512,
+                                    'remote': declarative_artifact.remote,
+                                }
+                                rel_path = content_artifact.relative_path
+                                remote_artifact_map[rel_path] = remote_artifact_data
```
jortel: Still correlated by rel_path.

plugin/pulpcore/plugin/stages/api.py
bmbouter: I think s/BaseStage/Stage/ is a great change. I'll make it on my next push along w/ the last remaining fixes.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: It's not, but I was too tired to remove it last night. It'll be removed in next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: It's not formally spelled out in Python, but *args and **kwargs should be accepted on all __init__() calls and used when calling super().__init__(*args, **kwargs). Otherwise the parent object can never receive its parameters through the subclass' __init__().

For places where *args and **kwargs are used, I'm going to add a docstring that it takes *args and **kwargs and it will say unused params passed on to `Stage` to stay DRY w/ the docstrings.

plugin/pulpcore/plugin/stages/artifact_stages.py
bmbouter: ok will do in next push

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: yup will fix in next push.

plugin/pulpcore/plugin/stages/association_stages.py
bmbouter: yup will fix in next push.

plugin/pulpcore/plugin/stages/content_unit_stages.py
bmbouter: yup this and the storage backend both are being fixed in next push.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -0,0 +1,287 @@
...
+                            to_download_count = 0
+                            for declarative_artifact in content.d_artifacts:
+                                if declarative_artifact.artifact.pk is None:
+                                    new_artifact = Artifact(
+                                        **download_result.artifact_attributes,
+                                        file=download_result.path
+                                    )
+                                    declarative_artifact.artifact = new_artifact
+                                    to_download_count = to_download_count + 1
+                            pb.done = pb.done + to_download_count
```
jortel: Just FYI: using `+=` would be more concise.

plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -0,0 +1,189 @@
...
+                                    'md5': declarative_artifact.artifact.md5,
+                                    'sha1': declarative_artifact.artifact.sha1,
+                                    'sha224': declarative_artifact.artifact.sha224,
+                                    'sha256': declarative_artifact.artifact.sha256,
+                                    'sha384': declarative_artifact.artifact.sha384,
+                                    'sha512': declarative_artifact.artifact.sha512,
+                                    'remote': declarative_artifact.remote,
+                                }
+                                content_pk = content_artifact.content.pk
+                                remote_artifact_map[content_pk] = remote_artifact_data
```
jortel: After switching to *integer* PK, the `content_artifact.content.pk` will be (None) because the model has not yet been saved.  Right?

bmbouter: This is rebased onto the integer PK already so it uses them currently. The pk is set from L#151.

I verified I cannot use the models themselves because they are unsaved and Django raises an unhashable error. We could use sha512 instead. The only reason I didn't is to save memory. Should I switch it to that?

bmbouter:
Thank you @jortel , @dkliban, @dralley , and @gmbnomis for all the collaboration and help with this! :100: I really think the plugin writers are going to get a big benefit from it.

I'm merging it and https://github.com/pulp/pulp_file/pull/102/files

PR #3561, opened 2018-07-24
Problem: OpenAPI schema doesn't use URIs to identify resources

dkliban:
Solution: Create a custom OpenAPI generator that refers to resources by their URI. 

To see the Docs generated with this change, point the browser to http://<pulphostname>:8000/pulp/api/v3/docs/

The JSON schema that can be used to generate bindings can be downloaded at http://localhost:8000/pulp/api/v3/docs/api?format=openapi

### DIFF
--- a/pulpcore/pulpcore/app/openapigenerator.py
+++ b/pulpcore/pulpcore/app/openapigenerator.py
@@ -1,0 +1,162 @@
+from collections import OrderedDict
+import re
+
+from drf_yasg.generators import OpenAPISchemaGenerator
+from drf_yasg import openapi
+import uritemplate
+
+
+class Paths(openapi.SwaggerDict):
+    def __init__(self, paths, **extra):
+        """A listing of all the paths in the API.
+
+        :param dict[str,.PathItem] paths:
+        """
+        super(Paths, self).__init__(**extra)
+        for path, path_obj in paths.items():
+            if path_obj is not None:  # pragma: no cover
+                self[path] = path_obj
+        self._insert_extras__()
+
+
+class PulpOpenAPISchemaGenerator(OpenAPISchemaGenerator):
+
+    def get_paths(self, endpoints, components, request, public):
+        """Generate the Swagger Paths for the API from the given endpoints.
+
+        Args:
+            endpoints (dict): endpoints as returned by get_endpoints
+            components (ReferenceResolver): resolver/container for Swagger References
+            request (Request): the request made against the schema view; can be None
+            public (bool): if True, all endpoints are included regardless of access through
+                           `request`
+
+        Returns:
+            tuple[openapi.Paths,str]: The :class:`openapi.Paths` object and the longest common path
+                                      prefix, as a 2-tuple
+        """
+        if not endpoints:
+            return openapi.Paths(paths={}), ''
+
+        prefix = ''
+        resources = {}
+        resource_example = {}
+        paths = OrderedDict()
+        for path, (view_cls, methods) in sorted(endpoints.items()):
+            operations = {}
+            for method, view in methods:
+                if not public and not self._gen.has_view_permissions(path, method, view):
+                    continue
+
+                operation = self.get_operation(view, path, prefix, method, components, request)
+                if operation is not None:
+                    operation.operation_id = operation.operation_id.replace('pulp_api_v3_', '')
+                    operations[method.lower()] = operation
+
+            if operations:
+                path_param = None
+
+                if '}' in path:
+                    resource_path = '%s}/' % path.rsplit(sep='}', maxsplit=1)[0]
+                    resource_other_path = self.get_resource_from_path(path)
+                    if resource_other_path in endpoints:
+                        resource_model = endpoints[resource_other_path][0].queryset.model
+                        resource_name = self.get_parameter_name(resource_model)
+                        param_name = self.get_parameter_slug_from_model(resource_model)
+                        if resource_path in resources:
+                            path = path.replace(resource_path, '{%s}' % resources[resource_path])
+                        elif resource_other_path in resources:
+                            path = path.replace(resource_path, '{%s}' % resources[
+                                resource_other_path])
+                        else:
+                            resources[resource_path] = param_name
+                            resource_example[resource_path] = self.get_example_uri(path)
+                            path = path.replace(resource_path, '{%s}' % resources[resource_path])
+                        example = resource_example[resource_other_path]
+                        resource_description = self.get_resource_description(resource_name, example)
+                        path_param = openapi.Parameter(
+                            name=param_name,
+                            description=resource_description,
+                            required=True,
+                            in_=openapi.IN_PATH,
+                            type=openapi.TYPE_STRING,
+                        )
+                        paths[path] = openapi.PathItem(parameters=[path_param], **operations)
+                    else:
+                        if not path.startswith('/'):
+                            path = '/' + path
+                        paths[path] = self.get_path_item(path, view_cls, operations)
+                else:
+                    paths[path] = openapi.PathItem(parameters=[path_param], **operations)
+
+        return Paths(paths=paths), prefix
+
+    @staticmethod
+    def get_resource_from_path(path):
+        """
+        Returns a path for a resource nested in the specified path
+
+        Args:
+            path (str): Full path to be searched for a nested resource
+
+        Returns:
+            str: path of nested resource
+        """
+        resource_path = '%s}/' % path.rsplit(sep='}', maxsplit=1)[0]
+        if resource_path.endswith('_pk}/'):
+            resource_path = '%s{id}/' % resource_path.rsplit(sep='{', maxsplit=1)[0]
+        return resource_path
+
+    @staticmethod
+    def get_resource_description(name, example_uri):
+        """Returns a description of an *_href path parameter
+
+        Args:
+            name (str): Name of the resource referenced by the *_href path parameter
+            example_uri (str): An example of the URI that is a reference for a specific resource
+
+        Returns:
+            str: Description of an *_href path parameter
+        """
+        return "URI of %s. e.g.: %s" % (name, example_uri)
+
+    @staticmethod
+    def get_example_uri(path):
+        """Returns an example URI for a path template
+
+        Args:
+            path (openapi.Path): path object for a specific resource
+
+
+        Returns:
+            str: The path with concrete path parameters.
+        """
+        params = {}
+        for variable in uritemplate.variables(path):
+            params[variable] = '1'
+        return uritemplate.expand(path, **params)
+
+    @staticmethod
+    def get_parameter_slug_from_model(model):
+        """Returns a path parameter name for the resource associated with the model.
+
+        Args:
+            model (django.db.models.Model): The model for which a path parameter name is needed
+
+        Returns:
+            str: *_href where * is the model name in all lower case letters
+        """
+        return '%s_href' % '_'.join([part.lower() for part in re.findall('[A-Z][^A-Z]*',
+                                                                         model.__name__)])
+
+    @staticmethod
+    def get_parameter_name(model):
+        """Returns the human readable name of the resource associated with the model
+
+        Args:
+            model (django.db.models.Model): The model for which a name is needed
+
+        Returns:
+            str: name of the resource associated with the model
+        """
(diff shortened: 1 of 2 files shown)


pulpcore/pulpcore/app/openapigenerator.py
```diff
@@ -0,0 +1,88 @@
...
+                    path_suffix = '/' + path_suffix
+                paths[path_suffix] = openapi.PathItem(parameters=[path_param], **operations)
+
+        return openapi.Paths(paths=paths), prefix
+
+
+    def get_resource_description(self, name, example_uri):
+        """Return a description for the resource associated with the viewset.
+        :param type view: the view class for which a description is desired.
+        :return: description of the
```
dralley: Use google style docstrings

pulpcore/pulpcore/app/openapigenerator.py
```diff
@@ -0,0 +1,88 @@
...
+    def get_paths(self, endpoints, components, request, public):
+        """Generate the Swagger Paths for the API from the given endpoints.
+
+        :param dict endpoints: endpoints as returned by get_endpoints
+        :param ReferenceResolver components: resolver/container for Swagger References
+        :param Request request: the request made against the schema view; can be None
+        :param bool public: if True, all endpoints are included regardless of access through `request`
+        :returns: the :class:`.Paths` object and the longest common path prefix, as a 2-tuple
+        :rtype: tuple[openapi.Paths,str]
+        """
```
dralley: Use google/napolean style docstrings

bmbouter:
I don't know much about openapi, and I could be totally wrong about this.

I think the human interacting with the autogenerated client would know what should go there from the description like `description: 'URI of Artifact. e.g.: /artifacts/1/'`. Similarly a programmer interacting with the bindings. I believe the user would get that string from an earlier call when the Artifact was created and the server returned `/artifacts/1/`.

dkliban:
The operationId provides each method in the generated bindings with a name. So the user of the bindings knows exactly what to specify for the parameter. 

https://github.com/dkliban/pulp3-python-bindings/blob/href-as-identifier/swagger_client/api/artifacts_api.py#L373

Is that what you were asking?

dkliban:
@werwty Or were you talking about validation of the response? I think that is what the schema response objects are used for. Artifact does not currently have one specified, but most of the others do. Like hrere: https://github.com/dkliban/pulp3-python-bindings/blob/href-as-identifier/swagger_client/api/remotes_api.py#L127

werwty:
> The operationId provides each method in the generated bindings with a name. So the user of the bindings knows exactly what to specify for the parameter.
Ah, I just looked into the bindings code and you are right. The bindings does not use the path to generate the swagger_client.api objects. 

How would validation on the urls be done? Can I pass any url to {artifact_href}? 
From the generated bindings code I suspect yes:
```
    def artifacts_read_with_http_info(self, artifact_href, **kwargs):  # noqa: E501

        path_params = {}
        if 'artifact_href' in params:
            path_params['artifact_href'] = params['artifact_href']  # noqa: E501

        return self.api_client.call_api(
            '/{artifact_href}/', 'GET',
            path_params,
            query_params,
            header_params,
            body=body_params,
            post_params=form_params,
            files=local_var_files,
            response_type=None,  # noqa: E501
            auth_settings=auth_settings,
            async=params.get('async'),
            _return_http_data_only=params.get('_return_http_data_only'),
            _preload_content=params.get('_preload_content', True),
            _request_timeout=params.get('_request_timeout'),
            collection_formats=collection_formats)
```

I also ran the generated schema code through a validator, and swagger does not like equivalent paths. There are a couple of issues open discussing this, but it doesn't appear that they're going to support it in the schema specification:
https://github.com/OAI/OpenAPI-Specification/issues/576

daviddavis:
> How would validation on the urls be done? Can I pass any url to {artifact_href}?
From the generated bindings code I suspect yes

I agree. I found another RFE that worries me: https://github.com/OAI/OpenAPI-Specification/issues/577.

dkliban:
@werwty Thank you for looking into the schema validation. I agree that we should try to produce a schema that is OpenAPI compliant. 

I have done some reading in this area and have found that there is a desire (mostly from Google) to improve the OpenAPI 3 spec to include either an ability to have path parameters that have '/' in them[0] or add ability to specify path parameters of type 'hypermedia'[1]. Another proposal is to have OpenAPI be able to keep track of resources[2].

With all that said, these are only proposals at this time and we cannot use any of the methods from above to describe our API at this time. 

In reading all these proposals it became clear that the problem with the shema this PR is generating is that it it is too obscure for generating server side code that would be matching URLs. However, this schema is not a problem for generating docs and bindings. I also think that @timburks said it best here[3]. 

Perhaps we should be more liberal about what WE consider valid OpenAPI schema?

[0] https://github.com/OAI/OpenAPI-Specification/issues/1459
[1] https://github.com/OAI/OpenAPI-Specification/issues/1553
[2] https://github.com/OAI/OpenAPI-Specification/issues/1553#issuecomment-383626417
[3] https://github.com/OAI/OpenAPI-Specification/issues/892#issuecomment-355783005

bmbouter:
Maybe we can accept ids and hrefs. I was thinking about how that would work, and there are some questions I have. Let's assume we don't merge this PR and let the openAPI v2 schema describe the API with IDs.

Here's my question. Currently we return task data like:

```
{
    "created_resources": [
        "http://localhost:8000/pulp/api/v3/repositories/1/versions/3/"
    ]
}
```

For the user to use that resulting data with an API then need to parse the url to get the two expected parameters (repo=1, version=3). We could do something like this:

```
{
	"created_resources": [{
		"url": "http://localhost:8000/pulp/api/v3/repositories/1/versions/3/",
		"repository_id": 1,
		"version": 3
	}]
}
```

As yet a third alternative, we could do what the github API does which is return the entire sub-resource serialized. That would be like:

```
{
	"created_resources": [{
		"url": "http://localhost:8000/pulp/api/v3/repositories/1/versions/3/",
		"repository_id": 1,
                ....  lots more things ....
		"version": 3
	}]
}
```

In thinking about this, my primary goals are:
1. To provide clickable urls returned with every object we are referring to in responses.
2. To accept urls as references like we do today, even if the openAPI schema still describes them using IDs.

What do you think we should do?

daviddavis:
@bmbouter the more I think about the hrefs vs id debate, the more I think a solution like what you propose makes the best sense. 👍 from me.

werwty:
+1  option 2

> To provide clickable urls returned with every object we are referring to in responses.

:+1: 
> To accept urls as references like we do today, even if the openAPI schema still describes them using IDs.

:+1:

daviddavis:
Correct me if I am wrong but accepting id and href would mean we don't have to define generators to alter the schema nor the bindings? We could use them as they are?

bmbouter:
In all cases (option 1, 2, or 3) how does the user handle the multiple object types that could be returned by `created_resource`. It could get back a repository version or a publication? The attributes and/or href will need to be handed back to a binding, how does it know which one?

dkliban:
@bmbouter I think @daviddavis was talking about the generator class that I wrote in this PR. 

@daviddavis I think that we still need to write a custom schema generator to expose all the right docs. I am thinking of the 'repository_pk' and 'number' path parameters that don't have docs right now. https://docs.pulpproject.org/en/3.0/nightly/integration-guide/rest-api/index.html#get--repositories-repository_pk-versions-number-

Perhaps there is another way to fix it that I am not aware of.

dkliban:
@bmbouter In all cases the user will need process the response and decide what to do with it. 

For both options 1 and 2, I don't think we'll be able to describe using OpenAPI 2.0 a schema that returns 1 of 2 different responses. I don't think there is that flexibility in the spec. 

In the case of this PR, the bindings generated from this PR don't provide any machinery for converting an href into an API client for that resource type. The user will still have to write code that determines what kind of API client needs to be used to access this resource.

bmbouter:
@jortel I'm confused by the suggestion to create two openAPI schemas. How can we describe one API two ways? The plan just sent to pulp-dev says that we're not supporting IDs and only hrefs. Assuming that's the plan, this PR is the only way openAPI can describe an API like that that I know of.

Am I thinking about this right?

jortel:
It's my understanding that the custom schema generator proposed here generates a schema with paths look like:
```
paths:
  /{artifact_href}/:
```
instead of:
```
paths:
  /artifacts/:
```
to accommodate swagger.

Is my understanding incorrect?

daviddavis:
The `/artifacts/` path would remain the same (assuming you're talking about the list artifacts path). But for single resource paths (e.g. read artifact, delete artifact, etc) those would be like what you mentioned:

```
GET /artifacts/
POST /artifacts/
GET /{artifact_href}/
PUT /{artifact_href}/
DELETE /{artifact_href}/
```

jortel:
@daviddavis, gotcha.  /artifacts/ was a bad example but same point.

jortel:
After getting a better understanding of this from @bmbouter, I no longer have the concern about the proposed generated schema at the `/docs/api` end point.  

However, I'd like to suggest an improvement in the `description` field for paths (and parameters) like:

```
paths:
  /{artifact_href}/:
    get:
      operationId: artifacts_read
      description: ''
      parameters: []
      responses:
```
and

```
    parameters:
      - name: 'artifact_href'
        in: path
        description: 'URI of Artifact. e.g.: /artifacts/1/'
        required: true
        type: string
```

I think it would be clearer if the `description` contained the format of the href instead of an example.

For example:

```
    parameters:
      - name: 'artifact_href'
        in: path
        description: 'The URI of an Artifact: /artifacts/{id}/'
        required: true
        type: string
```

dkliban:
@jortel I am concerned that if we add a format of the href, we will then need to once again describe what the 'id' field is. Or what the 'repository_pk' and 'version_number' are for the repo version href.

dkliban:
I was able to generate the bindings using this schema. I then wrote the following script that uses the python bindings to create a remote, a repository, and then sync the remote into the repository. 

```
import swagger_client
from swagger_client.rest import ApiException
from pprint import pprint
from urllib.parse import urlparse

# Configure HTTP basic authorization: basic
configuration = swagger_client.Configuration()
configuration.username = 'admin'
configuration.password = 'admin'
configuration.safe_chars_for_path_param = '/'

client = swagger_client.ApiClient(configuration)

# create an instance of the API class
api = swagger_client.PulpApi(client)
remote_data = swagger_client.FileRemote(name='bar27',
                                        url='https://repos.fedorapeople.org/pulp/pulp/demo_repos'
                                            '/test_file_repo/PULP_MANIFEST')

try:
    api_response = api.pulp_api_v3_remotes_file_create(remote_data)
    remote_href = urlparse(api_response.href).path

    pprint(api_response)
except ApiException as e:
    print("Exception when calling RemotesApi->pulp_api_v3_remotes_file_create: %s\n" % e)

repository_data = swagger_client.Repository(name='foo27')

try:
    api_response = api.pulp_api_v3_repositories_create(repository_data)
    repository_href = urlparse(api_response.href).path
    pprint(api_response)
except ApiException as e:
    print("Exception when calling RepositoriesApi->pulp_api_v3_repositories_create: %s\n" % e)

repository_sync_data = swagger_client.RepositorySyncURL(repository=repository_href)

try:
    api_response = api.pulp_api_v3_remotes_file_sync(remote_href, repository_sync_data)
    pprint(api_response)
except ApiException as e:
    print("Exception when calling Api->pulp_api_v3_remotes_file_sync: %s\n" % e)
```

daviddavis:
@dkliban thanks for the code example. It's helpful. I see you calling this:

```python
remote_href = urlparse(api_response.href).path
```

After we return relative URI, it will just be this?:

```python
remote_href = api_response.href
```

dkliban:
@daviddavis Yes, that is exactly correct. urlparse is only needed because we are returning a full URL.

daviddavis:
Looks like there are problems with the pulp_ansible plugin. There's [a view that returns a static dict](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/views.py#L11-L28) the Galaxy CLI requires when looking at the API's versions. There's no queryset needed since the response is static and so [this line](https://github.com/pulp/pulp/pull/3561/files#diff-fc11f181de9c986c637d36a34152bf37R63) raises a exception: `AttributeError: 'AnsibleGalaxyVersionView' object has no attribute 'queryset'`.

PR #3567, opened 2018-07-25
Fixes #3883 - Log to console by default

ehelms:


pulpcore/pulpcore/etc/pulp/server.yaml
```diff
@@ -68,7 +68,7 @@
 #       formatter: simple
```
bmbouter: These lines also need to be removed. They show the defaults.

ehelms: I assumed this was showing users how to set the formatter just like it shows how to configure a syslog handler. I think it would be useful to the discussion to express what the intent for examples in server.yaml are as that would help guide the change needed here.

bmbouter: The convention so far is that commented sections are the place for the user to see the exact defaults. Examples are good though, what about moving it to the docs in the logging section?

bmbouter:
This looks good to me. Thanks @ehelms ! :+1:

PR #3569, opened 2018-07-26
When unique constraint is violated status code is 409

vdusek:
closes #3846
https://pulp.plan.io/issues/3846


vdusek:
The test failes in the [is_valid](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L221) method. It stops always on [line 237](https://github.com/encode/django-rest-framework/blob/f404fda29c3c88e344db20f85cab7c51ef44ccdf/rest_framework/serializers.py#L237)

vdusek:
This PR defines new exception to return 409 for uniqueness error. The test fails in the [is_valid](https://github.com/encode/django-rest-framework/blob/f404fda29c3c88e344db20f85cab7c51ef44ccdf/rest_framework/serializers.py#L221) method because it [catches](https://github.com/encode/django-rest-framework/blob/f404fda29c3c88e344db20f85cab7c51ef44ccdf/rest_framework/serializers.py#L237) or [re-raises](https://github.com/encode/django-rest-framework/blob/f404fda29c3c88e344db20f85cab7c51ef44ccdf/rest_framework/serializers.py#L244) only generic ValidationError. So the exception I introduced is not handled properly. One of the solutions is to define our own is_valid method. But it doesn't seem to be a good solution. 
Any other thoughts or suggestions?

dkliban:
@vdusek I would write a custom is_valid method tht checks for the 409 and reraises it. If not a 409 it should just call into the super is_valid.

vdusek:
Now, when unique constraint is violated the response looks like this:
```
HTTP/1.1 409 Conflict
...
{
    "detail": "Conflict with the current state of the target resource."
}

```
Before the commit, with the ValidationError exception, it looked like this:
```
HTTP/1.1 400 Bad Request
...
{
    "name": [
        "This field must be unique."
    ]
}
```
So the different format of a detail message is probably the reason why this [test](https://travis-ci.org/pulp/pulp/jobs/411217011#L958) didn't pass. I spent a lot of time on figuring out how to format it right way but unsuccessfully. So if anyone has any idea how to do it, I'll be glad.

dparalen:
Hi,

thanks for the PR!
Please rebase it on master, either after your fix commits ac96aaaafc510c3ad16cd3fe9411dc60b37d3344 or on head

Thanks,
milan

vdusek:
Updated, checks finally passed. :) However it's not a final version, I'm going to open a PR on rest framework for a better solution.

vdusek:
Opened issue - https://github.com/encode/django-rest-framework/issues/6124 and PR - https://github.com/encode/django-rest-framework/pull/6126 on DRF. However it doesn't seem like we're getting to the solution.

daviddavis:
I'm a bit hesitant on merging this PR. I don't think returning a 400 in cases where there is a uniqueness violation is incorrect. Moreover, I see a few downsides if we merge this:

- This adds complexity which we will have to carry and maintain in our codebase
- I think DRF upstream had a good point. What if there are two invalidations with one being a 409 and the other one being a 400?
- Lastly, I think plugin writers will overlook this and just end up returning 400s for uniqueness violations which will lead to inconsistencies. In fact, it doesn't look like this PR exposes `PulpUniqueValidator` as part of the plugin API.

daviddavis:
@vdusek thanks for your hard work on this issue. Glad you were able to learn something too. 👍

PR #3577, opened 2018-08-06
Make pytest tell why tests skip

Ichimonji10:
When a test skips, pytest doesn't print information about why that test
skipped. For example, if a test contains this line:

    self.skipTest('https://pulp.plan.io/issues/1')

...then pytest does *not* print `https://pulp.plan.io/issues/1` in its
report.

Make pytest print skip messages.


Ichimonji10:
The `-r s` option should cause a summary of skipped tests to be shown, but it does not.

Ichimonji10:
The `-r s` option should cause a summary of skipped tests to be shown, but it does not.

dralley:
@Ichimonji10 it worked the other day when I tried it?  https://travis-ci.org/pulp/pulp/jobs/411867356#L1005

Ichimonji10:
Ah. I was looking at some incorrect info. Thanks.

PR #3580, opened 2018-08-07
Fix storage path for published metadata.

jortel:
https://pulp.plan.io/issues/3878

With integer IDs the PK is not yet assigned (and no longer unique across tables).


daviddavis:
LGTM. Will test it out tomorrow.

PR #3583, opened 2018-08-08
Adds doc string and validation for mirror parameter

vdusek:
re # 3494
https://pulp.plan.io/issues/3494

closes #3944
https://pulp.plan.io/issues/3944

### DIFF
--- a/plugin/pulpcore/plugin/stages/declarative_version.py
+++ b/plugin/pulpcore/plugin/stages/declarative_version.py
@@ -1,5 +1,4 @@
 import asyncio
-from gettext import gettext as _
 
 from pulpcore.plugin.models import RepositoryVersion
 from pulpcore.plugin.tasking import WorkingDirectory
@@ -12,7 +11,7 @@ from .content_unit_stages import ContentUnitSaver, QueryExistingContentUnits
 
 class DeclarativeVersion:
 
-    def __init__(self, first_stage, repository, sync_mode='mirror'):
+    def __init__(self, first_stage, repository, mirror=True):
         """
         A pipeline that creates a new :class:`~pulpcore.plugin.models.RepositoryVersion` from a
         stream of :class:`~pulpcore.plugin.stages.DeclarativeContent` objects.
@@ -86,13 +85,9 @@ class DeclarativeVersion:
         Raises:
             ValueError: if 'sync_mode' is passed an invalid value.
         """
-
-        if sync_mode != 'mirror' and sync_mode != 'additive':
-            msg = _("'sync_mode' must either be 'mirror' or 'additive' not '{sync_mode}'")
-            raise ValueError(msg.format(sync_mode=sync_mode))
         self.first_stage = first_stage
         self.repository = repository
-        self.sync_mode = sync_mode
+        self.mirror = mirror
 
     def create(self):
         """
@@ -107,9 +102,9 @@ class DeclarativeVersion:
                     QueryExistingContentUnits(), ContentUnitSaver(),
                     ContentUnitAssociation(new_version)
                 ]
-                if self.sync_mode == 'additive':
-                    stages.append(EndStage())
-                elif self.sync_mode == 'mirror':
+                if self.mirror:
                     stages.extend([ContentUnitUnassociation(new_version), EndStage()])
+                else:
+                    stages.append(EndStage())
                 pipeline = create_pipeline(stages)
                 loop.run_until_complete(pipeline)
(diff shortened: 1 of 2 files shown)


plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -82,14 +81,8 @@ def __init__(self, first_stage, repository, sync_mode='mirror'):
...
                  'mirror' is the default.
-
-        Raises:
-            ValueError: if 'sync_mode' is passed an invalid value.
         """
-
-        if sync_mode != 'mirror' and sync_mode != 'additive':
-            msg = _("'sync_mode' must either be 'mirror' or 'additive' not '{sync_mode}'")
-            raise ValueError(msg.format(sync_mode=sync_mode))
+        assert(sync_mode == 'mirror' or sync_mode == 'additive')
```
daviddavis: I'm not sure I understand this change?

daviddavis: Ah, I see now. The value is being validated by the serializer.

bmbouter: I think a ValueError is more correct than an AssertionError here. Look at the description of ValueError in the standard library docs. https://docs.python.org/3/library/exceptions.html#ValueError

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -12,7 +11,7 @@
 
 class DeclarativeVersion:
 
-    def __init__(self, first_stage, repository, sync_mode='mirror'):
+    def __init__(self, first_stage, repository, sync_mode):
```
bmbouter: Do we want plugin writers to not have a default value for `sync_mode`?

Also if sync_mode isn't getting a default there are various docstrings that need updating.

plugin/pulpcore/plugin/stages/declarative_version.py
dparalen: This was based on my suggestion so I'm to blame ;) The case will be more obvious if proper constants and namespacing are used:

```python
class DeclarativeVersion:
    sync_mode_mirror = 'mirror'
    sync_mode_additive = 'additive'
    default_sync_mode = sync_mode_mirror

    def __init__(self, first_stage, repository, sync_mode=None):
         ...

        if sync_mode is None:
           sync_mode = self.default_sync_mode
       assert sync_mode in {self.sync_mode_additive, sync_mode_mirror}, 'Oops!'
```
 IMO a declarative_version object always expects just the values/switches it was declared with therefore the `assert` statement is just a debugging thing.

plugin/pulpcore/plugin/stages/declarative_version.py
vdusek: I deleted it because it seemed to me that the default value is assigned twice. Once here and once in the file plugin. But I didn't realize that there could be plugins that might not do that. So I made the changes back. Thanks.

daviddavis:
FYI, there's a new thread on sync modes on pulp-dev that we may want to wait on before merging this PR.

https://www.redhat.com/archives/pulp-dev/2018-August/msg00035.html

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -82,13 +81,8 @@ def __init__(self, first_stage, repository, sync_mode='mirror'):
...
                  pre-existing units in the :class:`~pulpcore.plugin.models.RepositoryVersion`.
                  'mirror' is the default.
-
-        Raises:
-            ValueError: if 'sync_mode' is passed an invalid value.
         """
-        if sync_mode != 'mirror' and sync_mode != 'additive':
-            msg = _("'sync_mode' must either be 'mirror' or 'additive' not '{sync_mode}'")
-            raise ValueError(msg.format(sync_mode=sync_mode))
+        assert(sync_mode == 'mirror' or sync_mode == 'additive')
```
bmbouter: I don't think we should use assert here. Instead we should raise a `ValueError` which is already there. I think this file should be left unedited.

vdusek:
@bmbouter You may overlooked my and @dparalen 's answers. However I made all the changes in ```declarative_version.py``` back. Thanks.

daviddavis:
I responded to [the mailing list thread about ditching sync modes](https://www.redhat.com/archives/pulp-dev/2018-August/msg00035.html). I'm going to wait until Monday and if there's no response, let's merge this PR.

bmbouter:
@daviddavis  +1

PR #3586, opened 2018-08-08
Remove usages of selectors.bug_is_fixed()

dralley:
Now that Smash tests are in the same repo as the code under test, we no
longer need to look to an external source (Redmine) to determine whether
an issue should be tested.

When adding a new test for funtionality which does not exist or is broken, decorate the test or test class with @unittest.skip("https://pulp.plan.io/issue/..."), and the patch which adds or fixes this functionality should remove the skip decorator to re-enable the test.

This has several benefits - it means the PR test runner will always run
the tests for the functionality being added/fixed so long as the
developer has properly re-enabled the test. It should also make the
tests faster by a small degree, and less reliant on external connections.

### DIFF
--- a/pulpcore/tests/functional/api/test_api_docs.py
+++ b/pulpcore/tests/functional/api/test_api_docs.py
@@ -4,7 +4,7 @@ import unittest
 
 from requests.exceptions import HTTPError
 
-from pulp_smash import api, config, selectors
+from pulp_smash import api, config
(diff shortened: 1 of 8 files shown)


Ichimonji10:
These changes look great quality. I haven't run `git grep bug_is_fixed` to check for lingering skips, nor have I run the modified tests.

dralley:
I searched for both and couldn't find any.  Looks like they pass on Travis

PR #3588, opened 2018-08-08
Support redirect in the content application.

jortel:
https://pulp.plan.io/issues/3696
https://pulp.plan.io/issues/3697

Support redirect to the streamer.  This includes adding REDIRECT settings to the CONTENT section.
I doubt the defaults are correct but we can change them later.

### DIFF
--- a/pulpcore/pulpcore/app/views/content.py
+++ b/pulpcore/pulpcore/app/views/content.py
@@ -5,8 +5,12 @@ from logging import getLogger, DEBUG
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.http import (HttpResponse, HttpResponseForbidden, HttpResponseNotFound,
-                         StreamingHttpResponse)
+from django.http import (
+    HttpResponse,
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/views/content.py
```diff
@@ -104,87 +148,127 @@ def _match(self, path):
...
+            django.http.HttpResponseNotFound: on not-found.
+            django.http.HttpResponseForbidden: on forbidden.
+            django.http.HttpResponseRedirect: on redirect to the streamer.
+        """
+        server = settings.CONTENT['WEB_SERVER']
+
+        try:
+            responder = self.RESPONDER[server]
+        except KeyError:
+            raise ValueError(_('Web server "{t}" not supported.'.format(t=server)))
```
bmbouter: I'm not able to do a full review right now, but I noticed this line. The .format() needs to run on str returned from _() not the string passed into _(). This won't i8ln as is.

jortel: Fixed.

PR #3589, opened 2018-08-09
Adds Pulp 3 nightly and beta builds to travis

dkliban:
closes: #3722
https://pulp.plan.io/issues/3722


.travis.yml
```diff
@@ -32,6 +36,10 @@ stages:
   if: tag =~ ^pulpcore-plugin*
 - name: deploy-common
   if: tag =~ ^pulpcore-common*
+- name: publish-beta-docs
+  if: tag =~ ^pulpcore-3.0.0*
```
daviddavis: Why not just `^pulpcore-3.*` or whatever?

.travis.yml
```diff
@@ -55,3 +63,17 @@ jobs:
       script: bash .travis/deploy.sh common
       on:
         tags: true
+  - stage: publish-beta-docs
```
daviddavis: Why a specific beta job and not just have one job for releases and one for nightly?

.travis.yml
daviddavis: I think this is fine for now.

PR #3594, opened 2018-08-13
Update smash-tests w/ changes made to plugin_template

dralley:


dralley:
Here's a question @kersommoura , currently these tests copy over ```constants.py``` and ```utils.py``` from ```pulp_file``` in order to remain independent of any changes made there, because if we tried to import things from those tests and variable names changed, they might make these tests break, etc.

That probably won't scale well to multiple plugins.  What do you think about importing constants and utilities directly from the plugins?

In order to prevent breakages, the plugins could also run pulpcore's ```using_plugin``` tests, so if they change something that would break those tests, they know to go and fix it.

kersommoura:
@dralley, I will talk with @rochacbruno about it as well.

kersommoura:
>That probably won't scale well to multiple plugins. What do you think about importing constants and utilities directly from the plugins?

@dralley, I think this approach for the time being. I was wondering about namespace conflicts. Perhaps we can evaluate add those constants to Pulp-Smash itself. If the user does not want them from Pulp-Smash then just do not import them.

PR #3595, opened 2018-08-13
Update SELinux policy

dralley:
This PR is currently disorganized and pending cleanup, but it should be functional.

### DIFF
--- a/server/selinux/server/pulp-celery.te
+++ b/server/selinux/server/pulp-celery.te
@@ -16,17 +16,19 @@ files_tmp_file(pulp_tmp_t)
 
 require {
         type celery_t;
+        type gpg_exec_t;
         type httpd_sys_rw_content_t;
         type pstore_t;
-        type pulp_var_run_t;
+        type pulp_cert_t;
         type pulp_tmp_t;
         type pulp_var_cache_t;
-        type pulp_cert_t;
+        type pulp_var_run_t;
         type puppet_etc_t;
         type qpidd_t;
         type qpidd_var_lib_t;
         type tmp_t;
         type tmpfs_t;
+        type shell_exec_t;
         type sysfs_t;
         class sock_file { create link unlink write };
         class unix_stream_socket connectto;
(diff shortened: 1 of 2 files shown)


server/selinux/server/pulp-server.te
```diff
@@ -7,10 +7,14 @@ require {
     type puppet_etc_t;
     type httpd_t;
     type cert_t;
+    type nfs_t;
```
bmbouter: I don't think we need `nfs_t`, the user should configure their NFS with the `context` option w/ the expected labels. Then the other `:map` statement should cover it.

server/selinux/server/pulp-celery.te
```diff
@@ -57,6 +60,12 @@ allow celery_t tmpfs_t:filesystem getattr;
 allow celery_t tmpfs_t:dir { write add_name remove_name };
 allow celery_t tmpfs_t:file { create open read write link getattr unlink };
 
+# add missing rules for file map
+allow celery_t pulp_var_cache_t:file map;
+allow celery_t shell_exec_t:file map;
+allow celery_t gpg_exec_t:file map;
+allow celery_t nfs_t:file map;
```
bmbouter: similar here, I don't think we need to interact w/ `nfs_t`

server/selinux/server/pulp-streamer.te
```diff
@@ -21,6 +21,9 @@ require {
 }
 
 
+unconfined_shell_domtrans(streamer_t)
```
bmbouter: This suggests that the streamer is running as unconfined_t and then calling to a shell. What process context is the streamer running as?

server/selinux/server/pulp-server.te
```diff
@@ -7,10 +7,14 @@ require {
...
     class dir { read remove_name rmdir write };
     class file { getattr unlink write };
 }
 
+# add missing rules for file map
+allow httpd_t nfs_t:file map;
```
bmbouter: Also should not be needed AIACT.

server/selinux/server/pulp-streamer.te
bmbouter: Which shell is this executing that needs to transition to unconfined_t?

PR #3599, opened 2018-08-14
Fix DeclarativeVersion timeout error

dralley:


bmbouter:
I think we need some mention of this in the docs of the HttpDownloader that the ClientSession it creates takes these options. Also maybe in the Factory itself since it's providing configuration there too.

plugin/pulpcore/plugin/download/http.py
```diff
@@ -117,7 +117,8 @@ def __init__(self, url, session=None, auth=None, proxy=None, proxy_auth=None,
             self.session = session
             self._close_session_on_finalize = False
         else:
-            self.session = aiohttp.ClientSession()
+            timeout = aiohttp.ClientTimeout()
```
gmbnomis: I am wondering whether this leads to the opposite extreme case that a download connection may hang forever if the peer disappears? (established TCP connections can live forever if no data is sent and keep-alives are disabled)

IMHO, we should set `sock_read` and `sock_connect` if there is no total timeout. Something needs to kill a dead connection or a connection attempt to a "black hole".

bmbouter: @gmbnomis I'm so glad you brought. I agree when keepalive is disabled at the TCP level or in a buggy proxy we would fall prey to [these issues](http://www.tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html).

What about `sock_connect` and `sock_read` both set to 5 minutes? If a server can't respond in 5 minutes or new data can't be read in 5 minutes I think the remote side is dead.

gmbnomis: @bmbouter Makes sense. I would have chosen shorter values, but this is only gut feeling. There is no "one value fits all situations" here anyway. The important part is that downloads do not hang forever; tuning can happen later if necessary.

bmbouter:
@gmbnomis I had similar feelings. I really want to avoid false_positives; perhaps too much. +1 to more tuning later. Thank you so much for the input. It's much appreciated.

plugin/pulpcore/plugin/download/http.py
bmbouter: @gmbnomis I had similar feelings. I really want to avoid false_positives; perhaps too much. +1 to more tuning later. Thank you so much for the input. It's much appreciated.

plugin/pulpcore/plugin/download/http.py
```diff
@@ -44,6 +44,13 @@ class HttpDownloader(BaseDownloader):
     then closed when the download is complete. A session that is passed in will not be closed when
     the download is complete.
 
+    If a session is not provided, the one created by HttpDownloader uses non-default timing values.
+    Specifically, the "total" timeout is set to None and the "sock_connect" and "sock_read" are both
+    5 minutes. For more info on these settings, see the aiohttp docs:
+    http://aiohttp.readthedocs.io/en/stable/client.html#ssl-control-for-tcp-sockets
```
daviddavis: Same link here.

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -20,14 +20,24 @@ class DownloaderFactory:
...
 
     Usage:
         >>> the_factory = DownloaderFactory(remote)
         >>> downloader = the_factory.build(url_a)
         >>> result = downloader.fetch()  # 'result' is a DownloadResult
+
+    For http and https urls, in addition to the remote settings, non-default timing values are used.
+    Specifically, the "total" timeout is set to None and the "sock_connect" and "sock_read" are both
+    5 minutes. For more info on these settings, see the aiohttp docs:
+    http://aiohttp.readthedocs.io/en/stable/client.html#ssl-control-for-tcp-sockets
```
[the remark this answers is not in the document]
bmbouter: yeah these links aren't right. I'm pushing fixes in my next commit.

plugin/pulpcore/plugin/download/http.py
bmbouter: yeah these links aren't right. I'm pushing fixes in my next commit

bmbouter:
+1 to @daviddavis ask for a squash and good commit message.

dralley:
I fixed it up really quick, @bmbouter or @daviddavis feel free to merge it if you're satisfied since I'm out ("out") today

bmbouter:
This looks good to me @dralley Thank you!

PR #3600, opened 2018-08-15
Serializer ignore the csrfmiddlewaretoken field

dparalen:
The CSRF protection middleware kicks-in when a browser is used to access
Pulp API, after the user having successfully logged into Pulp, i.e
whenever a Session-based authentication is used by the DRF/Django
The middleware uses a custom (and invisible) form field:
`csrfmiddlewaretoken` that the page in the browser populates from a CSRF
cookie previously generated by Django whenever issuing a
POST/PATCH/DELETE... request to prevent the cross-site-request-forgery.
The whole mechanism is nicely described on wikipedia.

This `csrfmiddlewaretoken` field unfortunately isn't stripped from the
POSTed data by either DRF or Django, and causes our serializers to fail
validate form data because of it being an unknown field to our models.

This patch just makes our validation ignore the unknow
`csrfmiddlewaretoken` field.

Fixes: #3906
https://pulp.plan.io/issues/3906


daviddavis:
Mind adding a test case to [our unknown fields unit tests](https://github.com/pulp/pulp/blob/master/pulpcore/tests/unit/serializers/test_base.py#L25)?

dparalen:
@daviddavis thanks for the review!
Just added the test case

--
milan

PR #3605, opened 2018-08-29
Adds possibility inject stages into DeclarativeVersion

vdusek:
closes #3935
https://pulp.plan.io/issues/3935

### DIFF
--- a/plugin/pulpcore/plugin/stages/declarative_version.py
+++ b/plugin/pulpcore/plugin/stages/declarative_version.py
@@ -11,7 +11,8 @@ from .content_unit_stages import ContentUnitSaver, QueryExistingContentUnits
 
 class DeclarativeVersion:
 
-    def __init__(self, first_stage, repository, mirror=True):
+    def __init__(self, first_stage, repository, mirror=True, inject_before=None,
+                 inject_after=None):
         """
         A pipeline that creates a new :class:`~pulpcore.plugin.models.RepositoryVersion` from a
         stream of :class:`~pulpcore.plugin.stages.DeclarativeContent` objects.
@@ -82,6 +83,14 @@ class DeclarativeVersion:
                  pre-existing units in the :class:`~pulpcore.plugin.models.RepositoryVersion`.
                  'True' is the default.
         """
+        self.inject_before = inject_before or []
+        self.inject_after = inject_after or []
+
+        if not isinstance(self.inject_before, list):
+            self.inject_before = [inject_before]
+        if not isinstance(self.inject_after, list):
+            self.inject_after = [inject_after]
+
         self.first_stage = first_stage
         self.repository = repository
         self.mirror = mirror
@@ -99,9 +108,13 @@ class DeclarativeVersion:
                     QueryExistingContentUnits(), ContentUnitSaver(),
                     ContentUnitAssociation(new_version)
                 ]
+                for elem in self.inject_before:
+                    stages.insert(stages.index(elem[0]), elem[1])
+                for elem in self.inject_after:
+                    stages.insert(stages.index(elem[0]) + 1, elem[1])
                 if self.mirror:
-                    stages.extend([ContentUnitUnassociation(new_version), EndStage()])
-                else:
                     stages.append(EndStage())
+                else:
+                    stages.extend([ContentUnitUnassociation(new_version), EndStage()])
                 pipeline = create_pipeline(stages)
                 loop.run_until_complete(pipeline)


plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -12,7 +12,7 @@
 
 class DeclarativeVersion:
 
-    def __init__(self, first_stage, repository, sync_mode='mirror'):
+    def __init__(self, first_stage, repository, sync_mode='mirror', **kwargs):
```
bmbouter: Please switch these to a named kwarg because that way the options show more explicitly in the options. So that would be s/**kwargs/inject_before=None, inject_after=None/. Since a variable in Python evaluates to False, you can then use:

```
self.inject_before = inject_before or []
self.inject_after = inject_after or []
```

bmbouter: FYI, the time that I do use the kwargs idiom is whenever the arguments are passed through to super() and not used in this class directly.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -86,10 +86,21 @@ def __init__(self, first_stage, repository, sync_mode='mirror'):
...
         """
-
         if sync_mode != 'mirror' and sync_mode != 'additive':
             msg = _("'sync_mode' must either be 'mirror' or 'additive' not '{sync_mode}'")
             raise ValueError(msg.format(sync_mode=sync_mode))
+
+        self.inject_before = []
+        self.inject_after = []
+
+        for key, value in kwargs.items():
```
dparalen: I guess a better way is:
```python
def __init__(self, first_stage, repository, sync_mode='mirror', inject_after=None, inject_before=None):
      self.inject_before = inject_before or []
      self.inject_after = inject_after or []
```

plugin/pulpcore/plugin/stages/declarative_version.py
vdusek: Thanks, updated.

plugin/pulpcore/plugin/stages/declarative_version.py
vdusek: Thanks, updated.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -86,10 +87,18 @@ def __init__(self, first_stage, repository, sync_mode='mirror'):
...
         """
-
         if sync_mode != 'mirror' and sync_mode != 'additive':
             msg = _("'sync_mode' must either be 'mirror' or 'additive' not '{sync_mode}'")
             raise ValueError(msg.format(sync_mode=sync_mode))
+
+        self.inject_before = inject_before or []
+        self.inject_after = inject_after or []
+
+        if not isinstance(self.inject_before, list):
```
bmbouter: This check will fail for other iterable types, e.g. tuples. Can we instead check if it's a subclass of Stage and if it isn't then listify it?

vdusek: I know, but what if there's just a single tuple, fox example `inject_before=('ContentUnitSaver', 'BeforeContentUnitSaver')`. Then the for loops in the `create()` will failed.

bmbouter:
@vdusek Would you feel like making a followup PR to pulp_rpm to work with this one? Because this wasn't available, I had to have pulp_rpm stop using DeclarativeVersion just so that it could add in a custom stage. Can you convert pulp_rpm back to using DeclarativeVersion and have it inject the stage using this feature?

See the pulp_rpm code here:  https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/synchronizing.py#L59-L69

It was introduced in pulp_rpm with this PR:  https://github.com/pulp/pulp_rpm/pull/1168

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: Yup I had forgotten the data format. :) Let's keep this exactly how you have it. Please disregard my request. Thanks for the info.

vdusek: Allright then. :)

vdusek:
@bmbouter Yes, I can try. Should I open a new issue for that or it's not necessary?

bmbouter:
@vdusek pulp_rpm is easy to install, just do a source install w/ `python setup.py develop`, makemigrations, then apply migrations and you should be good to go.

The sync code is runnable from a pulp-smash test that's already in the repo, so configure Pulp smash on this same box with pulp_rpm.

Then run the tests with:  `pytest -v --color=yes --pyargs functional` from within this folder in the repo.  https://github.com/pulp/pulp_rpm/tree/master/pulp_rpm/tests

bmbouter:
@vdusek yes can you open a refactor issue against pulp_rpm please?

gmbnomis:
I fear that the customization that is possible with these parameters is not sufficient for my use case (`pulp_cookbook` seems to be special in every aspect :-( )
As I don't get any digest for cookbook artifacts, I can't use the artifact stages first. What I currently have is:

```
                stages = [
                    self.first_stage,
                    QueryExistingContentUnits(), ExistingContentNeedsNoNewArtifacts(),
                    ArtifactDownloader(), ArtifactSaver(),
                    ContentUnitSaver(),
                    ContentUnitAssociation(new_version)
                ]
```
Of course I could duplicate `QueryExistingContentUnits` and probably leave `QueryExistingArtifacts` in the pipeline, but that would just be duplicated or wasted work.

However, I don't have a nice idea for an  interface at the moment. We could put the creation of the `stages` list in a method of its own that plugin writers could overwrite in derived classes. But this still looks fragile. If Pulp core changes the standard setup for stages, how do we ensure that plugins get this change as well?

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -107,6 +116,10 @@ def create(self):
                     QueryExistingContentUnits(), ContentUnitSaver(),
                     ContentUnitAssociation(new_version)
                 ]
+                for elem in self.inject_before:
+                    stages.insert(stages.index(elem[0]), elem[1])
```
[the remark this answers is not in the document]
bmbouter: @gmbnomis you're right about this. The idea was that it would specify the class object, not the instance. Thank you for pointing this out.

@vdusek the finding of the index to insert before/after needs to be searched using isinstance() and not for a specific instance of that class type, e.g. `stages.index(...)`. Does this make sense to you also?

bmbouter:
@gmbnomis I understand what you're pointing out about the `pulp_cookbook` use case. Thanks for posting your example pipeline.

I actually implemented the subclass design you mentioned a while ago, but I scrapped it. The subclasser wouldn't receives updates to the parent class because it's overriding it. Another option is to make a hook (what I did) that allows the plugin writer to modify the pipeline. The issue with that is that plugin code couldn't know that an upgraded DeclarativeVersion has altered the type/number/order of stages, and it would only learn about this at runtime.

Here's an idea (maybe not a great one). Add 3 keyword params to DeclarativeVersion, `artifact_stages`, `content_unit_stages`, and `associate_unassociate_stages`. Allow the user to specify pipeline segments that replace the default segments used by DeclarativeVersion. Specifying a `[]` would effectivly disable that "section" of the pipeline. To me, the pipeline is first_stage -> artifact_stages -> content_unit_stages -> associate_unassociate_stages. So for your pipeline you would call this new interface with:

If we do ^ instead of merging this PR it would be:

DeclarativeVersion(artifact_stages=[], content_unit_stages=[QueryExistingContentUnits(), ExistingContentNeedsNoNewArtifacts(), ArtifactDownloader(), ArtifactSaver(), ContentUnitSaver()])

Another option is to add ^ feature in addition to this PR, and that would be:

DeclarativeVersion(artifact_stages=[], inject_after=(QueryExistingContentUnits, [ExistingContentNeedsNoNewArtifacts()), ArtifactDownloader(), ArtifactSaver())])

What do you think about these options?

dparalen:
Decorated pipeline factory, anyone?

```python
def pipeline_factory(pipeline_factory):
    def inner(*args, **kwargs):
        def runner(repo):
            mirror = kwargs.pop('mirror', False)
            with WorkingDirectory():
                with RepositoryVersion.create(repo) as new_version:
                    pipeline_stages = pipeline_factory(new_version, *args, **kwargs)
                    pipeline = create_pipeline(pipeline_stages)
                    if mirror:
                         pipeline_stages.append(ContentUnitUnassociation(new_version))
                    pipeline_stages.append(EndStage())
                    loop = asyncio.get_event_loop()
                    loop.run_until_complete(pipeline)
      return runner
return outer
 
@pipeline_factory
def my_pipeline_factory(new_version, blah):
      return [MyStage(new_version), FooStage(blah), BarStage(new_version)]

# to run:
pipeline42mirror = my_pipeline_factory(42, mirror=True)
pipeline42mirror(repo)


# on a class:
class MyStatefulMirrorPipelineFactory:
     def __init__(self):
          self.runs = 0
          self.first_stage = _SecretStage()

     @pipeline_factory
     def __call__(new_version, self):
        self.runs += 1
        return [self.first_stage, FooStage(), BarStage(new_version)]

# to run
pipeline_factory = MyStatefulMirrorPipeline()
pipeline  = pipeline_factory(mirror=False)
pipeline(repo)
pipeline  = pipeline_factory(mirror=True)
pipeline(repo)
print("first stage runs: %s" % pipeline_factory.runs)
```

gmbnomis:
@dparalen: I think that decorator vs. overriding a method is a matter of taste (more or less). Both variants exhibit the problems @bmbouter  is concerned about (and I agree that his concerns are very valid)

@bmbouter: This is really a tough one. Your idea to have segments limits the flexibility somewhat, but essentially has the same problems. Taking

```
DeclarativeVersion(artifact_stages=[], content_unit_stages=[QueryExistingContentUnits(), ExistingContentNeedsNoNewArtifacts(), ArtifactDownloader(), ArtifactSaver(), ContentUnitSaver()])
```

as an example: This call needs intricate knowledge about what is in `artifact_stages` by default since it replicates a part of the artifact stages. Thus a change to those stages might not become part of this pipeline. 

Or, as another example, who is going to ensure that running `QueryExistingContentUnits()` directly after the initial stage will work? (In the default pipeline it is one of the later stages and someone might add a dependency to information gathered in previous stages.)

Basically, if one allows to do such radical changes to the pipeline, I think it might become hard to do any substantial changes to the stages without breaking the assumption some plugin might have made.

I wonder if we could "reverse" the approach we have taken in our discussion and try to unify the pipelines instead. I am thinking about this as the default setup:

```
                stages = [
                    self.first_stage,
                    QueryExistingContentUnits(),
                    QueryExistingArtifacts(), ArtifactDownloader(), ArtifactSaver(),
                    ContentUnitSaver(),
                    ContentUnitAssociation(new_version)
                ]
```

Does running `QueryExistingContentUnits` first hurt any of the use cases (file, rpm, ...) you are aware of? (currently it does not touch artifacts at all).

`QueryExistingArtifacts()` would be changed slightly: It should leave declarative artifacts that already point to a saved artifact alone (which allows a stage that runs before to identify existing artifacts by using content units). Moreover,  QueryExistingArtifacts() should not care about artifacts that have no digest, but that seems already to be the case.

Then, the "hooks" approach of this PR would work for my use case, too. I could add `ExistingContentNeedsNoNewArtifacts()` after `QueryExistingContentUnits()` which makes QueryExistingArtifacts() a no-op.

(In the future, we may find out that it is necessary to actually replace stages for some plugins, but that's perhaps something for later when we have a clearer picture).

dparalen:
> @dparalen: I think that decorator vs. overriding a method is a matter of taste (more or less). Both variants exhibit the problems @bmbouter is concerned about (and I agree that his concerns are very valid)

I hoped that perhaps addressing a problem with a fitting design pattern might be a taste that  would please pretty much everyone. I'm sorry if I missed some point but looking at the code and the suggested usage, the value-added of the `DeclarativeVersion` class are the nested context managers, the adjustment of the pipeline with the `mirror` option and creating and running the pipeline. Since there's no shared state involved (such as the pipe depending on the runtime state of a DeclarativeVersion instance or vice versa), a "compile-time" decorator implementation should do it. It's declarative, after all.

> Does running `QueryExistingContentUnits` first hurt any of the use cases (file, rpm, ...) you are aware of? (currently it does not touch artifacts at all).

I'm afraid that assuming a default pipeline, that will fit them all, will eventually hit the issues that brought us into this same discussion.

On the other hand, letting the plugin writer to fully customize the pipeline by adjusting the stages involved (run-time) gives the most flexibility. This would allow even sub-classing a default stage (to e.g decorate it with some custom code).

gmbnomis:
@dparalen: Don't get me wrong, I think the decorator factory, overwriting a method in a derived class, or a more "classical" factory would all allow full customization (i.e. arrange the pipeline as the plugin developer envisions using existing, new or derived stages).

The question @bmbouter is after (if I understood him correctly) is whether this is too much flexibility. In a plugin API like this, this flexibility might restrict the possibilities to develop the stages further in Pulp core. If the API supports to fully re-arrange/exchange/overwrite stages, almost any change to the existing stages can lead to some undesired side-effect in a plugin.

Basically, we need to find the "sweet spot". Your solution or mine probably allow too much customization, the current implementation is too rigid. 

Therefore, I was wondering whether we can get away with a default pipeline like the one I proposed and the possibility to add custom stages introduced in this PR. At least for the time being, until we know what further customization we actually need.

dparalen:
@gmbnomis I see, thanks for the explanation. I'd prefer a more open solution, with a smallest possible core and the responsibility on the plugin side, with emphasize on customization. This I believe leads to a lean core design with lower API update frequency and a minimum of quirks, which in turn all the plugins would benefit from.

bmbouter:
The more I think about this, the more I think we should do nothing, and instead encourage plugins to use custom pipelines at least for now. The main reasoning is that the downsides aren't that bad, the risks are somewhat concerning, and we will know a *lot* more later after there are 10-20 plugins about what the right pipeline is.

The main downside of carrying a custom pipeline is that plugin writers and their users won't benefit from improvements in DeclarativeVersion itself. I think this is not as bad as it sounds. In practice, the plugin writer likely would have to pass additional options to DeclarativeVersion to benefit from additional features, so they won't really get it for free in many or even most cases. I think this problem is mostly theoretical (after I've considered it a while).

Also the risks of "configuring" something that itself can change (DeclarativeVersion's pipeline order/count/type) just seem high relative to any benefit. If plugin writer's have their own pipelines, they know what they are getting. The Pulp plugin ecosystem will just be overall safer.

This also tells me that we need to ensure that all release notes of changes are really clearly available. With great release notes, any changes in DeclarativeVersion could easily be ported.

@gmbnomis I think we should continue to consider your pipeline idea. It makes sense to me and maybe that is more broadly useful. I think it will be easier after we've seen more examples of them.

@dparalen this seems in line with the smaller core code approach since we aren't adding in a bunch of customization code just yet.

So if we do this we would:
close this PR without merging
close https://pulp.plan.io/issues/3935
close https://pulp.plan.io/issues/3945

@gmbnomis @dparalen how does all this sound?

gmbnomis:
> @gmbnomis @dparalen how does all this sound?

Hmm, doing that means that we basically tell plugin writers: "just copy that code if it does not fit". I am perfectly fine with not giving guidance how to customize pipelines at the moment. But, if we don't give guidance, I would prefer giving the plugin writer the option to choose instead of just copying the thing (OTOH, it's not much code).

Coming back to my original proposal, I would prefer something like this https://github.com/gmbnomis/pulp/blob/pipeline_customization/plugin/pulpcore/plugin/stages/declarative_version.py#L89. (Of course it still has the problems we discussed above, but it offers multiple path for customization)

It splits off the "plugin writer customizable" parts of the pipeline (which do not include ContentUnit(Un)Association because they have the special QuerySet coupling). Now, the plugin writer can:

- Scrap the whole thing if it does not fit
- Overwrite `pipeline_stages()` to instantiate stock or customized stages in any order
- Do "surgery" on the stages list by calling `super().pipeline_stages()` and modifying the result (the idea of this PR).
- Use it as it is if it fits

And it's not much additional code in`declarative_version.py`.

That being said, doing nothing is fine for me as well. It's only a small improvement after all.

bmbouter:
@gmbnomis thank you for bringing this back to your proposal. I think we should do it for the reasons you outline. We can assess the general plugin needs like I was advocating later, just as well with plugin writers enjoying the flexibility you describe now.

Practically speaking, this would allow the 4+ plugins that would carry custom pipelines to switch back to DeclarativeVersion, so that's great too.

Would you be willing to make it into a PR against master?

dparalen:
@gmbnomis @bmbouter 
> Practically speaking, this would allow the 4+ plugins that would carry custom pipelines to switch back to DeclarativeVersion, so that's great too.

thus all our current plugins would have to customize the `DeclarativeVersion` in one way or the other.
@gmbnomis 's solution allows for this so is much better than doing nothing

vdusek:
I'm closing this PR, @gmbnomis can implement his solution and open new one. I'll also unassigned myself from the issue, so @gmbnomis can assign it and describe there his solution.

PR #3607, opened 2018-08-30
Problem: identity fields are serialized as full URLs

dkliban:
Solution: stop passing request to the serializer so it can only form relative URLs

This patch modifies the behavior of DetailIdentityField and it also adds an
additional

ItentityField
RelatedField
NestedRelatedField
NestedIdentityField

to replace

HyperlinkIdentityField
HyperlinkRelatedField
NestedHyperlinkedIdentityField
NestedHyperlinkedRelatedField

provided by DRF and DRF nested.

Required PR: pulp/pulp_file#114

re: #3850
https://pulp.plan.io/issues/3850


dkliban:
I updated the task serializer.

PR #3611, opened 2018-08-31
Fixes broken unit test around TaskStatus model

dkliban:
The test uses a list of various objects as invalid inputs for a Task status. One of the objects is a dict
which mongoengine does not support as a value because it's not hashable. This patch removes the dict from
the list of possible bad inputs.

closes: #3955
https://pulp.plan.io/issues/3955


dkliban:
ok test

PR #3494, opened 2018-05-16
bumps version for beta 3

dkliban:

PR #3614, opened 2018-09-02
Have Travis validate commit messages

daviddavis:
Created a script to validate commit messages. This first pass checks for issues in the commit message and makes sure they're in the correct state.

fixes #3930
https://pulp.plan.io/issues/3930

### DIFF
--- a/.travis/check_commit.sh
+++ b/.travis/check_commit.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# skip this check for everything but PRs
+if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  exit 0
+fi
+
+if [ "$TRAVIS_COMMIT_RANGE" != "" ]; then
+  RANGE=$TRAVIS_COMMIT_RANGE
+elif [ "$TRAVIS_COMMIT" != "" ]; then
+  RANGE=$TRAVIS_COMMIT
+fi
+
+# Travis sends the ranges with 3 dots. Git only wants one.
+if [[ "$RANGE" == *...* ]]; then
+  RANGE=`echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../'`
+else
+  RANGE="$RANGE~..$RANGE"
+fi
+
+for sha in `git log --format=oneline --no-merges "$RANGE" | cut '-d ' -f1`
+do
+  python .travis/validate_commit_message.py $sha
+  VALUE=$?
+
+  if [ "$VALUE" -gt 0 ]; then
+    exit $VALUE
+  fi
+done
--- a/.travis/validate_commit_message.py
+++ b/.travis/validate_commit_message.py
@@ -1,0 +1,42 @@
+import re
+import requests
+import subprocess
+import sys
+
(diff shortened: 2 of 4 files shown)


dralley:
See merge conflict

daviddavis:
@dalley rebased.

.travis/check_commit.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Travis sends the ranges with 3 dots. Git only wants one.
+if [[ "$RANGE" == *...* ]]; then
+  RANGE=`echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../'`
+else
+  RANGE="$RANGE~..$RANGE"
+fi
+
+EXIT=0
```
dralley: ```EXIT``` doesn't get used again, what is it for?

dralley:
I tested this against commit ```e7233c62441875b512bb0903bdfc481fda37f63d``` and got a false negative, presumably because a colon was used (i.e. ```closes: #3963```).

.travis/validate_commit_message.py
```diff
@@ -0,0 +1,44 @@
...
+STATUSES = ['NEW', 'ASSIGNED', 'POST']
+
+sha = sys.argv[1]
+message = subprocess.check_output(['git', 'log', '--format=%B', '-n 1', sha]).decode('utf-8')
+
+
+def check_status(issue):
+    response = requests.get(
+        'https://pulp.plan.io/issues/{}.json'.format(issue)
+    )
```
dralley: This can fit on one line

daviddavis:
Good catch. Thinking about this some, I'm wondering if we need to allow everything that redmine allows. I checked [the regex that redmine uses](https://github.com/redmine/redmine/blob/master/app/models/changeset.rb#L128) hoping to copy it here but it does some other things we don't use like time logging and moreover, it uses Ruby character classes that aren't available in Python. It also allows some things that I think we should forbid (e.g. `fixes!?; #1234` would probably work with the redmine regex). I guess I am more concerned about false positives (i.e. commit messages that pass our checks but don't work with redmine) as we can always tweak the regex as needed later to be more forgiving. I went ahead though and updated the regex to be more lenient. It should now work against `e7233c6` and allow any number of whitespace characters or colon characters.

dralley:
When testing against commit ```c43834c109ab9c627ee35bc4492bb52e8be71bca```, I get the following

```
Error: issue #3494 has invalid status of CLOSED - WONTFIX. Status must be one of NEW, ASSIGNED, POST.
```

daviddavis:
I think this is a valid result and the commit message probably shouldn't include `re #3494`. I can't think of a good reason that `c43834c` should be attached to issue #3494. If a developer wants to reference the issue from the commit, they can do so manually from the commit message (e.g. `This commit relates to redmine issue #3494`). If they want to reference the commit from the redmine issue, they can do that with a comment.

My concern is that a person might create a commit against an issue that's already been released using the `re #1234` syntax and then this fix might not get included in the next release. I'd rather err on the side of caution in such a case. Worst case scenario, the issue can be reopened until the PR is resolved.

PR #3617, opened 2018-09-04
Small usability improvement in docs

bmbouter:
It's good to provide a snippet because everyone won't have to look it
up, and then users won't be inclined to put their password into
https://www.base64encode.org/


dkliban:
This looks good. You can also use the login feature of the browsable docs. I believe @daviddavis added that recently.

PR #3620, opened 2018-09-06
Problem: OpenAPI schema for RepositoryVersion 'retrieve' is wrong

dkliban:
Solution: Create separate serializers for 'create' operation

The problem was that we had a single serializer for both GET and POST. DRF was able to
distinguish which fields were to be used for accepting POST body params and which fields to 
use when serializing for a GET response. DRF knows this because some fields are 'write_only'.
However, the drf_yasg library we use for generating OpenAPI schema does not take into account
the 'write_only' attribute of a field. As a result when it was using the serializer to
generate a schema for the GET response it was including the 'write_only' fields.

This patch introduces a new serializer that is only used for the 'create' operation of the
RepositoryVersion ViewSet. This serializer is only used to inspect the inbound data in the POST
body. It is not used for the response. The response is generated using the OperationPostponed
serializer.

closes: #3966
https://pulp.plan.io/issues/3966

### DIFF
--- a/pulpcore/pulpcore/app/serializers/repository.py
+++ b/pulpcore/pulpcore/app/serializers/repository.py
@@ -413,6 +413,25 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
         help_text=_('A list of counts of each type of content in this version.'),
         read_only=True
     )
+    base_version = NestedRelatedField(
+        required=False,
+        help_text=_('A repository version whose content was used as the initial set of content '
+                    'for this repository version'),
+        queryset=models.RepositoryVersion.objects.all(),
+        view_name='versions-detail',
+        lookup_field='number',
+        parent_lookup_kwargs={'repository_pk': 'repository__pk'},
+    )
+
+    class Meta:
+        model = models.RepositoryVersion
+        fields = ModelSerializer.Meta.fields + (
+            '_href', '_content_href', '_added_href', '_removed_href', 'number',
+            'content_summary', 'base_version'
+        )
+
+
+class RepositoryVersionCreateSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/serializers/repository.py
```diff
@@ -435,5 +427,20 @@ class Meta:
...
+        write_only=True
+    )
+    remove_content_units = serializers.ListField(
+        help_text=_('A list of content units to remove from the latest repository version'),
+        write_only=True
+    )
+
+    class Meta:
+        model = models.RepositoryVersion
+        fields = ['add_content_units', 'remove_content_units']
```
bmbouter: So when creating, are these the only fields that will be serialized? Is that a usability problem?

dkliban: I don't understand the question. Please rephrase.

bmbouter: Also how does moving these fields here resolve the `write_only` issue? I don't understand this unfortunately.

dkliban: The problem is that we had 1 serializer for both GET and POST. However, DRF, was able to distinguish which fields were to be used for accepting POST body params and which fields to use when serializing for a GET. DRF knows this because some fields are 'write_only'. However, the drf_yasg library we use for generating OpenAPI schema does not take into account the 'write_only' attribute of a field. As a result when it was using the serializer to generate a schema for the GET response it was including the 'write_only' fields.

bmbouter: Does it make any sense to include the other fields also on this serializer?

dkliban: this serializer is only used to inspect data in the POST body. the response is the 202 OperationPosteponed

PR #3621, opened 2018-09-06
Add better support for related model saving

bmbouter:
Adds pre_save() and post_save() handlers to ContentUnitSaver that plugin
writers can override on subclasses.

https://pulp.plan.io/issues/3953
closes #3953


bmbouter:
@jortel good call. will fix before I merge and fix in the rpm pr too.

PR #3622, opened 2018-09-07
Problem: REST API docs are built from a static asset

dkliban:
Solution: install and run Pulp for doc testing


.travis/before_script.sh
```diff
@@ -14,3 +14,20 @@ sudo chown -R travis:travis /var/lib/pulp
 sudo cp .travis/server.yaml /etc/pulp/server.yaml
 
 echo "SECRET_KEY: \"$(cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(\-_=+)' | head -c 50)\"" | sudo tee -a /etc/pulp/server.yaml
+
```
dralley: Why move this before?  If there's a flake8 issue, it will increase the time before the failure is apparent.

dkliban: I wanted to move into before because I wanted to separate the install, migrations and starting services (before), and running of tests into 3 distinct scripts. I then use the install and before scripts in the stage that builds and publishes docs using cron.

.travis/publish_docs.sh
```diff
@@ -6,9 +6,13 @@ sudo chmod 600 ~/.ssh/pulp-infra
 echo "docs.pulpproject.org,8.43.85.236 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBGXG+8vjSQvnAkq33i0XWgpSrbco3rRqNZr0SfVeiqFI7RN/VznwXMioDDhc+hQtgVhd6TYBOrV07IMcKj+FAzg=" >> /home/travis/.ssh/known_hosts
 chmod 644 /home/travis/.ssh/known_hosts
 
-pip3 install -r doc_requirements.txt
+export TEST='doc'
```
dralley: I'm a little confused about this.  I would expect this script only to run if the ```docs``` flag is set in the environment by the matrix, in which case why re-export it?

Also, is there a distinction between ```doc``` here and the ```docs``` env variable used elsewhere?

.travis/before_script.sh
```diff
@@ -14,3 +14,20 @@ sudo chown -R travis:travis /var/lib/pulp
...
+# Run migrations.
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
```
dralley: I think all of this should be moved underneath the if statement.  We don't need migrations or workers for a docs test.  I think? that in it's original location, it was gated by an if statement that would just exit earlier in the script if it was a docs test.

.travis/publish_docs.sh
dkliban: Good catch. This was supposed to be 'docs'. I want to simulate the same code path in install.sh and before_script.sh that is taken for testing if the docs can build. Except this time I want to run the publish script so they get published to docs.pulpproject.org.

.travis/before_script.sh
dkliban: The 'make html' command downloads the OpenAPI schema from a running pulp instance on localhost. This requires a running web server.

PR #3623, opened 2018-09-07
Added Dynaconf POC

rochacbruno:
This address the issue 3981

Having the mandatory `DJANGO_SETTINGS_MODULE=pulpcore.app.settings`

This adds dynaconf to manage settings allowing:

On the shell (or in a .env file)
```bash
export PULP_SETTINGS=/path/to/settings.yaml
```

The above can be a list of files as `PULP_SETTINGS=/path/settings.yml,/path/another.toml,other.json` Dynaconf supports `json,ini,toml,yaml,py`

The dynaconf settings files must be `environment` based, that means **at least one** of `default, development, production, global` must be defined.

Ex: `/path/to/settings.toml`
```toml
[default]
FOO = 1
[development]
FOO = 2
[production]
FOO = 3
```

Or `/path/to/settings.yaml`

```yaml
default:
  foo: 1
development:
  foo: 2
production:
  foo: 3
```

To change the working env `PULP_ENV=production` - With no change the `[default]` is loaded then the `[development]` is loaded to override it. **it is possible to use only the default** or completely ignore the settings file loading anything from `envvars` prefixed by `PULP_` (it is also possible to use global **[pulp]** environment in settings files).

Example:

```bash
export PULP_TEST=HelloPulp
export PULP_FOO=4  # overrrides the `foo` defined in config file
export PULP_SECRET_KEY=blablabla  # overrides django secret key
export PULP_DATABASES={....}   # overdides django database settings
export PULP_LANGUAGE_CODE='pt-BR'  # overrides django.conf.settings.LANGUAGE_CODE
```

With some of the above it is possible to check if `dynaconf `loaded vars with dynaconf cli

```bash
$ dynaconf list
.. all settings prints to the shell

$ dynaconf list -k foo
Django app detected
Working in development environment 
FOO: 4

$ dynaconf list -k databases
Django app detected
Working in development environment 
DATABASES: {'default': {'CONN_MAX_AGE': 0,
             'ENGINE': 'django.db.backends.postgresql_psycopg2',
             'NAME': 'pulp',
             'USER': 'pulp'}}
```

Optionally Dynaconf can also **validate** the config file, this is useful to ensure users will write good config files.

```bash
 $ dynaconf validate
Django app detected
/pulp/dynaconf_validators.toml not found
```

Then we can write a `dynaconf_validators.toml` settings specs of rules that users may follow when writing  config files. examples: https://github.com/rochacbruno/dynaconf/tree/master/example/validators/with_toml

### DIFF
--- a/pulpcore/pulpcore/app/settings.py
+++ b/pulpcore/pulpcore/app/settings.py
@@ -14,8 +14,6 @@ from contextlib import suppress
 from importlib import import_module
 from pkg_resources import iter_entry_points
 
-import yaml
-
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/settings.py
```diff
@@ -216,64 +265,7 @@
...
-
-    return settings
-
-
-# Read PULP_SETTINGS environment variable to find the location of server.yaml,
-# defaults to /etc/pulp/server.yaml
-PULP_SETTINGS = os.getenv('PULP_SETTINGS', '/etc/pulp/server.yaml')
-load_settings([PULP_SETTINGS])
+# load pulp default settings
+# maybe it is better to define it directly in this settings file?
```
bmbouter: I think it would be better to define the settings in this settings file. To load them into another structure and then pull them out again seems unnecessary.

pulpcore/tests/unit/test_settings.py
```diff
@@ -1,62 +1,64 @@
-from unittest import TestCase, mock
+# TODO: Implement a dynaconf test case here
```
bmbouter: We've planned to add tests later in many areas so I'm ok to remove this with that same thinking. Otherwise someone will need to add them before we can merge. @rochacbruno what should we do?

pulpcore/pulpcore/app/settings.py
```diff
@@ -36,10 +34,61 @@
...
+That means that `export PULP_FOO=1` will be loaded to `django.conf.settings.FOO
+On command line it is possible to check it with `dynaconf list -k foo`
+"""
+
+ENVVAR_FOR_DYNACONF = "PULP_SETTINGS"
+"""This defines which path dynaconf will look to load config files
+example: `export PULP_SETTINGS=/path/to/settings.toml` and the format can be
+.ini, .json, .yaml or .toml
+
+e.g::
```
bmbouter: I think these things would be more useful in the docs themselves. This file won't be read by many users in practice since it's installed w/ the pulp code in the python path. Can this be moved to the docs?

dkliban:
Thanks for the PR. I made some changes on top of it and it got merged as part of https://github.com/pulp/pulp/pull/3663

PR #3625, opened 2018-09-11
Bumps version for Beta 7 release notes

dkliban:


plugin/setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    'pulpcore',
+    'pulpcore>=3.0.0b7',
```
asmacdo: Non-blocking question: Shouldn't we pin to pulpcore 3.0.0b7? Presumably the next beta is likely to be incompatible, and we won't know that until after we release this, and it would be too late to change.

dkliban: I don't think we can do that. In the future, if pulpcore is updated and pulpcore-plugin is not updated in the next release, pip won't allow for pulpcore to get updated because pulpcore-plugin requires that older version.

PR #3631, opened 2018-09-13
Reduction of information in response to operation postponed

vdusek:
Required PR: PulpQE/pulp-smash#1137

closes #3978
https://pulp.plan.io/issues/3978

### DIFF
--- a/pulpcore/pulpcore/app/response.py
+++ b/pulpcore/pulpcore/app/response.py
@@ -23,6 +23,5 @@ class OperationPostponedResponse(Response):
                 the response.
             request (rest_framework.request.Request): Request used to generate the _href urls
         """
-        task = {"_href": reverse('tasks-detail', args=[result.id], request=None),
-                "task_id": result.id}
+        task = {"task": reverse('tasks-detail', args=[result.id], request=None)}
         super().__init__(data=task, status=202)
--- a/pulpcore/tests/functional/api/test_tasks.py
+++ b/pulpcore/tests/functional/api/test_tasks.py
@@ -41,7 +41,7 @@ class TasksTestCase(unittest.TestCase):
         self.addCleanup(self.client.delete, repo['_href'])
         attrs = {'description': utils.uuid4()}
         response = self.client.patch(repo['_href'], attrs)
-        self.task.update(self.client.get(response['_href']))
+        self.task.update(self.client.get(response['task']))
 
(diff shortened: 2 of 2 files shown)


daviddavis:
LGTM. You have some test failures though.

daviddavis:
Also, we should probably update all the READMEs. Here's pulp_file for example:

https://github.com/pulp/pulp_file/blob/master/README.rst#use-the-bar-publisher-to-create-a-publication

vdusek:
One of the solution is rename `task` field back to `_href`, as it's done now. The second solution is to leave field named `task` and edit the tests. Specifically edit keys from `'_href'` to `'task'` here - [pulp_smash/api.py#L505](https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L505) and here - [pulpcore/tests/functional/api/test_tasks.py#L44](https://github.com/pulp/pulp/blob/master/pulpcore/tests/functional/api/test_tasks.py#L44). However it could seem kind of confusing from the perspective of creating tests, because key `'_href'` is used almost everywhere. What do you think @daviddavis, @dkliban?

daviddavis:
I lean towards updating the tests.

Also, ping @pulp/qe for their thoughts.

vdusek:
I added to the commit message reference to the required PR from pulp-smash, according [docs](https://docs.pulpproject.org/en/3.0/nightly/contributing/continuous-integration.html#requiring-other-pull-requests) (thanks @dkliban). I updated and rebased both PRs, just for sure, but travis-ci still didn't pass. So please, if anyone could tell me where's the problem or what I've been doing wrong, I'll be glad. Because locally checks pass without any trouble.
```
======================================== short test summary info ========================================
SKIP [1] tests/functional/api/test_auth.py:57: https://pulp.plan.io/issues/3248
SKIP [1] tests/functional/api/test_auth.py:48: https://pulp.plan.io/issues/3248
SKIP [1] tests/functional/api/test_auth.py:108: https://pulp.plan.io/issues/3248
SKIP [1] tests/functional/api/test_auth.py:94: https://pulp.plan.io/issues/3248

================================ 88 passed, 4 skipped in 317.55 seconds =================================
```

daviddavis:
It looks we removed the functionality that lets you require pulp-smash PRs when we moved the pulp-smash tests into the pulp repo. I think we need to merge the pulp-smash change first and then re-run this PR.

PR #3639, opened 2018-09-16
Problem: href values in browsable API are not clickable

dkliban:
Solution: add a filter that makes links out of href

This patch adds a copy of the rest_framework api template to this repository. This was necesary in order to use
the new custom "urlize_quoted_links" filter. The filter turns any string that starts with '/pulp/api/v3' into a
anchor tag with that value.

closes #3987
https://pulp.plan.io/issues/3987

### DIFF
--- a/pulpcore/pulpcore/app/settings.py
+++ b/pulpcore/pulpcore/app/settings.py
@@ -90,7 +90,7 @@ ROOT_URLCONF = 'pulpcore.app.urls'
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': [os.path.join(BASE_DIR, 'templates')],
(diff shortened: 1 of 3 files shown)


bmbouter:
This works really well for me. The only usability issue is on the schema page itself `localhost:8000/pulp/api/v3/`. There are two interesting things about that page:

* urls with `{` or `}` are clickable like:   ` /pulp/api/v3/publications/{id}/`.
* The url in the `_meta` section is not clickable.  Specifically on the `/pulp/api/v3/` page there is a section that looks like:

```
"_meta": {
        "url": "http://localhost:8000/pulp/api/v3/",     #  <--- not clickable
        "title": "Pulp API"
    },
```

These aren't really that important, but if they are easy to fix, that would be good. @dkliban what do you want to do? I can lgtm because it's working great on all pages but that one.

PR #3640, opened 2018-09-16
Actually show members in Stages API documentation

gmbnomis:
Apparently, `:special-members:` does nothing if `:members:` is not given as
well.  Add `:members:` where required.

[noissue]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ Vagrantfile
 # Built documentation
 docs/_build
 docs/_diagrams
+docs/api.yaml
 
 # Django migrations
 # Ignored during early development to prevent accidental commits;
--- a/docs/plugins/plugin-api/stages.rst
+++ b/docs/plugins/plugin-api/stages.rst
@@ -34,9 +34,11 @@ Stages API
 .. autofunction:: pulpcore.plugin.stages.create_pipeline
 
 .. autoclass:: pulpcore.plugin.stages.Stage
+   :members:
    :special-members: __call__
 
 .. autoclass:: pulpcore.plugin.stages.EndStage
+   :members:
    :special-members: __call__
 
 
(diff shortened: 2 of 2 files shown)


bmbouter:
I built the docs with and without this change, but I didn't see a difference. Without this change I can see the `:member:` attributes because we have that enabled as a global default with the [`autodoc_default_options`](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options). We specify it in our Sphinx conf file [here](https://github.com/pulp/pulp/blob/master/docs/conf.py#L107). Did this make a visible change for you?

gmbnomis:
The current publication (https://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-api/stages.html#stages-api) looks fine now. But yesterday (with the old layout) the `__call__' methods weren't there and my local rendering looked the same. Perhaps that was a side effect of the change to Sphinx 1.8 (rendering with 1.7 but using the 1.8 config).

Anyway, I am closing this PR.

bmbouter:
@gmbnomis cool; good looking out anyway. Thanks!

dkliban:
@gmbnomis - I don't think the content changed since yesterday. The only
difference is that we merged a PR[0] that once again installs the RTD theme
when building docs.

[0]
https://github.com/pulp/pulp/pull/3641/commits/55101fc4553faba18813c2876d4a31ca04e173e7

gmbnomis:
@dkliban: I think the problem with my locally built documentation originated from building with a Sphinx 1.7 using the new Sphinx 1.8 config (which changed the way o specify default options for autodoc).

I think I checked against the nightly version published on the web before posting this PR, but I may also simply have mixed up things.

After upgrading, my local build looks fine now and the published documentation looks fine as well. 

Sorry for the confusion.

PR #3643, opened 2018-09-17
Add async batches iterator

gmbnomis:
Implementation of batches iterator and refactoring of existing stages. Changes to ArtifactDownloader  (https://pulp.plan.io/issues/4018) will be a separate PR..

### DIFF
--- a/plugin/pulpcore/plugin/stages/api.py
+++ b/plugin/pulpcore/plugin/stages/api.py
@@ -23,6 +23,59 @@ class Stage:
         """
         raise NotImplementedError(_('A plugin writer must implement this method'))
 
+    @staticmethod
+    async def batches(in_q, minsize=1):
+        """
+        Asynchronous iterator yielding batches of :class:`DeclarativeContent` from `in_q`.
+
+        The iterator will try to get as many instances of
+        :class:`DeclarativeContent` as possible without blocking, but
+        at least `minsize` instances.
+
+        Args:
+            in_q (:class:`asyncio.Queue`): The queue to receive
+                :class:`~pulpcore.plugin.stages.DeclarativeContent` objects from.
+            minsize (int): The minimum batch size to yield (unless it is the final batch)
+
+        Yields:
+            A list of :class:`DeclarativeContent` instances
+
+        Examples:
+            Used in stages to get large chunks of declarative_content instances from
+            `in_q`::
+
+                class MyStage(Stage):
+                    async def __call__(self, in_q, out_q):
+                        async for batch in self.batches(in_q):
+                            for declarative_content in batch:
+                                # process declarative content
+                                await out_q.put(declarative_content)
+                        await out_q.put(None)
+
+        """
+        batch = []
+        shutdown = False
+
+        def add_to_batch(batch, content):
+            if content is None:
+                return True
+            batch.append(content)
+            return False
+
+        while not shutdown:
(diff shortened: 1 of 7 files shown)


plugin/pulpcore/plugin/stages/api.py
```diff
@@ -23,6 +23,56 @@ class Stage:
...
+            minsize (int): The minimum batch size to yield (unless it is the final batch)
+
+        Yields:
+            A list of :class:`DeclarativeContent` instances
+
+        Examples:
+            Used in stages to get large chunks of declarative_content instances from
+            `in_q`::
+
+                class MyStage(Stage):
```
bmbouter: What about using the >>> syntax here? I think that would give it python highlighting iirc. I use it here:

https://github.com/gmbnomis/pulp/blob/pipeline_customization/plugin/pulpcore/plugin/stages/declarative_version.py#L42-L56
Which renders like this:  https://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-api/stages.html?highlight=myfirststage#declarativeversion

plugin/pulpcore/plugin/stages/api.py
```diff
@@ -23,6 +23,56 @@ class Stage:
         """
         raise NotImplementedError(_('A plugin writer must implement this method'))
 
+    @staticmethod
+    async def batches(in_q, minsize=1):
+        """
+        Asynchronous iterator yielding batches of :class:`DeclarativeContent` from `in_q`. The
```
bmbouter: Consider adding a single sentence description at the top of this docstring. Some docstring parsers use the single-top line meaningfully.

gmbnomis:
> Can this be added somewhere in the docs? Probably somewhere here? https://github.com/pulp/pulp/blob/master/docs/plugins/plugin-api/stages.rst

As "batches" is a member of Stages, it will appear automatically. Or do you mean additional documentation in addition to the docstring?

plugin/pulpcore/plugin/stages/api.py
gmbnomis: This is a [literal block](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/basics.html#literal-blocks). `::` is the nicest thing in reST IMHO: No overhead for all practical purposes, looks completely natural in the source and gives nice Python highlighting in the docs (or whatever the default language is).

`>>>` is a [doctest block](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/basics.html#doctest-blocks) and serves a different purpose: "interactive Python sessions cut-and-pasted into docstrings"

plugin/pulpcore/plugin/stages/api.py
gmbnomis: done

plugin/pulpcore/plugin/stages/api.py
bmbouter: @gmbnomis Thank you for explaining this difference to me.

simon-baatz:
@bmbouter Thanks for the review! Published docs are missing members because the config is wrong still. The parameter has to be renamed to the 1.7 name (writing on my mobile, not so easy to look it up right now)

bmbouter:
@gmbnomis I now understand what you were telling me. I agree the reason it's not showing up in the docs is due to a config sphinx issue.

PR #3645, opened 2018-09-18
Rebuild content unit counts after unassociating units from a repo.

goosemania:
To avoid mistake sin counting removed units, especially in the cases
when there are dependant units which are removed as well.

re #3985
https://pulp.plan.io/issues/3985


dparalen:
@goosemania thanks for the fix!
LGTM; FWIW the [function]( https://github.com/pulp/pulp/blob/04a70ea6820f11f9e23f24cb31cd48e837ea3555/server/pulp/server/controllers/repository.py#L293) used now should have a speedup effect even

PR #3654, opened 2018-09-20
Add support for pass-through publications.

jortel:
https://pulp.plan.io/issues/4020

Adds support for pass-through publications.  

Publishers would still use the `Publication` _context_ and may (optionally) set pass-through=True. In combination with this, the publisher can (optionally) add `PublishedMeta` and possibly `PublishedArtifacts` to customize how the publication is composed.

I still need to run EXMPLAIN to see what django has translated the content-artifact query to be sure it's efficient and using the indexes.

I'm not sure where/how to update the docs.  I'm inclined to capture pass-through in planned more comprehensive docs rewrite stores.

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -16,6 +16,9 @@ class Publication(Model):
 
     Fields:
         complete (models.BooleanField): State tracking; for internal use. Indexed.
+        pass_through (models.BooleanField): Indicates that the publication is a pass-through
+            to the repository version. Enabling pass-through has the same effect as creating
+            a PublishedArtifact for all of the content (artifacts) in the repository.
 
     Relations:
         publisher (models.ForeignKey): The publisher that created the publication.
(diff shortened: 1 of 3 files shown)


daviddavis:
I was thinking that pass_through was going to only be available to plugin writers. I have two concerns about this solution. First, I wonder if users (as opposed to plugin writers) can understand the intricacies of pass-through vs not. How will they know whether they need to create a pass-through publication vs not? Also, what if I have plugin and I don't want to allow my content to be published via pass-through publications?

daviddavis:
I think to kind of address the concerns in my previous comment, we could remove the `POST /pulp/api/v3/publications/` and instead let plugin writers define their own endpoints to utilize the pass_through. So for example, the Ansible plugin might define its own `POST /pulp/api/v3/ansible/publications` endpoint.

bmbouter:
This PR is exactly as I imagined it, including the create that only allows `pass_through=True`. I hadn't considered what @daviddavis was saying though. Now I agree though that we should remove the ability to create Publications via the API and leave it for plugin writers to do via the plugin API. For me it's a safety issue. For plugins where its author expects to only use PublishedMetadata and/or PublishedArtifact, its users cannot use pass_through=True safely.

It's also not needed for the immediate use case. With pulp_ansible the plugin writer is making the Publication via plugin-code so the core API call isn't used there.

jortel:
I will remove the publication POST endpoint and supporting changes.

daviddavis:
LGTM

PR #3655, opened 2018-09-20
Clarify language around publishing in the planning guide

daviddavis:


docs/plugins/plugin-writer/planning-guide.rst
```diff
@@ -147,8 +147,8 @@ for more information.
 Publishing Your Content Units
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Pulp organizes content units into repositories, and the user publishes an entire repository. It's
-not possible to publish a content unit by itself, or a partial repository of content units.
+In order to make content unit files available to clients, users must publish these files. Typically,
+users will publish a repository which will make all the units in that repository available.
```
jortel: The term `units` should be used sparingly, not capitalized and only when proceeded by the word _content_.
Suggest replacing:

_make all the units in that repository available_

with:

_make the content in the repository available_

PR #3657, opened 2018-09-20
Problem: ArtifactFileField can delete an existing artifact's file

dkliban:
Solution: Add a check to see if the source file is already in artifact storage

Prior to this patch it was possible to break an existing artifact by passing a path to it's file
as the file for a new Artifact.

This patch also removes extraneous code in the ArtifactSaver stage. The code was present because
an assumption was made that bulk_save() on Artifacts would not call the presave handler on the
ArtifactFileField. However, this is not the case and the pre_save handler is getting called.

closes: #3915
https://pulp.plan.io/issues/3915

### DIFF
--- a/pulpcore/pulpcore/app/models/fields.py
+++ b/pulpcore/pulpcore/app/models/fields.py
@@ -1,3 +1,7 @@
+import os
+from gettext import gettext as _
+
+from django.conf import settings
 from django.db.models import FileField
 
 from pulpcore.app.files import TemporaryDownloadedFile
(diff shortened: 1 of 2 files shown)


pulpcore/pulpcore/app/models/fields.py
```diff
@@ -23,6 +26,11 @@ def pre_save(self, model_instance, add):
             Field's value just before saving.
 
         """
+        file_name = str(model_instance.file)
+        if file_name.startswith(os.path.join(settings.MEDIA_ROOT, 'artifact')):
+            raise Exception('The file referenced by the Artifact is already present in Artifact '
```
bmbouter: Can this get gettext translated please?

pulpcore/pulpcore/app/models/fields.py
```diff
@@ -23,6 +27,11 @@ def pre_save(self, model_instance, add):
             Field's value just before saving.
 
         """
+        file_name = str(model_instance.file)
+        if file_name.startswith(os.path.join(settings.MEDIA_ROOT, 'artifact')):
+            raise Exception(_('The file referenced by the Artifact is already present in '
```
jortel: I think raising base `Exception` should be avoided.  Perhaps `ValueError` would be appropriate.

https://docs.python.org/3/library/exceptions.html#ValueError

PR #3661, opened 2018-09-21
Implement single blocking asyncio.wait for ArtifactDownloader

gmbnomis:
The ArtifactDownloader had two locations where it could block: dequeue from
the upstream stage queue and waiting for downloads to finish.

When downloads were ongoing, no new work could be accepted (even when below
the limit of concurrent downloads).

Fix this by waiting on both upstream content units and downloads at the same
time.  Additionally, refactor the huge method into a "run" class with state.
This allows to split up the functionality into smaller methods without
passing a large set of parameters around.

Restructure the tasks: Tasks are handling one content instance in the form
of a single coroutine now.  The coroutine computes the artifacts to
download, downloads them and updates the content instance.  The overall limit
on the number of concurrent downloads is restricted by a common semaphore.

Add unit tests using simulated downloads and clocks to verify the dynamic
behavior of ArtifactDownloader.

closes #4018
https://pulp.plan.io/issues/4018

### DIFF
--- a/plugin/pulpcore/plugin/stages/artifact_stages.py
+++ b/plugin/pulpcore/plugin/stages/artifact_stages.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 
 from django.core.files import File
 from django.core.files.storage import DefaultStorage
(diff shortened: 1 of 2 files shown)


gmbnomis:
@bmbouter: Here is the stage API ArtifactDownloader change we talked about in https://pulp.plan.io/issues/4018. The change is larger than I originally intended, but the downloader now has both, a limit on concurrent downloads and a limit on content units currently being worked on. I would love to have your feedback on this.

PR #3662, opened 2018-09-21
Problem: pulpcore-common is not used outside of pulpcore

dkliban:
Solution: move constants into pulpcore.constants and remove pulpcore-common package

closes: #3652
https://pulp.plan.io/issues/3652


docs/contributing/platform-api/common.rst
```diff
@@ -1,11 +1,5 @@
-pulp.common
-===============
+pulp.constants
```
daviddavis: Should this file be renamed to constants.rst?

CodeHeeler: +1 renaming to constants.rst for consistency

dkliban:
I renamed the file.

PR #3663, opened 2018-09-21
Problem: Pulp can't be configured using environment variables

dkliban:
Solution: use dynaconf to manage all of Pulp configuration settings

This patch also adds pulpcore.rqconfig module that is dynamically generated from the Django settings
generated by dynaconf. The rq workers use this module for their configuration.

The default location of the config file is now at `/etc/pulp/settings.py`. Pulp can operate without by relying
on environment variables.

closes: #3981
https://pulp.plan.io/issues/3981

closes: #3879
https://pulp.plan.io/issues/3879

closes: #3980
https://pulp.plan.io/issues/3980

closes #3618
https://pulp.plan.io/issues/3618

closes: #3943
https://pulp.plan.io/issues/3943


docs/installation/configuration.rst
```diff
@@ -0,0 +1,101 @@
...
+=============
+
+.. _configuration:
+
+-----------
+
+Pulp uses `dynaconf <https://dynaconf.readthedocs.io/en/latest/>`_ for its settings. dynaconf
+allows storing `settings in multiple file formats <https://dynaconf.readthedocs
+.io/en/latest/guides/examples.html>`_. The location of the settings file is specified by setting
+the ``PULP_SETTINGS`` environment variable. Each of the settings can also be set by prepending
```
daviddavis: I feel like there should be a default settings file location so a user won't necessarily have to muck with env vars if they don't want to?

docs/installation/configuration.rst
```diff
@@ -0,0 +1,101 @@
...
+
+     Below is the default configuration written in Python.
+
+.. code-block:: python
+   :linenos:
+
+   CONTENT = {
+      'HOST': None,
+      'WEB_SERVER': 'django',
+      'REDIRECT': {
```
daviddavis: I'm curious about how to define CONTENT and CONTENT['REDIRECT'] as environment vars?

daviddavis:
I'm a bit hesitant that we're removing existing unit tests here and not adding new ones.

docs/installation/configuration.rst
rochacbruno: @daviddavis @dkliban there are 2 ways.

1. TOML  (recommended)

When reading environment variables dynaconf will always parse as a TOML value, so 

```bash
$ export PULP_FOO="{ name = 'Bruno', number =  42 }"
$ dynaconf list -k foo
{
   'name': 'Bruno',
   'number': 42
}
```

https://github.com/toml-lang/toml#inline-table 

2. `@` casting  (deprecated)

Previous versions of dynaconf used a specisl `@` casting (that was before the toml implementation) but still works

```bash
export PULP_FOO="@json {"name": "Bruno", "number": 42}"
export PULP_NUMBER="@float 73.5"
```

rochacbruno: More details: https://dynaconf.readthedocs.io/en/latest/guides/environment_variables.html#precedence-and-type-casting

docs/installation/configuration.rst
dkliban: I file a RFE with dynaconf for this feature. https://github.com/rochacbruno/dynaconf/issues/87

PR #3664, opened 2018-09-23
Allow simpler pipeline customization in DeclarativeVersion

gmbnomis:
For most custom pipelines it should be sufficient to override
`DeclarativeVersion.pipeline_stages()` now.

fixes #3935
https://pulp.plan.io/issues/3935


bmbouter:
Also looks good to me! Thank you @gmbnomis !

PR #3670, opened 2018-09-26
Confine symlink paths under the build_dir

dparalen:
The file distributor doesn't ensure the path of a symlink being created is
contained under the `build_dir`. As a result, a rogue input such as an ISO
Manifest that contains relative paths, could make Pulp write content to an
arbitrary system folder upon publish.

This patch prevents the issue by always confining symlink paths under the
`build_dir`.

Fixes: #3841
https://pulp.plan.io/issues/3841

### Pull Requests

Please be sure you have read our documentation regarding contributing
here:

There you can read a Contribution Checklist and learn more about
appropriate branches and rebasing among other things.

Once you are ready to create a pull request:

- [ ] For properly formatting commit messages, please see the
documentation here:
- [ ] Once you have created your pull request, please add a link to
your pull request into the notes/comments for the relevant bug in the
issue tracker and change the status to POST.
- [ ] Continuous integration tests will be automatically triggered
by your pull request.  You can see the status of these tests on the
page for your pull request.  If there are failures, please try to
fix them and update your pull request.

Thank you so much for contributing to Pulp!


dparalen:
works like this right now:
```
(pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo list
                            ISO Repositories

Id:                  alice
Display Name:        None
Description:         None
Content Unit Counts:
  Iso: 1


(pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo create --repo-id mallory --feed  http://localhost/~vagran
t/pulp_iso_repo/m1/m2/m3/m4/m5/m6/m7/m8/m9/ --relative-url mallory  --validate true --serve-https true
--remove-missing true
Successfully created repository [mallory]

(pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo sync run --repo-id mallory
                   Synchronizing Repository [mallory]

This command may be exited via ctrl+c without affecting the request.


The Pulp Manifest was downloaded successfully.

Downloading 1 ISOs...
[==================================================] 100%
ISOs: 1/1       Data: 2.3 MB/2.3 MB     Avg: 1.2 MB/s


Successfully downloaded 1 ISOs.


Task Succeeded



The repository was successfully published.


Task Succeeded


(pulp) [vagrant@pulp2 ~]$ sudo find /var/lib/pulp/published/ -name *.iso
/var/lib/pulp/published/https/isos/mallory/lib/pulp/published/https/isos/alice/pulp.iso
/var/lib/pulp/published/https/isos/alice/pulp/python.iso
pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo list
                            ISO Repositories

Id:                  alice
Display Name:        None
Description:         None
Content Unit Counts:
  Iso: 1


(pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo create --repo-id mallory --feed  http://localhost/~vagran
t/pulp_iso_repo/m1/m2/m3/m4/m5/m6/m7/m8/m9/ --relative-url mallory  --validate true --serve-https true
--remove-missing true
Successfully created repository [mallory]

(pulp) [vagrant@pulp2 ~]$ pulp-admin iso repo sync run --repo-id mallory
                   Synchronizing Repository [mallory]

This command may be exited via ctrl+c without affecting the request.


The Pulp Manifest was downloaded successfully.

Downloading 1 ISOs...
[==================================================] 100%
ISOs: 1/1       Data: 2.3 MB/2.3 MB     Avg: 1.2 MB/s


Successfully downloaded 1 ISOs.


Task Succeeded



The repository was successfully published.


Task Succeeded


(pulp) [vagrant@pulp2 ~]$ sudo find /var/lib/pulp/published/ -name *.iso
/var/lib/pulp/published/https/isos/mallory/lib/pulp/published/https/isos/alice/pulp.iso
/var/lib/pulp/published/https/isos/alice/pulp/python.iso
```

server/pulp/plugins/file/distributor.py
```diff
@@ -212,9 +213,13 @@ def _symlink_unit(self, build_dir, unit, target_paths):
         :param target_paths: The list of paths the unit should be symlinked to.
         :type  target_paths: list of L{str}
         """
+        cwd = os.path.abspath(os.path.curdir)
         for target_path in target_paths:
-            # symlink_filename = os.path.join(build_dir, unit.unit_key['name'])
-            symlink_filename = os.path.join(build_dir, target_path)
+            # any two absolute paths always share at least the root element
```
jortel: Wouldn't it be simpler to just ensure the target path is not absolute (leading /)?

dparalen: I'm wondering whether that wouldn't introduce backwards incompatibility within a bugfix; I wanted to make sure if there are users who legitimately used absolute or dotted path elements not to break them. But chances are I'm over complicating it.

server/pulp/plugins/file/distributor.py
```diff
@@ -212,9 +213,13 @@ def _symlink_unit(self, build_dir, unit, target_paths):
         :param target_paths: The list of paths the unit should be symlinked to.
         :type  target_paths: list of L{str}
         """
+        cwd = os.path.abspath(os.path.curdir)
```
gmbnomis: This looks just wrong: Why is all this path computation done with `cwd` and then, the resulting path is appended to `build_dir`?

Anyway, both `cwd` and `build_dir` are implementation details (and may depend on configuration, I don't know). The validation of `target_path` should not depend on the location of `cwd` nor `build_dir`.

server/pulp/plugins/file/distributor.py
gmbnomis: As said in https://pulp.plan.io/issues/3841#note-6, there is no valid use cases for absolute paths I know of.

Forcing absolute paths under `build_dir` at places depending on how much they happen to share with `cwd` does not make sense to me. Why should `/var/my_file` end up at `<build_dir>/my_file` just because it happens to share the `/var` with `cwd`?

server/pulp/plugins/file/distributor.py
```diff
@@ -212,9 +213,13 @@ def _symlink_unit(self, build_dir, unit, target_paths):
...
         for target_path in target_paths:
-            # symlink_filename = os.path.join(build_dir, unit.unit_key['name'])
-            symlink_filename = os.path.join(build_dir, target_path)
+            # any two absolute paths always share at least the root element
+            abs_target_path = os.path.abspath(target_path)
+            target_path_cwd_prefix = os.path.commonprefix([cwd, os.path.dirname(abs_target_path)])
+            confined_target_path = abs_target_path.replace(target_path_cwd_prefix, '').lstrip('/')
```
gmbnomis: This is buggy. Consider what happens to `/var../one/up` (if `cwd`start with "/var") or `/../b/o/g/u/s`.

If  we really want to support relative paths, I am thinking of the following checks:
- target_path must not be absolute
- target_path must not end with `/` (indicating a directory, for which we can't have a  valid digest or length)
- os.path.normpath(target_path) must not begin with `../` (which indicates that the relative path leaves the working directory)
- os.path.normpath(target_path) must not be `..` (which indicates that the relative path leaves the working directory)

A path that violates these requirement should raise an error. 

If such a path happens to work with the current implementation, it writes a file at a location outside of the build directory and the resulting published repo is broken as this file is not within the published tree (and has a manifest file pointing to a non-existent or wrong location).

I think it is justifiable to raise an error in this case.

server/pulp/plugins/file/distributor.py
dparalen: The reason is that `abspath`, when e.g expanding `..`, calculates based on `cwd` rather than the `build_dir` thus the prefix calculation has to follow suit or one has to `$ cd build_dir` first; I chose the former.
The semantics of an absolute `target_path` here is as if it was a `chroot` into the repository (`build_dir`):
```
In [22]: build_dir = '/var/lib/pulp/build_dir/my_repo'

In [23]: cd ~
/home/vagrant

In [24]: cwd = os.path.abspath(os.path.curdir)

In [25]: cwd
Out[25]: '/home/vagrant'

In [26]: target_path = '/foo/bar/baz'

In [27]: abs_target_path = os.path.abspath(target_path)

In [28]: target_path_cwd_prefix = os.path.commonprefix([cwd, os.path.dirname(abs_target_path)])

In [29]: confined_target_path = abs_target_path.replace(target_path_cwd_prefix, '', 1).lstrip('/')

In [30]: os.path.join(build_dir, confined_target_path)
Out[30]: '/var/lib/pulp/build_dir/my_repo/foo/bar/baz'
```

server/pulp/plugins/file/distributor.py
dparalen: @gmbnomis the bug is in my `replace` use; it has to be limited to a single occurence. Your example paths then get confined as expected:
```
In [1]: import os                                                                                                                                                                                          [2/1679]
                                                                                                                                                                                                                   
In [2]: cwd = os.path.abspath(os.path.curdir)                                                                                                                                                                      

In [3]: cwd                                                                                        
Out[3]: '/home/vagrant/devel/pulp'                        
                                                        
In [4]: cd /var/                                                                
/var                                        
            
In [5]: cwd = os.path.abspath(os.path.curdir) 
                                         
In [6]: cwd                                                                     
Out[6]: '/var'                                                      
                                                                            
In [7]: target_path = '/var/../one/up'                     
                                                                                                   
In [8]: abs_target_path = os.path.abspath(target_path)                                             
                                                                             
In [9]: abs_target_path                                                              
Out[9]: '/one/up'                                                                                  
                                                                                                    
In [10]: target_path_cwd_prefix = os.path.commonprefix([cwd, os.path.dirname(abs_target_path)])
                                     
In [11]: target_path_cwd_prefix
Out[11]: '/'

In [12]: confined_target_path = abs_target_path.replace(target_path_cwd_prefix, '').lstrip('/')
                                    
In [13]: confined_target_path                                                                                                                                                                                     
Out[13]: 'oneup'                                                               
                                              
In [14]: abs_target_path.replace(target_path_cwd_prefix, '')  
Out[14]: 'oneup'
                                                
In [15]: abs_target_path.replace(target_path_cwd_prefix, '', 1)
Out[15]: 'one/up'            
                                      
In [16]: abs_target_path.replace(target_path_cwd_prefix, '', 1).lstrip('/')
Out[16]: 'one/up'
                 
In [17]: target_path = '/../b/o/g/u/s'
                                             
In [18]: abs_target_path = os.path.abspath(target_path)
           
In [19]: target_path_cwd_prefix = os.path.commonprefix([cwd, os.path.dirname(abs_target_path)])

In [20]: confined_target_path = abs_target_path.replace(target_path_cwd_prefix, '', 1).lstrip('/')
    
In [21]: confined_target_path
Out[21]: 'b/o/g/u/s'
```
>  A path that violates these requirement should raise an error.

I'd prefer this sort of checking happened during the import of metadata rather than during the publishing of the repository.

> if such a path happens to work with the current implementation, it writes a file at a location outside of the build directory and the resulting published repo is broken as this file is not within the published tree (and has a manifest file pointing to a non-existent or wrong location).
>
> I think it is justifiable to raise an error in this case.

I thought it would be simpler to assume any absolute `target_path` meaning a `chroot` into the repository i.e treat both `/foo/bar/baz` and `foo/bar/baz` the same

gmbnomis: > In [7]: target_path = '/var/../one/up'

My example was:

`target_path = '/var../one/up'`

> > A path that violates these requirement should raise an error.
> 
> I'd prefer this sort of checking happened during the import of metadata rather than during the publishing of the repository.

I agree on the import part. From a security point of view, it is best to reject wrong data when entering the system. But my conclusion fpr publication is different: If we have proper rejection at importing/syncing, there is no need for publication to silently drop errors by forcing invalid paths into the build_dir. To the contrary, this should result in an error as something with the validation is very wrong in this case.

> 
> > if such a path happens to work with the current implementation, it writes a file at a location outside of the build directory and the resulting published repo is broken as this file is not within the published tree (and has a manifest file pointing to a non-existent or wrong location).
> > I think it is justifiable to raise an error in this case.
> 
> I thought it would be simpler to assume any absolute `target_path` meaning a `chroot` into the repository i.e treat both `/foo/bar/baz` and `foo/bar/baz` the same

But if you want that, why don't you do exactly that: Treat  `/foo/bar/baz` as `foo/bar/baz` (i.e. get rid of leading slashes) and treat it as a relative path into `build_dir`. As said, `target_path` has nothing to do with the current working directory IMHO.

**But:** This is a new feature changing the semantics of the "name" attribute (and we must look into the "name"s written into PULP_MANIFEST as well.) Additionally, the "../some_file" -> "<build_dir>/some_file" case still makes no sense to me.

Currently, almost all absolute paths lead to an error when publishing (because the location is usually not writable). With your change, they have a well defined meaning. I don't think this is the right place to introduce such a feature. Especially, if the next change you plan it to reject absolute paths at import/sync.

dparalen: >> I'd prefer this sort of checking happened during the import of metadata rather than during the publishing of the repository.

> I agree on the import part. From a security point of view, it is best to reject wrong data when entering the system. But my conclusion for publication is different: If we have proper rejection at importing/syncing, there is no need for publication to silently drop errors by forcing invalid paths into the build_dir. To the contrary, this should result in an error as something with the validation is very wrong in this case.

With this patch I'm not silently dropping errors of invalid paths, I'm rather introducing a convention.

> My example was:
>
> target_path = '/var../one/up'

Sorry I assumed it was a typo; I guess the resulting path is OK nevertheless: `/var/lib/pulp/build_dir/my_repo/var../one/up`

> But: This is a new feature changing the semantics of the "name" attribute (and we must look into the "name"s written into PULP_MANIFEST as well.)

I agree, it worries me a bit; if we have users with correct absolute path in the manifest, this patch will break them. So will the publish-time validation.

> Additionally, the "../some_file" -> "<build_dir>/some_file" case still makes no sense to me.

wit this patch it's the same semantics as `$ cd / ; ls ../mnt` it's odd but not broken 

> Currently, almost all absolute paths lead to an error when publishing (because the location is usually not writable). With your change, they have a well defined meaning. I don't think this is the right place to introduce such a feature. Especially, if the next change you plan it to reject absolute paths at import/sync.

Even then, I'm not preventing publishing of what was allowed to be imported. I find this a slightly better option.

gmbnomis:
Again, matching `target_path` against cwd which is not related to the `build_dir` makes no sense. I have demonstrated multiple problems with this approach and you seem to ignore them but one. With a cwd below `/var/cache/` the current code maps:

```
/var../one/up -> /build_dir/../one/up  (this is outside of the build dir!)
/var/some/file -> /build_dir/some/file (var is gone, why?)
```

Again, `../some/file -> /build_dir/some/file` creates the problem that the resulting PULP_MANIFEST is broken. It lists a file "../some/file" which is simply not there. Nevertheless, the publication succeeds with your patch. 

Additionally, you imply that a user giving "../some/file" actually meant "some/file". Why should this be the case? What if the repo contains "../README.md" and "README.md"?

Again, there is no such thing as "users with correct absolute path in the manifest". Please demonstrate a legitimate use of absolute paths in a content unit's "name" with the current Pulp implementation if you think that such a use exists.

The current situation in Pulp is just this: Names resulting in paths outside of build dir will cause the publication to fail if the location is not writable. If all locations happen to be writable the publish will succeed, but the resulting published repo will be broken (incomplete content, PULP_MANIFEST with entries pointing nowhere). For me the obvious _security fix_ is to make this consistent and let the publish fail in all of these cases.

We must not remap names to sometimes arbitrary locations, write a broken PULP_MANIFEST, and let the publish succeed. So please, let us just end this here and do something different.

gmbnomis:
@dparalen: Sure, I put together a short python script demonstrating the problems constructing the paths:

````python
import os.path


def symlink_filename(target_path):
    build_dir = "/build_dir"
    cwd = os.path.abspath(os.path.curdir)
    # any two absolute paths always share at least the root element
    abs_target_path = os.path.abspath(target_path)
    target_path_cwd_prefix = os.path.commonprefix([cwd, os.path.dirname(abs_target_path)])
    confined_target_path = abs_target_path.replace(target_path_cwd_prefix, '', 1).lstrip('/')
    return os.path.join(build_dir, confined_target_path)

for target_path in ["/var../one/up", "/var/some/file", "/novar/some/file", "some/file", "../some/file"]:
    print target_path, "->", symlink_filename(target_path)
````

Running this script in cwd `/var/cache/man` yields:

````
...man]$ python ~/sources/pulp/confine_path.py
/var../one/up -> /build_dir/../one/up
/var/some/file -> /build_dir/some/file
/novar/some/file -> /build_dir/novar/some/file
some/file -> /build_dir/some/file
../some/file -> /build_dir/some/file
````

Re the reproducer: I don't doubt that your with your change one is unable to change the "alice" repository. But as pointed out above, the published PULP_MANIFEST and the content of the "malice" repository won't fit together (because the paths do not match). Today, Pulp's publish fails for most invalid paths pointing outside of 'build_dir' (because the apache user is not able to write to these location). I strongly suggest that we should make it fail for _all_ invalid paths instead of squeezing everything below build_dir, creating problems with PULP_MANIFEST in the process.

**Comment by dparalen** [comment]
@gmbnomis thanks for the screen log. I see now what you're saying. My point was to avoid raising an exception late, during the publish but it's not going to work this way for the reasons you've mentioned before. Let me update with your suggestions of the check.

File: server/test/unit/plugins/file/test_distributor.py [hunk 4]
```diff
@@ -345,3 +345,83 @@ def test__symlink_units_EINVAL_os_error(self, mock_readlink):
...
+    @patch('os.path.normpath')
+    def test__target_symlink_path_OK_path(self, mock_normpath, mock_isabs, mock_join):
+        """
+        Check an OK target path passes the symlink path validation
+        """
+        distributor = self.create_distributor_with_mocked_api_calls()
+        build_dir = '/foo/bar'
+        target_path = 'baz'
+        # set up mock return values according to an OK path
+        norm_mock = mock_normpath.return_value
```
gmbnomis: I like the way you patch the os.path functions while retaining their original implementation below (using `side_effect`). Couldn't we use it here as well? i.e.:

```python
@patch('os.path.join', side_effect=os.path.join)
@patch('os.path.isabs', side_effect=os.path.isabs)
@patch('os.path.normpath', side_effect=os.path.normpath)
def test__target_symlink_path_OK_path(self, mock_normpath, mock_isabs, mock_join):
...
```

Then you would not need to mock all these values and could still used the `assert_called_...`functions like you do in the other tests.

server/test/unit/plugins/file/test_distributor.py
```diff
@@ -345,3 +345,83 @@ def test__symlink_units_EINVAL_os_error(self, mock_readlink):
...
+    @patch('os.path.join')
+    @patch('os.path.isabs')
+    @patch('os.path.normpath')
+    def test__target_symlink_path_OK_path(self, mock_normpath, mock_isabs, mock_join):
+        """
+        Check an OK target path passes the symlink path validation
+        """
+        distributor = self.create_distributor_with_mocked_api_calls()
+        build_dir = '/foo/bar'
+        target_path = 'baz'
```
gmbnomis: Suggestion: Should we also test a valid relative path with `..` and `.`, e.g. `foo/.././..baz/.bar`? (Does not need to be such a detailed test like here, just verifying that the expected path is returned by `_target_symlink_path()`

dparalen: I think the `os.path.normpath()` used in `target_symlink_path()` removes the "middle (double) dots" of the path, moving them in front (reducing other directories on the way). A testcase like that would therefore check that `os.path.normpath()` works as expected. The other tests rather assert particular code path in `_target_symlink_path()` were exercised. I can add that tho.

server/test/unit/plugins/file/test_distributor.py
dparalen: I stole the idea from the [test code above](https://github.com/pulp/pulp/blob/ae11e67274db41792054ffb8682c7057ab97b4bd/server/test/unit/plugins/file/test_distributor.py#L268), I like it too :) My goal here was to assert that all the "gears wheeled" as expected, esp. the `normpath()` return value equality with the expected `..` parameter,  thru the `norm_mock.__eq__` magic method. The `side_effect=os.path.normpath` would make the `mock_normpath` return a string instead of a mock, I wouldn't be able to check the gears on this level anymore.

dparalen:
@gmbnomis thanks for the review!
Please, see my answers in-line.

server/test/unit/plugins/file/test_distributor.py
gmbnomis: Ok, I see where you are coming from.

server/test/unit/plugins/file/test_distributor.py
gmbnomis: > A testcase like that would therefore check that os.path.normpath() works as expected.

Just to elaborate where I am coming from:

Yes. But more precisely, it would check what you or I _expect_ `os.path.normpath()` (and the remaining logic and functions) to do. Perhaps it does not do what we expect. (The how 'gears are wheeling' test does not verify those assumptions because the assumptions are encoded into the test.)

But more importantly, if someone wants to refactor or change the function in the future, it may not be based on `os.path.normpath()` anymore. The tests below describe when the function is expected to raise an exception. The proposed test for the positive case(s) would describe what the function is supposed to do.

In conclusion, I think both tests have their merit, the "inner gears working" one and the "outer expectation" one.

server/pulp/plugins/file/distributor.py
```diff
@@ -200,6 +200,16 @@ def post_repo_publish(self, repo, config):
         """
         pass
 
+    @staticmethod
+    def _target_symlink_path(build_dir, target_path):
+        norm_path = os.path.normpath(target_path)
+        if os.path.isabs(norm_path):
+            raise ValueError('Symlink target path must not be absolute: %s' % target_path)
+        if norm_path == '..' or norm_path.startswith('../'):
```
jortel: I think `startswith('..')`  will cover both cases.

gmbnomis: I don't think so: `..afile` is a perfectly valid path (i.e. filename) that starts with `..`

dparalen:
@jortel, @gmbnomis  thanks for the review guys!
@jortel please merge (I can't).

Cheers,
milan

server/pulp/plugins/file/distributor.py
jortel: An odd case but fair enough.

dralley:
@dparalen Please rebase and I can merge it (or Jeff)

dparalen:
@dralley @jortel, rebased, please let me know if I can do anything else.
Sorry for a later response due to my travelling.

PR #3671, opened 2018-09-26
Make publisher for publications optional

cutwater:
Set `publisher` field at `Publication` model nullable.

fixes #4038

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -43,7 +43,7 @@ class Publication(Model):
     complete = models.BooleanField(db_index=True, default=False)
     pass_through = models.BooleanField(default=False)
 
-    publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE)
+    publisher = models.ForeignKey('Publisher', on_delete=models.CASCADE, null=True)
     repository_version = models.ForeignKey('RepositoryVersion', on_delete=models.CASCADE)
 
     @classmethod


bmbouter:
This looks right to me. I'd like to hear from either @daviddavis or @jortel on if it looks right to them also.

jortel:
I think the model should also have `blank=True` and the serializer needs:
```
required=False,
allow_blank=True
```

jortel:
@bmbouter, I updated my comment.  I forgot that Publication cannot be created with the API so the serializer change isn't necessary.  I was thinking of my previous PR for the pass-through when it (temporarily) added support for POST on the Publication endpoint for pass-through but then decided to pull it out.

daviddavis:
I can't figure out what `blank=True` does:

```
# with blank=True
>>> repo_version = RepositoryVersion.objects.first()
>>> publication = Publication(repository_version=repo_version); publication.save() # saves
>>> publication = Publication(repository_version=repo_version, publisher=None); publication.save() # saves
>>> publication = Publication(repository_version=repo_version, publisher='') # raises: "Publication.publisher" must be a "Publisher" instance.

# with blank=False
>>> repo_version = RepositoryVersion.objects.first()
>>> publication = Publication(repository_version=repo_version); publication.save() # saves
>>> publication = Publication(repository_version=repo_version, publisher=None); publication.save() # saves
>>> publication = Publication(repository_version=repo_version, publisher='') # raises: "Publication.publisher" must be a "Publisher" instance.
```

I thought maybe it might have some impact in the API but I know we use `allow_blank` in serializers to customize this behavior. Any idea?

bmbouter:
`blank` is only used for validation purposes  https://docs.djangoproject.com/en/2.0/ref/models/fields/#blank

I've been reading the DRF docs to determine what `blank=True` does to the validation DRF uses. If anything it would be at the serialization field layer I think. @daviddavis all of your tests are at the DB layer so it won't matter there I don't think.

I wonder if you make a Publication in the db with no publisher, leave blank off of the model definition like it is now, and then ask the DRF serializer to serve it to a user doing a GET on it.

daviddavis:
So doing a GET on the publications I made in my previous comment seems like `blank=True` makes no difference in terms of behavior.

To further test, I removed `blank=True` from repository's `description` field and was able to create repos:

```
$ http POST :8000/pulp/api/v3/repositories/ name=TEST
{
    "_href": "/pulp/api/v3/repositories/2/",
    "_latest_version_href": null,
    "_versions_href": "/pulp/api/v3/repositories/2/versions/",
    "created": "2018-09-26T19:24:49.695726Z",
    "description": "",
    "name": "TEST",
    "notes": {}
}
```

I noticed this though when I removed `blank=True` from the `description` field and re-ran `makemigrations`:

```
Migrations for 'pulp_app':
  pulpcore/pulpcore/app/migrations/0007_auto_20180926_1926.py
    - Alter field description on repository
```

But it didn't seem to make a difference.

bmbouter:
From this testing I'm thinking indeed that `blank` is a form validation thing and we shouldn't use it. For this PR I think we don't need it. Also as another piece of work I think we should remove all the `blank` usage today.

dkliban:
@daviddavis does it make a difference if you try using the browsable API form to make the request?

daviddavis:
@dkliban not as far as I can tell. I removed `blank=True` from the repository `description` db field and I can still create repos without entering in any description.

PR #3672, opened 2018-09-26
Add tests for base_version and repository versions

kersommoura:
Add tests to verify that `base_version` can be used to create another
repository version. Using the same repository, and a different one.

Besides that, add a test to assure that an HTTP exception is raised when using
a non-exitent `base_version`.

closes: #4035
https://pulp.plan.io/issues/4035

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
@@ -24,6 +24,7 @@ from pulp_smash.pulp3.utils import (
 from tests.functional.api.using_plugin.constants import (
     FILE_CONTENT_PATH,
     FILE_FIXTURE_COUNT,
+    FILE_FIXTURE_MANIFEST_URL,
     FILE_LARGE_FIXTURE_MANIFEST_URL,
     FILE_PUBLISHER_PATH,
     FILE_REMOTE_PATH,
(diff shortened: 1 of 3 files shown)


kersommoura:
Tested  against installation from source, F28.

ping @pulp/qe

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -532,3 +533,113 @@ def test_all(self):
...
+        repo = self.client.get(repo['_href'])
+
+        self.client.post(
+            repo['_versions_href'],
+            {'base_version': get_versions(repo)[0]['_href']}
+        )
+        repo = self.client.get(repo['_href'])
+        added_content.append(get_content(repo))
+
+        self.assertEqual(added_content[0], added_content[1], added_content)
```
daviddavis: This test was hard to read. I think part of the reason is you are comparing added_content from repo version 1 to the content of repo version 2. I would recommend not dealing with added content and just examine the content of both repo versions. Something like:

```python
        # create repo version 1
        repo = self.create_sync_repo()
        version_1_content = get_content(repo)
        self.assertIsNone(get_versions(repo)[0]['base_version'])

        # create repo version 2
        self.client.post(
            repo['_versions_href'],
            {'remove_content_units': [version_1_content[0]['_href']]}
        )
        repo = self.client.get(repo['_href'])

       # create repo version 3 from version 1
        self.client.post(
            repo['_versions_href'],
            {'base_version': get_versions(repo)[0]['_href']}
        )
        repo = self.client.get(repo['_href'])
        version_3_content = get_content(repo)

        self.assertEqual(version_1_content, version_3_content)
````

File: pulpcore/tests/functional/api/using_plugin/test_repo_versions.py [hunk 2]
```diff
@@ -532,3 +533,113 @@ def test_all(self):
...
+
+        1. Create a new repository A and sync it.
+        2. Create a new repository B and a new version for this repository
+           specify repository A version 1 as the ``base_version``.
+        3. Check that repository A version 1 and repository B version 1 have
+           the same content.
+        """
+        repo = self.create_sync_repo()
+        added_content = []
+        added_content.append(get_added_content(repo))
```
daviddavis: Same here. Suggest examining the repo's content instead of added_content.

daviddavis:
Also, you may want to assert in your tests that `base_version` is set correctly.

PR #3675, opened 2018-09-27
Problem: Distribution is not available as an abstract model

dkliban:
Solution: Add a BaseDistribution for plugin writers to use

This patch adds an abstract BaseDistribution model to the plugin api.

re: 3991
https://pulp.plan.io/issues/3991

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -2,6 +2,7 @@ from django.db import models, transaction
 
 from . import storage
 from .base import Model
+from .repository import Publisher, Repository
 from .task import CreatedResource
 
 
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/models/publication.py
```diff
@@ -208,3 +208,28 @@ class Distribution(Model):
 
     class Meta:
         default_related_name = 'distributions'
+
+
+class BaseDistribution(Distribution):
+    """
+    A distribution defines how a publication is distributed by pulp.
```
bmbouter: s/pulp/Pulp's webserver/

dkliban:
I tested this change with https://github.com/pulp/pulp_docker/pull/264 and I was able to generate migrations that contained a Distribution table for pulp_app and a separate DockerDistribution table for pulp_docker. They have the exact same fields.

PR #3678, opened 2018-09-28
Problem: database settings are not being used

dkliban:
Solution: load dynaconf to find out the DATABASES setting

closes: #4046
https://pulp.plan.io/issues/4046


pulpcore/pulpcore/app/manage.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import sys
+from dynaconf.contrib import django_dynaconf  # noqa
```
rochacbruno: Maybe this line should go after `os.environ.setdefault("DJANGO_SETTINGS_MODULE", ...`

pulpcore/pulpcore/app/settings.py
```diff
@@ -48,8 +48,6 @@
 # Application definition
 
 INSTALLED_APPS = [
-    # Dynamic configuration with Dynaconf
-    'dynaconf.contrib.django_dynaconf',
```
asmacdo: It might be nice commented out rather than remove because this is the logical place to see which apps are being used.

dkliban: I did that and added a link to the dyanconf issue everywhere. Thanks for the suggestion.

PR #3683, opened 2018-09-30
Add test to assert that error should be raised when passing invalid parameters to CRUD API

rochacbruno:
Add test to assert that error should be raised when passing invalid extraneous parameters to CRUD APIs.

closes PulpQE/pulp-smash#1085

[noissue]


rochacbruno:
@pulp/qe

rochacbruno:
@daviddavis

PR #3685, opened 2018-10-01
Allow commits to require pulp-smash PRs

daviddavis:
fixes #4037
https://pulp.plan.io/issues/4037


daviddavis:
I tested this out with this PR: https://github.com/PulpQE/pulp-smash/pull/1143

And here is the expected failure: https://travis-ci.org/pulp/pulp/jobs/435692227

PR #3693, opened 2018-10-03
Expand base_version test case

kersommoura:
Test whether `base_version` can be used in conjunction with `add_content_unit`
and `remove_content_unit`.

See: https://pulp.plan.io/issues/3360

'[noissue]'


kersommoura:
ping @pulp/qe

PR #3694, opened 2018-10-03
Add test to ensure specific fields are returned from API

rochacbruno:
Tests to ensure specified fields as in `endpoint?fields=id,other,another` are strictly returned from the APIs

closes PulpQE/pulp-smash#1124

[noissue]


rochacbruno:
@dralley @pulp/qe

PR #3695, opened 2018-10-04
Pass DeclarativeArtifact.extra_data to downloader

asmacdo:
Note: I'm not sure what the intended use of
DeclarativeArtifact.extra_data, so I am not sure if this goes outside
the intended scope.

Use Case: During Docker sync, certain requests need specific headers
that are not shared with the entire session. Since Artifact downloads
happen in pulpcore rather than the plugin, there is no way to pass extra
arguments to downloader.run(). I suspect that this problem would exist
for any plugin that needs specially formed request to download Artifacts.

This pattern would allow Artifact downloaders to:
  1. Add arbitrary headers to certain requests (like Docker V1 vs V2
      artifact downloads)
  2. Add arbitrary query parameters
  3. dynamically set http verbs

1 is my use case, though 2 and 3 are hypothetical.

[noissue]

### DIFF
--- a/plugin/pulpcore/plugin/download/file.py
+++ b/plugin/pulpcore/plugin/download/file.py
@@ -32,12 +32,15 @@ class FileDownloader(BaseDownloader):
         self._path = os.path.abspath(os.path.join(p.netloc, p.path))
         super().__init__(url, **kwargs)
 
-    async def run(self):
+    async def run(self, extra_data=None):
         """
         Read, validate, and compute digests on the `url`. This is a coroutine.
 
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader.run`.
+
+        Args:
+            extra_data (dict): Extra data passed to the downloader.
         """
         async with aiofiles.open(self._path, 'rb') as f_handle:
             while True:
--- a/plugin/pulpcore/plugin/download/http.py
+++ b/plugin/pulpcore/plugin/download/http.py
@@ -159,7 +159,7 @@ class HttpDownloader(BaseDownloader):
                               url=self.url)
 
     @backoff.on_exception(backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=giveup)
-    async def run(self):
+    async def run(self, extra_data=None):
         """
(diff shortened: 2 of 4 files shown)


plugin/pulpcore/plugin/download/http.py
```diff
@@ -156,7 +156,7 @@ def __init__(self, url, session=None, auth=None, proxy=None, proxy_auth=None,
                               url=self.url)
 
     @backoff.on_exception(backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=giveup)
-    async def run(self):
+    async def run(self, **kwargs):
```
bmbouter: I figured these kwargs would need to get passed on to `get() in here`.

asmacdo:
Updated to add extra_data as an explicit optional kwarg in the downloaders.

plugin/pulpcore/plugin/download/file.py
```diff
@@ -32,12 +32,15 @@ def __init__(self, url, **kwargs):
...
-    async def run(self):
+    async def run(self, extra_data=None):
         """
         Read, validate, and compute digests on the `url`. This is a coroutine.
 
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader.run`.
+
+        Args:
+            extra_data (dict): Extra data needed by the downloader.
```
bmbouter: s/needed by/passed to/

plugin/pulpcore/plugin/download/http.py
```diff
@@ -169,6 +169,9 @@ def __init__(self, url, session=None, auth=None, proxy=None, proxy_auth=None,
 
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader.run`.
+
+        Args:
+            extra_data (dict): Extra data needed by the downloader.
```
bmbouter: s/needed by/passed to/
s/downloader/downloader as kwargs/

plugin/pulpcore/plugin/download/file.py
bmbouter: s/downloader/downloader as kwargs/

PR #3696, opened 2018-10-04
Add connection_limit to a Remote

goosemania:
Use aiohttp's `limit` option to limit a total number of TCP connections,
instead of using `asyncio.Semaphore` directly.

closes #4040
https://pulp.plan.io/issues/4040

### DIFF
--- a/plugin/pulpcore/plugin/stages/artifact_stages.py
+++ b/plugin/pulpcore/plugin/stages/artifact_stages.py
@@ -87,16 +87,13 @@ class ArtifactDownloaderRunner():
                 :class:`~pulpcore.plugin.stages.DeclarativeContent` objects from.
         out_q (:class:`asyncio.Queue`): The queue to put
                 :class:`~pulpcore.plugin.stages.DeclarativeContent` objects into.
-        max_concurrent_downloads (int): The maximum number of concurrent downloads this stage will
-            run.
         max_concurrent_content (int): The maximum number of
             :class:`~pulpcore.plugin.stages.DeclarativeContent` instances to handle simultaneously.
     """
 
-    def __init__(self, in_q, out_q, max_concurrent_downloads, max_concurrent_content):
+    def __init__(self, in_q, out_q, max_concurrent_content):
         self.in_q = in_q
         self.out_q = out_q
-        self.max_concurrent_downloads = max_concurrent_downloads
         self.max_concurrent_content = max_concurrent_content
 
     @property
@@ -118,9 +115,6 @@ class ArtifactDownloaderRunner():
         #: (:class:`asyncio.Task`): The task that gets new content from `in_q`.
         #    Set to None if stage is shutdown.
         self._content_get_task = self._add_to_pending(self.in_q.get())
-
-        #: (:class:`asyncio.Semaphore`): Semaphore controlling the number of concurrent downloads
-        self._download_semaphore = asyncio.Semaphore(value=self.max_concurrent_downloads)
 
         with ProgressBar(message='Downloading Artifacts') as pb:
             try:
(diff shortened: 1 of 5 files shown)


plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -174,17 +168,14 @@ def _downloaders_for_content(self, content):
...
         Returns:
             List of downloader coroutines (may be empty)
         """
-        async def download_with_limit(semaphore, downloader):
-            async with semaphore:
-                log.debug("ArtifactDownloader: Start download of '%s'", declarative_artifact.url)
-                result = await downloader.run()
-                log.debug("ArtifactDownloader: Downloaded: '%s'", declarative_artifact.url)
-                return result
+        async def download(downloader):
```
bmbouter: What about renaming this to `download_synchronously_and_log`?

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -103,6 +103,7 @@ def tls_storage_path(self, name):
     username = models.TextField(blank=True)
     password = models.TextField(blank=True)
     last_synced = models.DateTimeField(blank=True, null=True)
+    connection_limit = models.PositiveIntegerField(default=5)
```
bmbouter: This needs to be added to the docstring above.

bmbouter:
The tests are failing and need fixing.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -174,17 +168,14 @@ def _downloaders_for_content(self, content):
...
             List of downloader coroutines (may be empty)
         """
-        async def download_with_limit(semaphore, downloader):
-            async with semaphore:
-                log.debug("ArtifactDownloader: Start download of '%s'", declarative_artifact.url)
-                result = await downloader.run()
-                log.debug("ArtifactDownloader: Downloaded: '%s'", declarative_artifact.url)
-                return result
+        async def download_and_log(downloader):
+            log.debug("ArtifactDownloader: Start download of '%s'", declarative_artifact.url)
```
gmbnomis: I am wondering whether these log entries still make sense: If the stage work on 200 concurrent content units and 5 connections are allowed, the user will see 200 "Start download of..." logs, but only 5 of those will actually be downloading at a time. With the new approach, only the downloader may know what is actually downloading. Should we get rid of these messages here? Alternatively, we could change the message to "scheduling download..." or so.

bmbouter: That is a good point about the logs not being exactly correct . I'm +1 on getting rid of the messages here, but if you do that the method does nothing which means I'm +1 on removing it altogether.

gmbnomis: Yes, I agree.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -205,9 +190,8 @@ def _downloaders_for_content(self, content):
                     declarative_artifact.url,
                     **validation_kwargs
                 )
-                downloaders_for_content.append(
-                    download_with_limit(self._download_semaphore, downloader)
-                )
+                downloaders_for_content.append(downloader.run())
```
bmbouter: I don't think this works because you need to `await downloader.run()`. Calling downloader.run() returns the object of type `coroutine` which then needs to be scheduled "synchronously" via `await`. I didn't run the code, that is just what it seems like.

goosemania: @bmbouter, [`asyncio.gather`](https://github.com/pulp/pulp/blob/19b0c26bdd17e51fd468b332103bd6b9d96878ff/plugin/pulpcore/plugin/stages/artifact_stages.py#L162) is used for those.
I ran the code, it works for me. I still can miss something, so let me know :)

bmbouter: Nope you are exactly correct! Thank you for pointing this out.

gmbnomis:
> @gmbnomis , your tests are great :)

@goosemania Thank you!

> If you can take a look at my changes in the unit tests, that would be helpful.

Changes look great! I feared that the tests would be of limited value given the simplification of the stage, but they still make sense.

> With no limit for concurrent downloads in the artifact downloader, it seems like the `test_download_stall` doesn't make sense any more, so it's removed.

Hmm, the purpose of this test is still valid I think. In all other tests, the stage never runs out of work until the final "None" is seen. The purpose of this test was to ensure that if there is a "pause", the stage schedules `self._content_get_task` correctly and re-starts working on new content immediately when it arrives. Basically the test should prevent that something stupid is happening in this "no work" edge case. Now that I am looking at it, the test's name clearly is a misnomer...

Anyway, I think the test is not crucial, more like a "nice to have" test.

goosemania:
@gmbnomis , thanks for clarification, I'll add it back. Even if it's just nice-to-have one, it's done and has its value. Thanks!

PR #3698, opened 2018-10-04
Add tests for creating resources using invalid data 

kersommoura:
Add a test to verify messages when attempting to create a user with a long
`username` and `username` with invalid characters.

Besides that, refactor the test for create repos with the same name.

See: https://github.com/PulpQE/pulp-smash/issues/882

'[noissue]'

### DIFF
--- a/pulpcore/tests/functional/api/test_crud_repos.py
+++ b/pulpcore/tests/functional/api/test_crud_repos.py
@@ -34,11 +34,17 @@ class CRUDRepoTestCase(unittest.TestCase):
     def test_02_create_same_name(self):
         """Try to create a second repository with an identical name.
 
-        See: `Pulp Smash #1055
+        * `Pulp Smash #882 <https://github.com/PulpQE/pulp-smash/issues/882>`_.
+        * `Pulp Smash #1055
         <https://github.com/PulpQE/pulp-smash/issues/1055>`_.
         """
-        with self.assertRaises(HTTPError):
-            self.client.post(REPO_PATH, gen_repo(name=self.repo['name']))
+        self.client.response_handler = api.echo_handler
+        response = self.client.post(
+            REPO_PATH,
+            gen_repo(name=self.repo['name'])
+        )
+        self.assertIn('unique', response.json()['name'][0])
+        self.assertEqual(response.status_code, 400)
 
     @skip_if(bool, 'repo', False)
     def test_02_read_repo(self):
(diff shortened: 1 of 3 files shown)


pulpcore/tests/functional/api/test_crud_repos.py
```diff
@@ -34,11 +34,17 @@ def test_01_create_repo(self):
...
         """Try to create a second repository with an identical name.
 
-        See: `Pulp Smash #1055
+        * `Pulp Smash #882 <https://github.com/PulpQE/pulp-smash/issues/882>`_.
+        * `Pulp Smash #1055
         <https://github.com/PulpQE/pulp-smash/issues/1055>`_.
         """
-        with self.assertRaises(HTTPError):
-            self.client.post(REPO_PATH, gen_repo(name=self.repo['name']))
+        self.client.response_handler = api.echo_handler
```
kersommoura: This can be modified later when a custom HTTPError exception would be added to Pulp-Smash.

kersommoura:
@pulp/qe

pulpcore/tests/functional/api/test_crud_repos.py
```diff
@@ -34,11 +34,17 @@ def test_01_create_repo(self):
...
+        response = self.client.post(
```
rochacbruno: I suggest 

```py
reponse = api.Client(self.cfg, api.echo_handler).post(...)
```
instead of mutating the existing client handler. (or moving the handler back to its original handler after the test)

kersommoura: @rochacbruno, thanks for your review.  I did not update the client to the original `response_handler` since a new client is create for each test.

```python
    def setUp(self):
        """Create an API client."""
        self.client = api.Client(self.cfg, api.json_handler)
```

pulpcore/tests/functional/api/test_crud_users.py
```diff
@@ -134,6 +135,42 @@ def test_negative_create_user_with_invalid_parameter(self):
...
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.echo_handler)
+
+    def test_long_username(self):
+        """Create a user with long username."""
+        attrs = _gen_verbose_user_attrs()
+        attrs['username'] = gen_username(151)
+        response = self.client.post(USER_PATH, attrs)
```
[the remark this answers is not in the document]
kersommoura: I will update properly.

kersommoura:
@daviddavis can you review/merge this one?

PR #3700, opened 2018-10-05
Problem: created resources OpenAPI schema is not specific enough

dkliban:
Solution: strongly type created resources as an array of strings

This patch changes the CreatedResource serializer to inherit from the RelatedField serializer. This lets the
OpenAPI schema generator know that created resources are URI strings.

closes: #4053
https://pulp.plan.io/issues/4053

### DIFF
--- a/pulpcore/pulpcore/app/serializers/task.py
+++ b/pulpcore/pulpcore/app/serializers/task.py
@@ -9,7 +9,7 @@ from pulpcore.app.serializers import IdentityField, RelatedField, ModelSerialize
 from .base import viewset_for_model
 
 
-class CreatedResourceSerializer(ModelSerializer):
+class CreatedResourceSerializer(RelatedField):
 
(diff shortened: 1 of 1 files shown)


pulpcore/pulpcore/app/serializers/task.py
```diff
@@ -79,7 +79,8 @@ class TaskSerializer(ModelSerializer):
     created_resources = CreatedResourceSerializer(
         help_text=_('Resources created by this task.'),
         many=True,
-        read_only=True
+        read_only=True,
+        view_name='NotActuallyUsed'
```
bmbouter: What about calling this `'None`' instead?

bmbouter: Also can a brief (1-2 lines maybe) comment be added about how this is a polymorphic field so we don't know the view name.

PR #3701, opened 2018-10-05
Retry in case of some 5XX errors

goosemania:
closes #4036
https://pulp.plan.io/issues/4036

### DIFF
--- a/plugin/pulpcore/plugin/download/http.py
+++ b/plugin/pulpcore/plugin/download/http.py
@@ -19,6 +19,9 @@ def giveup(exc):
     Do not give up when the status code is one of the following:
 
         429 - Too Many Requests
+        502 - Bad Gateway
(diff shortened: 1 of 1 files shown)


goosemania:
@bmbouter , I added some 5XX errors for which it makes sense, imo, to retry.
I'm not sure we need to retry for all the 4XX/5XX errors. 
What do you think?

bmbouter:
@goosemania Yes this is probably just as good of an approach since many of the unlisted 4XX and 5XX response codes aren't emitted often in practice.

The only change I want to suggest is to disinclude 500. Typically 500 is a fatal server exception, which usually tells me the thing is totally broken. Since we're being thoughtful about which codes should retry, that one probably shouldn't be retried. What do you think?

goosemania:
Good point. Changes pushed.

PR #3707, opened 2018-10-10
Allow Artifact init and validation from a file

bmbouter:
There wasn't a way to have the upload handler involved so I had to open
and compute the hashes manually.

https://pulp.plan.io/issues/4072
re #4072


pulpcore/pulpcore/app/models/content.py
```diff
@@ -111,26 +113,40 @@ def delete(self, *args, **kwargs):
...
             An in-memory, unsaved :class:`~pulpcore.plugin.models.Artifact`
         """
+        if isinstance(file, str):
+            class File:
+                pass
+            with open(file, 'rb') as f:
+                file = File()
+                file.size = 0
+                file.hashers = {n: hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
+                data = f.read()
```
dkliban: could you chunk the bits please?

bmbouter: will fix in next version.

PR #3711, opened 2018-10-11
Removes custom User model

bmbouter:
https://pulp.plan.io/issues/4077
closes #4077


bmbouter:
I bet I could bring back the viewset and serializer correctly with the built-in generic User object. Should I do that?

bmbouter:
I pushed a version that brings back the viewsets and serializers stuff

PR #3717, opened 2018-10-16
Better support for auto-distribution.

jortel:
https://pulp.plan.io/issues/4045

### DIFF
--- a/pulpcore/pulpcore/app/models/publication.py
+++ b/pulpcore/pulpcore/app/models/publication.py
@@ -107,21 +107,54 @@ class Publication(Model):
             CreatedResource.objects.filter(object_id=self.pk).delete()
             super().delete(**kwargs)
 
+    def update_distributions(self, model=None):
+        """
+        Update distributions with this publication.
+
+        This supports the auto-distribution feature by updating the publication on
+        distributions with matching repository and publisher. By doing this, this
+        publication will be distributed.  This method may only be called after
+        setting complete=True and saved.
+
+        Args:
+            model (pulpcore.app.models.BaseDistribution): A distribution model used.
+                The Distribution model is used when not specified.
+
+        """
+        model = model or Distribution
+        distributions = model.objects.filter(
+            publisher=self.publisher,
+            repository=self.repository)
+        for distribution in distributions:
+            distribution.publication = self
+            distribution.save()
+
     def __enter__(self):
+        """
+        Enter context.
+
+        Returns:
+            Publication: self
+        """
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if not exc_val:
-            with transaction.atomic():
-                self.complete = True
-                self.save()
+        """
+        Exit the context.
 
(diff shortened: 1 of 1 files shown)


dralley:
Docs are failing to build with

```/home/travis/build/pulp/pulp/pulpcore/pulpcore/app/models/publication.py:docstring of pulpcore.app.models.Publication.update_distributions::more than one target found for cross-reference 'BaseDistribution': pulpcore.app.models.BaseDistribution, pulpcore.plugin.models.BaseDistribution```

**Comment by jortel** [comment]
ok test

**Comment by dralley** [comment]
@jortel Could you rebase this branch? Flake8 had an update w/ new linting rules, that is causing failures. I just fixed it in master.

File: pulpcore/pulpcore/app/models/publication.py [hunk 1]
```diff
@@ -107,21 +107,36 @@ def delete(self, **kwargs):
...
+            distribution.publication = self
+            distribution.save()
+
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if not exc_val:
             with transaction.atomic():
+                self.update_distributions()
```
dkliban: You should probably call this after self.save()

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -107,21 +107,36 @@ def delete(self, **kwargs):
             CreatedResource.objects.filter(object_id=self.pk).delete()
             super().delete(**kwargs)
 
+    def update_distributions(self, model=None):
```
dkliban: In case of Docker, where we have a DockerDistribution, this method will need to be called explicitly after a publication is created. The doc string should make that clear.

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -107,21 +107,36 @@ def delete(self, **kwargs):
...
+        """
+        Update distributions with this publication.
```
dkliban: s/Update distributions with this publication./Update distributions associated with this publication's repository and publisher./

pulpcore/pulpcore/app/models/publication.py
```diff
@@ -107,21 +107,36 @@ def delete(self, **kwargs):
...
+
+        This supports the auto-distribution feature by update the publication on
```
dkliban: How about changing this to: "Plugins that provide their own distribution models can call this method to implement the auto-distribution feature. This method has to be called after a publication has been saved to the database and marked as 'complete'. "

pulpcore/pulpcore/app/models/publication.py
jortel: Done.

pulpcore/pulpcore/app/models/publication.py
jortel: I think the combination of the summary (first line) and the following paragraph makes this clear.

jortel:
Re-Tested.

PR #3721, opened 2018-10-18
fix pulp-admin error when running in no tty : IOError: [Errno 25] Inappropriate ioctl for device

Poil:
Ref. https://pulp.plan.io/issues/4068


dralley:
@Poil Please change the target branch to ```2-master```.  All Pulp 2 development happens on ```2-master``` - we only modify release branches when backporting patches to a release, which is something we do later on.

Also please change the commit message to this format

```
Fix pulp-admin error when not running in tty

 fixes https://pulp.plan.io/issues/4068 
```

PR #3724, opened 2018-10-19
Django requirement (<2.0.0) limit for python 2.7

tkopecek:
dev environment is failing on trying to install django 2.x which ends with 'not supported on python 2.7' error message. Limiting upper version leads to successful installation.


server/setup.py
```diff
@@ -7,6 +7,8 @@
 # Django 1.8.0 requires Python >= 2.7
 if PYTHON_MAJOR_MINOR < '2.7':
     DJANGO_REQUIRES = 'django >=1.4.0, <1.8.0'
+elif PYTHON_MAJOR_MINOR < '3.0':
+    DJANGO_REQUIRES = 'django>=1.4.0, <2.0.0'
```
dralley: Since Pulp 2 is Python 2-only, I think you can just add this under the "else" instead.

dralley: Although it's interesting that python 2 is attempting to install a python 3-only package at all.  Django should have their ```python_requires``` set to avoid this issue

PR #3725, opened 2018-10-19
Use /pulp/api/v3/users to test auth

daviddavis:
I don't think using the /pulp/api/v3/ is a good test for authentication.
It doesn't require a user necessarily and it's not something we've
intentionally created in our application. I think something like
/pulp/api/v3/users/ might be better.

cc @pulp/qe


rochacbruno:
LGTM 

is this solving the `'Content-Type': 'application/vnd.oai.openapi'` error we had before for this test case?

daviddavis:
@rochacbruno kind of. the route /pulp/api/v3/ was returning json before DRF 3.9.0 but in the new release it converted this endpoint to yaml.

PR #3733, opened 2018-10-31
Use force_close and concurrency restriction

bmbouter:
Use force_close and concurrency restriction

The use of force_close was inspired by:

https://github.com/aio-libs/aiohttp/issues/2867#issuecomment-375278533

- Causes Factory and Downloaders to use force_close which causes the TCP
  connection to close with each request.
- Ups the default concurrency to 20
- Moves the concurrency restriction feature to the Downloaders and
  updates HttpDownloader and FileDownloader to use it.
- Stops using aiohttp for concurrency restriction because it's done in
  the downloaders.
- The Factory now configures the concurrency restriction using the value
  on the remote.
- creates a _run() method that subclassed Downloaders now need to
  implement instead of run().

https://pulp.plan.io/issues/4036
https://pulp.plan.io/issues/4075

closes #4036
closes #4075

### DIFF
--- a/plugin/pulpcore/plugin/download/base.py
+++ b/plugin/pulpcore/plugin/download/base.py
@@ -61,7 +61,8 @@ class BaseDownloader:
             ``custom_file_object`` option was specified, otherwise None.
     """
 
-    def __init__(self, url, custom_file_object=None, expected_digests=None, expected_size=None):
+    def __init__(self, url, custom_file_object=None, expected_digests=None, expected_size=None,
+                 semaphore=None):
         """
         Create a BaseDownloader object. This is expected to be called by all subclasses.
 
@@ -73,6 +74,8 @@ class BaseDownloader:
             expected_digests (dict): Keyed on the algorithm name provided by hashlib and stores the
                 value of the expected digest. e.g. {'md5': '912ec803b2ce49e4a541068d495ab570'}
             expected_size (int): The number of bytes the download is expected to have.
+            semaphore (asyncio.Semaphore): A semaphore the downloader must acquire before running.
+                Useful for limiting the number of outstanding downloaders in various ways.
         """
         self.url = url
         if custom_file_object:
@@ -83,6 +86,10 @@ class BaseDownloader:
             self.path = self._writer.name
         self.expected_digests = expected_digests
         self.expected_size = expected_size
+        if semaphore:
+            self.semaphore = semaphore
+        else:
+            self.semaphore = asyncio.Semaphore()  # This will always be acquired
         self._digests = {n: hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
         self._size = 0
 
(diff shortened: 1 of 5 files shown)


dralley:
You mention "force_full" by accident in the commit message but nbd

plugin/pulpcore/plugin/download/base.py
```diff
@@ -83,6 +86,10 @@ def __init__(self, url, custom_file_object=None, expected_digests=None, expected
             self.path = self._writer.name
         self.expected_digests = expected_digests
         self.expected_size = expected_size
+        if semaphore:
+            self.semaphore = semaphore
+        else:
+            self.semaphore = asyncio.Semaphore()  # This will always be acquired
```
dralley: This could be simplified to ```self.semaphore = semaphore or asyncio.Semaphore()``` but this way might be more clear to readers, up to you.

dralley:
Without having tested it, LGTM.  I'll let someone else review + test though since I don't want to tear up my environment.

dralley:
Do we have a way to regression-test this?  Would it make sense to do so?  Perhaps we could write a test and just have it be skipped by default, if it would take too long to run every time?

plugin/pulpcore/plugin/download/base.py
bmbouter: I was trying to remember that idiom! Thanks @dralley

bmbouter:
@dralley when you say regression test do you mean a pulp smash test? Also this PR has several apsects, so we need to think about testing each one at some point. Overall, one option to do that with would be to make a dummy aiohttp server that would serve HTTP 1.1 with session continuation back to Pulp's downloaders at test time and then verify that the client disconnects even though the server tells it to keep going. I'm going to file two Pulp Test issues to have those functional tests tracked.

bmbouter:
Now that concurrency restriction is in the downloaders themselves, I believe `connection_limit` doesn't make sense as the name anymore. Specifically it doesn't make sense with the FileDownloader which uses the option, but doesn't even have a "connection". Since this PR introduces this issue, we should resolve it with this PR by renaming s/connection_limit/concurrency_limit/. I will do that now.

plugin/pulpcore/plugin/download/base.py
asmacdo: I like your simplified idiom personally :)

plugin/pulpcore/plugin/download/http.py
```diff
@@ -54,6 +54,10 @@ class HttpDownloader(BaseDownloader):
     allow for an active download to be arbitrarily long, while still detecting dead or closed
     sessions even when TCPKeepAlive is disabled.
 
+    If a session is not provided, the one created will disable HTTP session force TCP closure with
```
dkliban: s/will disable HTTP session force TCP closure with each request/will force TCP connection closure after each request/

plugin/pulpcore/plugin/download/base.py
bmbouter: Actually I'm going to keep it as is because even though I like the shortened idiom, this for lets me keep the comment which I think is helpful since creating a random Semaphore might not be obvious why.

plugin/pulpcore/plugin/download/http.py
bmbouter: fixed in next push, ty

bmbouter:
download concurrency rename ticket is here:  https://pulp.plan.io/issues/4114

plugin/pulpcore/plugin/download/http.py
```diff
@@ -158,7 +163,8 @@ def __init__(self, url, session=None, auth=None, proxy=None, proxy_auth=None,
         return DownloadResult(path=self.path, artifact_attributes=self.artifact_attributes,
                               url=self.url)
 
-    @backoff.on_exception(backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=giveup)
+    @backoff.on_exception(backoff.expo, aiohttp.ClientResponseError,
```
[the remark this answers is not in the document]
bmbouter: @gmbnomis indeed! Thank you again for your astute observation. I'm going to rework this now, and I'm going to file a test to ensure we're testing backoff correctly in core's functional tests.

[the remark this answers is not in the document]
bmbouter: I filed this test issue so that we would ensure this with each pulp_smash run on core.  https://pulp.plan.io/issues/4118

bmbouter:
I pushed what I rewrote and just tested. It performed well for me and also implements the suggested changes. Feel free to take a look.

PR #3734, opened 2018-11-04
Reuse existing artifact for download without digest

mdellweg:
### Pull Requests

Please be sure you have read our documentation regarding contributing
here:

There you can read a Contribution Checklist and learn more about
appropriate branches and rebasing among other things.

Once you are ready to create a pull request:

- [ ] For properly formatting commit messages, please see the
documentation here:
- [ ] Once you have created your pull request, please add a link to
your pull request into the notes/comments for the relevant bug in the
issue tracker and change the status to POST.
- [ ] Continuous integration tests will be automatically triggered
by your pull request.  You can see the status of these tests on the
page for your pull request.  If there are failures, please try to
fix them and update your pull request.

Thank you so much for contributing to Pulp!

### DIFF
--- a/plugin/pulpcore/plugin/stages/artifact_stages.py
+++ b/plugin/pulpcore/plugin/stages/artifact_stages.py
@@ -56,6 +56,7 @@ class QueryExistingArtifacts(Stage):
                         if digest_value:
                             key = {digest_name: digest_value}
                             one_artifact_q &= Q(**key)
+                            declarative_artifact.digest_present = True
                     if one_artifact_q:
                         all_artifacts_q |= one_artifact_q
 
@@ -297,6 +298,16 @@ class ArtifactSaver(Stage):
             artifacts_to_save = []
             for declarative_content in batch:
                 for declarative_artifact in declarative_content.d_artifacts:
+                    if not declarative_artifact.digest_present:
+                        # Artifact may exist in the database,
+                        # but we did not know the digest before downloading
+                        digest_name = declarative_artifact.artifact.DIGEST_FIELDS[0]
+                        digest_value = getattr(declarative_artifact.artifact, digest_name)
+                        if digest_value:
+                            existing_artifact = Artifact.objects.filter(
+                                Q(**{digest_name: digest_value}))
+                            if existing_artifact:
+                                declarative_artifact.artifact = existing_artifact[0]
                     if declarative_artifact.artifact.pk is None:
                         declarative_artifact.artifact.file = str(declarative_artifact.artifact.file)
                         artifacts_to_save.append(declarative_artifact.artifact)
--- a/plugin/pulpcore/plugin/stages/models.py
+++ b/plugin/pulpcore/plugin/stages/models.py
@@ -27,7 +27,7 @@ class DeclarativeArtifact:
         ValueError: If `artifact`, `url`, `relative_path`, or `remote` are not specified.
     """
 
-    __slots__ = ('artifact', 'url', 'relative_path', 'remote', 'extra_data')
+    __slots__ = ('artifact', 'url', 'relative_path', 'remote', 'extra_data', 'digest_present')
 
     def __init__(self, artifact=None, url=None, relative_path=None, remote=None, extra_data=None):
         if not url:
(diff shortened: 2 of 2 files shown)


bmbouter:
@mdellweg I think your branch isn't using the pep8speaks config which is causing it leave those epic comments. I'm like 80% sure.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -297,6 +298,16 @@ class ArtifactSaver(Stage):
             artifacts_to_save = []
             for declarative_content in batch:
                 for declarative_artifact in declarative_content.d_artifacts:
+                    if not declarative_artifact.digest_present:
```
bmbouter: Is it possible to get a similar resolution by having the QueryExistingArtifacts stage come after this which has a similar implementation and would do the replacement similarly?

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -297,6 +298,16 @@ class ArtifactSaver(Stage):
...
+                        # Artifact may exist in the database,
+                        # but we did not know the digest before downloading
+                        digest_name = declarative_artifact.artifact.DIGEST_FIELDS[0]
+                        digest_value = getattr(declarative_artifact.artifact, digest_name)
+                        if digest_value:
+                            existing_artifact = Artifact.objects.filter(
```
bmbouter: I think the QueryExistingArtifacts uses a batch implementation so using that one will also avoid a one-by-one DB implementation.

bmbouter:
@mdellweg Are you using a custom pipeline for this? We've been talking w/ @gmbnomis about his content type that also does not have digest or size info on it. Here is [what he wrote](https://github.com/pulp/pulp/pull/3605#issuecomment-417020346) about the pipeline he is using to get around that, which I think is a similar problem.

plugin/pulpcore/plugin/stages/artifact_stages.py
mdellweg: This solves a slightly different problem, and is really meant as the last resort, where you cannot decide whether the upstream file changed before downloading it (and you expect it to change regularly).
Would be great to try the `If-Not-Modified-Since` and `If-None-Match` headers first.

mdellweg:
@bmbouter I thought this was the part that is not easy to prototype in the plugin, but i just realized how to.
If you prefer i can do just that and present this change in a broader context.

mdellweg:
@bmbouter It's in a very early stage (no pun intended), but you can have a look at what i came up with at https://github.com/ATIX-AG/pulp_deb/blob/78578eca68c66f7c7c82b1806a4986989a683ec6/pulp_deb/app/tasks/synchronizing.py#L58

gmbnomis:
@mdellweg There is one problem I am aware of: If the upstream meta-data provides no digest information, sync may decide to download an artifact and find that it is already present locally after the fact. In this case, we should use the artifact that is already present (that's a scenario that is also possible in pulp_cookbook which I haven't addressed). If I got this right, this is one aspect of your proposal.

However, I do not get the "is really meant as the last resort, where you cannot decide whether the upstream file changed before downloading it (and you expect it to change regularly)." thing.

Could you illustrate this using an example (what happens upstream and how should the stages behave)?

mdellweg:
@gmbnomis Well, it is the last resort, because it will always work (catch all mechanism), but there are a lot of ways that can avoid downloading the (entire) artifact in most of the situations.
The example i am using this for is pulp_deb (see above). We decided, we want to save the original metadata as artifacts to be able to publish the whole repository unchanged like a mirror. And the first thing to download now, is the Release file. We do not have a digest now, and it is expected to change bidaily.
I hope, this clears it up a little bit.

gmbnomis:
@mdellweg Yes, makes sense. Of course, downloading an artifact that is already present locally should be avoided. But if it can't, then your proposal is not the last resort, but the most sensible thing to do IMHO (i.e. don't let the sync fail, but use the existing artifact and discard the newly created one).

Regarding implementation, I am wondering whether https://github.com/pulp/pulp/pull/3744 does exactly that and is more optimized. @dkliban?

mdellweg:
@gmbnomis, @dkliban After fixing #3744 with #3777 this seems not to be needed anymore. Closing this one.

PR #3737, opened 2018-11-05
Fix SSL error and correct handling of Remote.ssl_validation.

jortel:
https://pulp.plan.io/issues/4062

Corrected passing the certificate and key arguments when building the ssl_context.

The `Remote.ssl_validation` documentation indicates that it controls whether or not the sever (peer) certificate is verified by the downloader.  Discovered this was not implemented correctly in the factory during testing and corrected.

### DIFF
--- a/plugin/pulpcore/plugin/download/factory.py
+++ b/plugin/pulpcore/plugin/download/factory.py
@@ -73,22 +73,22 @@ class DownloaderFactory:
             sslcontext = ssl.create_default_context(cafile=self._remote.ssl_ca_certificate.name)
             if self._remote.ssl_client_key.name and self._remote.ssl_client_certificate.name:
                 sslcontext.load_cert_chain(
-                    self._remote.ssl_client_key.name,
-                    self._remote.ssl_client_certificate.name
+                    self._remote.ssl_client_certificate.name,
+                    self._remote.ssl_client_key.name
                 )
         else:
             if self._remote.ssl_client_key.name and self._remote.ssl_client_certificate.name:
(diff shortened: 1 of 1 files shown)


plugin/pulpcore/plugin/download/factory.py
```diff
@@ -73,22 +73,23 @@ def _make_aiohttp_session_from_remote(self):
             sslcontext = ssl.create_default_context(cafile=self._remote.ssl_ca_certificate.name)
             if self._remote.ssl_client_key.name and self._remote.ssl_client_certificate.name:
                 sslcontext.load_cert_chain(
-                    self._remote.ssl_client_key.name,
-                    self._remote.ssl_client_certificate.name
+                    self._remote.ssl_client_certificate.name,
+                    self._remote.ssl_client_key.name
```
goosemania: Heh, nice :)

plugin/pulpcore/plugin/download/factory.py
```diff
@@ -73,22 +73,23 @@ def _make_aiohttp_session_from_remote(self):
...
                 )
 
         if sslcontext:
             tcp_conn_opts['ssl_context'] = sslcontext
 
-        if self._remote.ssl_validation:
-            tcp_conn_opts['verify_ssl'] = self._remote.ssl_validation
+        if not self._remote.ssl_validation:
```
goosemania: @jortel ,I think we should move this condition above the preceding one, or `sslcontext` may not get into `tcp_conn_opts`.
So the suggestion is to put L90-92  before L87.

PR #3738, opened 2018-11-05
Adds 'policy' to Remote

bmbouter:
Required PR: https://github.com/pulp/pulp_file/pull/132

This includes a serializer that validates input correctly.

https://pulp.plan.io/issues/3763
closes #3763

### DIFF
--- a/plugin/pulpcore/plugin/models/remote.py
+++ b/plugin/pulpcore/plugin/models/remote.py
@@ -1,3 +1,6 @@
+from gettext import gettext as _
+
+from pulpcore.app.models import Artifact as PlatformArtifact
 from pulpcore.app.models import Remote as PlatformRemote
 
 from pulpcore.plugin.download import DownloaderFactory
@@ -34,7 +37,7 @@ class Remote(PlatformRemote):
 
         Returns:
             DownloadFactory: The instantiated DownloaderFactory to be used by
-                get_downloader()
+                get_downloader().
         """
         try:
             return self._download_factory
@@ -42,20 +45,43 @@ class Remote(PlatformRemote):
             self._download_factory = DownloaderFactory(self)
             return self._download_factory
 
-    def get_downloader(self, url, **kwargs):
+    def get_downloader(self, remote_artifact=None, url=None, **kwargs):
         """
-        Get an asyncio capable downloader that is configured with the remote settings.
+        Get a downloader from either a RemoteArtifact or URL that is configured with this Remote.
+
+        This method accepts either `remote_artifact` or `url` but not both. At least one is
+        required. If neither or both are passed a ValueError is raised.
 
         Plugin writers are expected to override when additional configuration is needed or when
         another class of download is required.
 
         Args:
-            url (str): The download URL.
+            remote_artifact (:class:`~pulpcore.app.models.RemoteArtifact`): The RemoteArtifact to
+                download.
+            url (str): The URL to download.
             kwargs (dict): This accepts the parameters of
                 :class:`~pulpcore.plugin.download.BaseDownloader`.
+
+        Raises:
+            ValueError: If neither remote_artifact and url are passed, or if both are passed.
 
         Returns:
             subclass of :class:`~pulpcore.plugin.download.BaseDownloader`: A downloader that
             is configured with the remote settings.
         """
+        if remote_artifact and url:
+            raise ValueError(_("get_downloader() cannot accept both 'remote_artifact' and 'url'."))
+        if remote_artifact is None and url is None:
+            raise ValueError(_("get_downloader() requires either 'remote_artifact' and 'url'."))
+        if remote_artifact:
+            url = remote_artifact.url
+            expected_digests = {}
+            for digest_name in PlatformArtifact.DIGEST_FIELDS:
+                digest_value = getattr(remote_artifact, digest_name)
+                if digest_value:
+                    expected_digests[digest_name] = digest_value
+            if expected_digests:
+                kwargs['expected_digests'] = expected_digests
+            if remote_artifact.size:
+                kwargs['expected_size'] = remote_artifact.size
         return self.download_factory.build(url, **kwargs)
--- a/plugin/pulpcore/plugin/stages/declarative_version.py
+++ b/plugin/pulpcore/plugin/stages/declarative_version.py
@@ -11,7 +11,7 @@ from .content_unit_stages import ContentUnitSaver, QueryExistingContentUnits
 
 class DeclarativeVersion:
 
-    def __init__(self, first_stage, repository, mirror=True):
+    def __init__(self, first_stage, repository, mirror=True, download_artifacts=True):
         """
         A pipeline that creates a new :class:`~pulpcore.plugin.models.RepositoryVersion` from a
         stream of :class:`~pulpcore.plugin.stages.DeclarativeContent` objects.
@@ -20,7 +20,7 @@ class DeclarativeVersion:
         :class:`~pulpcore.plugin.stages.DeclarativeContent` object for each Content unit that should
         exist in the new :class:`~pulpcore.plugin.models.RepositoryVersion`.
 
-        The pipeline stages perform the following steps:
+        The pipeline stages perform the following steps by default:
 
         1. Create the new :class:`~pulpcore.plugin.models.RepositoryVersion`
         2. Query existing artifacts to determine which are already local to Pulp
@@ -45,7 +45,7 @@ class DeclarativeVersion:
         >>>         self.remote = remote
         >>>
         >>>     async def __call__(self, out_q):
-        >>>         downloader = remote.get_downloader(remote.url)
+        >>>         downloader = remote.get_downloader(url=remote.url)
         >>>         result = await downloader.run()
         >>>         for entry in read_my_metadata_file_somehow(result.path)
         >>>             unit = MyContent(entry)  # make the content unit in memory-only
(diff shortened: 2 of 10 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -105,6 +107,7 @@ def tls_storage_path(self, name):
     password = models.TextField(blank=True)
     last_synced = models.DateTimeField(blank=True, null=True)
     connection_limit = models.PositiveIntegerField(default=20)
+    policy = models.TextField(choices=POLICY_CHOICES, default=POLICY_MODE.IMMEDIATE)
```
daviddavis: Is it possible for a plugin writer to override these fields somehow?

bmbouter: They mostly could right now by defining the policy attribute on the subclass with another set of `choices`. There is one issue in that the helptext is duplicated because our use of choices isn't correct afaict. This PR kind of does the same thing since that problem already exists w/ the TASK_STATES. Other than that though they could do that, and we should fix that too but I'm not sure how yet. I filed it as this issue here:  https://pulp.plan.io/issues/4127

Note that these values affect the behavior of the streamer so if a plugin writer starts modifying the CHOICES values they are enabling/disabling features for the streamer. They need to be aware of that.

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -81,33 +81,36 @@ def __init__(self, first_stage, repository, mirror=True):
...
+                 content unit query and saving. 'False' will only handle content units and will not
+                 perform Artifact downloading and saving as part of the pipeline. 'False' is the
+                 default.
         """
         self.first_stage = first_stage
         self.repository = repository
         self.mirror = mirror
+        self.download_artifacts = download_artifacts
 
-    def pipeline_stages(self, new_version):
```
gmbnomis: Actually, `new_version` is here by intention if a plugin writer wants to insert a stage that uses it.

And I might even have a use case for it now: As said in my email on pulp-dev, I would like to avoid that content from other repositories show up during a sync of a repo (unless it is really the same content of course). If the upstream does not provide digests, one way to achieve this is to look at the current content in the base version when querying for existing content (like the sync in pulp_ansible does using ChangeSet https://github.com/pulp/pulp_ansible/blob/83f377602897ebf58261ca0d92473dc57a64ac25/pulp_ansible/app/tasks/synchronizing.py#L72-L75).

bmbouter: This makes good sense to me. I'm bringing this back now. Thanks!

bmbouter:
There is an issue with this because the 'policy' field is not optional. A default of 'immediate' is intended and is applied to the model layer [here](https://github.com/pulp/pulp/pull/3738/files#diff-a2fc7bce27efd36175122241684965e0R110). I can't see how to set a default as a ChoiceField in the drf docs [here](https://www.django-rest-framework.org/api-guide/fields/#choicefield).

When I create a remote without setting it I get:

```
{
    "policy": [
        "This field is required."
    ]
}
```

Or should this be a required field with no default?

pulpcore/pulpcore/app/models/repository.py
bmbouter: Actually I'm removing this default because it doesn't work w/ DRF.

pulpcore/pulpcore/constants.py
```diff
@@ -36,4 +37,22 @@
...
 )
 
+
+POLICY_MODE = SimpleNamespace(
+    IMMEDIATE='immediate',
+    ON_DEMAND='on_demand',
+    CACHE_ONLY='cache_only',
+)
+
+POLICY_CHOICES = (
```
dralley: Perhaps we should also have these descriptions in the docs somewhere?

bmbouter: Yes I agree, and an explanation of the whole lazy thing in general. Maybe I should go ahead and write this page?  https://docs.pulpproject.org/en/3.0/nightly/workflows/deferred-download.html

What about also me renaming that page to "Lazy Download"? Deferred Download just sounds so formal to me.

@asmacdo what do you think? ^

plugin/pulpcore/plugin/stages/declarative_version.py
```diff
@@ -103,11 +110,11 @@ def pipeline_stages(self, new_version):
...
 
         """
-        return [
-            self.first_stage,
-            QueryExistingArtifacts(), ArtifactDownloader(), ArtifactSaver(),
-            QueryExistingContentUnits(), ContentUnitSaver(),
-        ]
+        pipeline = [self.first_stage]
+        if self.download_artifacts:
+            pipeline.extend([QueryExistingArtifacts(), ArtifactDownloader(), ArtifactSaver()])
```
gmbnomis: Wondering whether we should still have `QueryExistingArtifacts()` in the non-download case. I think this makes sense for on demand, we might not want to do it for cache only(?).

plugin/pulpcore/plugin/models/remote.py
```diff
@@ -34,28 +37,51 @@ def download_factory(self):
...
+                get_downloader().
         """
         try:
             return self._download_factory
         except AttributeError:
             self._download_factory = DownloaderFactory(self)
             return self._download_factory
 
-    def get_downloader(self, url, **kwargs):
+    def get_downloader(self, remote_artifact=None, url=None, **kwargs):
```
gmbnomis: Doesn't this break (some) existing uses of `get_downloader`? This MR should have a breaking tag.  Should `url` come first?

bmbouter: Yes it does, but I forgot to set that. I added it just now. I actually put url second to make it super backwards-incompatible so that errors from this change won't pass silently. Thank you @gmbnomis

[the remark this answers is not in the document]
bmbouter: `url` only is used for downloading the very-first metadata files from the remote where you don't yet have a RemoteArtifact for it since it's the first thing.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: From a correctness interpretation of the name `download_artifacts` not including `QueryExistingArtifacts` is probably wrong. Query+Replace of Artifacts in the pipeline is not the same as downloading. So I see what you are saying.

My concern with adding it is the extra runtime cost; from our testing AIUI that stage adds a non-trivial amount of extra time to the overall runtime from the DB querying when processing say 70K Artifacts.

I think we should either do 1 of a few things.

a) Add back QueryExistingArtifacts and accept the runtime outcome
b) s/download_artifacts/handle_artifacts/ so that the name is more correct.
c) another idea

pulpcore/pulpcore/constants.py
```diff
@@ -36,4 +37,22 @@
     (SYNC_MODES.MIRROR, 'Add new content and remove content is no longer in the remote repository.')
 )
 
+
+POLICY_MODE = SimpleNamespace(
```
jortel: Suggest renaming to `DOWNLOAD_POLICY`.  Things are usually a _policy_ or a _mode_ so combining them seems both redundant and in conflict at the same time while not describing what the set of constants pertain to.

pulpcore/pulpcore/constants.py
jortel: Same here.

Suggest: `DOWNLOAD_POLICIES` or `DOWNLOAD_POLICY_CHOICES`

pulpcore/pulpcore/constants.py
bmbouter: Agree. fixed in next push.

pulpcore/pulpcore/constants.py
bmbouter: fixed in next push.

plugin/pulpcore/plugin/stages/declarative_version.py
gmbnomis: Is there a way to find out later that downloading a remote artifact is not necessary (e.g. when the artifact is requested; I don't know the full picture) Then, not replacing artifacts here would not hurt too much.

bmbouter:
@jortel I pushed those changes.

plugin/pulpcore/plugin/stages/declarative_version.py
jortel: I'm leaning toward leaving it the way it is.  When artifacts are not downloaded they should not be query/replaced in the pending-artifact either.  Mainly because the replacement isn't deterministic (may or may not be replaced) so logic further down the pipeline should not rely on it.  Leaving `QueryExistingArtifacts` in the pipeline without the downloading stage seems to add overhead  without adding any value.

jortel: As for the renaming (b), I'm not convinced it add clarity.  I'm not a fan of names that include _"handle"_ because it's such a vague term.

pulpcore/pulpcore/constants.py
asmacdo: Sounds good to me. The only hesitation I have is if this workflow differs plugin to plugin. For now, I think what you propose makes sense, if we get variation, we can move those docs to the plugin_template and each plugin.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: I agree the handle name is bad. What about `include_artifact_stages=True` and then it would include/disinclude all 3 of them?

jortel: What if the core, or a plugin includes an artifact related stage (in the future) that needs to be left in the pipeline even when artifacts are not downloaded?  This naming assumes that will never happen. 

The name `download_artifacts` is toggling a specific behavior that has the side-effect of omitting some of the stages.  The renaming to `include_artifact_stages` changes the toggled behavior to something that has the side-effect of toggling downloading which seems less clear.  I think.

pulpcore/pulpcore/constants.py
bmbouter: @asmacdo I wrote this story which I can also do today.  https://pulp.plan.io/issues/4218

Can you help me groom that story? Then I can write the docs in another PR very soon.

plugin/pulpcore/plugin/stages/declarative_version.py
bmbouter: Sounds good. I'm leaving the 3 stages being omitted and the name as is.

pulpcore/pulpcore/constants.py
bmbouter: I'm taking the other PR as assigned so that we can merge this one. Thanks for grooming it @asmacdo

PR #3739, opened 2018-11-06
Permit asking for nonexistent fields in new Pulp [DELIVERY-5012].

ipanova:

### DIFF
--- a/server/pulp/server/managers/content/query.py
+++ b/server/pulp/server/managers/content/query.py
@@ -38,6 +38,19 @@ class ContentQueryManager(object):
         @return:    list of content unit instances
         @rtype:     list
         """
+        serializer = units_controller.get_model_serializer_for_type(type_id)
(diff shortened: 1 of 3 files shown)


ipanova:
https://paste.fedoraproject.org/paste/YQINVU0YdtrPK1TnEpqyTA

@rohanpm this is the initial pr. couple of things:
1) turns out unit_association criteria needed to be patched since it is used in the content search.
2) i patched the query on for the --fields as was requested in the jira, but keep in mind that query in combination with, for example, sort would still raise an error.

```
$ pulp-admin rpm repo search --fields=display_name,repo_id,purr --sort=purr,ascending
Warning: path should have mode 0700 because it may contain sensitive information: /home/ipanova/.pulp/

                              Repositories

Invalid properties: ['Field: <purr> does not exist on objects in the <repos>
collection']

```

PR #3742, opened 2018-11-07
Handle building on CentOS

ekohl:
On CentOS you get the following output:

    >>> import platform
    >>> platform.dist()
    ('centos', '7.5.1804', 'Core')


streamer/setup.py
```diff
@@ -4,7 +4,7 @@
 
 # In EL twisted is still has sub-packages.
 dist = platform.dist()
-if dist[0] == 'redhat' and int(float(dist[1])) <= 7:
+if dist[0] in ('redhat', 'centos') and int(float(dist[1])) <= 7:
```
ekohl: This still has an issue with the float: ```ValueError: invalid literal for float(): 7.5.1804```

ehelms:
I saw this got approved, but the tests appear to be hung?

daviddavis:
ok test

PR #3744, opened 2018-11-07
Problem: bulk_create can fail with IntegrityError

dkliban:
Solution: add a manager that provides a bulk_get_or_create() method

closes #4060
https://pulp.plan.io/issues/4060

### DIFF
--- a/pulpcore/pulpcore/app/models/content.py
+++ b/pulpcore/pulpcore/app/models/content.py
@@ -4,14 +4,66 @@ Content related Django models.
 import hashlib
 
 from django.core import validators
-from django.db import models
+from django.db import IntegrityError, models
+from django.forms.models import model_to_dict
+
 from itertools import chain
 
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation, storage, fields
 from pulpcore.exceptions import DigestValidationError, SizeValidationError
 
 
-class Artifact(Model):
+class BulkCreateManager(models.Manager):
+    """
+    A manager that provides a bulk_get_or_create()
+    """
+
+    def bulk_get_or_create(self, objs, batch_size=None):
+        """
+        Insert the list of objects into the database and get existing objects from the database.
+
+        Do *not* call save() on each of the instances, do not send any pre/post_save signals,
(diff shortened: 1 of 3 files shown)


pulpcore/pulpcore/app/models/content.py
```diff
@@ -4,14 +4,64 @@
...
+        """
+        try:
+            return super().bulk_create(objs, batch_size=batch_size)
+        except IntegrityError:
+            objs = list(objs)
+            for i in range(len(objs)):
+                try:
+                    objs[i].save()
+                except IntegrityError:
+                    objs[i] = objs[i].objects.filter(objs[i].q())
```
gmbnomis: I didn't test this, but this looks like it should be a `get()`.

dkliban: Both work actually, but I did switch it to get().

gmbnomis:
I like this a lot! For pulp_cookbook, it solves a problem similar to the Docker problem mentioned in the issue.

pulpcore/pulpcore/app/models/content.py
```diff
@@ -4,14 +4,64 @@
...
 from pulpcore.app.models import Model, MasterModel, Notes, GenericKeyValueRelation, storage, fields
 from pulpcore.exceptions import DigestValidationError, SizeValidationError
 
 
-class Artifact(Model):
+class BulkCreateManager(models.Manager):
+    """
+    A manager that provides a bulk_get_or_create()
+    """
+    def bulk_get_or_create(self, objs, batch_size=None):
```
bmbouter: a nit, but can a line of whitespace be added above this  for readability?

pulpcore/pulpcore/app/models/content.py
```diff
@@ -4,14 +4,64 @@
...
+                except ValueError or IntegrityError:
+                    objs[i] = objs[i].__class__.objects.get(objs[i].q())
+        return objs
+
+
+class QueryMixin():
+    """
+    A mixin that provides models with querying utilities
+    """
+    def q(self):
```
bmbouter: adding a line above here too if you don't mind

PR #3746, opened 2018-11-08
Fixing missing hostname for deferred downloads

daviddavis:
A change was made in #4092 to allow users to use the server request
hostname as the redirect_host for on-demand repos by leaving host in the
lazy config as blank. However, the defered_download code doesn't have a
server request and therefore should default to socket.getfqdn() if
redirect_host is unset.

fixes #4120
https://pulp.plan.io/issues/4120


goosemania:
Niiice :tada:

daviddavis:
ok test

PR #3754, opened 2018-11-15
Bumps version for beta 14

dkliban:
[noissue]


daviddavis:
It looks like beta 14 was already tagged without this change. Shouldn't we do a beta 15 release then?

dkliban:
@daviddavis I tagged it wrong so nothing got published. I'll tag it correctly after this change is merged.

PR #3755, opened 2018-11-15
Pass extra_data as kwarg

asmacdo:
This is necessary if the overrides of this function have other kwargs
that need to be passed.

[noissue]


asmacdo:
example: https://github.com/pulp/pulp_docker/pull/274/files#diff-096737b594d36df447a28501228e9d89R35

PR #3756, opened 2018-11-15
Filtering the batch based on model type

daviddavis:
I'm pretty sure this is probably not the right fix. @gmbnomis or @bmbouter, hoping you can suggest what to do.

fixes #4165
https://pulp.plan.io/issues/4165

### DIFF
--- a/plugin/pulpcore/plugin/stages/content_unit_stages.py
+++ b/plugin/pulpcore/plugin/stages/content_unit_stages.py
@@ -53,6 +53,8 @@ class QueryExistingContentUnits(Stage):
             for model_type in content_q_by_type.keys():
                 for result in model_type.objects.filter(content_q_by_type[model_type]):
                     for declarative_content in batch:
+                        if type(declarative_content.content) is not model_type:
+                            continue
                         not_same_unit = False
                         for field in result.natural_key_fields():
                             in_memory_digest_value = getattr(declarative_content.content, field)


plugin/pulpcore/plugin/stages/content_unit_stages.py
```diff
@@ -52,7 +52,8 @@ class QueryExistingContentUnits(Stage):
 
             for model_type in content_q_by_type.keys():
                 for result in model_type.objects.filter(content_q_by_type[model_type]):
-                    for declarative_content in batch:
+                    model_batch = filter(lambda dcont: type(dcont.content) == model_type, batch)
```
gmbnomis: I only provided the `batch` iterator for this code, so I am not the expert here.

These nested loops look like they have the word "SLOW" imprinted on them (but let's wait for the outcome of the performance testing).

Given the current structure, it looks like the right thing to do. However, I would prefer staying imperative style here. There is already a `continue` in this loop, so I would prefer:

```python
                    for declarative_content in batch:
                        if type(dcont.content) is not model_type:
                            continue
                        not_same_unit = False
...
```
 An alternative would be to sort the batch into buckets by content type in the loop above (analogous to how the query is built) and iterate through the specific bucket here.

bmbouter:
@daviddavis thank you for working on this.

PR #3759, opened 2018-11-16
Delegate content authorization to optional content-guard.

jortel:
https://pulp.plan.io/issues/3970

Added ContentGuard to the plugin API.  Updated the Content application to delegate authorization to the guard associated with the distribution.  This required some minor re-factoring.


pulpcore/pulpcore/app/views/content.py
```diff
@@ -285,18 +288,56 @@ def _dispatch(self, request, path):
...
+                {
+                    'p': request.path,
+                    'g': guard.name,
+                    'r': str(pe)
+                })
+            raise
+        except Exception:
+            reason = _('Guard "{g}" failed:').format(g=guard.name)
+            log.debug(reason, exc_info=True)
+            raise PermissionError(reason)
```
daviddavis: I wonder if it would be better to let the exception bubble up and return a 500 error rather than a 403?

jortel: Perhaps.  I had trouble deciding this myself and could go either way.  In the past, we have tried to limit 500 codes to unanticipated and un-handled errors.  In other words, bugs were usually filed for 500 codes.  An exception in the content-guard _could_ be seen as anticipated and handled by core.  That said, the 403 seems a little inappropriate as well.

daviddavis: Yea, I worry that a 403 would be returned for something like an IndexError or KeyError. I'll leave it up to you though. The rest of the PR looks good so I'll approve.

jortel: My thinking behind returning the 403 is based on the fact that the user intends for the content to be protected.  An exception raised by the content guard can be interpreted (by the content app) as the guards' failure to authorize the request.  Although the authorization failure is not an explicit rejection, the end result is that the request is unauthorized.

pulpcore/pulpcore/app/views/content.py
```diff
@@ -285,18 +288,56 @@ def _dispatch(self, request, path):
...
+        Permit the request.
+
+        Authorization is delegated to the optional content-guard associated with the distribution.
+
+        Args:
+            request (django.http.HttpRequest): A request for a published file.
+            distribution (Distribution): The matched distribution.
+
+        Raises:
+            PermissionError: When not permitted.
```
bmbouter: Showing the full class path to `PermissionError` here would be helpful I think since its not a standard Exception.

PR #3760, opened 2018-11-18
Removing gen_file_publisher

rochacbruno:
The function `gen_file_publisher` is useless so removing it in favor of `gen_publisher` which does the same.

[noissue]


rochacbruno:
@pulp/qe

kersommoura:
@bmbouter, please, can you merge this one?

dralley:
@rochacbruno This is a perfectly fine change IMO, my only motivation for including it was that if we were to start adding options to the file publisher in the future it would be trivial to extend.

PR #3764, opened 2018-11-19
Problem: Artifact Saver fails to associate existing Artifacts

dkliban:
Solution: Update Declarative Artifacts with Artifacts returned by bulk_get_or_create()

[noissue]

### DIFF
--- a/plugin/pulpcore/plugin/stages/artifact_stages.py
+++ b/plugin/pulpcore/plugin/stages/artifact_stages.py
@@ -289,15 +289,17 @@ class ArtifactSaver(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches(in_q):
-            artifacts_to_save = []
+            da_to_save = []
             for declarative_content in batch:
                 for declarative_artifact in declarative_content.d_artifacts:
                     if declarative_artifact.artifact.pk is None:
                         declarative_artifact.artifact.file = str(declarative_artifact.artifact.file)
-                        artifacts_to_save.append(declarative_artifact.artifact)
+                        da_to_save.append(declarative_artifact)
 
-            if artifacts_to_save:
-                Artifact.objects.bulk_get_or_create(artifacts_to_save)
+            if da_to_save:
+                for da, artifact in zip(da_to_save, Artifact.objects.bulk_get_or_create(
+                        da.artifact for da in da_to_save)):
+                    da.artifact = artifact
 
             for declarative_content in batch:
                 await out_q.put(declarative_content)
--- a/plugin/pulpcore/plugin/stages/content_unit_stages.py
+++ b/plugin/pulpcore/plugin/stages/content_unit_stages.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 
-from django.db import transaction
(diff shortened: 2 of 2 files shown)


bmbouter:
@dkliban do you want to merge this and fix that separately or file it?

dkliban:
@bmbouter I want to merge and submit another PR

dkliban:
@bmbouter I take that back. I'll update tihs PR.

mdellweg:
I experienced, that assigning the pk in bulk_get_or_create acutally works.
Maybe if we update all properties, we don't break any expectations.

bmbouter:
@mdellweg I also pitched something similar to that as well; specifically to assign the pk and all attributes. I was suggesting all attributes because all attributes are present in the db but it's likely not-all attributes will be present on the in-memory model which is likely sparse from metdata).

The re-assign-the-attributes approach has downsides though, such as the in-memory model being in an "unsaved" state, even though it's data is actually all saved. That downside seemed risky enough for me to think this zip() approach is better, even with its memory and cpu downsides.

Other thoughts, ideas, criticisms are welcome.

plugin/pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -289,15 +289,17 @@ class ArtifactSaver(Stage):
...
                 for declarative_artifact in declarative_content.d_artifacts:
                     if declarative_artifact.artifact.pk is None:
                         declarative_artifact.artifact.file = str(declarative_artifact.artifact.file)
-                        artifacts_to_save.append(declarative_artifact.artifact)
+                        da_to_save.append(declarative_artifact)
 
-            if artifacts_to_save:
-                Artifact.objects.bulk_get_or_create(artifacts_to_save)
+            if da_to_save:
+                for da, artifact in zip(da_to_save, Artifact.objects.bulk_get_or_create(
```
mdellweg: Are we sure, that the order is preserved here?

bmbouter: Yeah this has been a key question we've been asking. It is not documented in their docs as something we can rely on. So that makes me a little nervous also. @dralley and I were talkinga bout this and he reviewed the bulk_create implementation to ensure it is order-preserving. We also checked that our implementation of bulk_get_or_create is also order-preserving.

We don't have a good way of guaranteeing this from django going forward. Maybe we should have a unit test so we can assert bulk_create in latest always returns items in-order. That would be one way to raise our confidence.

mdellweg: OK, let's hope for the best.
But i still think, we rely heavily on bulk_create to return the pk's here.
And if i read https://docs.djangoproject.com/en/2.1/ref/models/querysets/#bulk-create correctly, this is not a given on anything but postgres.

bmbouter: @mdellweg this is also concerning to me. We need to have a convo on the mailing list about if this is acceptable. This issue was introduced when we switched from UUID to ints since the uuid's were client side generated.

PR #3766, opened 2018-11-20
Add priority to DeclarativeContent

mdellweg:
This allows DeclarativeContent objects to bypass the batching process.

[noissue]

### DIFF
--- a/plugin/pulpcore/plugin/stages/api.py
+++ b/plugin/pulpcore/plugin/stages/api.py
@@ -68,10 +68,16 @@ class Stage:
 
         while not shutdown:
             content = await in_q.get()
+            if getattr(content, 'priority', False):
+                yield [content]
+                continue
             shutdown = add_to_batch(batch, content)
             while not shutdown:
                 try:
                     content = in_q.get_nowait()
+                    if getattr(content, 'priority', False):
+                        yield [content]
+                        continue
                 except asyncio.QueueEmpty:
                     break
                 else:
--- a/plugin/pulpcore/plugin/stages/models.py
+++ b/plugin/pulpcore/plugin/stages/models.py
@@ -66,11 +66,12 @@ class DeclarativeContent:
         ValueError: If `content` is not specified.
     """
 
-    __slots__ = ('content', 'd_artifacts', 'extra_data')
+    __slots__ = ('content', 'd_artifacts', 'extra_data', 'priority')
 
-    def __init__(self, content=None, d_artifacts=None, extra_data=None):
+    def __init__(self, content=None, d_artifacts=None, extra_data=None, priority=False):
         if not content:
             raise ValueError(_("DeclarativeContent must have a 'content'"))
         self.content = content
         self.d_artifacts = d_artifacts or []
         self.extra_data = extra_data or {}
+        self.priority = priority


mdellweg:
BTW, i was wondering, whether we really want the bunching to be greedy. It adds a lot of complexity in the bunching algorithm and allows (theoretically) for infinitely large bunches. IMHO Specifiying a constant bunch size gives more control over reducing unneeded database calls while at the same time limiting the sql-query size. It might even be better, not to bunch up the DeclarativeContent objects at all, but collect a certain number of Artifacts / ContentUnits to be created before calling bulk_create.
Just my two cents...

gmbnomis:
This looks a little 'ad-hoc' to me. What is the underlying problem? 

If there is no benefit in batching, then you just don't need to use the batches iterator. Just get single units from the queue directly (the ArtifactDownloader does this). Or is it because you have order constraints between content units that need to be preserved? Would it help to:

- guarantee that stages always preserve the order? (currently, at least the ArtifactDownloader does not do that)

- have one queue per content type? (or just more than one queue)

- be able to declare that some set of content unit instances belong together? (The stages would ensure that each set is handled "atomically".)

Regarding the bunch size:

> allows (theoretically) for infinitely large bunches. IMHO Specifiying a constant bunch size gives more control over reducing unneeded database calls while at the same time limiting the sql-query size. 

I think the bunch size is limited by the input queue size. Thus, the length of a batch is between min_size and the queue size (with the exception of the last batch, of course). If not, we have a problem indeed.

mdellweg:
Well the use case is, that i want to have metadata content_units. That means, i want to download and save them as every other content_type, but on the other hand, they decide (after downloading), what else to fetch. Therefore, i cannot wait for them to batch up, because the pipeline gets stuck that way. As i see it, the design is one pipeline that ends in one repository_version. Even if possible to install one pipeline per content_type, the benefit of this solution should be, that we do not break the maximal parallel downloads config option.

This preserved metadata is one of the requirements on the roadmap for pulp_deb and it allows to publish a synched repo unmodified (like a real mirror). That could actually be a _thing_ for other content types as well.

gmbnomis:
@mdellweg Does this mean that you have two Download stages? (First stage -> Download (metadata) -> ... -> Enrich with artifacts from metadata -> Download -> ...) That's a nice pattern to download metadata concurrently.

bmbouter:
One interest content type to think about is Maven, which for each artifact tells you more urls to download (it's dependencies). So you are discovering a graph that you don't know ahead of time and each time you explore a leaf node you download+parse it. What's interesting about that is that you can't know how many "Download->parse->" segments you need in the pipeline.

So to keep the pipeline really simple, if you add a custom stage after the downloader stage that does the parsing and instead of sending new declarative_content down the line, loop it back to the artifact downloader stage which also pulls from a "side queue". @mdellweg this is where you could send the "new" content. Would showing a solution like this be helpful?

The challenging aspect of this approach is at shutdown time. I have a plan to resolve this, and I can show the details in a prototype. The gist is to hold the segment of pipeline open until the "loopback" queue has also been shutdown which ensures no additional processing is required from loopback or the normal in_q which was already shutdown.

In terms of the batch starving the pipeline, how does it avoid starvation today? I imagined there would be a timer aspect to the batching implementation that prevents it from waiting for its batch size that will never come? For example, consider a downloader stage emits downloads really slowly to the next stage, it would be likely more efficient for it to process some things instead of waiting for a long time to handle 1 batch. What do you think?

mdellweg:
@bmbouter I think, the Maven challenge you describe here is solvable with the same mechanism i am using in pulp_deb right now. It associates Futures to DeclarativeContent objects that resolve to the Content object once created. Those futures can be awaited in the FirstStage and the artifacts in their results analyzed to fill more DeclarativeContent into the first queue. Only after handling all those futures, the terminal None is sent.
I guess for the problem you described, a flush marker to thaw all waiting batches on the way might be even better.

gmbnomis:
> I imagined there would be a timer aspect to the batching implementation that prevents it from waiting for its batch size that will never come? For example, consider a downloader stage emits downloads really slowly to the next stage, it would be likely more efficient for it to process some things instead of waiting for a long time to handle 1 batch.

@bmbouter No, there is no time dependency currently. I thought about that and decided against a timer in the iterator. The reasoning is that the processing time for a batch must be short (we don't have real concurrency). In your scenario from above (slow downloads) the difference is in the last batch (of course there will be "starvation" with previous batches, but that does not make a difference in the overall wall time for the sync). Even if processing the last batch takes 1 second per stage, we are talking about a couple of seconds we could save.

This reasoning applies for linear pipelines (i.e. no loops, content is generated in the first stage only). If, for example, a single content unit creates 10.000 other units, the processing time of these 10.000 other units is not negligible anymore and starvation will hurt.

However, I would like to avoid a timer here if possible:
- It is another parameter to tune and probably there is no "one size fits all" value
- As said, we don't have real concurrency and most stages are synchronous still. Creating more work than required in these stages by smaller batches might not come for free and lead to more blocking in the asynchronous stages.
- It is no real solution for the non-linear pipelines

mdellweg:
Succeeded by his rightful heir https://github.com/pulp/pulpcore-plugin/pull/2.
The PR is dead. Long live the PR.

daviddavis:
@mdellweg thank you for opening a PR against pulpcore-plugin 👍

PR #3773, opened 2018-11-27
Add test to content promotion

kersommoura:
Do the following:

 1. Create a repository that has at least one repository version.
 2. Create a publisher, and publication.
 3. Create 2 distributions - using the same publication. Those
    distributions will have different ``base_path``.
 4. Assert that distributions have the same publication.
 5. Select a content unit. Download that content unit from Pulp using
    the two different distributions.
    Assert that content unit has the same checksum when fetched from
    different distributions.

closes#4186
https://pulp.plan.io/issues/4186


kersommoura:
ping @pulp/qe 

Tested against source installation.

PR #3774, opened 2018-11-27
API Changes

dralley:
These are the API changes from the single-table-content work that were discussed.

### DIFF
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -2,9 +2,11 @@ from gettext import gettext as _
 
 from django.core import validators
 from django.db.models import Q
+from django.urls import reverse
 
 from rest_framework import serializers, fields
 from rest_framework.validators import UniqueValidator
+from rest_framework_nested.serializers import NestedHyperlinkedModelSerializer
 
 from pulpcore.app import models
 from pulpcore.app.serializers import (
@@ -20,7 +22,7 @@ from pulpcore.app.serializers import (
     ModelSerializer,
 )
 from pulpcore.app.serializers import validate_unknown_fields
-from rest_framework_nested.serializers import NestedHyperlinkedModelSerializer
+from pulpcore.app.util import get_view_name_for_model
 
 
 class RepositorySerializer(ModelSerializer):
@@ -407,17 +409,23 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
         view_name='versions-detail',
         lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
     )
-    _content_href = NestedIdentityField(
-        view_name='versions-content',
-        lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
+    _content = serializers.SerializerMethodField(
+        help_text=_('A mapping of the types of content in this version, and the HREF to view '
+                    'them.'),
+        read_only=True,
+        method_name='get_content'
     )
-    _added_href = NestedIdentityField(
-        view_name='versions-added-content',
-        lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
+    _content_added = serializers.SerializerMethodField(
+        help_text=_('A mapping of the types of content added in this version, and the HREF to '
+                    'view them.'),
+        read_only=True,
+        method_name='get_content_added'
     )
-    _removed_href = NestedIdentityField(
-        view_name='versions-removed-content',
-        lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
+    _content_removed = serializers.SerializerMethodField(
+        help_text=_('A mapping of the types of content removed from this version, and the HREF '
+                    'to view them.'),
+        read_only=True,
+        method_name='get_content_removed'
     )
     number = serializers.IntegerField(
         read_only=True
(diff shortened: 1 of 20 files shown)


pulpcore/pulpcore/app/models/repository.py
```diff
@@ -247,6 +247,20 @@ def content(self):
         )
         return Content.objects.filter(version_memberships__in=relationships)
 
+    @property
+    def content_by_type(self):
```
dralley: I'm not actually sure this function is necessary or even particularly useful, I may remove it before merging.  It seemed like an obvious thing to have at first but I realized that I can't think of anywhere it would actually be used.

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -60,3 +63,99 @@ class IsoDateTimeFilter(DateTimeFilter):
     def __init__(self, *args, **kwargs):
         kwargs.setdefault('help_text', _('ISO 8601 formatted dates are supported'))
         super().__init__(*args, **kwargs)
+
+
+class ContentRepositoryVersionFilter(Filter):
```
dralley: I want to DRY this but when I tried, it wasn't passing the "value" correctly.  I'll get to it later if I can.

pulpcore/pulpcore/app/models/repository.py
dralley: Should the ```added()``` and ```removed()``` methods be made properties, to match ```content```?  Or should we make ```content``` a method?  The latter would provide maximal flexibility in terms of being able to add functionality later on in the form of kwargs.

pulpcore/pulpcore/app/models/repository.py
```diff
@@ -247,6 +249,62 @@ def content(self):
         )
         return Content.objects.filter(version_memberships__in=relationships)
 
+    @property
```
dralley: These 3 properties are probably temporary, pending a better solution.  This is ugly but it was fairly simple for an initial implementation.

pulpcore/pulpcore/app/viewsets/__init__.py
```diff
@@ -1,10 +1,16 @@
 from .base import AsyncRemoveMixin, AsyncUpdateMixin, BaseFilterSet, NamedModelViewSet  # noqa
+# These need to be imported prior to ContentViewSet - do not shuffle this by alphabetical order
+from .custom_filters import (  # noqa
```
dralley: I struggled a little bit with circular imports here -- should we just move these to viewsets.content since they're only used in that one place?

pulpcore/pulpcore/app/viewsets/custom_filters.py
```diff
@@ -60,3 +63,101 @@ class IsoDateTimeFilter(DateTimeFilter):
     def __init__(self, *args, **kwargs):
         kwargs.setdefault('help_text', _('ISO 8601 formatted dates are supported'))
         super().__init__(*args, **kwargs)
+
+
+class RepoVersionHrefFilter(Filter):
```
dralley: Should I rename this to "BaseRepoVersionHrefFilter" since it's meant to be subclassed and not used directly?

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -110,13 +110,13 @@ def test_02_sync_content(self):
 
         self.assertIsNotNone(repo['_latest_version_href'])
 
-        content = get_content(repo)
+        content = get_file_content(repo)
```
dralley: As mentioned earlier, we could just always do ```get_content(repo)['file']``` instead.  Open to changing this if QE would prefer it that way.

dralley:
TODO: Docs changes.  I'll handle that after the code review is hashed out.

plugin/pulpcore/plugin/viewsets.py
```diff
@@ -2,7 +2,7 @@
 from pulpcore.app.response import OperationPostponedResponse  # noqa
 
 # Import Viewsets in platform that are potentially useful to plugin writers
-from pulpcore.app.viewsets import (  # NOQA
+from pulpcore.app.viewsets import (  # noqa
```
daviddavis: I fixed this [here](https://github.com/pulp/pulpcore-plugin/pull/3).

dkliban:
Perhaps we should just get rid of 'content', 'content_added', and 'content_removed' from the RepsositoryVersion serializer. If users want this in the future we can always add it.

pulpcore/app/serializers/repository.py
```diff
@@ -407,25 +407,21 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
         view_name='versions-detail',
         lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
     )
-    _content_href = NestedIdentityField(
-        view_name='versions-content',
-        lookup_field='number', parent_lookup_kwargs={'repository_pk': 'repository__pk'},
+    content = serializers.SerializerMethodField(
```
dralley: As I mentioned on IRC, we should be consider leaving some space for future expansion should it be necessary. 

I suppose if we ever needed to add back in the content urls, we could use "_hrefs" as we did previously.  But we could consider other alternatives.

bmbouter:
re: if urls should be included in responses or not. I think they should be because that is more usable for humans even though the bindings can't consume it in any reasonable way.

dralley:
@dkliban @bmbouter Alright, I've added back URLs and also added counts for added and removed content.  So we can subtract from this if we decide we don't want something.

pulpcore/app/serializers/repository.py
```diff
@@ -435,12 +421,154 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
...
+    )
+
+    def get_content_summary(self, obj):
+        """
+        The summary of contained content.
+
+        Returns:
+            dict: of {<type>: <count>}
+        """
+        annotated = obj.content.values('type').annotate(count=Count('type'))
```
dralley: We should consider whether this should be made a method like ```added()``` and ```removed()``` to be more consistent, and also leave open the possibility of adding kwargs later.  But that can be a different PR.

pulpcore/app/serializers/repository.py
```diff
@@ -435,12 +421,154 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
...
+        Args:
+            obj (pulpcore.app.models.RepositoryVersion): The RepositoryVersion being serialized.
+
+        Returns:
+            dict: {<type>: <url>}
+        """
+        content_urls = {}
+
+        for ctype in obj.content.values_list('type', flat=True):
+            ctype_model = obj.content.filter(type=ctype).first().cast().__class__
```
dralley: It would be good if we could find a better way to do this

PR #3777, opened 2018-11-29
Prevent failing bulk_create to consume objects from generator

mdellweg:
A failed bulk_create might have already iterated into the generator for
desired objects. This can be prevented by casting into a list first.

[noissue]


mdellweg:
@dkliban You might want to have a look at this.
There are two commits. The first one solves the problem i encountered. The second one tries to improve the q methods.

PR #3779, opened 2018-11-30
Adds the content app to pulpcore.content

bmbouter:
The Settings can be a lot simpler now that the architecture no longer
required redirection. It does need to account for the content being
served on a different HOST. This PR handles that.

Since the settings are now single values, they are much easier to
override using Dynaconf, which is great.

The content app can be removed now that the streamer code is handling it
all.

This PR is a prototype and has no docs. Use Pulp as normal , but run the
streamer from github.com/bmbouter/pulp_streamer/ and run the streamer
with:

`gunicorn pulpcore.streamer:server --bind localhost:8080 --worker-class aiohttp.GunicornWebWorker -w 2`

This PR needs pulp-smash changes from pulp_file so it requires the PR
below:

Required PR: https://github.com/pulp/pulp_file/pull/140

https://pulp.plan.io/issues/4239
closes #4239


.travis/install.sh
```diff
@@ -6,6 +6,12 @@ pip install "Django<=$DJANGO_MAX"
 pip install -r test_requirements.txt
 pip install -e .
 
+
+# temporarily install the pulpcore-content app from bmbouter's repo
```
dkliban: move this to after pulpcore-plugin is installed

dkliban: i was mistaken about the previous suggestion. please disregard.

bmbouter:
FYI since this PR removes the apache and nginx file serving integration I'm making another issue to bring that back. I'll post a link to that other issue here today. We can prioritize that together and I'm happy to bring it back immediately in another PR if that is what we should do. I'll link to the issue when I make it.

dralley:
@bmbouter The Travis config changes in this PR need to be copied over to plugins and the plugin template, since without the content app *or* the streamer, the "download content" tests fail.

dralley:
Also our docs on how to run Pulp

bmbouter:
@dralley I'm working on fixing Travis for the plugins and after that I'm fixing the docs. Hopefully all will get done today.

PR #3785, opened 2018-12-05
Don't fail migration in case there is no exception code

goosemania:
Not all MongoDB versions report a specific code when index is not found.

closes #4225
https://pulp.plan.io/issues/4225

### DIFF
--- a/server/pulp/server/db/migrations/0029_applicability_schema_change.py
+++ b/server/pulp/server/db/migrations/0029_applicability_schema_change.py
@@ -51,7 +51,7 @@ def migrate(*args, **kwargs):
     try:
         rpa_collection.drop_index("profile_hash_-1_repo_id_-1")
     except pymongo.errors.OperationFailure as e:
-        if e.code == 27:
+        if not e.code or e.code == 27:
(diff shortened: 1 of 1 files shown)


server/pulp/server/db/migrations/0029_applicability_schema_change.py
```diff
@@ -51,7 +51,7 @@ def migrate(*args, **kwargs):
     try:
         rpa_collection.drop_index("profile_hash_-1_repo_id_-1")
     except pymongo.errors.OperationFailure as e:
-        if e.code == 27:
+        if not e.code or e.code == 27:
             # index not found - good, it's been removed before
             pass
         else:
```
dparalen: I think that in case the `e.code` being `None` isn't unique to the index not found case, this might never be executed on the older versions of Mongo, that don't populate the `e.code`, silencing other issues.

goosemania: @dparalen , I agree completely but we need it to work on old Mongo, in my opinion, until we are ready to require mongo 3

PR #3789, opened 2018-12-11
Unpin the redis requirement

daviddavis:
We're requiring rq >= 0.12 and redis < 3.0 but the new release of rq (0.13.0) requires redis >= 3.0 so our package requirements raise a conflict. I suggest we unpin the redis requirement and let pip install the right version of redis based on what rq needs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ requirements = [
     'drf-yasg',
     'psycopg2-binary',
     'PyYAML',
-    'redis >= 2.7.0, <3.0',
+    'redis',
     'rq>=0.12.0',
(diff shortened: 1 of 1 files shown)


rochacbruno:
I just got the same error, thanks.

dralley:
@daviddavis Should we get rid of the dependency entirely?  It was only added for the purpose of pinning it, but rq will pull it in either way.

daviddavis:
@dralley yea I think that makes sense actually. Want to open a PR or should I?

dralley:
I can do it

PR #3792, opened 2018-12-11
Problem: pulpcore packaging is missing parts of the app 

dkliban:
Solution: add the migrations package

This patch adds the pulpcore.app.migrations package to the built asset.

This patch also updates CI install script to ensure that we are testing the same way users will be
installing pulpcore.

re: #4245
https://pulp.plan.io/issues/4245


dkliban:
Closing this in favor of the actual migrations being committed.

PR #3800, opened 2018-12-14
Create a new Task field job_id to store rq job ids

daviddavis:
fixes #3871
https://pulp.plan.io/issues/3871

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -266,7 +266,7 @@ class Task(Model):
         parent (models.ForeignKey): Task that spawned this task (if any)
         worker (models.ForeignKey): The worker that this task is in
     """
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    job_id = models.UUIDField(unique=True, default=uuid.uuid4, editable=False)
     state = models.TextField(choices=TASK_CHOICES)
 
     started_at = models.DateTimeField(null=True)
@@ -287,11 +287,11 @@ class Task(Model):
             pulpcore.app.models.Task: The current task.
         """
         try:
-            task_id = get_current_job().id
+            job_id = get_current_job().id
         except AttributeError:
             task = None
         else:
-            task = Task.objects.get(pk=task_id)
+            task = Task.objects.get(job_id=job_id)
         return task
 
     def set_running(self):
--- a/pulpcore/app/response.py
+++ b/pulpcore/app/response.py
@@ -1,6 +1,8 @@
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 
+from .models import Task
+
 
 class OperationPostponedResponse(Response):
     """
@@ -23,5 +25,6 @@ class OperationPostponedResponse(Response):
                 the response.
             request (rest_framework.request.Request): Request used to generate the _href urls
         """
-        task = {"task": reverse('tasks-detail', args=[result.id], request=None)}
-        super().__init__(data=task, status=202)
+        task = Task.objects.get(job_id=result.id)
+        resp = {"task": reverse('tasks-detail', args=[task.pk], request=None)}
+        super().__init__(data=resp, status=202)
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -48,7 +48,7 @@ def _acquire_worker(resources):
     return Worker.objects.get_unreserved_worker()
 
 
-def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwargs, options):
+def _queue_reserved_task(func, inner_job_id, resources, inner_args, inner_kwargs, options):
     """
     A task that encapsulates another task to be dispatched later.
 
@@ -63,7 +63,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
 
     Args:
         func (basestring): The function to be called
-        inner_task_id (basestring): The UUID to be set on the task being called. By providing
+        inner_job_id (basestring): The job_id to be set on the task being called. By providing
             the UUID, the caller can have an asynchronous reference to the inner task
             that will be dispatched.
         resources (basestring): The urls of the resource you wish to reserve for your task.
@@ -74,7 +74,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
         options (dict): For all options accepted by enqueue see the RQ docs
     """
     redis_conn = connection.get_redis_connection()
-    task_status = Task.objects.get(pk=inner_task_id)
+    task_status = Task.objects.get(job_id=inner_job_id)
     task_name = func.__module__ + '.' + func.__name__
     while True:
         if task_name == "pulpcore.app.tasks.orphan.orphan_cleanup":
@@ -86,7 +86,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                 task_status.state = TASK_STATES.RUNNING
                 task_status.save()
                 q = Queue('resource_manager', connection=redis_conn, is_async=False)
-                q.enqueue(func, args=inner_args, kwargs=inner_kwargs, job_id=inner_task_id,
+                q.enqueue(func, args=inner_args, kwargs=inner_kwargs, job_id=inner_job_id,
                           timeout=TASK_TIMEOUT, **options)
                 task_status.state = TASK_STATES.COMPLETED
                 task_status.save()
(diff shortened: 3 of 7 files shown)


pulpcore/app/models/task.py
```diff
@@ -287,11 +287,11 @@ def current():
             pulpcore.app.models.Task: The current task.
         """
         try:
-            task_id = get_current_job().id
+            task_job_id = get_current_job().id
```
jortel: `task_job_id` was confusing the first time I read it.  Recommend `job_id` would be better.

pulpcore/app/models/task.py
```diff
@@ -287,11 +287,11 @@ def current():
...
         except AttributeError:
             task = None
         else:
-            task = Task.objects.get(pk=task_id)
+            task = Task.objects.get(job_id=task_job_id)
```
dralley: ```job_id``` is a UUID, ```id``` is an integer AutoField, so wouldn't this + the next line be invalid?

pulpcore/app/response.py
```diff
@@ -23,5 +25,6 @@ def __init__(self, result, request):
                 the response.
             request (rest_framework.request.Request): Request used to generate the _href urls
         """
-        task = {"task": reverse('tasks-detail', args=[result.id], request=None)}
-        super().__init__(data=task, status=202)
+        task = Task.objects.get(job_id=result.id)
```
dralley: Same here

pulpcore/tasking/tasks.py
```diff
@@ -183,9 +183,9 @@ def enqueue_with_reservation(func, resources, args=None, kwargs=None, options=No
     current_job = get_current_job(connection=redis_conn)
     parent_kwarg = {}
     if current_job:
-        current_task = Task.objects.get(id=current_job.id)
+        current_task = Task.objects.get(job_id=current_job.id)
```
dralley: Same here

pulpcore/tasking/tasks.py
```diff
@@ -141,7 +141,7 @@ def _release_resources(task_id):
         exc = RuntimeError(msg.format(task_id=task_id))
         task.set_failed(exc, None)
 
-    Task.objects.get(pk=task_id).release_resources()
+    Task.objects.get(job_id=task_id).release_resources()
```
jortel: Should this be get(job_id=job_id)?  Or, should task_id be renamed?

pulpcore/tasking/tasks.py
```diff
@@ -131,7 +131,7 @@ def _release_resources(task_id):
 
     """
     try:
-        task = Task.objects.get(pk=task_id, state=TASK_STATES.RUNNING)
+        task = Task.objects.get(job_id=task_id, state=TASK_STATES.RUNNING)
```
jortel: Same here.  `job_id`=`task_id` seems wrong.

pulpcore/tasking/tasks.py
```diff
@@ -74,7 +74,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
         options (dict): For all options accepted by enqueue see the RQ docs
     """
     redis_conn = connection.get_redis_connection()
-    task_status = Task.objects.get(pk=inner_task_id)
+    task_status = Task.objects.get(job_id=inner_task_id)
```
jortel: Same here.  `job_id=inner_task_id` seems wrong.

pulpcore/tasking/tasks.py
```diff
@@ -183,9 +183,9 @@ def enqueue_with_reservation(func, resources, args=None, kwargs=None, options=No
     current_job = get_current_job(connection=redis_conn)
     parent_kwarg = {}
     if current_job:
-        current_task = Task.objects.get(id=current_job.id)
+        current_task = Task.objects.get(job_id=current_job.id)
         parent_kwarg['parent'] = current_task
-    Task.objects.create(pk=inner_task_id, state=TASK_STATES.WAITING, **parent_kwarg)
+    Task.objects.create(job_id=inner_task_id, state=TASK_STATES.WAITING, **parent_kwarg)
```
jortel: Same here.  `job_id=inner_task_id` seems wrong.

pulpcore/app/models/task.py
bmbouter: +1 on this

bmbouter:
Do we still want `class Task(Model):` or was this supposed to switch it to inherit from Pulp's parent Model?

dralley:
> Do we still want `class Task(Model):` or was this supposed to switch it to inherit from Pulp's parent Model?

We decided to keep it this way.

pulpcore/tasking/tasks.py
```diff
@@ -86,7 +86,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                 task_status.state = TASK_STATES.RUNNING
                 task_status.save()
                 q = Queue('resource_manager', connection=redis_conn, is_async=False)
-                q.enqueue(func, args=inner_args, kwargs=inner_kwargs, job_id=inner_task_id,
+                q.enqueue(func, args=inner_args, kwargs=inner_kwargs, job_id=inner_job_id,
```
bmbouter: This is making more sense :)

daviddavis:
@bmbouter I fixed it. I think should add task cancel unit tests since pulp-smash doesn't/can't test it.

dralley:
+1 to adding tests for task cancellation

pulpcore/app/response.py
```diff
@@ -23,5 +25,6 @@ def __init__(self, result, request):
...
+        resp = {"task": reverse('tasks-detail', args=[task.id], request=None)}
```
CodeHeeler: Same issue with 'id'

pulpcore/tasking/tasks.py
```diff
@@ -113,35 +113,35 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
...
+        task = Task.objects.get(job_id=job_id, state=TASK_STATES.RUNNING)
     except Task.DoesNotExist:
         pass
     else:
         msg = _('The task {task_id} exited immediately for some reason. Marking as '
                 'failed. Check the logs for more details')
-        _logger.error(msg.format(task_id=task_id))
-        exc = RuntimeError(msg.format(task_id=task_id))
+        _logger.error(msg.format(task_id=task.id))
+        exc = RuntimeError(msg.format(task_id=task.id))
```
CodeHeeler: Same issue with 'id' as above.

pulpcore/tasking/tasks.py
```diff
@@ -178,15 +178,15 @@ def enqueue_with_reservation(func, resources, args=None, kwargs=None, options=No
...
 
     resources = {util.get_url(resource) for resource in resources}
-    inner_task_id = str(uuid.uuid4())
+    inner_job_id = str(uuid.uuid4())
     redis_conn = connection.get_redis_connection()
     current_job = get_current_job(connection=redis_conn)
     parent_kwarg = {}
     if current_job:
-        current_task = Task.objects.get(id=current_job.id)
+        current_task = Task.objects.get(job_id=current_job.id)
```
CodeHeeler: Same issue with 'id'

pulpcore/app/models/task.py
```diff
@@ -287,11 +287,11 @@ def current():
             pulpcore.app.models.Task: The current task.
         """
         try:
-            task_id = get_current_job().id
+            job_id = get_current_job().id
```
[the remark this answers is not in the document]
bmbouter: We're not in control of this name `id` because this is the return type from rq's method `get_current_job()`. So we can't switch this.

pulpcore/app/response.py
bmbouter: This one is ours, so this could become `pk` possible. I don't have a preference.

pulpcore/tasking/tasks.py
bmbouter: This one is ours, so this could become `pk` possible. I don't have a preference.

pulpcore/tasking/tasks.py
bmbouter: This is the `rq` one from `get_current_job` so this one is not in our control.

pulpcore/app/models/task.py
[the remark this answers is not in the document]
CodeHeeler: Ah, gotcha.  Unfortunate, but I understand.  Thanks for clarifying.

pulpcore/app/response.py
CodeHeeler: I think this really should be pk then given the redis 'id' to ease confusion where we are able.

pulpcore/app/models/task.py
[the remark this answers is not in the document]
jortel: Also, think of it this way.  get_current_job() returns (presumably) a `Job` object which naturally has an `id` attribute.  So reading like:
```
job = job_id = get_current_job()
job_id = job.id
```
Seems very intuitive to me.

It would also be clearer if the code just used `job.id` instead of `job_id`

PR #3802, opened 2018-12-17
Align default settings with new content app

mdellweg:
https://pulp.plan.io/issues/4239
re #4239


mdellweg:
You could have asked me to delete it. ;)

mdellweg:
@bmbouter You might also need to re'MODIFIED' the plan.io ticket.

bmbouter:
@mdellweg I moved the ticket back to MODIFIED. I do think that we need to delete this file every since we switched to dynaconf. I'll email pulp-dev because I want to highlight the reasoning there.

bmbouter:
I wrote it up here:  https://www.redhat.com/archives/pulp-dev/2018-December/msg00065.html

PR #3803, opened 2018-12-17
Do not fail if file already moved to storage

mdellweg:
https://pulp.plan.io/issues/4272
closes #4272

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -44,7 +44,7 @@ class BulkCreateManager(models.Manager):
             for i in range(len(objs)):
                 try:
                     objs[i].save()
-                except ValueError or IntegrityError:
+                except IntegrityError:
                     objs[i] = objs[i].__class__.objects.get(objs[i].q())
         return objs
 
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/fields.py
```diff
@@ -28,7 +32,11 @@ def pre_save(self, model_instance, add):
 
         """
         file_name = str(model_instance.file)
-        if file_name.startswith(os.path.join(settings.MEDIA_ROOT, 'artifact')):
+        upload_to = self.upload_to(model_instance, '')
+        log.info('file_name: {}'.format(file_name))
+        log.info('upload_to: {}'.format(upload_to))
+        if file_name != upload_to and file_name.startswith(
```
mdellweg: Not sure, whether we should remove the test altogether...

pulpcore/app/models/content.py
```diff
@@ -44,7 +44,7 @@ def bulk_get_or_create(self, objs, batch_size=None):
             for i in range(len(objs)):
                 try:
                     objs[i].save()
-                except ValueError or IntegrityError:
+                except (ValueError, IntegrityError):
```
dkliban: I think we can remove the ValueError exception handling. If it gets raised now, it means something did go wrong.

mdellweg: ... took me a while to realize that the syntax for catching manifold exceptions was wrong...

PR #3804, opened 2018-12-17
Add docs on lazy to the workflows section

bmbouter:
https://pulp.plan.io/issues/4218
closes #4218

### DIFF
--- a/docs/workflows/lazy-downloading.rst
+++ b/docs/workflows/lazy-downloading.rst
@@ -1,0 +1,63 @@
+Lazy Downloading
+================
+
+Overview
+--------
+
+Pulp can sync content in a few modes: 'immediate', 'on_demand', and 'streamed'. Each provides a
+different behavior on how and when Pulp acquires content. These are set as the `policy` attribute
+of the :term:`Remote` performing the sync. Policy is an optional parameter and defaults to
+`immediate`.
+
+immediate
+  When performing the sync, download all :term:`Artifacts<artifact>` now. Also download all metadata
+  now to create the content units in Pulp, associated with the
+  :term:`repository version<RepositoryVersion>` created by the sync. `immediate` is the default, and
+  any plugin providing a sync is expected to implement the `immediate` mode.
+
+on_demand
+  When performing the sync, do not download any :term:`Artifacts<artifact>` now. Download all
+  metadata now to create the content units in Pulp, associated with the
+  :term:`repository version<RepositoryVersion>` created by the sync. Clients requesting content
+  trigger the downloading of :term:`Artifacts<artifact>`, which are saved into Pulp to be served to
+  future clients.
+
+  This mode is ideal for saving disk space because Pulp never downloads and stores
+  :term:`Artifacts<artifact>` that clients don't need. Units created from this mode are
+  :term:`lazy content units<lazy content>`.
+
+streamed
+  When performing the sync, do not download any :term:`Artifacts<artifact>` now. Download all
+  metadata now to create the content units in Pulp, associated with the
+  :term:`repository version<RepositoryVersion>` created by the sync. Clients requesting content
+  trigger the downloading of :term:`Artifacts<artifact>`, which are *not* saved into Pulp. This
+  content will be re-downloaded with each client request.
+
+  This mode is ideal for content that you especially don't want Pulp to store over time. For
+  instance, syncing from a nightly repo would cause Pulp to store every nightly ever produced which
+  is likely not valuable. Units created from this mode are :term:`lazy content units<lazy content>`.
+
+
+Does Plugin X Support Lazy?
+---------------------------
+
+Plugins do have to enable support for these modes, so check that plugin's documentation to be sure.
+See the `Pulp Plugin API <../../pulpcore-plugin/nightly/>`_ documentation for more details on how to
+add lazy support to a plugin.
+
+
+Associating Lazy Content with Additional Repository Versions
+------------------------------------------------------------
+
+A :term:`lazy content unit<lazy content>` can be associated and unassociated from a
+:term:`repository version<RepositoryVersion>` just like a normal unit. Note that the original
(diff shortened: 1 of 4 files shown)


docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+  When performing the sync, do not download any Artifacts now. Download all metadata now to create
+  the content units in Pulp, associated with the RepositoryVersion created by the sync. Clients
+  requesting content trigger the downloading of Artifacts, which are saved into Pulp to be served to
+  future clients.
+
+  This mode is ideal for saving disk space because Pulp never downloads and stores packages that
+  clients don't need.
+
+streamed
+  When performing the sync, do not download any Artifacts now. Download all metadata now to create
```
asmacdo: Could you add links to the glossary for anything in there? Artifacts, RepositoryVersion, etc

    :term:`artifacts<artifact>`

docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+  the content units in Pulp, associated with the RepositoryVersion created by the sync. Clients
+  requesting content trigger the downloading of Artifacts, which are *not* saved into Pulp. This
+  content will be re-downloaded with each client request.
+
+  This mode is ideal for content that you especially don't want Pulp to store over time. For
```
asmacdo: s/over time/unless it is requested/

docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+---------------------------
+
+Plugins do have to enable support for Lazy, so check with your plugin docs to be sure. See the
+plugin docs for more details on how to add lazy support to a plugin.
+
+
+Terminology
+-----------
+
+lazy content
```
asmacdo: Maybe it would be best to move this section to the glossary?

docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+
+  This mode is ideal for content that you especially don't want Pulp to store over time. For
+  instance, syncing from a nightly repo would cause Pulp to store every nightly ever produced which
+  is likely not valuable.
+
+
+Does Plugin X Support Lazy?
+---------------------------
+
+Plugins do have to enable support for Lazy, so check with your plugin docs to be sure. See the
```
asmacdo: s/check with your plugin/check that plugin's/
s/see the plugin docs/see the plugin writer docs/ +link

docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+normal unit. Note that the original `Remote` will be used to download content should a client
+request it, even as that content is made available in multiple places.
+
+
+.. warning::
+
+    Deleting a `Remote` that was used in a sync with either the `on_demand` or `streamed` options
+    can break published data. Specifically, clients who want to fetch content that a Remote was
+    providing access to would begin to 404. Recreating a remote and re-triggering a sync will cause
+    these broken units to recover again.
```
asmacdo: This surprises me a little bit. Wouldn't "recreating the remote" actually create a new remote with a different pk, which wouldn't be found by the RemoteArtifact? In fact, doesn't CASCADE mean that if the Remote is deleted, so are the RemoteArtifacts?

https://github.com/pulp/pulp/blob/79a741359b37cd0ac0f48cf0f72499e451f26b5b/pulpcore/app/models/content.py#L335

docs/workflows/lazy-downloading.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Plugins do have to enable support for Lazy, so check with your plugin docs to be sure. See the
+plugin docs for more details on how to add lazy support to a plugin.
+
+
+Terminology
+-----------
+
+lazy content
+  A Content unit that was sync'd into Pulp does not have all of its Artifacts downloaded and stored.
```
asmacdo: nitpick: s/sync'd/synchronized/

docs/workflows/lazy-downloading.rst
bmbouter: yes fixed in next push.

bmbouter: this came out really nicely I think.

docs/workflows/lazy-downloading.rst
bmbouter: yes a great idea. fixed in next push.

docs/workflows/lazy-downloading.rst
bmbouter: sounds good. fixed in next push.

docs/workflows/lazy-downloading.rst
bmbouter: Removing the Remote will cascade delete the `RemoteArtifact` objects so the content app will 404 instead of lazily fetching when that happens. To recover from this, the user could make another Remote and sync it w/ either `on_demand` or `streamed` and this would recognize the content unit is already in Pulp but it would make a new `RemoteArtifact` associated with the `ContentArtifact` that already existed. Then when the client fetches again, the new `RemoteArtifact` will be used and invoke the new Remote to serve the already-existed-in-pulp content unit. Does that make sense? Is that the behavior your expected?

docs/workflows/lazy-downloading.rst
bmbouter: fixed in next push.

docs/workflows/lazy-downloading.rst
asmacdo: Ah yeah, triggering a new sync would work. I think the only change is that I would expect that deleting the remote would be blocked if there are associated remote artifacts unless a force=True or similar was added. But that is definitely outside the scope of this PR, and probably not even worth filing an issue unless someone complains. LGTM.

bmbouter:
Thanks for the review @asmacdo

PR #3807, opened 2018-12-18
Adds importing of the 'content' module

bmbouter:
The 'content' module needs to be auto-imported when the content app
starts so that loading is added to its startup sequence with this PR.
Importing is all we have to do and aiohttp's registration mechanisms
take over.

https://pulp.plan.io/issues/4273
re #4273

### DIFF
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -1,14 +1,24 @@
-from aiohttp import web
+from contextlib import suppress
+from importlib import import_module
 
+from aiohttp import web
 from django.conf import settings
+from pulpcore.app.apps import pulp_plugin_configs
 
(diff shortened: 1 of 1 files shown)


bmbouter:
Here is the diff I used to "duplicate" the Content API onto a second root URL called `/another/`.

```
diff --git a/pulp_file/app/content.py b/pulp_file/app/content.py
index e69de29..7148627 100644
--- a/pulp_file/app/content.py
+++ b/pulp_file/app/content.py
@@ -0,0 +1,5 @@
+from aiohttp import web
+from pulpcore.content import app, Handler
+
+
+app.add_routes([web.get('/another/{path:.+}', Handler().stream_content)])

```

PR #3810, opened 2018-12-20
Checking in migrations

daviddavis:
fixes #4067
https://pulp.plan.io/issues/4067


daviddavis:
No idea why pep8speaks is raising errors for migrations. Should be ignored:

https://github.com/pulp/pulp/blob/master/.pep8speaks.yml#L8

asmacdo:
@daviddavis could you also remove the lines from https://github.com/pulp/pulp/blob/master/docs/installation/instructions.rst that instruct users to make migrations?

daviddavis:
@asmacdo am I just removing the makemigrations line?

asmacdo:
daviddavis yeah remove makemigrations at a minimum, but if you wanted to pick up https://pulp.plan.io/issues/3062 you might be able to remove the `migrate --noinput auth` line too

PR #3812, opened 2018-12-20
Add lots more content app docs

bmbouter:
https://pulp.plan.io/issues/4218
closes #4218


bmbouter:
The `pulp-content` doesn't work because I'm not including the script yet. Do we want to do that in this PR or can I do it in a followup one?

bmbouter:
Thanks! I'm also going to add the script in a followup PR

PR #3813, opened 2018-12-21
Add support for testing against mariadb

daviddavis:
fixes #4270
https://pulp.plan.io/issues/4270


daviddavis:
Found one error already:

```
pulp_app.ContentGuard.name: (mysql.E001) MySQL does not allow unique CharFields to have a max_length > 255.
```

daviddavis:
Another error on Worker name:

```
_mysql_exceptions.OperationalError: (1170, "BLOB/TEXT column 'name' used in key specification without a key length")
```

daviddavis:
At this point it seems that the error is caused by the code relying on bulk_create returning PKs.

bmbouter:
I agree, and I think we need to fix all that probably before we merge this. I think of all of this as RC blocker stuff.

bmbouter:

pulpcore/app/models/base.py
```diff
@@ -14,7 +16,7 @@ class Model(models.Model):
         * https://docs.djangoproject.com/en/1.8/topics/db/models/#automatic-primary-key-fields
 
     """
-    _id = models.AutoField(primary_key=True)
+    _id = models.UUIDField(default=uuid.uuid4, primary_key=True)
```
dralley, line 19: @daviddavis When you reopen this issue, please add ```editable=False``` here.

dralley:
@daviddavis Additionally, please copy these two ```.q()``` method changes from my branch here

https://github.com/pulp/pulpcore/compare/master...dralley:uuid?expand=1#diff-5308662d441ee7c5774295277cba139cL63

As well as the ```object_id``` change, and then regenerate migrations.

You will also need to merge this PR https://github.com/pulp/pulpcore-plugin/pull/69

PR #3814, opened 2018-12-22
Combine test coverage

mdellweg:
[noissue]

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -1,21 +1,38 @@
 #!/usr/bin/env bash
 # coding=utf-8
 
-set -veuo pipefail
+set -mveuo pipefail
+
+wait_for_pulp() {
+  TIMEOUT=${1:-5}
+  while [ "$TIMEOUT" -gt 0 ]
+  do
+    echo -n .
+    sleep 1
+    TIMEOUT=$(($TIMEOUT - 1))
+    if [ $(http :8000/pulp/api/v3/status/ | jq '.database_connection.connected and .redis_connection.connected') = 'true' ]
+    then
+      echo
+      return
+    fi
+  done
+  echo
+  return 1
+}
 
 if [ "$TEST" = 'docs' ]; then
+  pulp-manager runserver >> ~/django_runserver.log 2>&1 &
+  sleep 5
   cd docs
   make html
-  set +euo pipefail
-  return "$?"
+  exit
 fi
-
 
 # check the commit message
 ./.travis/check_commit.sh
 
 # Lint code.
-flake8 --config flake8.cfg || exit 1
+flake8 --config flake8.cfg
 
 # Run unit tests.
 coverage run manage.py test ./pulpcore/tests/unit/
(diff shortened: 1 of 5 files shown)


.travis/script.sh
```diff
@@ -29,16 +46,25 @@ show_logs_and_return_non_zero() {
...
+PULP_SERVER_PID=$!
+echo "CONTENT_APP_PID=$CONTENT_APP_PID"
+echo "PULP_SERVER_PID=$PULP_SERVER_PID"
+
+wait_for_pulp 20
 pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional || show_logs_and_return_non_zero
 pytest -v -r sx --color=yes --pyargs pulp_file.tests.functional || show_logs_and_return_non_zero
-codecov
 
-# Travis' scripts use unbound variables. This is problematic, because the
```
mdellweg: I think, this is only a problem because we used `source` to call the scripts. Is there any (still valid) reason to do so?

mdellweg:
@bmbouter I think it is worth looking now. The rq jobs still nee coverage (whence WIP).

mdellweg:
@bmbouter now ;)

.travis/script.sh
bmbouter: I don't know of a reason.

pulpcore/app/management/commands/runserver-coverage.py
```diff
@@ -0,0 +1,17 @@
+import atexit
```
bmbouter: Putting this here will cause it to also be on production systems which doesn't seem right to me. That being said all test environments would benefit from it so reusability is good in those environments.

Does moving this into a Travis script make sense? Is my concern a real one?

.travis/script.sh
```diff
@@ -29,16 +46,24 @@ show_logs_and_return_non_zero() {
...
-# changes made to this script's environment appear to persist when Travis'
-# scripts execute. Perhaps this script is sourced by Travis? Regardless of why,
-# we need to reset the environment when this script finishes.
-#
-# We can't use `trap cleanup_function EXIT` or similar, because this script is
-# apparently sourced, and such a trap won't execute until the (buggy!) calling
-# script finishes.
-set +euo pipefail
+# Stop services to write coverage
+kill -SIGINT %?runserver
```
bmbouter: Nice. :+1:

.travis/script.sh
```diff
@@ -1,21 +1,38 @@
...
+      return
+    fi
+  done
+  echo
+  return 1
+}
 
 if [ "$TEST" = 'docs' ]; then
+  pulp-manager runserver >> ~/django_runserver.log 2>&1 &
+  sleep 5
```
bmbouter: what about using wait_for_pulp here too? We've had issues in Travis runs with this exact issue but we never had `wait_for_pulp`!

pulpcore/tests/functional/content_with_coverage.py
```diff
@@ -0,0 +1,17 @@
+import atexit
```
bmbouter: Kind of like what is here. If we don't ship tests this won't ship either, which seems right.

bmbouter:
I agree the RQ coverage isn't working. I always check the _queue_reserved_tasks which runs with every task run by Pulp's tasking system in the resource_manager [here](https://github.com/pulp/pulp/blob/202bfddbd3bc0c51e7afcb5b70a0c22d13d819b9/pulpcore/tasking/tasks.py#L76).

I am very interested in getting RQ working. I did some work on it mid 2018 and I still have a [WIP PR](https://github.com/pulp/pulp/pull/3496/files) that is not as good as yours so I'm going to close it.

In my testing I was getting RQ to generate the report locally I couldn't get it to show coverage from any RQ tasks locally. Let me know what you learn because I'm very interested in this working. We maybe should consider merging this as is and doing that separately since this does bring the coverage up 17% already. What do you think?

.travis/script.sh
mdellweg: Yes, there is room for improvement: In it's actual state `wait_for_pulp` checks the availability of the worker threads (and could also check the content app if it were present in the `status` endpoint). So, it is not usable here unmodified.

pulpcore/app/management/commands/runserver-coverage.py
mdellweg: Using `coverage run pulp-manager runserver` did not work, i think. That is why i searched for a way to insert it into the command. If we can manipulate the corresponding search path, we can place it elsewhere. I do not know how.

mdellweg:
Still, the rq worker does not report coverage. I think this is due to rq using `fork()` for every single Job. If we can run the queue with `SimpleWorker` instead of `Worker` it could work. But right now, i have now idea how.
@bmbouter I am ok with merging now. I can take back the rq parts if you like, as they do not work...

.travis/script.sh
bmbouter: OK we can leave this as is for now

mdellweg:
@bmbouter I'd prefer to not see this PR go to waste. But i see, we are both not entirely happy with it. I think i can split it to add the coverage for individual services. That way, it should also be easier, to see, which solution has an impact. (I would start with the content app, because i am rather confident with it.)

bmbouter:
@mdellweg I'm also hoping that we don't loose this work. Splitting it out like you are saying sounds great. I'm also ok to just have the workers coverage aspects removed if it isn't working. Let me know if you want me to look at it again for review. Thank you!

mdellweg:
@bmbouter Ping

bmbouter:
I'm reviewing now, ty!

bmbouter:
@mdellweg This looks exactly correct to me, ty. I'm lgtm-ing it. Are you able to also port something like this to github.com/pulp/pulpcore-plugin/ repo also? Or maybe file an issue for someone else to do it linking to this PR?

@pulp/qe this causes the runserver to also capture it's coverage output so I wanted someone to ack before merging.

mdellweg:
@bmbouter I took the runserver out, too, because you had concerns about the new command to `pulp-manager`. I'll make a new PR to discuss it once this is finished.

bmbouter:
@mdellweg yes I noticed you pulled that out. I appreciate that being discussed separately.

kersommoura:
Thanks for informing us about the changes.

bmbouter:
@mdellweg do you have any interest in porting these changes to this other repo https://github.com/pulp/pulpcore-plugin/blob/master/.travis/before_script.sh or should I file a refactor issue for someone else to do that? Thanks again!

bmbouter:
Actually I'm filing the issue anyway because it should be done.

bmbouter:
Here is the issue:   https://pulp.plan.io/issues/4416

mdellweg:
I guess, my next step is to try to extend this to `runserver`.
So a refactor ticket might be better (if you do not assign it to me...).

bmbouter:
Excellent! Thanks @mdellweg . We'll send this ticket through Sprint planning for anyone to pick up.

PR #3817, opened 2019-01-02
MySQL doesn't support text fields as indexes

daviddavis:
Changing fields from text to char since MySQL doesn't support text
fields as indexes. Also, recreating the migrations since they blow up
when you run them on MySQL.

fixes #4288
https://pulp.plan.io/issues/4288


dralley:
Should we write guidelines about this?  It's easy to forsee plugin writers going down the same path, resulting in plugins that only work on Postgres, even if they don't use any Postgres-specific features.

pulpcore/app/migrations/0001_initial.py
```diff
@@ -436,7 +436,7 @@ class Migration(migrations.Migration):
         ),
         migrations.AlterUniqueTogether(
             name='publishedmetadata',
-            unique_together={('publication', 'file'), ('publication', 'relative_path')},
+            unique_together={('publication', 'relative_path'), ('publication', 'file')},
```
goosemania: Interesting... it seems like the order is random and it's confusing because there were no changes to the uniqueness constraints :\

daviddavis:
Opened https://github.com/pulp/pulpcore-plugin/pull/26.

pulpcore/app/migrations/0001_initial.py
bmbouter: Can we just not have this line included in the diff?

bmbouter: I'm LGTM if we can remove this 1 line.

PR #3818, opened 2019-01-02
Improve Artifact generation and saving for content app

gmbnomis:
- Refactor Artifact generation for "on_demand" policy into method of its own.
  This allows derived plugin content handlers to customize the saving part.

- Don't fail if the Artifact already exists.  This may happen if multiple
  content instances share the same artifact or if the content type does not
  have enough metadata to ensure uniqueness without downloading the actual
  artifact (i.e. no digest in metadata).

fixes #4293
https://pulp.plan.io/issues/4293


gmbnomis:
@bmbouter Here is the PR I mentioned today on IRC

dralley:
Looks good to me, but I will let @bmbouter review also.

PR #3820, opened 2019-01-08
Fix Content.bulk_get_or_create() not protected by nested transaction.

jortel:
The `Manager.bulk_create()` and `Model.save()` need to be called within a nested transaction so that it can be safely used in an outer transaction.  Even though an `IntegrityError` is caught, it will still cause the transaction to be in an _ERROR_ state and can no longer be used.


jortel:
Replaced with #3821

PR #3822, opened 2019-01-08
Do not search for dupes by pk

asmacdo:
The ContentUnitSaver stage uses content.q() to retrieve the query for a
duplicate content unit. However, there are cases where the pk of the
content is set even when the save fails with an IntegrityError. This
means that content.q() will always search for content that does not
exist. Instead, we should always query by the natural key.

[noissue]

NOTE:
I'm not confident in the testing because because solves a problem that only happens in a WIP Docker PR. After making this change there were other problems in that PR (ContentArtifacts were not created during sync). Hopefully this PR will be helpful for https://github.com/pulp/pulp_docker/pull/294, and when finished @ipanova can confirm that this works.

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -58,8 +58,6 @@ class QueryMixin():
         """
         Returns a Q object that represents the model
         """
-        if self.pk:
-            return models.Q(pk=self.pk)
         try:
             kwargs = self.natural_key_dict()
         except AttributeError:


pulpcore/app/models/content.py
```diff
@@ -58,8 +58,6 @@ def q(self):
         """
         Returns a Q object that represents the model
         """
-        if self.pk:
```
dralley: I don't know what the performance ramifications of this are, perhaps it should "try" to search by PK first, and if it fails, do the natural key search?

asmacdo: I wondered about that too. That seems like a decent option to me, its cleaner than https://github.com/pulp/pulpcore-plugin/pull/29 and seems like the correct behavior in all circumstances.

dralley: @bmbouter What should we do about this PR?

bmbouter: This commit message describes the issue well. It's probably safe to do this. I'm not sure of the performance implications.

dralley: @bmbouter would this fix the issue? https://github.com/pulp/pulp/compare/master...dralley:uuid#diff-5308662d441ee7c5774295277cba139cL63

bmbouter:

pulpcore/app/models/content.py
bmbouter: @dralley This diff seems empty?

PR #3825, opened 2019-01-11
Add diagrams to Concepts and Terminology

pavelpicka:
Diagrams that illustrates all
of the concepts on the Concepts and Terminology page.

closes: #4100
https://pulp.plan.io/issues/4100

Signed-off-by: Pavel Picka <ppicka@redhat.com>

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -46,7 +46,7 @@ clean:
 
 diagrams:
 	mkdir -p $(DIAGRAM_BUILD_DIR)
-	plantuml diagrams_src/*.dot
+	python -m plantuml diagrams_src/*.dot
 	mv diagrams_src/*.png $(DIAGRAM_BUILD_DIR)/
 
 html:
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -15,13 +15,24 @@ deb) is defined by a plugin.  Files that belong to a content unit are called
 :term:`artifacts<artifact>`. Each content unit can have 0 or many artifacts and artifacts can be
 shared by multiple content units.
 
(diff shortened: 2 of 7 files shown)


bmbouter:
This is great because it will get the diagrams on the site which is important.

I had an idea I wanted to share from Pulp2. In pulp2's docs builder the source was checked into version control instead of the built png files. Then the docs builder would build the plantuml diagrams from the source with each docs build. This provides a few benefits:

1. It would keep the source in an area so it could be edited in the future
2. The images are always up to date after the source does change.

That code is here:  https://github.com/pulp/pulp/blob/2-master/docs/Makefile#L47-L50

I wanted to share the idea. It's not the smallest piece of work, so I'm really happy w/ this PR as is.

pavelpicka:
This is a good idea, going to rewrite it to have the sources with our code.

bmbouter:
@pavelpicka awesome! In the pulp2 code you can see sphinx including the images in their built location. Post here on the issue if you want me to find one specifically for you to see.

pavelpicka:
Please share, I went through Makefile in pulp2 (https://github.com/pulp/pulp/blob/2-master/docs/Makefile#L47) and that is why I used it same way here.

dralley:
@bmbouter @pavelpicka ping, is this PR currently complete?

bmbouter:
It looks right to me, but it needs tessting. It would be great if someone else could see if the diagrams make correctly as part of the docs build and the sphinx links render them correctly.

bmbouter:
Oh and I'll be out until Monday so +1 to continuing without me

PR #2507, opened 2016-04-01
Adds patch to python-kombu to fix pulp_celerybeat deadlock

bmbouter:
closes #1801
https://pulp.plan.io/issues/1801

PR #3826, opened 2019-01-11
Content app tries all RemoteArtifacts

bmbouter:
The Content app would raise an exception if there were multiple
RemoteArtifacts. It also would not try additional ones if the first ones
failed.

https://pulp.plan.io/issues/2507
closes #2507

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -8,6 +8,7 @@ from dynaconf.contrib import django_dynaconf  # noqa
 import django  # noqa otherwise E402: module level not at top of file
 django.setup()  # noqa otherwise E402: module level not at top of file
 
+from aiohttp.client_exceptions import ClientResponseError
 from aiohttp.web import FileResponse, StreamResponse
 from aiohttp.web_exceptions import HTTPForbidden, HTTPNotFound
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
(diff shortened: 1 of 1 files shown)


bmbouter:
The thing to know about this PR is that almost all of the diff is from an indention of existing code. Other than that, it just iterates through the list of Remotes instead of picking the first one. If the content can't be downloaded from an existing remote it raises a 502 exception. That means BadGateway which I think is accurate for the situation the streamer is in at that time.

bmbouter:
@jortel The version I just pushed includes the changes, can you review once more please?

pulpcore/content/handler.py
```diff
@@ -215,42 +216,58 @@ def _permit(request, distribution):
...
+
+            downloader = remote.get_downloader(remote_artifact=remote_artifact,
+                                               headers_ready_callback=handle_headers)
+            original_handle_data = downloader.handle_data
+            downloader.handle_data = handle_data
+            original_finalize = downloader.finalize
+            downloader.finalize = finalize
+            try:
+                download_result = await downloader.run()
+            except ClientResponseError:
```
gmbnomis: @bmbouter Sorry for the late (again) comment: Do you know what happens if the ClientResponseError occurs in an ongoing download (e.g. server closes connection unexpectedly)?

The code seems to look for the next remote artifact and put it into the same response (which may cause an exception for `response.prepare` or, even worse, just append the content to the broken content of the first artifact.

(I have not tested this, I may be wrong)

bmbouter: I finally got around to filing this https://pulp.plan.io/issues/4434

PR #3828, opened 2019-01-14
Adding names to tasks

pavelpicka:
User can see and filter the class path
and function name which run inside the task.

closes: #2889 
https://pulp.plan.io/issues/2889

Signed-off-by: Pavel Picka <ppicka@redhat.com>


pavelpicka:
As discussed I add to name full class_path

`"results": [                                                                   
  {                                                                            
    ...                                               
    "state": "waiting",                                                        
    "name": "pulp_file.app.tasks.synchronizing.synchronize",                   
    ...                                                
  }                                                                            
]`

but it is possible to have little shorter version as `pulp_file.synchronize` as function name should expose action just by its name. What do you think?

daviddavis:
I'm torn. `pulp_file.synchronize` seems shorter but what if I have two tasks with the same name in my plugin (e.g. `pulp_foo.bar.synchronize` and `pulp_foo.baz.synchronize`)?

pavelpicka:
+1 to keep full class path version, as keep space for plugin writers and extending plugins when will be needed.

PR #3829, opened 2019-01-14
Disallow user creation/update

goosemania:
One admin user will exist until the full support for user
management is added

closes #4313
https://pulp.plan.io/issues/4313

### DIFF
--- a/pulpcore/tests/functional/api/test_crud_users.py
+++ b/pulpcore/tests/functional/api/test_crud_users.py
@@ -12,6 +12,7 @@ from pulpcore.tests.functional.utils import set_up_module as setUpModule  # noqa
 from pulpcore.tests.functional.utils import skip_if
 
 
+@unittest.skip("User creation via API is not allowed until Pulp has proper user management")
 class UsersCRUDTestCase(unittest.TestCase):
     """CRUD users."""
 
@@ -135,6 +136,7 @@ class UsersCRUDTestCase(unittest.TestCase):
         assert response.json()['foo'] == ['Unexpected field']
 
 
+@unittest.skip("User creation via API is not allowed until Pulp has proper user management")
 class InvalidUserCreateTestCase(unittest.TestCase):
     """Invalid user test creation.
 
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -19,8 +19,6 @@ class UserFilter(BaseFilterSet):
 
 
 class UserViewSet(NamedModelViewSet,
-                  mixins.CreateModelMixin,
-                  mixins.UpdateModelMixin,
                   mixins.RetrieveModelMixin,
                   mixins.ListModelMixin,
                   mixins.DestroyModelMixin):


goosemania:
@pulp/qe

kersommoura:
Is there a related pulp.plan.io issue to when those tests will be enabled again?
I would suggest to add an extra item to this issue - enable tests, then we do not forget about those later on.

pulpcore/tests/functional/api/test_crud_users.py
```diff
@@ -25,6 +25,7 @@ def setUp(self):
         """Create an API client."""
         self.client = api.Client(self.cfg, api.json_handler)
 
+    @unittest.skip("User creation via API is not allowed until Pulp has proper user management")
```
kersommoura: We can skip the whole class, instead of individual methods. Then we will have just one change to adjust later. It is fine as it is, just a suggestion.

goosemania:
@kersommoura , no issue exists at the moment. I think we won't forget about/miss those tests, since we will run into them any way with the first attempt to write tests for user management ;)

pulpcore/tests/functional/api/test_crud_users.py
goosemania: @bmbouter , if we remove them and not just skip them, they will be definitely forgotten :(, in my opinion.

bherrin3: Agree. While I like version control, `out of sight, out of mind` is still a thing.

+1 to not deleting the tests and then accidentally re-writing them later.

The removal of create and update remove feature parity when compared to Pulp 2, correct?

If so, wouldn't feature parity be on the list for the RC? 

My apologies if I am off-base.

goosemania: @bherrin3 , there will be no Pulp 2 feature parity in Pulp 3.0 release, expect it in later Y releases and only for the features we decided not to drop.

bmbouter: I totally see what you are saying if we are bringing back user management. I was convinced from several discussion w/ devs that Pulp3 can't offer Users of any kind for probably a long time, possibly years. This PR isn't the best place to bring up that concern so I think I'll need to mail pulp-dev to see what the community thinks.

To share knoelwedge on the pulp2/pulp3 feature parity, the feature set of Pulp3 has many things missing from Pulp2. Some were conspicuously left behind, others just aren't a priority based on feedback from the community and various stakeholders. In many ways the features themselves are so different it's hard to even match them up. So for the RC the only remaining items are those on the ever-changing RC blockers list here:  https://pulp.plan.io/issues?query_id=121 This is my understanding at least, but others could see it differently.

bmbouter: For clarity, I'm lgtm on this as-is while we figure out the future of Users some other time. Thanks for putting this together!

daviddavis: I could see user management not coming back at all (at least not in its current form) so +1 to removing these tests from me.

PR #3830, opened 2019-01-14
Problem: static files are not available on user installs

dkliban:
Solution: collect static files before releasing to PyPI

This patch adds the 'whitenoise' middleware that serves static content for Pulp. 'whitenoise' is configured
to serve static content found in pulpcore/app/static directory. This means that after installing pulp, the user
will need to run the 'pulp-manager collectstatic' command to get all the static content into the directory.

closes #4180
https://pulp.plan.io/issues/4180

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -16,7 +16,7 @@ from pkg_resources import iter_entry_points
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
-
+STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/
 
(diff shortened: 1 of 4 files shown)


goosemania:
@dkliban , how about using it in development as well? http://whitenoise.evans.io/en/stable/django.html#using-whitenoise-in-development

pulpcore/app/settings.py
```diff
@@ -88,6 +88,7 @@
 
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
+    'whitenoise.middleware.WhiteNoiseMiddleware',
```
bmbouter: Is it supposed to go here versus after all the django contribute dones?

setup.py
```diff
@@ -16,7 +16,8 @@
     'PyYAML',
     'rq>=0.12.0',
     'setuptools',
-    'dynaconf>=1.0.4'
+    'dynaconf>=1.0.4',
+    'whitenoise'
```
bmbouter: awww yissss

bmbouter:
I think we need to fixup [these docs](https://docs.pulpproject.org/en/3.0/nightly/components.html#static-content) also since we don't really recommend these things anymore right?

PR #3833, opened 2019-01-15
Fix handling of PermissionError raised by ContentGuard.

jortel:
403 when `PermissionError` (or anything) is raised by the content guard.

Updated imports for consistency.

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -8,7 +8,8 @@ from dynaconf.contrib import django_dynaconf  # noqa
 import django  # noqa otherwise E402: module level not at top of file
 django.setup()  # noqa otherwise E402: module level not at top of file
 
-from aiohttp import web, web_exceptions
+from aiohttp.web import FileResponse, StreamResponse
+from aiohttp.web_exceptions import HTTPForbidden, HTTPNotFound
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.db import IntegrityError, transaction
 from pulpcore.app.models import Artifact, ContentArtifact, Distribution, Remote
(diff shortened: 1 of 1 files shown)


pulpcore/content/handler.py
```diff
@@ -135,11 +136,11 @@ def _permit(request, distribution):
                     'g': guard.name,
                     'r': str(pe)
                 })
-            raise
+            raise HTTPForbidden(reason=str(pe))
```
bmbouter: Is there any chance this will be used outside of the Content app?

pulpcore/content/handler.py
```diff
@@ -135,11 +136,11 @@ def _permit(request, distribution):
                     'g': guard.name,
                     'r': str(pe)
                 })
```
bmbouter: The docstring needs the exception type updated

pulpcore/content/handler.py
jortel: Really can't think of a use case.  Can you?

pulpcore/content/handler.py
jortel: Ah, good catch.

pulpcore/content/handler.py
bmbouter: I can't think of a reason, but if a plugin writer decided to offer content validation through the rest API for whatever reason, they would be using Django there and catching an aiohttp exception would be strange, but a PermissionError wouldn't. Transforming it in the content app handler, I think makes the ContentGuard feature a bit more portable.

bmbouter: Actually what if we wanted a feature that validated X509 certs as they are created that they are in turn valid w/ some uber cert. This would be a possible future feature that could make sense. That validation would happen in Django.

jortel: In both these cases, a plugin writer using the `Handler._permit()` separately for this _seems_  inappropriate (or at least not a good fit) because:
- It's a protected method (which could be changed)
- There is no guarantee that the associated content guard is X.509.
- I would think that anything leveraging the content guard would be using (or subclassing) the `Handler` anyway.

bmbouter: That's a good point, why is `_permit()` private? Is there a public version of it?

bmbouter: I was confusing which object permit() was on. After looking at it again I can see the implementation here is all correct.

PR #3837, opened 2019-01-21
Add assertions about `content_hrefs` to AddRemoveContentTestCase

nixocio:
Add assertions about `content_hrefs` to AddRemoveContentTestCase.

`[noissue]`


pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -117,6 +117,13 @@ def test_02_sync_content(self):
 
         self.assertIsNotNone(repo['_latest_version_href'])
 
+        content_hrefs = self.client.get(
+            repo_versions[0]['content_hrefs']['pulp_file.file']
```
dralley: Why not use ```get_content(repo_version)['pulp_file.file']```?

PR #3839, opened 2019-01-22
Support resource URLs in enqueue_with_reservation().

jortel:
https://pulp.plan.io/issues/4331

I think keeping the `resources` parameter as `list` and not also support it also being a `str` as suggested in the issue, helps reduce API complexity.  It also seemed better to translate the user supplied `resources` list instead of mutating it. 

The `resources` could simply be a `collections.Iterable` to support passing `tuple` or `generators` but leaving this way for now.

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -4,6 +4,7 @@ import uuid
 from gettext import gettext as _
 
 from django.db import IntegrityError
+from django.db.models import Model
 from rq import Queue
 from rq.job import get_current_job, Job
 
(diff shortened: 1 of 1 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -163,7 +164,8 @@ def enqueue_with_reservation(func, resources, args=None, kwargs=None, options=No
     Args:
         func (callable): The function to be run by RQ when the necessary locks are acquired.
         resources (list): A list of resources to reserve guaranteeing that only one task
-            reserves these resources
+            reserves these resources. Each resource can be either a (str) resource URL
+             or a (django.models.Model) resource instance.
         args (tuple): The positional arguments to pass on to the task.
         kwargs (dict): The keyword arguments to pass on to the task.
         options (dict): The options to be passed on to the task.
```
bmbouter: Can the ValueError be added to the `Raises` section in the docstring please?

jortel: Added.

PR #3840, opened 2019-01-23
Allow validation of _artifacts field if empty

mdellweg:
[noissue]


mdellweg:
Maybe this should validate to invalid if _artifacts is empty, but in the current state, `is_valid()` throws an Exception, which it should not.

PR #3846, opened 2019-01-24
Let core decide whether to use "_artifact" or "_artifacts"

dralley:
Remove boilerplate from the plugins. Plugins must define a
"multiple_artifacts_per_content" attribute on their Serializer subclass
Meta class, and core will look at that to decide which to use.

re: 4340
https://pulp.plan.io/issues/4340

### DIFF
--- a/pulpcore/tests/unit/content/test_handler.py
+++ b/pulpcore/tests/unit/content/test_handler.py
@@ -13,9 +13,9 @@ class HandlerSaveContentTestCase(TestCase):
 
     def setUp(self):
         self.f1 = FileContent.objects.create(relative_path='f1', digest='1')
-        self.f1.artifact = None
+        ContentArtifact.objects.create(artifact=None, content=self.f1, relative_path='f1')
         self.f2 = FileContent.objects.create(relative_path='f2', digest='1')
-        self.f2.artifact = None
+        ContentArtifact.objects.create(artifact=None, content=self.f2, relative_path='f2')
 
     def download_result_mock(self, path):
         dr = Mock()
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -14,7 +14,12 @@ from .base import (  # noqa
     AsyncOperationResponseSerializer
 )
 from .fields import BaseURLField, ContentRelatedField, LatestVersionField  # noqa
-from .content import ArtifactSerializer, ContentSerializer  # noqa
+from .content import (  # noqa
+    ArtifactSerializer,
+    NoArtifactContentSerializer,
+    SingleArtifactContentSerializer,
+    MultipleArtifactContentSerializer,
+)
 from .progress import ProgressReportSerializer  # noqa
 from .publication import (  # noqa
     ContentGuardSerializer,
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -11,9 +11,32 @@ from pulpcore.app.serializers import base, fields
 UNIQUE_ALGORITHMS = ['sha256', 'sha384', 'sha512']
 
 
-class ContentSerializer(base.MasterModelSerializer):
+class BaseContentSerializer(base.MasterModelSerializer):
     _href = base.DetailIdentityField()
 
+    class Meta:
+        model = models.Content
+        fields = base.MasterModelSerializer.Meta.fields
+
+
+class NoArtifactContentSerializer(BaseContentSerializer):
+
+    class Meta:
+        model = models.Content
+        fields = base.MasterModelSerializer.Meta.fields
+
+
+class SingleArtifactContentSerializer(BaseContentSerializer):
+    _artifact = fields.SingleContentArtifactField(
+        help_text=_("Artifact file representing the physical content"),
+    )
+
+    class Meta:
+        model = models.Content
+        fields = base.MasterModelSerializer.Meta.fields + ('_artifact',)
+
+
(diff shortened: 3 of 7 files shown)


gmbnomis:
@dralley I didn't think this through, but couldn't we have different base classes for the single Artifact case? (`SingleArtifactContent` model probably needs to be derived from `Content` as a proxy model). If so, plugin writers would simply decide which model/serializer to derive from.

Btw. we already had an issue for this in the past: https://pulp.plan.io/issues/3678

dralley:
@gmbnomis That sounds like a reasonable (probably better) approach.  @daviddavis What do you think?

daviddavis:
When I went to work on this problem, I was thinking of having two separate base classes as well. I lean more towards that solution I suppose.

dralley:
@daviddavis What worries me a little is that we do pass around Content objects directly in many places (we need to), so we still ultimately need to have a universal base class, and I think making the inheritance hierarchy more complex would be a mistake.

Would you be OK with having a single Content base model which contains all the functionality, retain the warnings about when it can be used, and then have two separate serializers?  Or do you think that would be too messy?

daviddavis:
I think the latter option seems simpler but the former option seems more technically correct. I would maybe ask @jortel to weigh in since he had some thoughts on this when I worked on https://pulp.plan.io/issues/3678 before.

jortel:
At the risk of blowing things up .. I've never understood the value of `_artifact` or `SingleContentArtifact` at all.  Dealing with `_artifacts` (plural) seems trivial and straight forward.  What problem does it solve? What am I missing?

dralley:
I actually completely agree on that front, but was under the impression that at least a couple of others disagreed.

My preferred solution would be to do exactly that, and then perhaps have a "SingleArtifactContentMixin" that provides the helpful getter/setter properties so that we don't have boilerplate in 95% of our plugins.

jortel:
> My preferred solution would be to do exactly that, and then perhaps have a "SingleArtifactContentMixin" that provides the helpful getter/setter properties so that we don't have an extra 12 lines of boilerplate in 95% of our plugins.

Can you paste an example of the 12 lines?  And which class it's in?

dralley:
https://github.com/pulp/pulp_file/blob/master/pulp_file/app/models.py#L28-L44

https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models.py#L180-L196

https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/models.py#L38-L60

We have lines like this in every single plugin currently, and the only difference between them is related to relative path.  So we could do this: 

```
class SingleArtifactContentMixin:

    def _get_relative_path(self):
        raise NotImplementedError("This method must be defined by the subclass")

    @property
    def artifact(self):
        """
        Return the artifact id (there is only one for this content type).
        """
        try:
            return self._artifacts.get().pk
        except Artifact.DoesNotExist:
            return None
        except Artifact.MultipleObjectsReturned:
            raise ValueError("Multiple artifacts exist for Content, yet SingleArtifactContentMixin should not be used if this is possible.")

    @artifact.setter
    def artifact(self, artifact):
        """
        Set the artifact for this Ansible Role version.
        """
        if self.pk:
            ca = ContentArtifact(
                artifact=artifact,
                content=self,
                relative_path=self._get_relative_path()
            )
        ca.save()
```

```
class AnsibleRoleVersion(Content, SingleArtifactContentMixin):

    def _get_relative_path(self):
        return "{namespace}/{name}/{version}.tar.gz".format(
            namespace=self.role.namespace,
            name=self.role.name,
            version=self.version
        )


class RpmPackage(Content, SingleArtifactContentMixin):

    def _get_relative_path(self):
        return self.filename
```

jortel:
Few initial comments.

1. IMHO, I really don't think _mixins_ are good practice unless no other pattern is (at all) practical.  Multiple inheritance is a well known anti-pattern.
2. The getter is misnamed and not symetrical to the setter because it returns a `pk` and not an `Artifact` object.
3. Is the getter/setter really needed if the serializer just deals with plural artifacts?
4. Can `_artifacts.first()` be leveraged to reduce the boilerplate code?  I don't know if the related manager supports it but I suspect it does.

dralley:
We *can* probably get rid of the getter and setter entirely, and just inline that code in the once place (each) that they're needed.

pulpcore/tests/unit/content/test_handler.py
```diff
@@ -13,9 +13,7 @@ class HandlerSaveContentTestCase(TestCase):
 
     def setUp(self):
         self.f1 = FileContent.objects.create(relative_path='f1', digest='1')
-        self.f1.artifact = None
```
gmbnomis: This creates and saves the corresponding ContentArtifact instance. I think this is needed.

pulpcore/tests/unit/content/test_handler.py
```diff
@@ -13,9 +13,7 @@ class HandlerSaveContentTestCase(TestCase):
...
         self.f2 = FileContent.objects.create(relative_path='f2', digest='1')
-        self.f2.artifact = None
```
gmbnomis: same here

pulpcore/tests/unit/content/test_handler.py
gmbnomis: Ah, the setter is gone now, then it should become:
```
ContentArtifact.objects.create(artifact=None, content=self.f1, relative_path='f1')
```

or so

daviddavis:
I've reviewed all the PRs and they LGTM. I'll let @jortel or @bmbouter have final approval.

PR #3853, opened 2019-02-01
Defer the setup instructions for pulplift

mdellweg:
In order to avoid duplicated (and partially outdated) documentation, the
details of how to utilize pulplift should be written down at pulplift
and only referenced elsewhere.

[noissue]


mdellweg:
@asmacdo I think, it was your suggestion...

mdellweg:
Thanks!

PR #3856, opened 2019-02-04
Add support for Amazon S3

daviddavis:
fixes #3900
https://pulp.plan.io/issues/3900

### DIFF
--- a/docs/installation/configuration.rst
+++ b/docs/installation/configuration.rst
@@ -42,6 +42,24 @@ DATABASES
    `Django documentation on databases <https://docs.djangoproject.com/en/2
    .1/ref/settings/#databases>`_
 
+DEFAULT_FILE_STORAGE
+^^^^^^^^^^^^^^^^^^^^
+
+   By default, Pulp uses the local filesystem to store files. The default option which
+   uses the local filesystem is ``pulpcore.app.models.storage.FileSystem``.
+
+   This can be configured though to alternatively use `Amazon S3 <https://aws.amazon.com/s3/>`_. To
+   use S3, set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``. For more
+   information about different Pulp storage options, see the `storage documentation <storage>`_.
+
+MEDIA_ROOT
+^^^^^^^^^^
+
+   The location where Pulp will store files. By default this is `/var/lib/pulp/`.
+
+   If you're using S3, point this to the path in your bucket you want to save files. See the
+   `storage documentation <storage>`_ for more info.
+
 LOGGING
 ^^^^^^^
 
--- a/docs/installation/storage.rst
+++ b/docs/installation/storage.rst
@@ -1,0 +1,46 @@
+Storage
+=======
+
+.. _storage:
+
+-----------
+
+  Pulp uses `django-storages <https://django-storages.readthedocs.io/>`_ to support multiple storage
+  backends. If no backend is configured, Pulp will by default use the local filesystem. If you want
+  to use another storage backend such as Amazon Simple Storage Service (S3), you'll need to
+  configure Pulp.
+
+Amazon S3
+^^^^^^^^^
+
+Setting up S3
+-------------
+
+  In order to use Amazon S3, you'll need to set up an AWS account. Then you'll need to create a
+  bucket for Pulp to use. Then you'll need to go into Identity and Access Management (IAM) in AWS to
+  create a user that Pulp will use to access your S3 bucket. Save the access key id and secret
+  access key.
+
+Configuring Pulp
+----------------
+
+  To have Pulp use S3, you'll need to install the optional django-stroages Python package::
+
+      pip install django-storages
+
+  Next you'll need to set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``
+  in your Pulp settings. At a minimum, you'll also need to set ``AWS_ACCESS_KEY_ID``,
+  ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see
+  the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
+
+  You will also want to set the ``MEDIA_ROOT`` configuration option. This will be the path in your
+  bucket that Pulp will use. An empty string is acceptable as well if you want Pulp to create its
+  folders in the top level of the bucket.
+
+  Here is an example configuration that will use a bucket called ``pulp3``::
+
+        AWS_ACCESS_KEY_ID = 'AKIAIT2Z5TDYPX3ARJBA'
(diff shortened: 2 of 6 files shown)


pulpcore/app/models/fields.py
```diff
@@ -35,7 +35,11 @@ def pre_save(self, model_instance, add):
                                'Artifact storage. Files must be stored outside this location '
                                'prior to Artifact creation.'))
         file = super().pre_save(model_instance, add)
-        if file and file._committed and add:
+        if file and file._committed and add and str(file) != upload_to:
+            # the file needs to be moved into place
```
daviddavis: This change is a bit questionable in my mind. The problem is that the following lines of code was being executed for uploads that were already in place (via upload_to). The read failed in the case where we're using S3 because of course the file is already stored in S3. So I'm adding a check to see if the file path is equal to upload_to.

Not sure if there's a better way to do this?

docs/installation/configuration.rst
```diff
@@ -42,6 +42,25 @@ DATABASES
    `Django documentation on databases <https://docs.djangoproject.com/en/2
    .1/ref/settings/#databases>`_
 
+DEFAULT_FILE_STORAGE
+^^^^^^^^^^^^^^^^^^^^
+
+   By default, Pulp uses the local filesystem to store files. This can be configured though to
```
bmbouter: This would be clearer if the first paragraph was all about the defaults and the second paragraph was all about S3. That way more backends could be added as their own paragraphs later too. What do you think?

pulpcore/app/models/fields.py
```diff
@@ -35,7 +35,11 @@ def pre_save(self, model_instance, add):
...
             file._file = TemporaryDownloadedFile(open(file.name, 'rb'))
             file._committed = False
```
bmbouter: It seems like this strangeness was already here, but we're calling super() twice with the same arguments. Should we reorder these statements so we are calling it only once?

docs/installation/configuration.rst
```diff
@@ -42,6 +42,25 @@ DATABASES
...
+   Note that if you use S3 you should at a minimum also set ``AWS_ACCESS_KEY_ID``,
+   ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuratino options,
+   see the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
+
+MEDIA_ROOT
+^^^^^^^^^^
+
+   The location where Pulp will store files. By default this is `/var/lib/pulp/`.
+
+   If you're using S3, this will be the path within the bucket you're using.
```
bmbouter: Does the user need to set this or is it set for them? If they need to take action that could be clearer.

docs/installation/configuration.rst
```diff
@@ -42,6 +42,25 @@ DATABASES
    `Django documentation on databases <https://docs.djangoproject.com/en/2
    .1/ref/settings/#databases>`_
 
+DEFAULT_FILE_STORAGE
+^^^^^^^^^^^^^^^^^^^^
+
+   By default, Pulp uses the local filesystem to store files. This can be configured though to
+   alternatively use `Amazon S3 <https://aws.amazon.com/s3/>`_. To use S3, set
+   ``DEFAULT_FILE_STORAGE`` to ``pulpcore.app.models.storage.S3Storage``. The default option which
```
bmbouter: What about adding a sentence that identifies that the user needs to `pip install`?

pulpcore/app/models/storage.py
```diff
@@ -164,3 +164,14 @@ def get_tls_path(model, name):
         type(model).__name__,
         str(uuid4()),
         name)
+
+
+try:
+    # django-storages/boto3 are optional packages
+    from storages.backends.s3boto3 import S3Boto3Storage
+
+    class S3Storage(S3Boto3Storage):
```
bmbouter: Why don't we have users use `storages.backends.s3boto3.S3Boto3Storage` directly?

pulpcore/content/handler.py
```diff
@@ -286,3 +287,23 @@ def _save_content_artifact(self, download_result, content_artifact):
...
+        Handle response for file.
+
+        Depending on where the file storage (e.g. filesystem, S3, etc) this could be responding with
+        the file (filesystem) or a redirect (S3).
+
+        Args:
+            file (:class:`django.db.models.fields.files.FieldFile`: File to respond with
+
+        Raises:
+            :class:`aiohttp.web_exceptions.HTTPFound`: When we need to redirect to the file
```
bmbouter: This also needs a `NotImplementedError` in its Raises section.

pulpcore/content/handler.py
```diff
@@ -286,3 +287,23 @@ def _save_content_artifact(self, download_result, content_artifact):
...
+        """
```
bmbouter: I believe it also needs a Returns section with `FileResponse` named in it.

pulpcore/app/models/fields.py
daviddavis: I tried looking at how to reorder this to only call super() once but I am not sure. This code is confusing and hard to follow. Would it make sense to open a separate refactor issue in redmine to try to rewrite this logic?

pulpcore/app/models/storage.py
daviddavis: We could but then we couldn't introduce custom logic or settings later on without having users change their configuration. If we're sure that we're not going to need to do that later on, then I am fine with making that change.

bmbouter: I don't know if we would do this in the future, but when we have a reason to we can introduce it's usage as a configurable setting with that version. I think removing it would be good.

pulpcore/app/models/fields.py
bmbouter: I'm ok with an issue filed now to refactor this later.

docs/installation/storage.rst
```diff
@@ -0,0 +1,45 @@
+Storage
```
bmbouter: These new docs are legit!

docs/installation/storage.rst
```diff
@@ -0,0 +1,45 @@
...
+=======
+
+.. _storage:
+
+-----------
+
+  Pulp uses `django-storages <https://django-storages.readthedocs.io/>`_ to support multiple storage backends. If no backend is configured,
```
bmbouter: Can this be line wrapped at 100 chars?

pulpcore/app/models/fields.py
daviddavis: Ok, here we go: https://pulp.plan.io/issues/4382

bmbouter: Thanks!

pulpcore/content/handler.py
```diff
@@ -286,3 +287,27 @@ def _save_content_artifact(self, download_result, content_artifact):
...
+
+    def _handle_file_response(self, file):
+        """
+        Handle response for file.
+
+        Depending on where the file storage (e.g. filesystem, S3, etc) this could be responding with
+        the file (filesystem) or a redirect (S3).
+
+        Args:
+            file (:class:`django.db.models.fields.files.FieldFile`: File to respond with
```
bmbouter: I missing a paren at the end here.

PR #3858, opened 2019-02-06
Save ContentArtifacts in ContentSerializer

mdellweg:
Required PR: pulp/pulpcore-plugin#65
Required RP: pulp/pulp_file#166

re #4394
https://pulp.plan.io/issues/4394

### DIFF
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -1,6 +1,7 @@
 from gettext import gettext as _
 import hashlib
 
+from django.db import transaction
 from rest_framework import serializers
 from rest_framework.validators import UniqueValidator
 
@@ -23,7 +24,7 @@ class NoArtifactContentSerializer(BaseContentSerializer):
 
     class Meta:
         model = models.Content
-        fields = base.MasterModelSerializer.Meta.fields
+        fields = BaseContentSerializer.Meta.fields
 
 
 class SingleArtifactContentSerializer(BaseContentSerializer):
@@ -31,9 +32,33 @@ class SingleArtifactContentSerializer(BaseContentSerializer):
         help_text=_("Artifact file representing the physical content"),
     )
 
+    _relative_path = serializers.CharField(
+        help_text=_("Path where the artifact is located relative to distributions base_path"),
+        validators=[fields.relative_path_validator],
+        write_only=True,
+    )
+
+    @transaction.atomic
+    def create(self, validated_data):
+        """
+        Create the content and associate it with its Artifact.
+
+        Args:
+            validated_data (dict): Data to save to the database
+        """
+        artifact = validated_data.pop('_artifact')
+        relative_path = validated_data.pop('_relative_path')
+        content = self.Meta.model.objects.create(**validated_data)
+        models.ContentArtifact.objects.create(
+            artifact=artifact,
+            content=content,
+            relative_path=relative_path,
+        )
+        return content
+
     class Meta:
         model = models.Content
-        fields = base.MasterModelSerializer.Meta.fields + ('_artifact',)
+        fields = BaseContentSerializer.Meta.fields + ('_artifact', '_relative_path')
 
 
 class MultipleArtifactContentSerializer(BaseContentSerializer):
@@ -43,9 +68,27 @@ class MultipleArtifactContentSerializer(BaseContentSerializer):
                     "'/artifacts/1/'"),
     )
 
+    @transaction.atomic
+    def create(self, validated_data):
+        """
+        Create the content and associate it with all its Artifacts.
+
+        Args:
+            validated_data (dict): Data to save to the database
+        """
+        _artifacts = validated_data.pop('_artifacts')
+        content = self.Meta.model.objects.create(**validated_data)
+        for relative_path, artifact in _artifacts.items():
+            models.ContentArtifact.objects.create(
+                artifact=artifact,
+                content=content,
+                relative_path=relative_path,
+            )
+        return content
+
     class Meta:
         model = models.Content
-        fields = base.MasterModelSerializer.Meta.fields + ('_artifacts',)
+        fields = BaseContentSerializer.Meta.fields + ('_artifacts',)
 
 
 class ArtifactSerializer(base.ModelSerializer):
--- a/pulpcore/app/serializers/fields.py
+++ b/pulpcore/app/serializers/fields.py
@@ -10,6 +10,12 @@ from pulpcore.app import models
 from pulpcore.app.serializers import DetailRelatedField, RelatedField
 
 
+def relative_path_validator(relative_path):
+    if os.path.isabs(relative_path):
+        raise serializers.ValidationError(_("Relative path can't start with '/'. "
+                                            "{0}").format(relative_path))
+
+
 class ContentRelatedField(DetailRelatedField):
     """
     Serializer Field for use when relating to Content Detail Models
@@ -76,9 +82,7 @@ class ContentArtifactsField(serializers.DictField):
         """
         ret = {}
         for relative_path, url in data.items():
-            if os.path.isabs(relative_path):
-                raise serializers.ValidationError(_("Relative path can't start with '/'. "
-                                                    "{0}").format(relative_path))
+            relative_path_validator(relative_path)
             artifactfield = RelatedField(view_name='artifacts-detail',
                                          queryset=models.Artifact.objects.all(),
                                          source='*', initial=url)
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -13,7 +13,13 @@ from .base import (  # noqa
     validate_unknown_fields,
     AsyncOperationResponseSerializer
 )
-from .fields import BaseURLField, ContentRelatedField, LatestVersionField  # noqa
+from .fields import (  # noqa
+    BaseURLField,
+    ContentRelatedField,
+    LatestVersionField,
+    SingleContentArtifactField,
+    relative_path_validator,
+)
 from .content import (  # noqa
     ArtifactSerializer,
     NoArtifactContentSerializer,


pulpcore/app/serializers/content.py
```diff
@@ -18,12 +18,23 @@ class Meta:
         model = models.Content
         fields = base.MasterModelSerializer.Meta.fields
 
+    def create(self, validated_data):
+        _artifacts = validated_data.pop('_artifacts')
```
daviddavis: I think this method needs to check for three cases: the existence of _artifacts, _artifact, or no _artifact/_artifacts.

daviddavis:
Would you be willing to file an issue in the tracker and attach this work to it?

pulpcore/app/serializers/content.py
daviddavis:
```suggestion
        if '_artifacts' in validated_data:
            _artifacts = validated_data.pop('_artifacts')
        elif '_artifact' in validated_data:
            _artifacts = [validated_data.pop('_artifact')]
        else:
            _artifacts = []
```

daviddavis:
Actually I think the challenge here will be to set `relative_path` on the content artifact. Each plugin does it differently. In pulp_file, it's passed in. In pulp_python, it's a `filename` param. And in pulp_ansible, it's generated based on the AnsibleRole name and namespace values.

pulpcore/app/serializers/content.py
mdellweg: Yeah, figured that out the hard way. I WIPed the PR accordingly. ;)

mdellweg: Actually (i was discussing this with @asmacdo), we thought, the serializers validate should leave all validated_data in a common state. Another option is to move this create to the more specialized serializers.

mdellweg:
> Actually I think the challenge here will be to set `relative_path` on the content artifact. Each plugin does it differently. In pulp_file, it's passed in. In pulp_python, it's a `filename` param. And in pulp_ansible, it's generated based on the AnsibleRole name and namespace values.

In the actual form, i try to read `relative_path` from the `Content` model, and use `None` if it is not there. I think in the other plugins, that might be gapped by providing a `relative_path` property.

daviddavis:
So you could check if content has a relative_path and then set it on the ContentArtifact if it does? That's fine with me but I think others were opposed to adding this logic to the data model. 

CCing @dralley @jortel @bmbouter for visibility.

mdellweg:
> So you could check if content has a relative_path and then set it on the ContentArtifact if it does?

That is the solution, as it is now.
The other option would be to add a `_relative_path` field to the serializer, that is not supposed to be added to the model, but only used to create the `ContentArtifact`.

daviddavis:
It looks like the code uses validated_data though and not the content object. I am not sure how we'd set `relative_path` in plugins that generate this field. Here's [pulp_ansible as an example](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/viewsets.py#L80-L83).

What about a method on the serializer that will return `validated_data.get('relative_path', None)` or can be overridden by plugins?

mdellweg:
Would definitively work. I would not default it to relative_path then, but force all plugins to implement it for consistency. Is this a nice API then?

daviddavis:
👍 from me. I'll ping @jortel and @dralley to weigh in too.

gmbnomis:
> Would definitively work. I would not default it to relative_path then, but force all plugins to implement it for consistency. Is this a nice API then?

There is one additional challenge: In pulp_cookbook, there are two uses for `relative_path()` defined in the model:

1. in the `create()` method of the content serializer
2. in the first stage of the sync pipeline when creating DeclarativeArtifact instances

I reckon that this is a common pattern.

If we put the knowledge how to construct `relative_path` into the serializer, how do we access it in the first stage? (We can't use `create()` because we must not save the content at this point. Creating a Serializer and a 'validated_data' dict just to get the relative_path looks much more cumbersome than asking the in memory content model about its relative path.)

daviddavis:
> If we put the knowledge how to construct `relative_path` into the serializer, how do we access it in the first stage?

I've made the case before that this logic belongs in the model but @jortel disagreed.

mdellweg:
From time to time, I do think that we should also use the serializers in the pipeline for consistency.

gmbnomis:
> > If we put the knowledge how to construct `relative_path` into the serializer, how do we access it in the first stage?
> 
> I've made the case before that this logic belongs in the model but @jortel disagreed.

And there are good reasons not to put in the model depending how you look at it. Maybe I am opening a can of worms here, but our handling of relative paths in `ContentArtifact` is asymmetric:

-  In the multi-artifact case, relative path is a "first-class citizen". One needs to specify it for every artifact when creating a content model instance using the serializer. Conversely, it is part of the output when serializing.

- For single artifact content, we keep the relative path under the rug: only the href to the artifact is part of the output. We seem to assume that the relative path is of no interest when knowing the model, i.e. the relative path is essentially redundant information (in the sense that it can be derived easily from other information).

  The most blatant example of this is pulp_file, where relative_path is actually duplicated in the model. (another aspect is that it is part of the natural key for content)

For a single artifact content type, if path/filename information is important to users, the plugin writer is forced to include this information in the model. And then, it should be possible to derive `relative_path` from the information stored in the model.

If, on the other hand, we would treat `relative_path` like in the multi-artifact case, the model would have no idea how to generate the `relative_path` in the general case.

A `relative_path` method/property in the content model makes it very clear that the information must already be present in the model. Alternatively, we can handle `relative_path` as a distinct piece of information. But then, it should also be included in the representation when serializing.

bmbouter:
@gmbnomis I think you've characterized the situation well. I believe one of those two changes are necessary.

Generally adding things to the content model sounds right, but if we add a `relative_path` method/property in the content model, what would that do for 'multi-artifact' content models?

Also @dkliban FYI ^, maybe this is related to the relative_path issues we were talking about last week w.r.t your pass-through cache story.

dkliban:
@bmbouter Yes it is related. 

I am currently working on adding the pass-through cache functionality[0]. I've found that I need to add an interface to pulpcore.plugin.models.content. Content that can instantiate a piece of Content using an Artifact and a relative path. The content app then uses this interface to instantiate Content after it streams the Artifact to the client. This can only work for single artifact content. 

I am open to more ideas on how the instantiation of Content can be implemented. 

[0] https://pulp.plan.io/issues/3894

daviddavis:
> Generally adding things to the content model sounds right, but if we add a `relative_path` method/property in the content model, what would that do for 'multi-artifact' content models?

I'm imagining that the relative_path method either wouldn't be implemented in `Content` or (if it is) it would return `NotImplementedError`.

dralley:
Another consideration: I've found that the upload case is made more difficult by the fact that many of the tools that we use to parse the content metadata from package files (with Python, and RPM, for example), also validate that the name of the file matches up with the metadata it contains.  Which means that you upload the artifact

dralley:
whoops, wrong button

mdellweg:
I would prefer, to have the _serializers_ more symmetric:
In the single artifact case having two fields: `_relative_path` and `_artifact`
In the multiple artifact case having a dict `_artifacts` containing `relative_path` -> `artifact` pairs.

Then the plugin must decide whether to fill the `_relative_path` field with a corresponding field or a method of the content model.

Is it vaild, that the `relative_path` in the `ContentArtifact` can be `None`?

daviddavis:
@mdellweg thinking it over, I think that makes the most sense.

According to the code I think `relative_path` cannot but null. Also, it must be unique. I wonder if that makes sense.

jortel:
Reading up on this, I'm unclear about the problem being solved.  The `create()` method on the serializers (in this PR) seems to address the problem stated on the issue.  What continues to drive adding `_artifact` and `_relative_path` fields/methods/properties on the `Content` models?

jortel:
> @mdellweg thinking it over, I think that makes the most sense.
> 
> According to the code I think `relative_path` cannot but null. Also, it must be unique. I wonder if that makes sense.

I can't think of a case where the relative path would be NULL.  Can you give an example?

mdellweg:
I thought, `relative_path` was only used for `pass_through` publishing. If the plugin publishes "by hand" it could be NULL.

jortel:
> I thought, `relative_path` was only used for `pass_through` publishing. If the plugin publishes "by hand" it could be NULL.

The `ContentArtifact.relative_path` was initially/primarily designed to be used by `Publisher`.  As it turned out, it also supported pass-thru nicely.

For content with a single artifact, the publisher _could_ derive a path.  This would only be needed for cases whereby the path can only be calculated by the Publisher and not when the content is created.  Is there a use case for this?

bmbouter:
Perhaps this is unrelated, but believe our model layer is stronger and more usable when we have `NoArtifactContent`, `SingleArtifactContent`, and `MultiArtifactContent` model types that plugin writers can use. These specific names aren't important to me, just as examples.

asmacdo:
I don't understand the need to change this on the model layer. AFAICT, an `_artifacts` field is sufficient to store 0, 1, or many artifacts. Given that `_relative_path` is not a Content field (well, except for pulp_file), it seems incorrect to put this behavior on the Content model. The serializer is the [idiomatic DRF place to write nested fields](https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects)-- so I think the current state is how it should be done. For plugins with specific needs, overriding the serializer's `create` and `update` is allowed and idiomatic, and `_relative_path` can be removed from user-facing fields entirely in the `class Meta` (for pulp_ansible).

Heres the examples to use this PR:
with single artifacts: https://github.com/pulp/pulp_file/pull/166
with multi artifacts: https://github.com/ATIX-AG/pulp_gem/pull/19

daviddavis:
I'm running into a bit of a problem applying this to pulp_ansible. I have two choices for the fields on the `AnsibleRoleVersionSerializer`. I can either do:

```python
fields = tuple(set(SingleArtifactContentSerializer.Meta.fields) - {'_relative_path'}) + ('version',)

fields = SingleArtifactContentSerializer.Meta.fields + ('version',)
```

The first option raises `{"_relative_path": ["Unexpected field"]}` while the second one shows a `_relative_path` param in the api docs for the ansible role version create endpoint.

I think @jortel mentioned that I need to override the validation method in the serializer to not check `relative_path`?

daviddavis:
I tried playing around with the validation and it didn't work. The problem is that the unexpected field logic lives in [the validate method of ModelSerializer](https://github.com/pulp/pulp/blob/cf30f2e9a77f7ad935184c5cbe58dfad788febc3/pulpcore/app/serializers/base.py#L73-L76) so defining `validate_relative_path()` didn't help.

I tried also to override `validate()` and munge initial_data but that seemed like it affected what got passed to create. In other words, `_relative_path` wasn't set in `validated_data`.

The solution I found that worked was to have this code in the viewset:

```python
          data = request.data.copy()
          role = AnsibleRole.objects.get(pk=role_pk)
          relative_path = "{namespace}/{name}/{version}.tar.gz".format(
              namespace=role.namespace,
              name=role.name,
              version=data['version']
          )   
          serializer = self.get_serializer(data=data)
          serializer.is_valid(raise_exception=True)
          role_version = serializer.save(role=role, _relative_path=relative_path)
```

I guess another option would be to override `validate` and save a backup copy of `initial_data`, remove relative_path, and then set `self.initial_data = initial_data_backup` before returning from validate. However, this option seems more esoteric for plugin writers.

mdellweg:
@daviddavis can you try `data[_relative_path] = relative_path` before instanciating the serializer?
Is it bad to have `_relative_path` show up in the docs? It needs to be provided to create the content anyway.

daviddavis:
> @daviddavis can you try `data[_relative_path] = relative_path` before instanciating the serializer?

I get `{"_relative_path": ["Unexpected field"]}`.

> Is it bad to have `_relative_path` show up in the docs? It needs to be provided to create the content anyway.

I'm hesitant to show the field in the api docs if it's not something users are supposed to set. And it shows up as being required as well. Maybe others feel differently though?

mdellweg:
@daviddavis I know @asmacdo did not like to do the messy stuff in the validation, but more and more i get the impression, that is the DRF way. I updated the pulp_file PR accordingly. And i can confirm that the result is to have it ask for `artifact` and `relative_path` in the web_gui from which it derives `digest` and `_relative_path`. Maybe, you want to copy the idea in the ansible plugin.

daviddavis:
I tested out this PR, the pulpcore-plugin one, and the pulp_file and it worked. The code looks good to me. I'm hoping to get 1-2 more devs to review as well before I merge.

pulpcore/app/serializers/content.py
```diff
@@ -23,17 +24,43 @@ class NoArtifactContentSerializer(BaseContentSerializer):
...
+        validators=[fields.relative_path_validator],
+        write_only=True,
+    )
+
+    @transaction.atomic
+    def create(self, validated_data):
+        """
+        Create the content and associate it with its Artifact.
+
+        Use `relative_path` from the `Content` in the `ContentArtifact` if available.
```
asmacdo: This line isn't true any more.

mdellweg: You are right.

pulpcore/app/serializers/fields.py
```diff
@@ -10,6 +10,12 @@
 from pulpcore.app.serializers import DetailRelatedField, RelatedField
 
 
+def relative_path_validator(relative_path):
+    if os.path.isabs(relative_path):
+        raise serializers.ValidationError(_("Relative path can't start with '/'. "
```
daviddavis: nitpick - space after `.`.

mdellweg: The string continues on the next line. I think, the space is intentional.

```
>>> a = ("asdf"
...     "ghjk")
>>> a
'asdfghjk'
```

asmacdo:
Just wanted to give an explicit +1. We discussed and think perhaps @dralley should be the other reviewer.

PR #3864, opened 2019-02-08
Update release notes to use tags instead of dates

daviddavis:
fixes #4401
https://pulp.plan.io/issues/4401


daviddavis:
Note: the first couple links (pre 3.0.0b4) use the old date range links because we didn't tag before pulpcore-3.0.0b3.

bmbouter:
This PR is the legit! Thank you.

PR #3865, opened 2019-02-11
Update installation instructions

goosemania:
Both PyPI and source ones.

closes #4356
https://pulp.plan.io/issues/4356

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -43,20 +43,19 @@ PyPI Installation
 
    $ sudo apt-get install python3-venv
 
-3. Install Pulp::
+3. Install Pulp (install ``pulpcore-plugin``, and ``pulpcore`` itself will be installed as
+a dependency)::
 
-   $ pip install pulpcore
+   $ pip install pulpcore-plugin
 
 .. note::
 
-   To install from source, clone the git repository, navigate to the directory containing the
-   ``setup.py`` file, and do a local, editable pip installation::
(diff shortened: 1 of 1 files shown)


docs/installation/instructions.rst
```diff
@@ -77,15 +76,13 @@ PyPI Installation
...
 
    $ pulp-manager collectstatic --noinput
 
-10. Run Pulp:
-::
-
-   $ django-admin runserver
+10. Run Pulp::
 
+    $ pulp-manager runserver
```
goosemania: Changed to `pulp-manager` for consistency here (since it's used in all the other instructions above).

docs/installation/instructions.rst
```diff
@@ -43,20 +43,19 @@ PyPI Installation
 
    $ sudo apt-get install python3-venv
 
-3. Install Pulp::
+3. Install Pulp (install ``pulpcore-plugin``, and ``pulpcore`` itself will be installed as
+a dependency)::
```
asmacdo: Maybe we should suggest that the user install the plugin they want, and pulpcore-plugin will be picked up as a dep. Is there any reason a user would install pulp without a plugin?

daviddavis: That makes sense to me. Something like:

```rst
3. Installing a plugin will automatically install ``pulpcore`` and ``pulpcore-plugin``::

    pip install pulp_file
```

PR #3867, opened 2019-02-12
Clarify password change docs

asmacdo:
[noissue]


server/etc/pulp/server.conf
```diff
@@ -70,7 +70,7 @@
 # default_login:    default admin username of the Pulp server; this user will be
 #                   the first time the server is started
 # default_password: default password for admin when it is first created; this
-#                   should be changed once the server is operational
+#                   should be changed using pulp-admin once the server is operational
```
mdellweg: Is it `create_user` or another command?
Might be helpful to be a tiny bit more specific.

dralley:
@asmacdo Could you rebase?

PR #3869, opened 2019-02-12
Remove threading from pagination test_content

kersommoura:
Make the `test_content` from pagination test case synchronous. Thus
avoiding occasional test failures. Test will be a bit slower, but
less error prone. Later on, if necessary an alternative can
be evaluated.

'[noissue]'


kersommoura:
@daviddavis, can you please merge this one?

PR #3870, opened 2019-02-12
Distribution base_url no longer default to request netloc.

jortel:
https://pulp.plan.io/issues/4391

### DIFF
--- a/pulpcore/app/serializers/fields.py
+++ b/pulpcore/app/serializers/fields.py
@@ -193,8 +193,12 @@ class BaseURLField(serializers.CharField):
     """
 
     def to_representation(self, value):
-        if settings.CONTENT_HOST:
-            host = settings.CONTENT_HOST
-        else:
-            host = self.context['request'].get_host()
-        return ''.join([host, settings.CONTENT_PATH_PREFIX, value])
+        base_path = value
+        host = settings.CONTENT_HOST or ''
+        prefix = settings.CONTENT_PATH_PREFIX or ''
+        return '/'.join(
+            (
+                host.strip('/'),
+                prefix.strip('/'),
+                base_path.lstrip('/')
+            ))


pulpcore/app/serializers/fields.py
```diff
@@ -193,8 +193,12 @@ class BaseURLField(serializers.CharField):
...
 
     def to_representation(self, value):
-        if settings.CONTENT_HOST:
-            host = settings.CONTENT_HOST
-        else:
-            host = self.context['request'].get_host()
-        return ''.join([host, settings.CONTENT_PATH_PREFIX, value])
+        base_path = value
+        host = settings.CONTENT_HOST or ''
+        prefix = settings.CONTENT_PATH_PREFIX or ''
```
bmbouter: This will always be set because the default it's defined in settings.py as '/pulp/content/'. I suggest removing the `or ''` part since it won't be used.

pulpcore/app/serializers/fields.py
```diff
@@ -193,8 +193,12 @@ class BaseURLField(serializers.CharField):
     """
 
     def to_representation(self, value):
-        if settings.CONTENT_HOST:
-            host = settings.CONTENT_HOST
-        else:
-            host = self.context['request'].get_host()
-        return ''.join([host, settings.CONTENT_PATH_PREFIX, value])
+        base_path = value
+        host = settings.CONTENT_HOST or ''
```
bmbouter: This one defaults to `None` in settings.py. That default was chosen to align with this code. What about having this PR remove the `or ''` and instead setting `CONTENT_HOST = ''` in settings.py?

jortel: There isn't anything preventing a user to setting it to `None` which we should deal with gracefully.

pulpcore/app/serializers/fields.py
jortel: There isn't anything preventing a user to setting it to `None` which we should deal with gracefully.

pulpcore/app/serializers/fields.py
bmbouter: I don't think handling this gracefully is beneficial to the user. In fact I think it masks situations which should be explicit for the user. Here are some of the reasons I think about.

If the user is applying an incorrect config we should not translate it into another value and we doubly shouldn't do so behind the scenes. To translate for them allows their errors to pass unnoticed and silently. Say they used a None value instead of a string, they should be presented this mistake immediately so they can fix it.

Also if we're concerned about users setting invalid values we should have some sort of validation for our settings. If we do that, we would also not want it here because it wouldn't be centralized. Doing it everywhere it is accessed creates a DRY problem which spreads out the default value.

jortel: I think `None` and `''` are equivalent in this case.  Forcing the user to use `''` instead of `None` does not seem valuable.  I think we should focus on the user's intent and not bog them down in syntax.

jortel: +1 for validating the settings but raising an error on `None` instead of `''` seems pedantic.

daviddavis: So the question is whether for `CONTENT_HOST`, the value `None` is equivalent to `''` or whether it should be considered erroneous?

bmbouter: My request is that we don't do interpretation or adjustment of these variables here. I hope this code can only deal with how these config values are brought together without adjusting their possible values along the way.

daviddavis: > My request is that we don't do interpretation or adjustment of these variables here.

That seems reasonable to me.

jortel: For the purpose of generating the URL here, I don't see how `None` and `''` could be anything but equivalent.  So, not interpreting or adjusting but rather just adds robustness.

daviddavis: I don't disagree. I just wonder if there could be a more central place to do it?

jortel: Perhaps.  Where?
Do we really want to develop and maintain a settings validation/conditioning mechanism  to just handle `settings.CONTENT_HOST or ''`?

bmbouter:

PR #3871, opened 2019-02-12
Modify response_handler for populate_pulp function

kersommoura:
Modify `response_handler` used by populate_pulp to be `page_handler`.
For Pulp 3 `page_handler` should be the prefered one.

See: https://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.api.html?highlight=page_handler#pulp_smash.api.page_handler

'[noissue]'


kersommoura:
I will re-evaluate this change.

PR #3873, opened 2019-02-13
Add coverage to runserver

mdellweg:


mdellweg:
@bmbouter i am quite confident, that this approach has not worked like this the last time i tried.

mdellweg:
ping @pulp/qe

bmbouter:
This looks like it's working to me! @pulp/qe can you lgtm+merge since I'll be out.

@dkliban FYI on this change as well.

kersommoura:
ack.

PR #3875, opened 2019-02-14
Mark pass_through write_only on Serializer

mdellweg:
This hides the pass_through option when serializing Publications.

re #4421
https://pulp.plan.io/issues/4421

### DIFF
--- a/pulpcore/app/serializers/publication.py
+++ b/pulpcore/app/serializers/publication.py
@@ -23,9 +23,6 @@ class PublicationSerializer(ModelSerializer):
     _href = IdentityField(
         view_name='publications-detail'
     )
-    pass_through = serializers.BooleanField(
-        help_text=_('The publication is a pass-through for the repository version.')
-    )
     publisher = DetailRelatedField(
         help_text=_('The publisher that created this publication.'),
         queryset=models.Publisher.objects.all()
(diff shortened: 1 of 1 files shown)


daviddavis:
When I look at the api docs, the `pass_through` field shows up for `Publication` definition schema. I think we want to remove it.

mdellweg:
Do you have a link? I think in the api docs for the model we want to have the flag, since the plugin needs to provide it.
But as for the Serializer (that is only used one way atm) we can delete it.

daviddavis:
@mdellweg I just look at the json api docs via httpie (`http --pretty=all :8000/pulp/api/v3/docs/api.json | less -R`).

mdellweg:
Then it might just go away...

mdellweg:
It is still there in the docs. Do i need to rebuild them manually?

daviddavis:
It should be dynamically generated.

daviddavis:
[This line](https://github.com/pulp/pulp/blob/889bad578559298080a9bda5f61805872b554a05/pulpcore/app/serializers/publication.py#L47) needs to be removed as well.

mdellweg:
Right. It shall be gone.

mdellweg:
@daviddavis Thank you for helping.

PR #3876, opened 2019-02-14
WIP: Collect coverage from rq worker

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/worker_with_coverage.py
+++ b/pulpcore/tests/functional/worker_with_coverage.py
@@ -1,0 +1,16 @@
+import sys
+import coverage
+
+from ...tasking.worker import PulpWorker  # noqa
+
+
+class CoveragePulpWorker(PulpWorker):
+    def perform_job(self, job, queue):
+        cov = coverage.Coverage()
+        cov.start()
+
+        super().perform_job(job, queue)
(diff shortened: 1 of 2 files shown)


pulpcore/tests/functional/worker_with_coverage.py
```diff
@@ -0,0 +1,16 @@
+import sys
```
dralley: @bmbouter What do you think about having this located in /tests/ rather than in tasking?  I think we could put it in tasking... we could possibly have different variants for test coverage, cprofiling enabled, and so forth.

pulpcore/tests/functional/worker_with_coverage.py
```diff
@@ -0,0 +1,16 @@
...
+import coverage
+
+from ...tasking.worker import PulpWorker  # noqa
```
dralley: Use an absolute import here

```from pulpcore.tasking.worker import PulpWorker```

pulpcore/tests/functional/worker_with_coverage.py
bmbouter: I think this makes sense. :+1:

bmbouter:

PR #3891, opened 2019-02-21
Issue 3044 Make Distribution CUD async

CodeHeeler:
Makes Distribution CUD operations async

Uses resource locking protection to create/update/destroy distributions
asynchronously which prevents race conditions in base_path checking.

fixes #3044
https://pulp.plan.io/issues/3044

### DIFF
--- a/pulpcore/app/tasks/__init__.py
+++ b/pulpcore/app/tasks/__init__.py
@@ -1,3 +1,3 @@
-from pulpcore.app.tasks import base, repository  # noqa
+from pulpcore.app.tasks import base, distribution, repository  # noqa
 
 from .orphan import orphan_cleanup  # noqa
--- a/pulpcore/app/tasks/distribution.py
+++ b/pulpcore/app/tasks/distribution.py
@@ -1,0 +1,44 @@
+from django.core.exceptions import ObjectDoesNotExist
+from pulpcore.app.models import Distribution
+from pulpcore.app.serializers import DistributionSerializer
+
+
+def create(*args, **kwargs):
+    """
+    Creates a :class:`~pulpcore.app.models.Distribution`
+    """
+    data = kwargs.pop('data', None)
+    serializer = DistributionSerializer(data=data)
+    serializer.is_valid(raise_exception=True)
+    serializer.save()
+
+
+def update(instance_id, *args, **kwargs):
+    """
+    Updates a :class:`~pulpcore.app.models.Distribution`
+
+    Args:
+        instance_id (int): The id of the distribution to be updated
+    """
+    data = kwargs.pop('data', None)
+    partial = kwargs.pop('partial', False)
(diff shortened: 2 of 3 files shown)


daviddavis:
This looks great 👍 

Question: why is distribution delete asynchronous?

pulpcore/app/tasks/__init__.py
```diff
@@ -1,3 +1,3 @@
-from pulpcore.app.tasks import base, repository  # noqa
+from pulpcore.app.tasks import base, repository, distribution  # noqa
```
asmacdo: nitpick: please make these imports alphabetical order

pulpcore/app/tasks/distribution.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def delete(instance_id, *args, **kwargs):
+    """
+    Delete a :class:`~pulpcore.app.models.Distribution`
+
+    Args:
+        instance_id (int): The id of the Distribution to be deleted
+    """
+    instance = Distribution.objects.get(pk=instance_id)
```
asmacdo: This could raise an error if the Distribution has already been deleted, which needs to be caught. 

https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.get

This situation would occur if 2 tasks are enqueued before the first task is executed. I think it is enough to catch the exception and pass.

pulpcore/app/viewsets/publication.py
```diff
@@ -70,12 +78,70 @@ class Meta:
 
 
 class DistributionViewSet(NamedModelViewSet,
-                          mixins.CreateModelMixin,
                           mixins.UpdateModelMixin,
                           mixins.RetrieveModelMixin,
                           mixins.ListModelMixin,
                           mixins.DestroyModelMixin):
+    """
+    Provides CUD methods that dispatch tasks with reservation that lock all distributions
```
asmacdo: It also provides retrieve and list.

asmacdo:
I made a first pass, but since I helped with this work, someone else should do a final review.

CodeHeeler:
@daviddavis originally, the ticket called for all of CRUD, and we already had mixins for async update and delete so I needed to write create, but when scrapping my initial version and starting cleanly on replacing these methods instead of using those mixins while pairing with asmacdo, I failed to account for the updated ticket and did all but READ.

Do you think this is unnecessary and needs to be removed?

pulpcore/app/viewsets/publication.py
CodeHeeler: Good catch, was getting too myopic in adding docstrings and there were none for this before now.

daviddavis:
Actually thinking through it some more. I think it’s fine to leave this as is. Otherwise, you have to handle the case where a user could delete a distribution that has a queued update.

bmbouter:

PR #3892, opened 2019-02-22
Refactor ArtifactFileField

goosemania:
closes #4382
https://pulp.plan.io/issues/4382

### DIFF
--- a/pulpcore/app/models/fields.py
+++ b/pulpcore/app/models/fields.py
@@ -9,7 +9,7 @@ from pulpcore.app.files import TemporaryDownloadedFile
 
 class ArtifactFileField(FileField):
     """
-    A custom FileField that always saves files to location specified by 'upload_to'
+    A custom FileField that always saves files to location specified by 'upload_to'.
 
     The field can be set as either a path to the file or File object. In both cases the file is
     moved or copied to the location specified by 'upload_to' field parameter.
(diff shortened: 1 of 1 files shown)


goosemania:
@daviddavis , if you can try it on S3 that would be great. Thank you!

pulpcore/app/models/fields.py
```diff
@@ -9,37 +9,49 @@
...
+        if file.name != artifact_storage_path and file.name.startswith(
                 os.path.join(settings.MEDIA_ROOT, 'artifact')):
             raise ValueError(_('The file referenced by the Artifact is already present in '
                                'Artifact storage. Files must be stored outside this location '
                                'prior to Artifact creation.'))
-        file = super().pre_save(model_instance, add)
-        if file and file._committed and add and str(file) != upload_to:
-            # the file needs to be moved into place
+
+        move = (file._committed and file.name != artifact_storage_path)
```
daviddavis: Why define a `move` variable here?

daviddavis:
This looks like a huge improvement! Will test it out in S3.

pulpcore/app/models/fields.py
goosemania: I tried to make it more readable so it's clear what this conditions are deciding. 
I can drop it.

daviddavis: Either way works for me. I'll leave it up to you.

bmbouter:
This is a big improvement. Thanks @goosemania ! Thank you for testing on S3 @daviddavis ! :+1:

PR #3893, opened 2019-02-25
Add unique index for login

goosemania:
closes #4425
https://pulp.plan.io/issues/4425


goosemania:
There is no migration but when user runs pulp-manage-db, new index is created if it's not there yet.
If there are duplicates in the DB, pulp-manage-db fails and user needs to remove duplicates manually.

Just to make sure that everyone is aware of that ^ situation.

PR #3895, opened 2019-02-25
Problem: pulpcore does not have pass-through cache feature

dkliban:
Solution: extend the content app to support pass-through cache

This patch enables the content app handler to look for a remote on a distribution. When a remote
is associated with a distribution, that remote is used to download content missing from Pulp.

re: #3894
https://pulp.plan.io/issues/3894


bmbouter:

PR #3896, opened 2019-02-25
Adopts new content_summary layout

bmbouter:
The performance issue causes us to introduce a new model named
RepositoryVersionContentDetails. This is also a great opportunity to
update the content_summary of the RepositoryVersion serializer to match
the recent API changes.

Most of this code was inspired from a patch from @dalley.

Requires PR: https://github.com/PulpQE/pulp-smash/pull/1174

https://pulp.plan.io/issues/4283
closes #4283


bmbouter:
The API now looks like this:

screenshot from 2019-02-25 17-54-43

That matches this data in the DB:

```
pulp=> select * from pulp_app_repositoryversioncontentdetails;
 id | content_type | count | repository_version_id | count_type
----+--------------+-------+-----------------------+------------
  1 | file.file    |     3 |                     1 | A
  2 | file.file    |     3 |                     1 | P
(2 rows)
```

pulpcore/app/models/repository.py
```diff
@@ -474,3 +507,52 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        ctype_view = get_view_name_for_model(ctype_model, 'list')
+        try:
+            ctype_url = reverse(ctype_view)
+        except django.urls.exceptions.NoReverseMatch:
+            # We've hit a content type for which there is no viewset.
+            # There's nothing we can do here, except to skip it.
+            return
+        rv_href = "/pulp/api/v3/repositories/{repo}/versions/{version}/".format(
+            repo=self.repository_version.repository.pk, version=self.repository_version.number)
+        full_url = "{base}?repository_version={rv_href}".format(
```
dralley: This string needs to be toggled on the type of the count, e.g. if it's the href for content being added, it should be ```"{base}?repository_version_added={rv_href}"``` and if it's removed it should be ```"{base}?repository_version_removed={rv_href}"```.   For "present" it would stay as it is currently.

bmbouter: fixed in this push. Thanks @dralley !

bmbouter:

PR #3907, opened 2019-04-02
Problem: streamer decodes data while streaming

dkliban:
Solution: configure nectar downloaders to not decode responses

This patch relies on a new config option for the nectar downloaders.

re: #4603
https://pulp.plan.io/issues/4603


dkliban:
All of the failing tests are due to nectar 1.5.6 being used to test. This patch requires https://github.com/pulp/nectar/pull/65

goosemania (approved):
Thank you for fixing the issue, great job, @dkliban

PR #3911, opened 2019-04-17
Improve performance of publish with iso_distributor via fast-forward

zxiong:
Only symlinks for those incremental or changed files are generated
and copied to target directories, then the publish gets faster.

ref #4708
https://pulp.plan.io/issues/4708

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -11,6 +11,7 @@ from pulp.common.plugins.progress import ProgressReport
 from pulp.plugins.distributor import Distributor
 from pulp.server.managers.repo import _common as common_utils
 from pulp.server.util import copytree
+from pulp.server.db.model.criteria import UnitAssociationCriteria
 
 BUILD_DIRNAME = 'build'
 
@@ -69,6 +70,9 @@ class FileDistributor(Distributor):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
+        if not config.get("skip_fast_forward", True):
+            return self.publish_repo_fast_forward(repo, publish_conduit, config)
+
         progress_report = FilePublishProgressReport(publish_conduit)
         _logger.info(_('Beginning publish for repository <%(repo)s>') % {'repo': repo.id})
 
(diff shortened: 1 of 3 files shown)


server/pulp/plugins/file/distributor.py
```diff
@@ -117,6 +121,101 @@ def publish_repo(self, repo, publish_conduit, config):
...
+        """
+        Publish the repository.
+
+        :param repo:            metadata describing the repo
+        :type  repo:            pulp.plugins.model.Repository
+        :param publish_conduit: The conduit for publishing a repo
+        :type  publish_conduit: pulp.plugins.conduits.repo_publish.RepoPublishConduit
+        :param config:          plugin configuration
+        :type  config:          pulp.plugins.config.PluginConfiguration
+        :param config_conduit: Configuration Conduit;
```
rohanpm: There's no config_conduit parameter.

zxiong: Thanks, Rohan. Removed used description.

dkliban:
ok test

dkliban:
ok test

dkliban:
This change applies to anything that inherits from FileDistributor. I am only aware of PuppetFileDistributor and ISODistributor. 

https://github.com/pulp/pulp_puppet/blob/2-master/pulp_puppet_plugins/pulp_puppet/plugins/distributors/filedistributor.py#L20
https://github.com/pulp/pulp_rpm/blob/2-master/plugins/pulp_rpm/plugins/distributors/iso_distributor/distributor.py#L21

dkliban:
@zxiong There is a failing unit test.

zxiong:
@dkliban Where can I check test report to see which test case was failed? And is there any guide to setup unit test env to run the tests?

zxiong:
I found the failed place in the testcase[1], mock_makedirs.assert_has_calls([call('dst'), call('dst/dir1'), call('dst/dir2')])

But I don't find out the reason why it's failed. @dkliban, could you help to check it if it's possible?

[1]/home/jenkins/workspace/unittest-pulp-pr/node-type/rhel7-np/pulp/server/test/unit/server/test_util.py

zxiong:
If we look at the code change for server/pulp/server/util.py, "os.makedirs(dst)" will not be called in general, because the new implementation always copy incremental files from "src" to "dst" directory, it does not remove previous "dst" directory.

Since "os.makedirs(dst)" was not be called in the code changes, it makes sense the "failure" happened. So I guess the assertion[1] may not be suitable for the new implementation.

[1]mock_makedirs.assert_has_calls([call('dst'), call('dst/dir1'), call('dst/dir2')])

dkliban:
@zxiong The os.makedirs is now called conditionally so the test probably needs to be updated. Here is more info on running the tests. 

https://docs.pulpproject.org/dev-guide/policies/testing.html#coverage

zxiong:
@dkliban, thanks.

If the test will be update, the change will need to be squashed into this PR, right?

zxiong:
Squashed unit test update into this PR.

dkliban:
ok test

dkliban:
looks like there is still  a test failure @zxiong

zxiong:
@dkliban, could you help re-run unit test again?

I fixed the issue and re-run the unit test (test_util.py) in my local environment and passed.

goosemania:
ok test

dkliban (approved):

PR #3913, opened 2019-04-23
Problem: flake8 failures are present

dkliban:
Solution: fix real problems and add .travis dir to list of ignored

[noissue]


dkliban:
ok test

goosemania (approved):

flake8.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = ./docs/*, ./playpen/*, */build/*, */backports/*
+exclude = ./docs/*, ./playpen/*, */build/*, */backports/*, ./.travis/*
```
daviddavis: Why is the .travis directory being excluded?

PR #3915, opened 2019-05-01
Remove consumer profiles by id

goosemania:
Each profile is a dictionary. It's not reliable to use nested
dictionaries in criteria to remove a document from a collection
because Python doesn't guarantee key order in a dictionary
but it's important for MongoDB.
The solution is to remove by id, the rest of criteria is redundant.

closes #4654
https://pulp.plan.io/issues/4654


dkliban (approved):
Thank you!

PR #3916, opened 2019-05-06
Use timestamp and repo_id in the temporary directory name

goosemania:
To avoid race condition when multiple repositories are published
at the same time.

closes #3098
https://pulp.plan.io/issues/3098


dkliban (approved):
This makes sense to me. Thank you for fixing it.

PR #3917, opened 2019-05-09
Replace option "skip_fast_forward" with "force_full" for FileDistributor

zxiong:
In order to keep consistent with the option name "force_full" which is used in yum and rsync distributors, the option name "skip_fast_forward" will be replaced to with "force_full" in FileDistributor

ref #4799
https://pulp.plan.io/issues/4799

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -70,7 +70,7 @@ class FileDistributor(Distributor):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
-        if not config.get("skip_fast_forward", True):
+        if config.get("force_full", False):
             return self.publish_repo_fast_forward(repo, publish_conduit, config)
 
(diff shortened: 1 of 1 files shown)


dkliban:
ok test

dkliban (approved):

server/pulp/plugins/file/distributor.py
```diff
@@ -70,7 +70,7 @@ def publish_repo(self, repo, publish_conduit, config):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
-        if not config.get("skip_fast_forward", True):
+        if config.get("force_full", False):
```
rohanpm, line 73: This doesn't make sense: force_full is supposed to *disable* fast-forward, not *enable* it.

- force_full=False should mean "don't force a full publish - it's OK to not process already-published content"
- force_full=True should mean "force a full publish - process all content even if already published"

The behavior is now the wrong way around, where force_full *disables* fast-forward publish for rpm, rsync distributors, but *enables* it for file distributors.

PR #3918, opened 2019-05-10
Correct the opposite logic of "force_full"

zxiong:
It doesn't make sense that "force_full" enable fast-forward, so change
the opposite logic

ref #4810
https://pulp.plan.io/issues/4810

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -70,7 +70,7 @@ class FileDistributor(Distributor):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
-        if config.get("force_full", False):
+        if not config.get("force_full", False):
             return self.publish_repo_fast_forward(repo, publish_conduit, config)
 
         progress_report = FilePublishProgressReport(publish_conduit)
@@ -196,6 +196,8 @@ class FileDistributor(Distributor):
                         dir_name = os.path.dirname(unit_path)
                         if not os.listdir(dir_name):
                             os.removedirs(dir_name)
+                    elif os.path.islink(unit_path):
+                        os.unlink(unit_path)
 
             self.post_repo_publish(repo, config)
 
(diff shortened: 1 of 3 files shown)


zxiong:
Following up last comment from https://github.com/pulp/pulp/pull/3917, correct the opposite logic.

server/pulp/plugins/file/distributor.py
```diff
@@ -70,7 +70,7 @@ def publish_repo(self, repo, publish_conduit, config):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
-        if config.get("force_full", False):
+        if not config.get("force_full", True):
```
ipanova: force_full flag if unspecified should be set to False.
Consider changing to:
```
if not config.get("force_full", False)
```

ipanova (changes requested):
please address https://github.com/pulp/pulp/pull/3918/files#r283252715

zxiong:
@ipanova, could you help to review it again?

ipanova (approved):

dkliban:
ok test

zxiong:
The failed unit tests are caused by test data which is return from "publish_conduit.get_units(criteria=criteria)" in the method "publish_repo_fast_forward", so the mock method "get_units(criteria=None)" in "pulp/devel/mock_distributor.py" is modified to add a condition to make up test data for "publish_repo_fast_forward" when the criteria likes "{'checksum': {"$in": ["sum1",...]}})"

Previously, the method "publish_repo_fast_forward" should not be covered by existing unit tests, so adding five test cases which are referring to test cases for the method "publish_repo".

Besides that, improves "publish_repo_fast_forward" to delete symlinks that pointing files were removed, and the change also is covered by new unit tests.

zxiong:
@ipanova, @dkliban, @rohanpm, could you help to review the PR again?

dkliban:
ok test

PR #3919, opened 2019-05-13
Add an optional pre index part for DB migrations

quba42:
**Caused by**: https://pulp.plan.io/issues/4055
**Fixes:** https://pulp.plan.io/issues/4871

Since https://pulp.plan.io/issues/4055 was fixed, indexes are created prior to db migrations to ensure uniqueness constraints are observed.

However, this makes it pretty much impossible to change an existing key field with a new one.
(The key with the as of yet empty new field may not yet be able to guarantee uniqueness prior to the migration).

This PR adds the option to define an additional `pre_index_migrate()` function, which will be called before indexes are created.

### DIFF
--- a/server/pulp/server/db/manage.py
+++ b/server/pulp/server/db/manage.py
@@ -67,11 +67,18 @@ def migrate_database(options):
     Create indexes before running any migration to avoid duplicates, e.g. in case a new collection
     is created.
 
+    WARNING: If any unapplied migrations include a "prepare_reindex_migration" function, these
+    functions will be run before indexes are applied. This is before ANY additional regular
+    migrations are applied, irrespective of what migration version the DB is currently in.
+    In addition, execution of these functions is not tracked independently of the regular migration
+    they are associated with. This means they may be reapplied (even if successfull) until the
+    associated regular migration has been successfully applied.
+
+    prepare_reindex_migration functions should only be used, to prepare database fields and indexes
+    that would otherwise prevent a successfull index change.
+
     :param options: The command line parameters from the user
     """
-
-    if not options.dry_run:
-        ensure_database_indexes()
 
     migration_packages = models.get_migration_packages()
     unperformed_migrations = False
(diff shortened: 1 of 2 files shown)


quba42:
Should I create a pulp issue to go with this PR?

dkliban:
ok test

goosemania:
@quba42 , @dkliban , the PR looks good. @quba42 , thanks a lot for your work!
I have a concern which I'm not sure how to address. The fact that we have no idea if pre-index part was done fine and in case of any failures for the main migration we need to re-run it, so there is a strict requirement to the pre_index part to be re-runable even if it succeeded.
So after commenting on [debian pre-index part](https://github.com/pulp/pulp_deb/pull/83/files#r284401681), I have an idea and let me know if it's completely wrong or too crazy :)

What if we just make a flag whether to apply indexes or not to the migration. By default we should always do. But if you need a pre-index part, it will be a separate migration which is run first but without applying any index.
Pros:  a migration which used to be a pre-index part is not re-run if successful, it's registered as a migration which is applied; Pulp won't run and will push for running all the migrations, so the second migration with indexes will have to be run.
Cons: 2 migrations. If user removes file with the second migration, you end up in the inconsistent state (pre-migration part applied, migration is not).

What do you think?

quba42:
> What if we just make a flag whether to apply indexes or not to the migration. By default we should always do. But if you need a pre-index part, it will be a separate migration which is run first but without applying any index.

If I understand this correctly it would essentially be an out of order migration? (Some regular migrations may be dependent on the presence of indices to guarantee uniqueness and have to happen in the usual order. Others require indices to be gone, and would so have to happen before the very first regular migration.)

If there are out of order migrations, I am not sure how pulp can reliably record which of these have run and which haven't (is this currently done solely via the order?!). Perhaps there is also some way to independently track the number of pre index migrations and regular migrations that have been applied.

goosemania:
@quba42 , could you elaborate a bit more what do you mean by out-of-order migrations? I'm not sure I followed. Do you want to run a new migration before some old ones?

The example for what I suggested:
1. Say, you already have 5 migrations
2. Now you decide to change your unit_key
3. You write two migrations:
 - first one does everything which should be done without applying new indexes, so it will be migration_0006
 - second one applies indexes and does the rest what should be done when indexes are present, so it will be migration_0007

> (is this currently done solely via the order?!)

Order of migrations are defined by their number. So if 0006 succeeds, and you re-run pulp_manage_db, it will be skipped. The number of the last successfully applied migration is recorded in the database for each plugin, see `db.migration_trackers`.

Did I answer your question? Or you meant something else?

quba42:
@goosemania If I have understood everything correctly then I think in your example you could run into the following problem:

1. You already have 5 migrations
1. Now you decide to change your unit_key
1. You add two new ones, the first of which (0006_migration) needs to run before your new key is applied.
1. You now have a pulp user using a version of pulp with the first four migrations applied.
1. This user tries to update directly to the new version with the new key (perhaps they skipped some release).
1. Pulp tries to apply the 5th migration which expects the old unit key to be applied to be safe
1. Instead Pulp tries to apply the new unit key prior to this migration.
1. Things break...

The only way to catch this scenario is to make sure 0006_migration is run first (even before 0005_migration). It is an "out-of-order" migration. Even if there was some way to run it before migration 0005_migration, the writer of 0006_migration needs to make sure it can be run after any point during the first five migrations, and then you end up with something similar to my PR (except that the out-of-order part has a migration of it's own, and that you now need to track for each number whether that number has already been applied).

> Did I answer your question? Or you meant something else?

Yes, you answered my question, that was exactly what I meant.

goosemania:
@quba42 , we'll accept this PR to meet pulp_deb needs, however the ability to specify pre_index_migrate is quite worrying at the same time. E.g. there are limitations (the fields you can work with have to be in the initial model before any migration and can't be removed in the future) and complexities for developers (what if multiple migrations have pre_index_migrate, one would need to keep an eye on those and check what they do as well), which makes it very easy to break things, imo. My only hope that it won't be used much and it's needed one-time to solve the unit_key change in pulp_deb.

Could you, please, do the following:
 - update docstring of the `migrate_database` to reflect the change and add to the same docstring a warning about complexities/limitations of this change.
 - `pre_index_migrate` is a bit misleading name which may let someone think that it runs just before indexes are "ensured" for this particular migration. If there is a clearer name to reflect that it runs always before any unapplied migrations, that would be great. I don't have good ideas though. So maybe a clear docstring ^ is enough.
 - since the change is noticeable and arguably risky, please, file a redmine issue and refer to it in the commit message.
 - update the commit message to reflect the reasons for the change.

Thank you!

quba42:
@goosemania Thanks for the quick review. I think your analysis is spot on, and I will try to get around to the requested changes (docstrings, issue, and commit message) today.

I don't have any great ideas for a more descriptive function name  (`queue_jumping_pre_index_migrate` ? :stuck_out_tongue_winking_eye: ).
I guess the docstring may need to suffice.

quba42:
I created the following pulp issue to go with this PR: https://pulp.plan.io/issues/4871

quba42:
@goosemania The newest version of the PR seeks to address your requested changes.
Please let me know if you still notice anything in this version.

goosemania:
ok test

goosemania:
ok test

goosemania (approved):

PR #3923, opened 2019-05-22
Fix the issue that PULP_MANIFEST is not update correctly.

zxiong:
Fix the issue that PULP_MANIFEST is not update correctly to publish
files with iso_distributor via fast-forward when deleting a exsting
unit. Besides that, add two unit tests and improve an unit test to
check the fix.

fixes #4865

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -183,11 +183,6 @@ class FileDistributor(Distributor):
                     links_to_create = self.get_paths_for_unit(unit)
                     self._symlink_unit(build_dir, unit, links_to_create)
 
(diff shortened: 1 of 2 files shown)


zxiong:
@ipanova, @dkliban, could you help to review the PR?

goosemania:
ok test

goosemania:
Tested. It works and also fixes https://pulp.plan.io/issues/4857

goosemania (changes requested):
It looks good, thanks for fixing @zxiong !

Please add `fixes #4857` to your commit message, since it fixes that regression as well.

goosemania (approved):
Disregard the previous comment, I marked issues as duplicates.
Thanks for the fix!

PR #3924, opened 2019-05-22
Problem: no config for building 2.19.0 docs

dkliban:
Solution: add a config for building 2.19.0 docs

This is needed because I accidently put 2.19.1 beta docs into place of 2.19.0 docs on
docs.pulpproject.org

[noissue]


bmbouter (approved):
This looks right and I can see .travis already has files like this. Let's merge this I can see how this will fix the docs.

PR #3925, opened 2019-05-24
Add a User-Agent header to pulp downloaders

mdellweg:


dkliban (approved):

dkliban (changes requested):
We should probably add a release note for this.

mdellweg:
ping @dkliban

dkliban (approved):

PR #3934, opened 2019-07-01
Fixing ISO publish fails with BSON document too large

zxiong:
There's a limit to how large a single mongo query can be (16 Mbytes).
If too many (>134k) units are published, the query to fetch units
will exceed that limit and crash.

This fix sets a threshold to avoid generating too large query and
crash.

If exceeding the threshold, go back to regular full publish since
there won't be much benefit of fast-forward anyway.

ref #5058
https://pulp.plan.io/issues/5058

### DIFF
--- a/server/pulp/plugins/file/distributor.py
+++ b/server/pulp/plugins/file/distributor.py
@@ -70,11 +70,14 @@ class FileDistributor(Distributor):
         :return:                report describing the publish operation
         :rtype:                 pulp.plugins.model.PublishReport
         """
+        _logger.info(_('Beginning publish for repository <%(repo)s>') % {'repo': repo.id})
         if not config.get("force_full", False):
-            return self.publish_repo_fast_forward(repo, publish_conduit, config)
+            try:
+                return self.publish_repo_fast_forward(repo, publish_conduit, config)
+            except FastForwardUnavailable:
+                _logger.debug("Fast-forward publish bailed out, continuing normally")
 
         progress_report = FilePublishProgressReport(publish_conduit)
-        _logger.info(_('Beginning publish for repository <%(repo)s>') % {'repo': repo.id})
 
         try:
             progress_report.state = progress_report.STATE_IN_PROGRESS
@@ -173,11 +176,23 @@ class FileDistributor(Distributor):
                             unit_checksum_old_set.add(checksum)
                             if checksum not in unit_checksum_set:
                                 unit_over_path_map[checksum] = fields[0]
+                _logger.debug("%d items were in MANIFEST %s, which exists? %s." % (
+                              len(unit_checksum_old_set), metadata_filename,
+                              os.path.exists(metadata_filename)))
 
                 # Copy incremental files into publishing directories
                 checksum_absent_set = unit_checksum_set - unit_checksum_old_set
+                _logger.debug("Increasing %d units" % len(checksum_absent_set))
+
+                # If added too many units, then publish repo with force_full
+                max_increase_units = min(50000, len(units) / len(hosting_locations))
+                if len(checksum_absent_set) > max_increase_units:
+                    self._rmtree_if_exists(build_dir)
+                    raise FastForwardUnavailable
+
                 criteria = UnitAssociationCriteria(
-                    unit_filters={'checksum': {"$in": list(checksum_absent_set)}})
+                    unit_filters={'checksum': {"$in": list(checksum_absent_set)}},
+                    unit_fields={'name', 'checksum', '_storage_path', 'size'})
                 unit_absent_set = publish_conduit.get_units(criteria=criteria)
                 for unit in unit_absent_set:
                     links_to_create = self.get_paths_for_unit(unit)
@@ -382,3 +397,11 @@ class FilePublishProgressReport(ProgressReport):
         ProgressReport.STATE_NOT_STARTED: (STATE_IN_PROGRESS, ProgressReport.STATE_FAILED),
         STATE_IN_PROGRESS: (ProgressReport.STATE_FAILED, ProgressReport.STATE_COMPLETE),
     }
+
+
+class FastForwardUnavailable():
+    """
+    The excetopn will be raised once publish with fast forward failed, then go back to
+    publish with force full.
+    """
+    pass
--- a/server/test/unit/plugins/file/test_distributor.py
+++ b/server/test/unit/plugins/file/test_distributor.py
@@ -13,6 +13,7 @@ from pulp.common.plugins.distributor_constants import MANIFEST_FILENAME
 from pulp.devel.mock_distributor import get_publish_conduit
 from pulp.plugins.file.distributor import FileDistributor, FilePublishProgressReport, BUILD_DIRNAME
 from pulp.plugins.model import Repository, Unit
+from pulp.plugins.config import PluginCallConfiguration
 
 
 DATA_DIR = os.path.realpath("../../../data/")
@@ -66,8 +67,8 @@ class FileDistributorTest(unittest.TestCase):
     def test_repo_publish_api_calls(self, mock_get_working, force_full=True):
         mock_get_working.return_value = self.temp_dir
         distributor = self.create_distributor_with_mocked_api_calls()
-        result = distributor.publish_repo(self.repo, self.publish_conduit,
-                                          {'force_full': force_full})
+        config = PluginCallConfiguration({}, {}, {'force_full': force_full})
+        result = distributor.publish_repo(self.repo, self.publish_conduit, config)
         self.assertTrue(result.success_flag)
         self.assertTrue(distributor.get_hosting_locations.called)
         self.assertTrue(distributor.post_repo_publish.called)
@@ -84,7 +85,8 @@ class FileDistributorTest(unittest.TestCase):
     def test_repo_publish_files_placed_properly(self, mock_get_working, force_full=True):
         mock_get_working.return_value = self.temp_dir
         distributor = self.create_distributor_with_mocked_api_calls()
-        distributor.publish_repo(self.repo, self.publish_conduit, {'force_full': force_full})
+        config = PluginCallConfiguration({}, {}, {'force_full': force_full})
+        distributor.publish_repo(self.repo, self.publish_conduit, config)
         target_file = os.path.join(self.target_dir, SAMPLE_RPM)
         # test if the link was created
         self.assertTrue(os.path.islink(target_file))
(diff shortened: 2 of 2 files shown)


ipanova:
@zxiong @midnightercz  @rohanpm Thank you for the PR. Hard-coding the value of the distributor config `force_full` in the code is not very friendly. It seems like this would satisfy your specific situation, given that i suggest whenever you make an API  publish call then set force_full to true in the override_config parameter.

Did you consider to limit `criteria` to certain fields when the unit is returned? 
This way we might save on the query size.

zxiong:
> @zxiong @midnightercz @rohanpm Thank you for the PR. Hard-coding the value of the distributor config `force_full` in the code is not very friendly. It seems like this would satisfy your specific situation, given that i suggest whenever you make an API publish call then set force_full to true in the override_config parameter.

It's difficult to decide to choose better way (fast-forward vs force-full) before making an API publish call since it needs to estimate based on some data in Pulp.  And setting force_full does not impact original configuration, as the config object is copied from original "config" object then passed to "do_publish" as a parameter.

> Did you consider to limit `criteria` to certain fields when the unit is returned?
> This way we might save on the query size.

Limited `criteria` to certain fields as your suggestion. Tests shows it saves ~40% of returned size from MongoDB logs. It's great.

ipanova:
@zxiong you mentioned that the size of the query decreased 40%. Does it still hit the limit? if not, can we just continue publishing fast-forward way( you said you don't gain much, i assume you also do not loose anything), so we do not set the value of force_full option based on the number of units?

server/pulp/plugins/file/distributor.py
```diff
@@ -173,11 +174,27 @@ def publish_repo_fast_forward(self, repo, publish_conduit, config):
...
+                _logger.debug("%d items were in MANIFEST %s, which exists? %s." % (
+                              len(unit_checksum_old_set), metadata_filename,
+                              os.path.exists(metadata_filename)))
 
                 # Copy incremental files into publishing directories
                 checksum_absent_set = unit_checksum_set - unit_checksum_old_set
+                _logger.debug("Increasing %d units" % len(checksum_absent_set))
+
+                # If added too many units, then publish repo with force_full
+                max_increase_units = min(50000, len(units) / len(hosting_locations))
```
rohanpm, line 188: What's the idea with dividing by len(hosting_locations)?

server/pulp/plugins/file/distributor.py
```diff
@@ -173,11 +174,27 @@ def publish_repo_fast_forward(self, repo, publish_conduit, config):
...
+                if len(checksum_absent_set) > max_increase_units:
+                    config_tmp = copy.deepcopy(config)
+                    if not config_tmp.override_config:
+                        config_tmp.override_config = {}
+                    config_tmp.override_config['force_full'] = True
+                    self._rmtree_if_exists(build_dir)
+                    return self.publish_repo(repo, publish_conduit, config_tmp)
```
rohanpm: I'd probably have gone with raising an exception instead, that avoids having to rewrite override_config and overall I think the code path would be less convoluted.

I mean for example, instead of having:

`publish_repo` => `publish_repo_fast_forward` => `publish_repo` (with different config)

You would instead have:

`publish_repo` => try `publish_repo_fast_forward`, if it fails, continue normally

Like this:

```
    if not config.get("force_full", False):
        try:
          return self.publish_repo_fast_forward(repo, publish_conduit, config)
        except FastForwardUnavailable:
          _logger.debug("Fast-forward publish bailed out, continuing normally")

    # (...rest of the publish_repo code here as normal)
```

Probably less confusing than having a method call an inner method which might call back to the outer method again with different arguments.

rohanpm (commented):
I put some suggestions, but seems to me like this is going to work.

rohanpm:
> @zxiong you mentioned that the size of the query decreased 40%. Does it still hit the limit?

Requesting the specific needed fields is a good idea, but I don't see how it could help this particular problem.  The issue was never about how much data was returned by the mongo query, it's the *size of the query itself* which exceeds the maximum size mongo is able to handle.  Adding "fields" doesn't reduce the size of the query, rather it (very slightly) *increases* it.

server/pulp/plugins/file/distributor.py
zxiong: During publish, symlink generating and copying to hosting_locations take most of the time. so both are expected to do less.

For symlink generating, when doing with fast-forward,
`total number of generated symlinks` = `number of increase units` * `len(hosting_locations)`
When doing with force full,
`total number of generated symlinks` = `number of total units`

We expect to do publish with fast-forward to gain better performance, so,
`number of increase units` * `len(hosting_locations)` <= `number of total units`
Then, 
`number of increase units` <= `number of total units / len(hosting_locations)`

server/pulp/plugins/file/distributor.py
zxiong: Thanks for your suggestion, updated it.

ipanova:
@rohanpm isee what you're saying.  i am trying to understand where exactly are we reaching this limit? there was mentioned that issue occurs with 136k+ units being published.  Checksum size is 32bytes * 134k units, if we do the math here i cannot really understand where the 16mb+ come from?

zxiong:
> @rohanpm isee what you're saying. i am trying to understand where exactly are we reaching this limit? there was mentioned that issue occurs with 136k+ units being published. Checksum size is 32bytes * 134k units, if we do the math here i cannot really understand where the 16mb+ come from?

@ipanova, I got a query (just search 7 checksum in this case) from mongodb logs (part of the query is omitted since it's too long):

`2019-07-04T07:09:28.622+0000 [conn1018664] warning: log line attempted (6706k) over max size (10k), printing beginning and end ... query pulp_database.units_iso query: { $query: { $and: [ { _id: { $in: [ "b45f705b-dc0a-4f73-83f1-5d66d0de45ee", "704fbf3e-86bd-4410-b39f-f7f02da5f838", "4d68b10b-e675-47b9-9e4e-661000083022", ......, "4d68b10b-e675-47b9-9e4e-661000032663" ] } }, { checksum: { $in: [ "d83670fa63415b787b7d7c9b9d86947be36910f371426b83be511e3f55c02a3e", ......, "2cd6bd21400d20b70f991d1847048672f887af056b3b748dce6d65fdd53b71ea" ] } } ] }`

The filter of above query includes `_id` and `checksum`. Both both conditions are huge if too many units are queried, and query size increases as the number of checksum in the query and the number of existing units in the repo.

Regarding how to calculate query size exactly, we may need to go through `pymongo`.

zxiong:
@ipanova, following is the script to reproduce the issue:

	import pymongo

	myclient = pymongo.MongoClient("mongodb://xxx.xxx.xxx.xxx:27017/")`
	mydb = myclient["pulp_database"]
	units_iso = mydb["units_iso"]

	checksums = []
	unit_ids = []

	for i in range(135000):
	    a = 1000000000 + i

	    md5sum_fake = "38f51894ddeb181ae0cb4acedd4d34ceccbef999d0a4e5071a7d68%s" % (a)
	    checksums.append(md5sum_fake)

	    unit_fake = "0cc2db4f-8311-4905-886c-0b%s" % (a)
	    unit_ids.append(unit_fake)

	query = {'_id': {'$in': unit_ids} ,'checksum': {'$in': checksums}}
	res = units_iso.find(query)
	for r in res:
	    print r

ipanova:
ok test

ipanova:
@zxiong thank you for the reproducer, that was helpful.
I am looking at the code logic and i see that in the situation when we fail over to force-full publish we instantiate a new progress report as well as processing again each unit https://github.com/pulp/pulp/pull/3934/files#diff-d43dd2dd64469093b3a8e01b32735db3L92

If i read it correctly we will process 134k units, realize that we cannot go fast-forward and then process 134k units again?

Did you benchmark this scenario? Are there performance implications in this case? i.e how longer publish will be in case when we try to go fast-forward and then fail over force full publish in respect to publishing force-full only?

zxiong:
> @zxiong thank you for the reproducer, that was helpful.
> I am looking at the code logic and i see that in the situation when we fail over to force-full publish we instantiate a new progress report as well as processing again each unit https://github.com/pulp/pulp/pull/3934/files#diff-d43dd2dd64469093b3a8e01b32735db3L92
> 
> If i read it correctly we will process 134k units, realize that we cannot go fast-forward and then process 134k units again?
> 
> Did you benchmark this scenario? Are there performance implications in this case? i.e how longer publish will be in case when we try to go fast-forward and then fail over force full publish in respect to publishing force-full only?

There is a slight overhead, it's around 5% (+2 minutes / 40 minutes) in my test, which published 134k new increase units (~40k units in the repo originally). So I think it's acceptable.

ipanova:
@zxiong thank you for this info. Please run pulp-smash tests in order to reveal regressions if any. This step will save a lot of time for qe and devs.

ipanova (approved):
@zxiong thank you for the PR!

nixocio:
Tests related to Pulp 2 are here:
https://github.com/PulpQE/Pulp-2-Tests

I would suggest at least run the tests that have the `iso_importer`.
```sh
 git grep iso_importer
```
pulp_2_tests/tests/rpm/api_v2/test_duplicate_uploads.py:            'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_duplicate_uploads.py:            'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:    '_href': '/pulp/api/v2/repositories/{}/importers/iso_importer/',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:    'id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:    'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:        'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:            'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:                'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:                'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:                'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py:            {'importer_type_id': 'iso_importer'},
pulp_2_tests/tests/rpm/api_v2/test_iso_sync_publish.py:            'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_sync_publish.py:            'importer_type_id': 'iso_importer',
pulp_2_tests/tests/rpm/api_v2/test_iso_sync_publish.py:            'importer_type_id': 'iso_importer',

ipanova:
@zxiong please let us know if those tests passed and we are ready to merge!

zxiong:
Sure. I will try to run those tests as soon as possible.

zxiong:
Ran subset of smash tests:

`
pulp_2_tests/tests/rpm/api_v2/test_duplicate_uploads.py
pulp_2_tests/tests/rpm/api_v2/test_iso_crud.py
pulp_2_tests/tests/rpm/api_v2/test_iso_sync_publish.py
`

All passed.

PR #3941, opened 2019-07-11
Manually retry failing task upsert queries

daviddavis:
fixes #4428
https://pulp.plan.io/issues/4428

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -16,6 +16,7 @@ from celery import task, Task as CeleryTask, current_task, __version__ as celery
 from celery.app import control, defaults
 from celery.result import AsyncResult
 from mongoengine.queryset import DoesNotExist
+from mongoengine.errors import NotUniqueError
 
 from pulp.common.constants import RESOURCE_MANAGER_WORKER_NAME, SCHEDULER_WORKER_NAME
 from pulp.common import constants, dateutils, tags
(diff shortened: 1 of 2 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -514,11 +515,20 @@ def __call__(self, *args, **kwargs):
             worker_name = self.request.hostname
             # Using 'upsert' to avoid a possible race condition described in the apply_async method
             # above.
-            TaskStatus.objects(task_id=self.request.id).update_one(
-                set__state=constants.CALL_RUNNING_STATE,
-                set__start_time=start_time,
-                set__worker_name=worker_name,
-                upsert=True)
+            try:
```
dkliban, line 518: This is where the sleep statement would have helped.

dkliban (approved):
These changes make sense to me. I was misunderstanding the problem before. I commented on the line where we could try a sleep statement. However, I think these changes should suffice.

server/pulp/server/async/tasks.py
```diff
@@ -514,11 +515,20 @@ def __call__(self, *args, **kwargs):
...
+                TaskStatus.objects(task_id=self.request.id).update_one(
+                    set__state=constants.CALL_RUNNING_STATE,
+                    set__start_time=start_time,
+                    set__worker_name=worker_name,
+                    upsert=True)
+            except DuplicateKeyError:
```
daviddavis: This should be `NotUniqueError`

evgeni:
This patch fixed the E11000 errors I've been seeing when syncing multiple repos in bulk to a 2.18 install.

dkliban (approved):
Thank you!

PR #3942, opened 2019-07-11
Let's get some autonotification going for builds

ggainey:


daviddavis (approved):
I might consider setting on_success to "changed" but we can try this out first and see how much noise this creates first.

ggainey:
> I might consider setting on_success to "changed" but we can try this out first and see how much noise this creates first.

Exactly - planning to start with the defaults, and then tweak once we see how it works.

PR #3943, opened 2019-07-15
Use 'pulp-common' to retrieve the package version

mdellweg:
closes #5126
https://pulp.plan.io/issues/5126


dkliban (approved):
Thank you! This looks right to me.

PR #3944, opened 2019-07-15
Enhance irc-build-notification template

ggainey:


.travis.yml
```diff
@@ -26,4 +26,13 @@ before_install:
...
 notifications:
-  irc: "chat.freenode.net#pulp-build"
+  irc:
+    channels:
+      - "chat.freenode.net#pulp-build"
+    template:
+      - "%{repository_slug}#%{build_number} (%{branch} - %{commit} : %{author} : %{commit_subject})"
+      - "Change view : %{compare_url}"
+      - "Build details : %{build_url}"
+      - "%{message}"
```
daviddavis, line 36: Just curious: how is this different than the default and why change it up?

.travis.yml
```diff
@@ -26,4 +26,13 @@ before_install:
...
+  on_success: change
```
daviddavis, line 37: :+1:

.travis.yml
ggainey: Added the commit_subject - makes it easier to know what the PR is at a glance (imnsho, anyway). I expect the template will get tweaked over time as we discover what will make it most useful for us.

daviddavis (approved):

PR #3945, opened 2019-07-19
Add API to allow queueing an async task for multiple resources

ggainey:
Adds API to tasks.py:
  * _queue_reserved_task_list
  * get_worker_for_reservation_list
  * apply_async_with_reservation_list

Closes #5108
https://pulp.plan.io/issues/5108

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -108,6 +108,46 @@ class PulpTask(CeleryTask):
 
 
 @task(base=PulpTask, acks_late=True)
+def _queue_reserved_task_list(name, task_id, resource_id_list, inner_args, inner_kwargs):
+    """
+    A task that allows multiple resources to be reserved before dispatching a second, "inner", task.
+    See _queue_reserved_task for details on the inner workings.
+
+    :param name:                The name of the task to be called
+    :type name:                 basestring
+    :param inner_task_id:       The UUID to be set on the task being called. By providing
+                                the UUID, the caller can have an asynchronous reference to the inner
+                                task that will be dispatched.
+    :type inner_task_id:        basestring
+    :param resource_id_list:    A list of names of the resources you wish to reserve for your task.
+                                The system will ensure that no other tasks that want any of the same
+                                reservations will run concurrently with yours.
+    :type  resource_id_list:    list
+
+    :return: None
+    """
+    _logger.debug('_queue_reserved_task_list for task %s and ids [%s]' %
+                  (task_id, resource_id_list))
+    # Find a/the available Worker for processing our list of resources
+    worker = get_worker_for_reservation_list(resource_id_list)
+    # Reserve each resource, associating them with that Worker
+    for rid in resource_id_list:
+        _logger.debug('...saving RR for RID %s' % rid)
+        ReservedResource(task_id=task_id, worker_name=worker['name'], resource_id=rid).save()
+
+    # Dispatch the Worker
+    inner_kwargs['routing_key'] = worker.name
+    inner_kwargs['exchange'] = DEDICATED_QUEUE_EXCHANGE
+    inner_kwargs['task_id'] = task_id
+    try:
+        celery.tasks[name].apply_async(*inner_args, **inner_kwargs)
+    finally:
+        # Arrange to release all held reserved-resources
+        _release_resource.apply_async((task_id, ), routing_key=worker.name,
+                                      exchange=DEDICATED_QUEUE_EXCHANGE)
+
+
+@task(base=PulpTask, acks_late=True)
 def _queue_reserved_task(name, task_id, resource_id, inner_args, inner_kwargs):
     """
     A task that encapsulates another task to be dispatched later. This task being encapsulated is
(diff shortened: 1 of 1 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -373,20 +452,110 @@ def serialize(self):
...
+                    args=[task_name, inner_task_id, reservation, args, kwargs],
+                    queue=RESOURCE_MANAGER_QUEUE
+                )
+        except Exception:
+            TaskStatus.objects(task_id=task_status.task_id).update(state=constants.CALL_ERROR_STATE)
+            raise
+
+        return AsyncResult(inner_task_id)
+
+    def apply_async_with_reservation_list(self, resource_type, resource_id_list, *args, **kwargs):
```
dkliban: It would be better if we let the user specify a dictionary of resources where the type is the key and resource id is the value. What do you think about that?

dkliban (commented):
This looks really good overall. I left one question that I'd like to discuss before approving. @bmbouter, could you also provide feedback?

server/pulp/server/async/tasks.py
```diff
@@ -198,6 +238,45 @@ def get_worker_for_reservation(resource_id):
...
+    """
+
+    _logger.debug('get_worker_for_reservation_list [%s]' % resources)
+    # We leave this loop once we find a Worker to return - otherwise, sleep and try again
+    while True:
+        reservation_workers = set(
+            [reservation['worker_name'] for reservation in
+                ReservedResource.objects(resource_id__in=resources)])
+        _logger.debug('...num-RR is %d' % len(reservation_workers))
+        if len(reservation_workers) == 1:  # Exactly one worker holds any of the desired resources
```
bmbouter, line 263: s/any/all/

server/pulp/server/async/tasks.py
ggainey: A couple of things 
- do you mean the/a key would be the type ('list' or 'str') and the value would be a list-of-resource-id-strs, or a single resource-id, depending on the type?
- do we want to handle the case where the dict has *both* list and str keys?
- Do we want even notice if the user passes in a type that is neither list nor str? (as written, if it's not-a list, we *assume* it's a string and safe to just pass along to the existing singler-resource method)

Consider that `reservation` could look like this:

`{ 'list': ['res_Id-1', 'res_id_2', 'res_id_3'], 'str': 'res_id_4', 'object': 'we-ignore-this'}`

In that case, we would have to 'notice' the list-and-str case, append the solo 'str' to the list, and then call `_queue_reserved_task_list` with the resulting combined list. (And whatever you passed in as not-list-or-str-type we would ignore. Or throw a "what are thinking?' exception?)

server/pulp/server/async/tasks.py
bmbouter: Actually it's all, nm.

server/pulp/server/async/tasks.py
dkliban: `resources = {'repository': 'repo-id-foo', 'importer': 'importer-id-bar', 'repository': 'repo-id-shoe'}`

bmbouter: @dkliban +1 to the simplicity of that, but I expected a list since it's has `_list` in the name.

dkliban: Maybe a list of tuples? [(resource_type, resource_id)]

bmbouter: I'm not convinced we should make another dispatch interface with the `_list` name. What if Can't we generalize the existing one to coerce the single type case to a list with something like:

```
if type(resource_id) != 'list':
    list(resource_id)
```

What do you think about something like that?

dkliban: if we do that, then will we assume that all the resources are of the same type? Or will we assume that the resource type is a list also and each item corresponds to an item in the list of resource ids?

ggainey: I went back and forth about that with myself and w/@dkliban. Making a new API makes it clearer where we're grabbing multiple resource locks. Also makes the proposal above (list-of-tuples) possible, without the confusion of "why do I have to provide a resource-type arg when I built all of them into the list I'm passing in?"
I mean, the *cleanest* thing to do is have one API that takes a list-of-tuples - but my first concern w/Pulp2 is "do no harm", and I didn't think going back and finding/changing all the existing code (which works) to use a new API just to be marginally cleaner, was a good choice - more risk, low return. So instead opted to be explicit about what's going on, when we call this API, and leave the existing code as unchanged as I could. Make any sense?

ggainey: Concur w/list-of-tuples

bmbouter: @dkliban I had assumed the latter. That is how Pulp3 works at least. Resource is a list of strings:

resource_id_list = ['lock1', 'lock2', 'lock3']

What is written here is ok too. I think I just imagined it simpler with one interface.

ggainey: async_with_reservation took a type and an id and made a combined-key out of them, in order to essentially namespace the reservations, so _list needs to take a list of (type,id) tuples to accomplish the same goal - since we don't want to push the combined-key-creation out onto the caller.

bmbouter: @ggainey I see. Thank you. +1 to the list of tuples described here https://github.com/pulp/pulp/pull/3945/files#r307866063

ggainey: @dkliban @bmbouter tuple-accepting-API pushed as a new commit - feel free to throw moar rocks

dkliban (approved):

dkliban:
ok test

dkliban:
ok test

PR #3946, opened 2019-07-23
Add verify_unit callback to GetLocalUnitsStep

mdellweg:
re #5170
https://pulp.plan.io/issues/5170


server/pulp/plugins/util/publish_step.py
```diff
@@ -1256,12 +1263,18 @@ def process_main(self, item=None):
             available_units = self.parent.available_units
 
         for units_group in misc.paginate(available_units, self.unit_pagination_size):
+            # any units that are already in pulp
+            units_we_already_had = set()
```
mdellweg, line 1267: Moving this around should not impact functionality, but improve performance.

dkliban (approved):

PR #3949, opened 2019-08-19
Fix Warning Indentaion

amayer5125:
Fixes issue where numbering starts over on [user-guide/installation/rhel_centos](https://docs.pulpproject.org/user-guide/installation/rhel_centos.html). On step labelled "Start Apache httpd and set it to start on boot. For Upstart based systems:"


dkliban:
ok test

dkliban (approved):

PR #3963, opened 2019-10-15
Add --ignore-existing to remote_mkdir

midnightercz:
To avoid recreating existing parts of directory structure --ignore-existing is
added to arguments passed to rsync when calling remote_mkdir


ipanova (approved):

ipanova:
ok test

PR #3964, opened 2019-10-15
Allow for MONGODB-X509 authentication to Mongo

ggainey:
Adds a config-setting, 'x509_auth: true/false', to the [database] stanza. Can
be used only if ssl: true and ssl_certfile: <path> are set, username: is the subject
of the specified client-PEM, and Mongo is set to clusterAuthMode = x509

closes #5512


daviddavis (approved):
LGTM :+1:

ipanova (approved):

PR #3966, opened 2019-11-15
Fix modulemd_defaults upload/copy collision traceback

ggainey:
closes #5658

### DIFF
--- a/server/pulp/server/db/model/__init__.py
+++ b/server/pulp/server/db/model/__init__.py
@@ -883,7 +883,15 @@ class FileContentUnit(ContentUnit):
         path = FileStorage.get_path(self)
         if filename:
             if not os.path.isabs(filename):
-                path = os.path.join(path, filename)
+                if not os.path.exists(path) or os.path.isdir(path):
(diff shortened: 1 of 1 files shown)


daviddavis:
Forgive my ignorance but I see that this problem is caused by two different directory structures being used for modulemd defaults (uploaded vs associated). Is it possible to resolve that problem? Or is it possible to resolve this problem in pulp_rpm somehow?

ggainey:
> Forgive my ignorance but I see that this problem is caused by two different directory structures being used for modulemd defaults (uploaded vs associated). Is it possible to resolve that problem? Or is it possible to resolve this problem in pulp_rpm somehow?

You're not wrong, and we thought about approaching this by modifying the upload code path. Fixing upload fixes future uses, but since this has been released 'into the wild already', would require work at upgrade-time to find and move/rename affected files.

The combination of "risk of cracking open upload", the impact on upgrade time, and the number of other things we need to be doing with the resources we have, made me decide to patch the problem at the symptom - especially since we're already inside a method that already describes itself with "This is a total hack"

daviddavis (approved):
Thanks for the explanation. LGTM.

PR #3969, opened 2019-12-11
Updates latest version to 3.0

dkliban:
This patch ensures that pulp 2 docs never publish to the root of docs.pulpproject.org again.
Docs will continue to be published to docs.pulpproject.org/en/2.Y/.

[noissue]


bmbouter (approved):
This is an important change. Thanks @dkliban !

PR #3971, opened 2019-12-17
Split filter items to avoid BSON document too large if source repo contains > 345,000 units of same type

zxiong:
The issue was caused by long query (>16M) if source repo contains too many
units ( > 345,000) of same type.

The PR splits unit_ids in the filter to several chunks, then fetches data
separately, eventually combines all results.

Note that the PR only handle the case when the query does not need to sort.
It means the PR will not help if the query contains sorting. So the patch does
not fix the issue pulp#2220 completely.

ref pulp#2220
https://pulp.plan.io/issues/2220

### DIFF
--- a/server/pulp/server/managers/repo/unit_association_query.py
+++ b/server/pulp/server/managers/repo/unit_association_query.py
@@ -23,6 +23,8 @@ SORT_DESCENDING = pymongo.DESCENDING
 
 _VALID_DIRECTIONS = (SORT_ASCENDING, SORT_DESCENDING)
 
+UNITS_BATCH_SIZE = 100000
+
 
 class RepoUnitAssociationQueryManager(object):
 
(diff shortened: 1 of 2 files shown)


rbikar:
@zxiong  Looks good to me, I'd rather leave +1 for pulp developer as I'm not that familiar with this part of code.

dkliban:
ok test

daviddavis:
This patch looks good to me. I've asked the Pulp team for a second review since I haven't worked much in this area of Pulp 2.

Also, I think we'll probably wait until @ipanova is back in January to merge this since we don't have any upcoming releases scheduled before her return.

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -23,6 +23,7 @@
 
 _VALID_DIRECTIONS = (SORT_ASCENDING, SORT_DESCENDING)
 
+UNITS_CHUNK_SIZE = 100000
```
ipanova: i would rather name this variable units_batch_size, chunk can be confusing

server/pulp/server/managers/repo/unit_association_query.py
```diff
@@ -115,9 +116,23 @@ def get_units(self, repo_id, criteria=None, as_generator=False):
...
         # collections once we've returned our limit of results.
         # Be sure to skip cursors that would otherwise return an empty result set.
-        units_cursors = (self._associated_units_by_type_cursor(t, criteria,
-                                                               associations_lookup[t].keys())
-                         for t in association_unit_types if t in associations_lookup)
+        if criteria.unit_sort:
+            units_cursors = (self._associated_units_by_type_cursor(t, criteria,
+                                                                   associations_lookup[t].keys())
+                             for t in association_unit_types if t in associations_lookup)
+        else:
```
ipanova, line 129: if you could leave here a comment with an explanation something along the lines https://pulp.plan.io/issues/2220#note-2 that would be very helpful whenever we'd look into this code in the future.

ipanova:
@zxiong thank you for the patch, it looks good, i left 2 minor comments.

server/pulp/server/managers/repo/unit_association_query.py
zxiong: Replaced UNITS_CHUNK_SIZE with UNITS_BATCH_SIZE.

server/pulp/server/managers/repo/unit_association_query.py
zxiong: Added comments to explain why and how to change.

ipanova:
ok test

ipanova (approved):

PR #3974, opened 2020-02-03
Respect last_unit_added when deciding whether to publish

rohanpm:
This is a follow-up to commits for issue #5951.
Previous commits made it so that mutating an erratum unit would set
last_unit_added on all affected repos; however, the code for deciding
whether to skip publish ignored this field and only looked at
RepositoryContentUnits, so the commits didn't fix the issue.

Fix this by checking last_unit_added as well, so we cover both the
"new RepositoryContentUnits created for the repo" and "unit shared
between multiple repos was updated" cases. Erratum units are the only
type known to fall into the latter case.

re: #5951
https://pulp.plan.io/issues/5951


ipanova (approved):

ipanova:
ok test

PR #3981, opened 2020-03-18
Fix last_publish method call

emilyzheng:
Method last_publish was not called properly so that a bound
method was used instead of its return. It made publish always
be fast-forward no matter repo has been published or not when
force_full=True is not set.

Added new case for testing this and fixed tests added before.

Closes #5659

### DIFF
--- a/devel/pulp/devel/mock_distributor.py
+++ b/devel/pulp/devel/mock_distributor.py
@@ -6,7 +6,7 @@ from pulp.plugins.model import PublishReport
 
 
 def get_publish_conduit(type_id=None, existing_units=None, pkg_dir=None, checksum_type="sha",
-                        repodata=None, last_published=None):
+                        repodata=None, last_published="2019-12-05 19:40:26.284627"):
     def build_success_report(summary, details):
         return PublishReport(True, summary, details)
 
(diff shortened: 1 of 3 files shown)


negillett:
@emilyzheng as part of this, could you check to see if changes to mock_distributor (devel/pulp/devel/mock_distributor.py), and there therefore test_distributor (server/test/unit/plugins/file/test_distributor.py), are necessary to support this behavior?

emilyzheng:
@nathanegillett test cases are updated.

negillett:
Thank you, @emilyzheng!

negillett (approved):

dkliban:
ok test

negillett:
Hey, @dkliban, it seems the build for this encountered an error. Could you check on it for us?

negillett:
@ipanova could someone take a look at this?

dkliban:
ok test

ipanova:
@nathanegillett @emilyzheng  this PR needs to refer to a new issue. the issue you have referring to  is already in the current_release state. I have opened a new one so please refer to it in the commit message so when we release we do not miss to include the commit in the release. Thanks https://pulp.plan.io/issues/6520

ipanova (approved):

ipanova:
@nathanegillett @emilyzheng  i have manually associated revisions with the proper commits no need to do anything :) thanks for the PR!

negillett:
Oh, thanks, @ipanova!

PR #3982, opened 2020-04-28
Handle emtpy consumer profiles properly

goosemania:
No longer create an applicability profile for empty consumer profiles.
Remove applicability profiles which refer to an empty consumer profile.

closes #6387
https://pulp.plan.io/issues/6387

### DIFF
--- a/server/pulp/server/managers/consumer/applicability.py
+++ b/server/pulp/server/managers/consumer/applicability.py
@@ -344,7 +344,7 @@ class ApplicabilityRegenerationManager(object):
 
         # Get all unit profiles associated with given consumers
         unit_profile_criteria = Criteria(
-            filters={'consumer_id': {'$in': consumer_ids}},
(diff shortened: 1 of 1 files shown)


server/pulp/server/managers/consumer/applicability.py
```diff
@@ -540,7 +532,16 @@ def remove_orphans():
         #
 
         # Find the profile hashes that exist in current UnitProfiles
-        active_profile_hashes = UnitProfile.get_collection().distinct('profile_hash')
+        unit_profile_coll = UnitProfile.get_collection()
+        active_profile_hashes = unit_profile_coll.distinct('profile_hash')
+
+        # Find if there are any empty consumer profiles, remove them from the active ones because
+        # applicability for those is not needed. Max is one profile per content type.
```
goosemania, line 539: 2 content types are supported, 'rpm' and 'modulemd', so max there will be 2 profiles to remove. In the majority of cases it will be 1 because 'rpm' profiles are almost never empty.

ipanova (approved):

PR #3984, opened 2020-05-08
Allow building docs with warnings

dralley:
A newer version of Sphinx warns about something that used to be ok,
but we run Sphinx with "treat warnings as errors" flag. At this
stage in Pulp 2's lifecycle, disabling this is fine.

closes: #6597
https://pulp.plan.io/issues/6597

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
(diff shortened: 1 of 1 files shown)


dralley:
Cherry pick PR https://github.com/pulp/pulp/pull/3985

dkliban (approved):

PR #3986, opened 2020-05-14
Fix batch regenerate applicability for repo

hao-yu:
Pulp calculates the applicability for repo in the batch size
of 10 but the profiles of the 10th consumer in the batch will
always be missed. This commit fixed it.

closes: #6724
https://pulp.plan.io/issues/6724

### DIFF
--- a/server/pulp/server/managers/consumer/applicability.py
+++ b/server/pulp/server/managers/consumer/applicability.py
@@ -149,9 +149,8 @@ class ApplicabilityRegenerationManager(object):
                         continue
                     seen_hashes.add(all_profiles_hash)
                     profiles = consumer_profile_map[consumer_id]['profiles']
-                    if len(profiles_to_process) < batch_size:
-                        profiles_to_process.append((repo_id, all_profiles_hash, profiles))
-                    else:
+                    profiles_to_process.append((repo_id, all_profiles_hash, profiles))
+                    if len(profiles_to_process) >= batch_size:
                         batch_regenerate_applicability_task.apply_async(
                             (profiles_to_process,), **{'group_id': task_group_id})
                         profiles_to_process = []


server/pulp/server/managers/consumer/applicability.py
```diff
@@ -152,6 +152,7 @@ def queue_regenerate_applicability_for_repos(repo_criteria):
                     if len(profiles_to_process) < batch_size:
                         profiles_to_process.append((repo_id, all_profiles_hash, profiles))
                     else:
+                        profiles_to_process.append((repo_id, all_profiles_hash, profiles))
                         batch_regenerate_applicability_task.apply_async(
                             (profiles_to_process,), **{'group_id': task_group_id})
                         profiles_to_process = []
```
goosemania: It seems like the same code is in both condition statements, should it just be outside?
```
profiles_to_process.append((repo_id, all_profiles_hash, profiles))
if len(profiles_to_process) >= batch_size:
    batch_regenerate_applicability_task.apply_async(
        (profiles_to_process,), **{'group_id': task_group_id})
￼   profiles_to_process = []
```
@hao-yu , what do you think?
And thanks for the patch!

goosemania (changes requested):

server/pulp/server/managers/consumer/applicability.py
hao-yu: Yeah, that looks good to me. I will make the change and re-submit the patch. Thanks for the suggestion.

hao-yu:
>
             for consumer_id in repo_consumer_map[repo_id]:
                if consumer_id in consumer_profile_map:
                    all_profiles_hash = consumer_profile_map[consumer_id]['all_profiles_hash']
                    if all_profiles_hash in seen_hashes:
                        continue
                    seen_hashes.add(all_profiles_hash)

@goosemania Looks like I found another bug here. The seen_hashes condition will cause all the consumers with exactly the same all_profile_hash to be skipped. For example, I have 100 hosts and all of them have exactly the same installed packages. Only the first 1 will get calculated. I think the fix is to remove the seen_hashes condition. What do you think? Do you want me to include it in this patch?

goosemania:
@hao-yu , I think it's a desired behaviour, let me know if I misunderstood you.

Applicability is calculated for a profile, not for a consumer, so the same calculation doesn't happen multiple times. When one queries the applicability results, Pulp looks at the consumer and finds applicability results for its profile.

hao-yu:
@goosemania , Ah sorry. You are right. Consumer looks for repo profile applicability by profile. Thanks.

hao-yu:
@goosemania done!

hao-yu:
@goosemania I think I found a regression of a previous commit. Please check this out:

https://github.com/pulp/pulp/pull/3990

Thanks.

goosemania:
@hao-yu , I was on a vacation, sorry for the delay.
Thanks for the changes, looks good!

goosemania (approved):

PR #3988, opened 2020-05-18
Fix $date operator to work with $and or $or

rajulkumar:
Queries with $date operator doesn't work as expected
with $and or $or operators. $date operator descends
only on dicts and ignores everything else. This fix
modifies $date operator to descend on list too so
it work with $and or $or operators.

Closes #5700

### DIFF
--- a/server/pulp/server/db/model/criteria.py
+++ b/server/pulp/server/db/model/criteria.py
@@ -499,9 +499,13 @@ class DateOperator(object):
             if matched:
                 query[key] = translated
                 continue
-            if not isinstance(value, dict):
+            if isinstance(value, dict):
+                DateOperator.apply(value)
                 continue
-            DateOperator.apply(value)
+            if isinstance(value, list):
+                for item in value:
+                    if isinstance(item, dict):
+                        DateOperator.apply(item)
 
     @staticmethod
     def translate(value):
--- a/server/test/unit/server/db/model/test_criteria.py
+++ b/server/test/unit/server/db/model/test_criteria.py
@@ -348,6 +348,21 @@ class TestDateOperator(unittest.TestCase):
         criteria.DateOperator.apply(query)
         self.assertTrue(isinstance(query['created']['$gt'], datetime))
 
+    def test_apply_multiple_conditions(self):
+        query = {
+            '$and': [
+                {'created': {
+                    '$gt': {'$date': '2013-12-01T14:35:00Z'}
+                }},
+                {'updated': {
+                    '$lt': {'$date': '2020-05-15T14:35:00Z'}
+                }},
+            ]
+        }
+        criteria.DateOperator.apply(query)
+        self.assertTrue(isinstance(query['$and'][0]['created']['$gt'], datetime))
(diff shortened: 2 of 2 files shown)


ipanova:
@rohanpm can you give it a look as well?

ipanova (approved):

ipanova:
@rajulkumar have you run pulp-smash against these changes?

server/pulp/server/db/model/criteria.py
```diff
@@ -499,9 +499,13 @@ def apply(query):
...
                 continue
-            if not isinstance(value, dict):
+            if isinstance(value, dict):
+                DateOperator.apply(value)
                 continue
-            DateOperator.apply(value)
+            if isinstance(value, list):
+                for item in value:
+                    if isinstance(item, dict):
+                        DateOperator.apply(item)
```
goosemania, line 508: It looks like it solves only one level of nesting.
More complex queries don't seem to be covered:
```
        query = {
￼            '$and': [
￼                {'created': {
￼                    '$gt': {'$date': '2013-12-01T14:35:00Z'}
￼                }},
                 {'$or' : [
￼                    {'updated': {
￼                        '$lt': {'$date': '2020-05-15T14:35:00Z'}
￼                    }},
                     {'updated': {
￼                        '$gt': {'$date': '2020-05-20T00:00:00Z'}
￼                    }},
                ]},
￼            ]
￼        }
```

rajulkumar: I think this will work as it's a recursive call to DateOperator.apply() for each item in the list that will go to any level of nesting. Also, tried this example too and it works fine.

rajulkumar:
> @rajulkumar have you run pulp-smash against these changes?

@ipanova No. How should I go about it?

server/pulp/server/db/model/criteria.py
rohanpm: It won't descend into lists of lists though.  If there is any list within a list, then we break out of recursion.  It seems slightly wrong but my guess is that Pulp never contains data where such a query makes sense anyway.

rohanpm (approved):

ipanova:
> > @rajulkumar have you run pulp-smash against these changes?
> 
> @ipanova No. How should I go about it?

here are the docs on how to install and run the test suite https://pulp-2-tests.readthedocs.io/en/latest/ Running platform and rpm tests would suffice.

server/pulp/server/db/model/criteria.py
rajulkumar: Agree. It's with the assumption that it will either be dict or list of dict.

ggainey:
I ran the pulp_rpm pulp-smash tests against this PR and got no new failures. Will fire off the platform tests and report back here. 

@rajulkumar def do read up on pulp_smash, once you're able to run it it will go along way to giving you some reassurance that a proposed change isn't going to Break Things elsewhere.

ggainey:
platform tests pass as well - @ipanova , looks good to me.

ipanova:
@ggainey thank you for looking into this!

rajulkumar:
@ggainey Sure, I'll check it out. Thanks

PR #3989, opened 2020-05-22
Fixed BSON too large when unassociating in large repos

midnightercz:
When unassociating from large repos, to connect repo unit with content unit
pulp queries all units in the repository and then query content units
according to the unit filters + _id in <list-of-ids> returned by query to
repo units. If this <list-of-ids> is too large, mongo will fail with
BSON document too large error.
This commit changes the approach of querying the db. First
unit_type ids are gathered from the query. Then for every unit type
appropriate collection is queried for the content units. For found
content units are queried repo units based on unit _ids.

closes #6806

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1,5 +1,5 @@
 from gettext import gettext as _
-from itertools import chain
+from itertools import chain, izip_longest
 import copy
 import logging
 import os
@@ -222,27 +222,31 @@ def find_repo_content_units(
 
     for unit_type, unit_ids in type_map.iteritems():
         _model = plugin_api.get_unit_model_by_id(unit_type)
-        qs = _model.objects(q_obj=units_q, __raw__={'_id': {'$in': list(unit_ids)}})
-        if unit_fields:
-            qs = qs.only(*unit_fields)
+        for ids_chunk in izip_longest(*[iter(unit_ids)] * 1000):
+            ids_chunk = filter(lambda x: x, ids_chunk)
 
-        for unit in qs:
-            if skip and skip_count < skip:
-                skip_count += 1
-                continue
+            qs = _model.objects(q_obj=units_q,
+                                __raw__={'_id': {'$in': list(ids_chunk)}})
+            if unit_fields:
+                qs = qs.only(*unit_fields)
 
-            if yield_content_unit:
-                yield unit
-            else:
-                cu = content_units[unit_type][unit.id]
-                cu.unit = unit
-                yield cu
+            for unit in qs:
+                if skip and skip_count < skip:
+                    skip_count += 1
+                    continue
 
-            if limit:
-                if yield_count >= limit:
-                    return
+                if yield_content_unit:
+                    yield unit
+                else:
+                    cu = content_units[unit_type][unit.id]
+                    cu.unit = unit
+                    yield cu
 
-            yield_count += 1
+                if limit:
+                    if yield_count >= limit:
+                        return
+
+                yield_count += 1
 
 
 def find_units_not_downloaded(repo_id):
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -222,6 +222,12 @@ def find_repo_content_units(
 
     for unit_type, unit_ids in type_map.iteritems():
         _model = plugin_api.get_unit_model_by_id(unit_type)
+        # do chunks with izip_longest which zips together passed arguments
+        # so from x1,x2,x3 ... xn arguments it returns [x1[0],x2[0],x3[0],..xn[0]] every iteration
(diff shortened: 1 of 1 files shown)


ipanova (commented):
This looks reasonable, but i would benefit from more reviewers @goosemania

ipanova:
@midnightercz have you run pulp-smash against these changes?

midnightercz:
I haven't. Also I forgot to CC our guys here as I promised
@chandwanitulsi  @querti @lipoja @JayZ12138 
please review

ggainey:
@midnightercz I've been looking at this PR today and running it thru the pulp-smash tests. In my dev-env, against 2-master I get **failures=3, errors=1**; with the PR, I am seeing **failures=10, errors=79**, with a lot of the failures being 'too many values to unpack' from associate_from_repo(). I suspect there is some problem that, once fixed, will clean up a lot of tests, but I haven't figured it out yet.

I attached the output from running the following pulp_smash invocation against this PR checked out against 2-master:

`time python3 -m unittest discover pulp_2_tests.tests.rpm.api_v2 |& tee bson-too-large-fix.txt`

to the issue as `bson-too-large-fix.txt.gz`

server/pulp/server/controllers/repository.py
```diff
@@ -205,26 +205,35 @@ def find_repo_content_units(
...
 
     for repo_content_unit in qs:
-        id_set = type_map.setdefault(repo_content_unit.unit_type_id, set())
-        id_set.add(repo_content_unit.unit_id)
-        content_unit_set = content_units.setdefault(repo_content_unit.unit_type_id, dict())
-        content_unit_set[repo_content_unit.unit_id] = repo_content_unit
+        type_list.append(repo_content_unit["_id"])
 
-    for unit_type, unit_ids in type_map.iteritems():
+    for unit_type, unit_ids in type_list:
```
midnightercz: @ggainey  Hi, I'm still in process of setuping pulp and pulp-smash. I assume the error could be caused by this line which I forgot to change properly.

Would it be possible to fix this and run pulp-smash on your side?

ggainey: Can do - will take a little while, my computer greeted me this morning by hard-locking and had to be rebooted, so it's going to take a few minutes to get everything working again :) Will comment here with results

ggainey: Ah, I see - yeah, get me a patch and I can re-run!

midnightercz:
```suggestion
    for unit_type in type_list:
```

ggainey: Much better - we're down to `FAILED (failures=13, errors=14, skipped=16)`. Added test-output to the issue as `midnightercz-bson-too-large-fix-changed.txt.gz`

You def want to be firing up pulp-smash for yourself - see https://pulp-2-tests.readthedocs.io/en/latest/ for details.

midnightercz:
I realized my previous patch could not have work. So I forced pushed new patch. Also I tested with pulp_smash however in middle of test - I assume after some test which requires restarting of workers - it starts failing due unable to connect to the server. @ggainey can I ask for you assistance with running pulp_smash on the new patch?

ggainey:
@midnightercz I will try to get to retesting this #soon - will let you know results when I have them.

ggainey:
@midnightercz My Pulp2 environment is broken - will have to take another shot at this on Mon/Tues.

ggainey:
@midnightercz or maybe tonight instead of Monday.

Current state looks **much** better - we now match the "FAILED (failures=3, errors=1, skipped=16)" I get just running against master. Do keep working on being able to run pulp_smash in your own env though - it catches a lot of problems.

@ipanova -  current state of the commit passes pulp_smash, but prob needs a re-review for changes.

ipanova:
thanks, i will take a look!

server/pulp/server/controllers/repository.py
```diff
@@ -222,27 +222,31 @@ def find_repo_content_units(
...
-                cu = content_units[unit_type][unit.id]
-                cu.unit = unit
-                yield cu
-
-            if limit:
-                if yield_count >= limit:
-                    return
-
-            yield_count += 1
+        for ids_chunk in izip_longest(*[iter(unit_ids)] * 1000):
```
midnightercz, line 231:
```suggestion
        # do chunks with izip_longest which zips together passed arguments
        # so from x1,x2,x3 ... xn arguments it returns [x1[0],x2[0],x3[0],..xn[0]] every iteration
        # until there's item at least in one of the x. Passing arugments as iter() * X will produces
        # zipping from same iterator object thus iterating over the same source.
        # So putting everything together behaves as source [x1,x2,x3,x4,...] is splitted
        # into chunks of size 1000
        for ids_chunk in izip_longest(*[iter(unit_ids)] * 1000):
```

ipanova (approved):

mdellweg (approved):

PR #3990, opened 2020-05-28
Fix regression of #6387

hao-yu:
Regenerate applicability task and retrieve applicability request
are not using the same all_profiles_hash to save and lookup the
repo profile applicability.

closes: #6851
https://pulp.plan.io/issues/6851

### DIFF
--- a/server/pulp/server/db/model/consumer.py
+++ b/server/pulp/server/db/model/consumer.py
@@ -277,9 +277,7 @@ class UnitProfile(Model):
     """
 
     collection_name = 'consumer_unit_profiles'
-    unique_indices = (
-        ('consumer_id', 'content_type'),
-    )
+    unique_indices = Model.unique_indices + (('consumer_id', 'content_type'),)
 
     def __init__(self, consumer_id, content_type, profile, profile_hash=None):
         """
--- a/server/pulp/server/managers/consumer/applicability.py
+++ b/server/pulp/server/managers/consumer/applicability.py
@@ -703,7 +703,7 @@ def _add_profiles_to_consumer_map_and_get_hashes(consumer_ids, consumer_map):
     :rtype:              list
     """
     profiles = UnitProfile.get_collection().find(
-        {'consumer_id': {'$in': consumer_ids}},
+        {'consumer_id': {'$in': consumer_ids}, 'profile': {'$ne': []}},
         projection=['consumer_id', 'profile_hash'])
     all_profiles_hashes = set()
     for p in profiles:


ianballou (commented):
I've recommended this PR as a quick fix for some users who are having applicability issues.  Looks like it's worked for one of them so far: https://community.theforeman.org/t/katello-3-15-doesnt-show-available-package-updates-anymore/18948/64

server/pulp/server/db/model/consumer.py
```diff
@@ -279,6 +279,7 @@ class UnitProfile(Model):
     collection_name = 'consumer_unit_profiles'
     unique_indices = (
         ('consumer_id', 'content_type'),
+        ('id'),
```
goosemania: @hao-yu , could you elaborate why it is needed?
I believe all models have a unique index for ids https://github.com/pulp/pulp/blob/2-master/server/pulp/server/db/model/base.py#L56

server/pulp/server/managers/consumer/applicability.py
```diff
@@ -344,7 +345,7 @@ def _get_consumer_profile_map(consumer_ids):
 
         # Get all unit profiles associated with given consumers
         unit_profile_criteria = Criteria(
-            filters={'consumer_id': {'$in': consumer_ids}, 'profile': {'$ne': []}},
+            filters={'consumer_id': {'$in': consumer_ids}},
```
goosemania: Do I understand correctly that, alternatively, you could have kept it here and add the` 'profile': {'$ne': []}` to the applicability retrieval part https://github.com/pulp/pulp/pull/3990/files#diff-497c4c3e2dfb13249190d0776ea94a5eL706?

I believe filtering out the empty profile works well the way you suggested, we just carry them and filter out at the last moment. FWIW, The goal was to get rid of them at early stages.

Thanks for the patch! I'm asking more to check my understanding of the problem. No need to fix this part.

goosemania (changes requested):
Please remove the id from the indices, I believe it's not necessary to have it there.

hao-yu:
Can you double check if the "id" index is really not needed here? I added it because I found that the id index is missing from the "consumer_unit_profiles" and explain() shows that the query is using colscan without the id index.

Please note that it is the "id" index not the "_id" index which is the default.

server/pulp/server/managers/consumer/applicability.py
goosemania: @hao-yu , actually, thinking more about it... Right now applicability is calculated for `all_profile_hash` which does not include an empty profile. So your fix wouldn't help to retrieve already calculated applicability. User will need to retrigger applicability calculation. 
If we fix the retrieval part, the `all_profile_hash` will be calculated correctly and users will get the already calculated applicability for consumers.

Secondly, someone might clean up consumer empty profiles (I saw support folks doing that) and that will make applicability not being retrieved if it's not recalculated for the same issue as we see now.

It seems that it's safer to fix the retrieval part https://github.com/pulp/pulp/pull/3990/files#diff-497c4c3e2dfb13249190d0776ea94a5eL706.

What do you think? Any flaw in my logic?
Let me know if my comment is unclear, the topic is complicated and not easy to follow :/

server/pulp/server/db/model/consumer.py
goosemania: The class rewrites indices, rather than extending them.
Could you extend the parent indices to make it as it was supposed to be? Thanks.
`unique_indices = Model.unique_indices + (('consumer_id', 'content_type'),)`

goosemania:
@hao-yu , checked on my installation, no index :/ 
Clearly (now), it's a bug, because parent class indices are not used. Commented on the corresponding line.

hao-yu:
@goosemania  done!

goosemania:
@hao-yu , thank you! Do you agree with the suggested logic? And do you have a reproducer to test since katello had troubles to reproduce the issue in the first place.

goosemania (approved):

hao-yu:
@goosemania 
1) Yeah, i agree with your suggestion and I also explained above. 
2) Sorry, my foreman reproducer is in my laptop.

I think you should be able to reproduce the issue through the below steps:
1) Install/Revert the Pulp version before the regression. 
2) Calculate the applicability for a rhel 7 consumer
3) Retrieve the applicability for consumer should return the correct result.
4) Upgrade Pulp to the latest version which has the regression commit.
5) Retrieve the applicability for consumer should get empty result.
6) Patch this commit and retrieve the applicability for consumer should get correct result.

Let me know if you have any question.

goosemania:
Thanks a lot!

PR #3997, opened 2020-07-22
Remove unnecessary criteria of force_full rsync publish

emilyzheng:
Rsync publish after units removal or failed predistributor publish
doesn't have to be force_full publish, so remove them from criteria.

RepoPublishResult doesn't have distributor_config, so the criteria
of predistributor_force_full doesn't work, remove it.

closes #7185
https://pulp.plan.io/issues/7185

### DIFF
--- a/server/pulp/plugins/rsync/publish.py
+++ b/server/pulp/plugins/rsync/publish.py
@@ -363,9 +363,7 @@ class Publisher(PublishStep):
         :return: Whether or not this publish should be in fast forward mode
         :rtype: bool
         """
-        force_full = False
-        config_force_full = self.get_config().get("force_full", False)
-        force_full = force_full | config_force_full
+        force_full = self.get_config().get("force_full", False)
         delete = self.get_config().get("delete", False)
 
         return not force_full and not delete and self.last_predist_last_published
--- a/server/pulp/plugins/rsync/publish.py
+++ b/server/pulp/plugins/rsync/publish.py
@@ -12,7 +12,6 @@ from pulp.common import dateutils
 from pulp.plugins.util.publish_step import PublishStep
 from pulp.server.config import config as pulp_config
 from pulp.server.exceptions import PulpCodedException
-from pulp.server.db.model.repository import RepoPublishResult
 from pulp.server.db.model import Distributor
 
 
@@ -334,19 +333,6 @@ class Publisher(PublishStep):
             scratchpad = self.distributor.scratchpad or {}
             self.last_predist_last_published = scratchpad.get("last_predist_last_published")
 
-        if self.last_published:
-            string_date = dateutils.format_iso8601_datetime(self.last_published)
-        else:
-            string_date = None
-
-        if self.predistributor:
-            search_params = {'repo_id': repo.id,
-                             'distributor_id': self.predistributor["distributor_id"],
-                             'started': {"$gte": string_date}}
-            self.predist_history = RepoPublishResult.get_collection().find(search_params)
-        else:
-            self.predist_history = []
-
         self.remote_path = self.get_remote_repo_path()
 
         if self.is_fastforward():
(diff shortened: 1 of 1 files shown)


midnightercz (approved):

ipanova:
@goosemania @daviddavis @dkliban

server/pulp/plugins/rsync/publish.py
```diff
@@ -378,30 +364,11 @@ def is_fastforward(self):
         :rtype: bool
         """
         force_full = False
-        for entry in self.predist_history:
-            predistributor_force_full = entry.get("distributor_config", {}).get("force_full",
```
goosemania, line 382: I agree I don't see a distributor_config there. However, isn't it a problem?
It seems to me that the intention is correct - to check if there is a force_full configuration on predistributor.
So maybe it makes sense to fix it, aka to get a distributor_id from the RepoPublishResult and then search for its config.

@ipanova, I'm not familiar with rsync distributor so take it with a grain of salt.

emilyzheng: Since it has been broken for a long time, I wonder if the original intention is really correct. If it is, I'd like to know the reason that rsync publish after a force_full predistributor publish has to be force_full.

goosemania: @emilyzheng , when is force_full publish needed for rsync_distributor?

emilyzheng: It can be used for fixing content (wrong/missing symlinks for example) in rsync target. It should also be done when publishing with delete=True which needs handling all units instead of incremental units.

ipanova: @emilyzheng as you said force_full fixes wrong/missing symlinks and at a minimum re-creates repodata. So if the predistributor was published in a force full mode you do want to rsync publish in a force full mode as well to bring in new repodata and fixed symlinks

server/pulp/plugins/rsync/publish.py
```diff
@@ -378,30 +364,11 @@ def is_fastforward(self):
...
-                                                                                False)
-            force_full |= predistributor_force_full
-            if entry.get("result", "error") == "error":
```
ipanova, line 385: i might agree on the fact that probably we do not need to rsync in a force_full mode if publish of the predistributor failed.

server/pulp/plugins/rsync/publish.py
```diff
@@ -378,30 +364,11 @@ def is_fastforward(self):
...
-                force_full = True
-
-        if self.last_published:
-            last_published = self.last_published.replace(tzinfo=None)
-        else:
-            last_published = None
-
-        if self.last_deleted:
```
ipanova, line 393: it feels like a sane logic and rsync should not happen in a force full mode if a unit got removed however I am afraid we might miss some other corner case. @rohanpm can you please review this PR? I feel like you would have the best insight here since my brain needed to be cleaned  from a layer of dust to remember this area.

ipanova:
@emilyzheng did you happen to run pulp2 tests with these changes?

server/pulp/plugins/rsync/publish.py
rohanpm: Yeah, I think this is right.  We looked into the history of this code and discussed it internally prior to this PR.

server/pulp/plugins/rsync/publish.py
rohanpm: I don't think this is needed for the sake of repodata.  If you look at the rsync distributor implementations, if the publish happens at all then they always sync all repodata regardless of whether force_full is used.

I don't think the code can actually know whether a force_full publish is needed.  To me, this is more of an API design question: if caller triggers predistributor with force_full, then can they expect the next rsync distributor to automatically do the same, or is the caller responsible for also specifying that?

If the commit message is right (that this code doesn't work) then the current behavior is that the caller is responsible for deciding whether force_full should be passed to rsync distributor.  The more cautious change is to leave the API behavior alone and not "fix" it to match the original developer's intent, as anyone currently using this must have designed their code to deal with the existing behavior.

emilyzheng:
> @emilyzheng did you happen to run pulp2 tests with these changes?

I did the same regression test with what we do for new release candidates of Pulp.

server/pulp/plugins/rsync/publish.py
ipanova: the last change in this area happened 4 years ago so it is really hard to remember what was the original intent.
I think it makes sense to let the caller explicitly specify `force_full` option as on any other distributor and not make it dependent from predistributor.
In that case we can remove the variable on line 380 and 399 @emilyzheng

emilyzheng: Removed.

ipanova (approved):

PR #3999, opened 2020-09-29
Ignore force_full in config override when deciding publish to be oper…

pmoravec:
…ational

Comparing config override to decide (non)operational publish can safely ignore
force_full option; this option itself determines the non-operational publish
independently.

closes #7526
https://pulp.plan.io/issues/7526

Signed-off-by: Pavel Moravec <pmoravec@redhat.com>

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1198,13 +1198,20 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
         skip_for_predistributor = published_after_predistributor or not \
             predistributor_last_published
 
-    same_override = dist.last_override_config == config_override
-    if not same_override:
+    if not (dist.last_override_config == config_override):
         # Use raw pymongo not to fire the signal hander
         model.Distributor.objects(
             repo_id=repo_obj.repo_id,
             distributor_id=dist_id).update(set__last_override_config=config_override)
 
+    # do not compare force_full from config override - this particular difference itself does not
+    # determine the need of repo publish; so remove the option from the overrides as they are
+    # not used anywhere else since now
+    last_override = dist.last_override_config.copy()
+    curr_override = config_override.copy()
+    last_override.pop('force_full', False)
+    curr_override.pop('force_full', False)
+    same_override = last_override == curr_override
     # Check if content has not changed since last publish and a predistributor is not defined.
     unchanged_content_and_no_predistributor = last_published and not last_updated and \
         not units_removed and not predistributor_id


server/pulp/server/controllers/repository.py
```diff
@@ -1198,13 +1198,18 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
...
         model.Distributor.objects(
             repo_id=repo_obj.repo_id,
             distributor_id=dist_id).update(set__last_override_config=config_override)
 
+    # do not compare force_full from config override - this particular difference itself does not
+    # determine the need of repo publish
+    config_override = {k: v for k, v in config_override.items() if k != 'force_full'}
+    dist.last_override_config = {k: v for k, v in dist.last_override_config.items()
+                                 if k != 'force_full'}
+    same_override = dist.last_override_config == config_override
```
dralley: I would prefer if we used the same code from the other place where we do this

https://github.com/pulp/pulp/blob/b8f5498f90869e14c0a72cfff5f04ae309d58885/server/pulp/server/controllers/repository.py#L947-L956

dralley: Or reuse the same code

goosemania: @dralley , the other place is for an importer override, here it's the distributor's override config. Or is your suggestion to unify it somehow? I agree it can be improved, however in support of touch/change less in pulp 2, this PR looks ok to me. Let me know if I misunderstood or if you are in favor of improving anything here.

dralley: I'm fine with not unifying it (to avoid changes) but I do think we should use the same strategy, since it's marginally cleaner, and would make it more obvious that the same thing is happening.

pmoravec:
Good idea to use `pop` here, PR updated accordingly. Thanks for the feedback!

goosemania:
@pmoravec , last small bit. It's usually not safe to modify dicts you get by reference since the changes are not isolated, and the original object is modified. That's the reason for the importer case both configs were copied before doing `pop`.

Adding something along those lines
```
 prev_config = dist.last_override_config.copy() 
 current_config = call_config.override_config.copy()
 ```
and doing `pop` for those objects would be a safer approach.

Thanks a lot for the patch and for finding the spot!

pmoravec:
Got it, thanks. PR updated.

goosemania (approved):

goosemania:
@dralley , waiting on you before merging. Feel free to merge if you think everything is ok.

dralley (approved):

PR #4000, opened 2020-10-05
Taught util.misc.mkdir() to set umask to allow group-w for makedirs c…

ggainey:
…alls.

Hunted down a LOT of makedirs calls and replaced with call to misc.mkdir()

fixes #7445

### DIFF
--- a/server/pulp/plugins/util/metadata_writer.py
+++ b/server/pulp/plugins/util/metadata_writer.py
@@ -12,7 +12,7 @@ from xml.sax.saxutils import XMLGenerator
 from pulp.common import error_codes
 from pulp.server.exceptions import PulpCodedValidationException, PulpCodedException
 from pulp.server.util import CHECKSUM_FUNCTIONS
-
+from pulp.plugins.util import misc
 _LOG = logging.getLogger(__name__)
 BUFFER_SIZE = 1024
 
@@ -124,7 +124,7 @@ class MetadataFileContext(object):
             parent_dir = os.path.dirname(self.metadata_file_path)
 
             if not os.path.exists(parent_dir):
-                os.makedirs(parent_dir, mode=0770)
+                misc.mkdir(parent_dir, mode=0770)
 
             elif not os.access(parent_dir, os.R_OK | os.W_OK | os.X_OK):
                 msg = _('Insufficient permissions to write metadata file in directory [%(d)s]')
--- a/server/pulp/plugins/util/misc.py
+++ b/server/pulp/plugins/util/misc.py
@@ -40,16 +40,20 @@ def mkdir(*args, **kwargs):
     """
     Create the specified directory.
     Tolerant of race conditions.
+    Sets umask to 002.
 
     :param args: path[, mode] that goes to os.makedirs
     :param kwargs: path
     :return:
     """
+    mask = os.umask(0002)
     try:
         os.makedirs(*args, **kwargs)
     except OSError, e:
         if e.errno != errno.EEXIST:
             raise
+    finally:
+        os.umask(mask)
 
 
 def get_parent_directory(path):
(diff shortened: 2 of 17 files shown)


server/pulp/plugins/util/metadata_writer.py
```diff
@@ -124,7 +124,7 @@ def _open_metadata_file_handle(self):
             parent_dir = os.path.dirname(self.metadata_file_path)
 
             if not os.path.exists(parent_dir):
```
ipanova, line 126: you don't need to check for presence of the path here, remove the if condition

server/pulp/server/content/storage.py
```diff
@@ -6,22 +6,7 @@
 from hashlib import sha256
 
 from pulp.server.config import config
-
-
-def mkdir(path):
```
ipanova, line 11: :+1:  let's say no to the code duplication!

server/pulp/plugins/util/metadata_writer.py
ggainey: The os.access() check in the elif is only needed if we're not creating the path in the mkdir() call. I tried a couple of iterations here, and decided that while this one is a little redundant around the mkdir(), it seemed more clear in terms of understanding the code. What do you think?

server/pulp/server/async/tasks.py
```diff
@@ -780,11 +782,7 @@ def _handle_cProfile(self, task_id):
         if config.getboolean('profiling', 'enabled') is True:
             self.pr.disable()
             profile_directory = config.get('profiling', 'directory')
-            try:
-                os.makedirs(profile_directory, 0755)
-            except OSError as exc:
-                if exc.errno != errno.EEXIST:
-                    raise
+            misc.mkdir(profile_directory, mode=0755)
```
ipanova, line 785: 0775

server/pulp/plugins/util/metadata_writer.py
ipanova: oooh right, i did not notice the elif statement, i am cool with leaving if

server/pulp/server/async/tasks.py
ipanova: i think this is important and i understand correctly how umask works - if having mode 0755 the result will be:
 usr grp others 
-rwx  r-x   r-x is represented in octal as 0755
-000 000 010 umask of 002 removes any permission where there is a 1
-rwx  r-x   r-x is the result achieved requesting 0755 with umask of 002...and we are screwed :)


if having mode 0775 the result will be:
 usr grp others 
-rwx  rwx  r-x is represented in octal as 0775
-000 000 010 umask of 002 removes any permission where there is a 1
-rwx  rwx  r-x   is the result achieved requesting 0775 with umask of 002

ggainey: I left a few things like certificate-directories and this profiling-directory alone - they're not 'content', and are more sensitive to permissions. Unless we're migrating profiling-information, this set of perms shouldn't affect migration, and I felt it better to leave the permissions as they had been specifically set.

(Your understanding of umask is correct - it turns off permission-bits where the umask has a 1. So 0755, with the mkdirs() umask of 0002, leaves 0755)

ipanova (approved):
changes look good to me, i will try to test them tomorrow.

ipanova:
@rohanpm please take a look to be sure that you are not impacted by this change in an unexpected way

server/pulp/plugins/util/misc.py
```diff
@@ -40,13 +40,16 @@ def mkdir(*args, **kwargs):
...
     Tolerant of race conditions.
+    Sets umask to 002.
 
     :param args: path[, mode] that goes to os.makedirs
     :param kwargs: path
     :return:
     """
     try:
+        mask = os.umask(0002)
         os.makedirs(*args, **kwargs)
```
rohanpm, line 51: The umask is apparently never restored if this raises an exception.  Should the original os.umask be above the "try", and the umask reset in a new "finally" block?

rohanpm:
> @rohanpm please take a look to be sure that you are not impacted by this change in an unexpected way

Thanks Ina, I don't see any reason this would be harmful for RHSM Pulp, though I believe there is a bug where umask may not be restored when intended.

server/pulp/plugins/util/misc.py
ggainey: ...I am an idiot. I had exactly that at some point, and apparently lost it somewhere along the way. Thank you for the great catch, that would have been *very* embarrassing to allow out into the wild!

ggainey: @rohanpm @ipanova fixed mkdir() - more eyes please, when you have a moment free

PR #4008, opened 2020-11-12
find_repo_content_units: skip db query if expected result is empty

rbikar:
For performance reasons, this change does server-side counting
of units that are expected to be in result before actual db query
happens. If expected count is zero, skip db query and continue to
the next chunk of unit ids.

The performance problem is noticable in RHSM-pulp for big repos
with 200k and more units associated. During associate action
results of the most of queries in affected code will be empty
so we can skip querying the db this way.

Without this fix, code will do eg. 200 unnecesary queries for repo
with 200k units which will take a lot time to finish that leads
to major performance regression.

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -234,6 +234,12 @@ def find_repo_content_units(
                                 __raw__={'_id': {'$in': list(ids_chunk)}})
             if unit_fields:
                 qs = qs.only(*unit_fields)
+            
+            # for performace reasons, do server-side counting of units the will be returned in result
+            # before we do actual query to db
+            # if expected count of units in result is zero, continue to next chunk of unit_ids 
+            if qs.count() == 0:
+                continue
 
             for unit in qs:
                 if skip and skip_count < skip:
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -234,10 +234,10 @@ def find_repo_content_units(
                                 __raw__={'_id': {'$in': list(ids_chunk)}})
             if unit_fields:
                 qs = qs.only(*unit_fields)
-            
-            # for performace reasons, do server-side counting of units the will be returned in result
-            # before we do actual query to db
-            # if expected count of units in result is zero, continue to next chunk of unit_ids 
+
+            # for performace reasons, do server-side counting of units the will be returned in
+            # result before we do actual query to db
+            # if expected count of units in result is zero, continue to next chunk of unit_ids
             if qs.count() == 0:
                 continue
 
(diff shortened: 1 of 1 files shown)


rbikar:
@rohanpm  @midnightercz  @ipanova 
Can you please review?

server/pulp/server/controllers/repository.py
```diff
@@ -235,6 +235,12 @@ def find_repo_content_units(
             if unit_fields:
                 qs = qs.only(*unit_fields)
 
+            # for performace reasons, do server-side counting of units the will be returned in
+            # result before we do actual query to db
+            # if expected count of units in result is zero, continue to next chunk of unit_ids
+            if qs.count() == 0:
```
rohanpm, line 241: The logic here looks right, no concerns from "is it functionally correct" point of view.

Can I ask though whether the performance impact is still just a theory or whether it was verified by testing?  It's really surprising to me if the mongo performance of doing a count with no results, vs doing a find with no results, has a notable difference.  It seems like the DB should be performing almost exactly the same work in both cases.

I tried some testing myself against one of the actual DBs involved (QA env), using scripts: https://gist.github.com/rohanpm/1652c648a135a22af34547f7faf19e05

It doesn't simulate the queries here exactly, but it deals with same number of ids in "$in" and uses a query which will return no results.  Timing it shows up pretty much as I expected, no notable difference between the two.

That test of course is not definitive, but if this hasn't been tested I'm a little skeptical about it.

midnightercz: My theory is that degradation happens due to instantiation of qs for every chunk - which I think creates new mongo cursor every time. If my theory is correct I think this won't help to resolve the problem

rbikar: I tested the solution with partial prod db on my laptop with pub push-advisory for e2e repos.
When I debugged where exactly execution waits incredible amount of time it was exactly on `for unit in qs:` 
With skipping on count==0, performance was almost back to normal.
It is true that qs is instantiated for every chunk but actual query is happening during the iteration over qs.
I'll try to come up with some reproducer script, it looks it needs more realistic context for replicating the problem.

rohanpm: Based on the testing done elsewhere (RHELDST-2771), although this is unintuitive and might point to some deficiency in mongo or one of the client libraries, it really is giving a major performance benefit.  Apparently, doing a search which finds nothing can be a lot slower than doing a count which finds nothing.

This PR gets a :+1: from me as we have evidence that it will significantly improve the performance.

rohanpm (approved):

goosemania:
@rbikar, thanks for the PR and clarifications around this change. Please squash the commits and we'll merge it.

ipanova (approved):
Based on the provided reproducer and write up, lgtm

PR #4009, opened 2020-11-25
Adds task failure handling for celery in bad state

bmbouter:
Sometimes celery's internal state is in such a bad position it causes
the on_failure handler itself to fail. This is very bad for Pulp as it
causes the tasking system to stop processing tasks due to on_failure not
releasing locks like it should.

This adds a defensive codepath for that situation which maintains the
current functionality, but if Celery is uncooperative it still performs
the on_failure cleanup.

fixes #5502

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -737,6 +737,21 @@ class Task(PulpTask, ReservedTaskMixin):
             self._handle_cProfile(task_id)
             common_utils.delete_working_directory()
 
+    def _handle_on_failure_cleanup(self, task_id, exc, einfo):
+        now = datetime.now(dateutils.utc_tz())
+        finish_time = dateutils.format_iso8601_datetime(now)
+        task_status = TaskStatus.objects.get(task_id=task_id)
+        task_status['state'] = constants.CALL_ERROR_STATE
+        task_status['finish_time'] = finish_time
+        task_status['traceback'] = einfo.traceback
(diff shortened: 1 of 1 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -758,7 +758,11 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
             # celery will log the traceback
         if kwargs.get('scheduled_call_id') is not None:
             utils.increment_failure_count(kwargs['scheduled_call_id'])
-        if not self.request.called_directly:
+        try:
+            called_directly = self.request.called_directly
+        except Exception:
```
ipanova: can we catch a more specific exception?

ipanova:
update please the commit message so it also contains the issue number

server/pulp/server/async/tasks.py
bmbouter: We could, but if we don't handle any exception from Celery by continuing to remove the locks Pulp will deadlock. I did rework this commit in two ways though:

1) to re-raise the Celery exception instead of using the `except Exception; pass` pattern
2) to also log the original exception since we're loosing that due to celery never logging it as we expect to.

dralley (approved):

ipanova (approved):

server/pulp/server/async/tasks.py
ipanova: The changes make sense to me, thank you.

PR #4011, opened 2021-01-13
Add index to _last_updated for ContentUnits

ggainey:
fixes #8044

### DIFF
--- a/server/pulp/server/db/migrations/0030_last_updated_indices.py
+++ b/server/pulp/server/db/migrations/0030_last_updated_indices.py
@@ -1,0 +1,19 @@
+import pymongo
+
+from pulp.server.db import connection
+
+
+def migrate(*args, **kwargs):
+    """
+    Add an index on ContentUnit._last_updated, to allow ordering by update
+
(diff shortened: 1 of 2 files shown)


server/pulp/server/db/migrations/0030_last_updated_indices.py
```diff
@@ -0,0 +1,20 @@
+import pymongo
+
+from pulp.server.db import connection
+
+
+def migrate(*args, **kwargs):
+    """
+    Add an index on last_updated and _last_updated, to allow ordering by update, on
```
ipanova: This sentence is a bit misleading - did you mean to add index also on Importers/Distributors?

ggainey: Ah, great catch! No, that's the last remains of an early assumption that everything that looked like last_updated needed to be indexed to fix the problem. Let me correct the comment and repush.

ipanova (approved):

goosemania (approved):

PR #4013, opened 2021-01-18
2.21.5 Cherrypicks

pcreech:


ggainey (approved):
These look correct to me

PR #4015, opened 2021-01-18
Update tests for find_repo_content_units() func

pcreech:
In (#4008) 033b2e656989e7ea2f68a10bf13d1a7c9c772bd6, the function
was updated to skip db query, if the expected results is empty.

It seems that it was forgotten to update tests, so this change
updates tests to reflect the change in #4008.

In this commit I've also fixed the repository.py file.
There was accidentally added unicode char to the end the repository.py file,
it caused failure during run of coverage tool in CI.

(cherry picked from commit 0e0b90587f6ff8587f103a58c14ae47a1fe62f64)

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1830,4 +1830,3 @@ class LazyUnitDownloadStep(DownloadEventListener):
         else:
             if not os.path.isfile(file_path):
                 raise IOError(_("The path '{path}' does not exist").format(path=file_path))
-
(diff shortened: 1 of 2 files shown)


ggainey:
> Hello @pcreech! Thanks for opening this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`server/test/unit/server/controllers/test_repository.py`](https://github.com/pulp/pulp/blob/9ab55b3bea8533811b6c04b9c3d8abce135c8627/server/test/unit/server/controllers/test_repository.py):
> 
> > [Line 233:98](https://github.com/pulp/pulp/blob/9ab55b3bea8533811b6c04b9c3d8abce135c8627/server/test/unit/server/controllers/test_repository.py#L233): [W291](https://duckduckgo.com/?q=pep8%20W291) trailing whitespace

Not a cherry-pick problem, it's this way in the source - ignore.

ggainey (approved):
Matches the commit being cherry-picked, lg2m

PR #4018, opened 2021-01-28
Set the auth private key size to be 2048 bits

mprahl:
Clients such as Fedora 33 have a default OpenSSL configuration
that doesn't allow the usage of 1024 bit keys.

See more information at:
https://fedoraproject.org/wiki/Changes/StrongCryptoSettings2

closes #8317

### DIFF
--- a/server/pulp/server/managers/auth/cert/cert_generator.py
+++ b/server/pulp/server/managers/auth/cert/cert_generator.py
@@ -178,7 +178,7 @@ class SerialNumber:
 
 
 def _make_priv_key():
-    cmd = 'openssl genrsa 1024'
+    cmd = 'openssl genrsa 2048'
     p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     p.wait()
     exit_code = p.returncode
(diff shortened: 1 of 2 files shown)


mprahl:
FYI, I did not run the tests locally since it wasn't immediately obvious how to run them. If a maintainer could verify them locally, that'd be appreciated.

ipanova:
@mprahl thank you for the PR and sorry for the delayed answer. Please open an upstream issue and reference it in the commit message in the format of ` closes #1234 ` where 1234 is the issue number. https://pulp.plan.io/issues/new

ipanova:
you need to add this new config option to the Security stanza https://github.com/pulp/pulp/blob/2-master/server/etc/pulp/server.conf#L110

mprahl:
Thanks for the review @ipanova. Could you please take another look?

mprahl:
@ipanova I've change it to hard code 2048 bit as previously discussed.

bmbouter:
I can't speak to its backwards compatibility so I won't LGTM myself, but if it is fully backwards compatible I'm +1 on merging and releasing in a 2.21.z.

ipanova (approved):

PR #4019, opened 2021-02-12
Fixes logging traceback in task.on_failure

bmbouter:
The bugfix for issue 5502 introduced a traceback in some failure
scenarios. This PR causes the logging of the traceback to occur in all
scenarios.

closes #8251

### DIFF
--- a/server/pulp/server/async/tasks.py
+++ b/server/pulp/server/async/tasks.py
@@ -764,21 +764,17 @@ class Task(PulpTask, ReservedTaskMixin):
         :param kwargs:  Original keyword arguments for the executed task.
         :param einfo:   celery's ExceptionInfo instance, containing serialized traceback.
         """
+        _logger.debug(traceback.format_exc())
         if isinstance(exc, PulpCodedException):
             _logger.info(_('Task failed : [%(task_id)s] : %(msg)s') %
                          {'task_id': task_id, 'msg': str(exc)})
-            _logger.debug(traceback.format_exc())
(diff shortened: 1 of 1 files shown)


server/pulp/server/async/tasks.py
```diff
@@ -764,20 +764,19 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
         :param kwargs:  Original keyword arguments for the executed task.
         :param einfo:   celery's ExceptionInfo instance, containing serialized traceback.
         """
+        original_formatted_traceback = traceback.format_exc()
```
ipanova: i don;t see traceback variable defined or passed anywhere, should it be einfo.traceback?

ipanova: i see, we have an imported traceback module, nvm then!

ipanova (approved):

server/pulp/server/async/tasks.py
```diff
@@ -764,20 +764,19 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
         if isinstance(exc, PulpCodedException):
             _logger.info(_('Task failed : [%(task_id)s] : %(msg)s') %
                          {'task_id': task_id, 'msg': str(exc)})
             _logger.debug(traceback.format_exc())
```
ipanova: @bmbouter since we have introduced this variable, can we make sure we also use it here? :)

PR #4020, opened 2021-02-25
Allow find_units method to include only certain fields

hao-yu:
closes: #8306
https://pulp.plan.io/issues/8306


goosemania (approved):

ipanova (approved):

dralley:
As discussed in the RPM meeting, this is something that we might consider merging independently of the other PR, if it would make hotfix distribution easier.

PR #4022, opened 2021-05-13
Set last_publish as a start of publish [RHELDST-5086, #8778] 

rbikar:
When a repository is published, set last_publish field on distributor
to the start time of publish instead of finish time.

With this change pulp shouldn't falsely skip publish of a repo when
last_unit_added is set to the time between publish start and finish.

### DIFF
--- a/server/pulp/server/controllers/repository.py
+++ b/server/pulp/server/controllers/repository.py
@@ -1234,7 +1234,7 @@ def check_publish(repo_obj, dist_id, dist_inst, transfer_repo, conduit, call_con
         # Use raw pymongo not to fire the signal hander
         model.Distributor.objects(
             repo_id=repo_obj.repo_id,
-            distributor_id=dist_id).update(set__last_publish=publish_end_timestamp)
+            distributor_id=dist_id).update(set__last_publish=publish_start_timestamp)
 
         result_code = RepoPublishResult.RESULT_SKIPPED
         _logger.debug('publish skipped for repo [%s] with distributor ID [%s]' % (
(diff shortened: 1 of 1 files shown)


rbikar:
@rohanpm  @midnightercz  @ipanova 
Please review.

rohanpm (approved):

goosemania:
@rbikar , could you share steps to reproduce or explain the situation when the publish is falsely skipped.
In theory, `last_unit_added` should not be set between start and finish of the publish because publish task locks on the repository it is publishing, so no other tasks should be run for that repository in parallel. I'm not sure how it happens that last_unit_added is set while publish is running.

rbikar:
@goosemania  
Sure. This is related ticket https://pulp.plan.io/issues/5951 for previous fix of the problem. The current fix relates to https://projects.engineering.redhat.com/browse/RHELDST-5086. The problematic content type here is erratum that is shared with more than one repository. When this erratum gets updated, pulp will set `last_unit_added` to all affected repos, including those currently in publish (afaict only erratum type units behave like this). So the subsequent fast-forward publish will be skipped because the erratum was updated during publish.

goosemania:
Thank you for the explanation!
Unfortunately, it makes sense.

goosemania (approved):

goosemania:
@jlsherrill , fyi, more times publish will be operational with that change, but it's a valid case.

rbikar:
@goosemania 
Do we need more reviews for this change? 
If not, can you please merge it?

ipanova:
@rbikar the changes make sense to me, especially after your explanation. Will it be too much of the trouble to file an issue on plan.io and basically copy/paste what you have written here https://github.com/pulp/pulp/pull/4022#issuecomment-843042306, also update the commit so it references the issue. Thank you!

rbikar:
@ipanova  
Created https://pulp.plan.io/issues/8778 and updated commit message. Is it ok now?

PR #4026, opened 2021-09-29
SELinux: Add support for pulp-2to3-migration plugin

mikedep333:
The rhsm interface was not used because it is not present on EL7.

Corresponds to https://pulp.plan.io/issues/9468
which has more info.

Closes: #9470

### DIFF
--- a/server/selinux/server/pulp-server.te
+++ b/server/selinux/server/pulp-server.te
@@ -29,6 +29,14 @@ corenet_tcp_connect_mongod_port(httpd_t)
 
 allow httpd_t cert_t:file write;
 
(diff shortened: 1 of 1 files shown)


mikedep333:
I can't set labels, but this is currently WIP. I am still testing/debugging.

mikedep333:
I was unable to reproduce the original error on my test system as I was testing https://github.com/pulp/pulpcore-selinux/pull/38 along with this..

But it does build successfully, even on EL7.

mikedep333:
@ekohl @zpytela Please review.

dkliban (approved):

PR #4027, opened 2021-10-30
Don't swallow the error when a call to finalize() fails

dralley:
closes: #9205
https://pulp.plan.io/issues/9205


goosemania:
I wonder if that can potentially help with katello issues related to the package count discrepancy. cc @ggainey 
At the same time we should have seen the logging statement about finalizing step failing, so I'mnot sure it's related.

goosemania (approved):

